PR #64, opened 2014-11-13
Makes BTDropInViewController conform more gracefully to a user provided theme

ggamecrazy:
As above. 

I needed finer integration with the theme property in BTDropInViewController. Also made BTDropInViewController's subviews handle more gracefully user provided themes.

Perhaps someone else will find these changes useful.

Cheers,

### DIFF
--- a/Braintree/Drop-In/BTDropInContentView.m
+++ b/Braintree/Drop-In/BTDropInContentView.m
@@ -207,16 +207,16 @@
 }
 
 - (void)setTheme:(BTUI *)theme {
-  [super setTheme:theme];
-  
-  self.cardFormSectionHeader.textColor = theme.sectionHeaderTextColor;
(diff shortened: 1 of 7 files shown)


egold:
Yeah it looks like although there is a `theme` property on the drop-in view controller, the way `initWithClient` sets up the drop-in view controller, providing a theme after instantiating the view controller will have no effect.

It seems like the only way to solve this is to subclass `BTDropInViewController` and setup a different designated initializer, and also to re-declare all of the private interface extensions such as `@property (nonatomic, strong) BTDropInContentView *dropInContentView;`

Is this intended behavior?

WangYaoYi:
paypal

demerino:
Take a look at `4.4.0`+. We've added a new Drop-In component and exposed our UI components (along with styling). More info [here](https://github.com/braintree/braintree_ios/blob/master/Docs/Drop-In-Update.md).

PR #69, opened 2014-12-04
Implementing remaining PKPaymentAuthorizationViewControllerDelegate methods

jurezove:
I haven't written any tests since this is basically just a passthrough of the delegate calls. We found it useful since we're calculating tax when a user selects an address on the Apple Pay sheet.

Not sure if defining methods in `BTPaymentMethodCreationDelegate.h` as `@optional` is the best way to go. Any suggestions?


demerino:
Closing. Reopen if the issue is still present in `4.x`.

PR #73, opened 2015-01-08
Switch delegate notification order

sberrevoets:
In our app, we slightly divert from the (probably common) pattern of presenting the PayPal view controller modally because we are already in a modal context. Instead, we push the PayPal view controller on the current navigation stack. When the user taps Cancel in the top left corner, we pop the view controller and return to our modal context. If the user successfully authenticates with PayPal, we dismiss the entire modal stack and return to our main app.

Right now, when our delegate gets sent `paymentMethodCreator:requestsDismissalOfViewController:`, we have no idea if the request happens because the user tapped Cancel or because the payment method had been created. We need to know this, because in case of a Cancel we need to pop, but in case of authentication we need to dismiss the entire modal stack. By sending `paymentMethodCreator:didCreatePaymentMethod:` first, we can distinguish these two events from each other and do the right thing based on that.

I ran unit tests, they still pass, so it looks like this is a safe change to make.

### DIFF
--- a/Braintree/Payments/@Public/BTPaymentMethodCreationDelegate.h
+++ b/Braintree/Payments/@Public/BTPaymentMethodCreationDelegate.h
@@ -7,6 +7,8 @@
 /// an object creating a payment method, such as BTPaymentButton or BTPaymentProvider.
 @protocol BTPaymentMethodCreationDelegate <NSObject>
 
+@optional
+
 /// The payment method creator requires presentation of a view controller in order to
 /// proceed.
 ///
--- a/Braintree/Payments/BTPaymentProvider.m
+++ b/Braintree/Payments/BTPaymentProvider.m
@@ -231,8 +231,8 @@
 }
 
 - (void)payPalViewController:(BTPayPalViewController *)viewController didCreatePayPalPaymentMethod:(BTPayPalPaymentMethod *)payPalPaymentMethod {
-    [self informDelegateRequestsDismissalOfAuthorizationViewController:viewController];
     [self informDelegateDidCreatePaymentMethod:payPalPaymentMethod];
+    [self informDelegateRequestsDismissalOfAuthorizationViewController:viewController];
 }
 
 - (void)payPalViewController:(__unused BTPayPalViewController *)viewController didFailWithError:(NSError *)error {
--- a/Braintree/Payments/BTPaymentProvider.m
+++ b/Braintree/Payments/BTPaymentProvider.m
@@ -241,8 +241,8 @@
 }
 
 - (void)payPalViewControllerDidCancel:(BTPayPalViewController *)viewController {
-    [self informDelegateRequestsDismissalOfAuthorizationViewController:viewController];
     [self informDelegateDidCancel];
+    [self informDelegateRequestsDismissalOfAuthorizationViewController:viewController];
 }
 
 #pragma mark BTAppSwitchingDelegate


burnto:
Hi @sberrevoets - thanks! We'll review and reply back here.

For what it's worth, the PayPal UI is intended to be presented modally, rather than within a navigation controller. `BTPayPalViewController` is actually just a thin wrapper around a `PayPalProfileSharingViewController` (provided by PayPal) that is a `UINavigationController`. I see your design issue with this restriction though, and if it's working for you to push the PayPal UI on your navigation stack, then great.

sberrevoets:
I just realized that I was pushing a navigation controller in another navigation controller while debugging another issue. Didn't realize this before, because everything did seem to work fine like you said.

I will discuss this internally as well, as what we want to do does not seem to be supported and I don't know if that's something we want then. My PR is only symptom fixing at that point and we may (and, in fact, are) run into other issues because of that.

In any case, I couldn't find this limitation in the documentation. Unless I missed it, it may be worth adding that to save time for other developers in the future.

Thanks for your quick response!

burnto:
@sberrevoets you're right, we should make this more explicit in the header. Thanks for the feedback.

mickeyreiss:
I might even go as far as saying that I would consider this to represent an design flaw in the `BTPaymentMethodCreationDelegate` protocol.

In a future major revision, I might consider changing `paymentMethodCreator:requestsPresentationOfViewController:`, such that it receives a `UINavigationController` and specifies that this view controller *must* be presented. 

An interesting side note: `PKPaymentAuthorizationViewController` (i.e. Apple Pay) is also a `UIViewController` subclass, but it seems to me as if it is designed to be presented modally and never pushed.

sberrevoets:
Well, if you're going to make a major change, might as well give users the option to push instead of present, right? :wink: 

In any case, after internal discussion, we will present it modally. However, this PR is still relevant, as I would like to be able to tell the difference between a request for dismissal because the user tapped Cancel, or because the authentication was successful. In the first case we will dismiss just `BTPayPalViewController`, in the second case we want to dismiss the entire modal stack.

Thanks @burnto and @mickeyreiss for your input.

mickeyreiss:
> Well, if you're going to make a major change, might as well give users the option to push instead of present, right? :wink:

We've learned a lot from 3.x release and intend to overhaul a number of [smells](http://en.wikipedia.org/wiki/Code_smell) we've learnt about in the future. 

> However, this PR is still relevant, as I would like to be able to tell the difference between a request for dismissal because the user tapped Cancel, or because the authentication was successful.

We're a little bit apprehensive about accepting this change (even if it may be a better design) since it may be breaking for other developers who rely on the _current_ ordering. Do you have any other ideas for being able to distinguish between successes and cancelations?

Thank you, @sberrevoets, for spurring the discussion!

sberrevoets:
I understand your apprehension. Another, arguably better, solution would be to add a parameter to `paymentMethodCreator:requestsDismissalOfViewController:` that indicates the reason for dismissal. A simple `enum` will do there, with reason `BTViewControllerDismissalReasonCanceled` and `BTViewControllerDismissalReasonAuthenticated`, or something like that.

Personally, I like that design better, but since it's a breaking change that's not backwards compatible, it would require a major version update (assuming you operate under the semantic versioning principles). You could choose to provide this new delegate method in addition to the old one and have it not break backwards compatibility, but then you're stuck with 2 methods.

Not sure what the best way to move forward is here.

Reflejo:
Another option that is backwards compatible would be: provide the status as a property or method in either the Payment Provider (which is the `sender` of that delegate) or the BTPayPalViewController. Probably the former makes more sense.

mickeyreiss:
@Reflejo I like that idea. Pull requests would be welcome; otherwise, we will add this to our backlog and try to add that property.

Reflejo:
@mickeyreiss I'm gonna take a look when I get some spare time. One quick question: Is there any reason why here (It's not the only place but just as an example): https://github.com/braintree/braintree_ios/blob/master/Braintree/Payments/BTPaymentProvider.m#L91

You guys are calling `[self.delegate paymentMethodCreator:self didFailWithError:error];` instead of `[self informDelegateDidFailWithError:error];` ?. The first one might crash if the delegate doesn't respond to the selector, right?. If that's intended let me know, if it's not I'm gonna change it as part of the PR.

mickeyreiss:
> You guys are calling [self.delegate paymentMethodCreator:self didFailWithError:error]; instead of [self informDelegateDidFailWithError:error]; ?

Yes, I would treat this as a bug; it was probably a refactoring mistake when we added the informer methods. 

One thing to note, which we will have to confirm internally: `-informDelegateDidFailWithError` also [posts an analytics event](https://github.com/braintree/braintree_ios/blob/979f4d63b5bb005794309d4ddcc1833ae1cb6927/Braintree/Payments/BTPaymentProvider.m#L213); I'll need to double check whether or not we should be bypassing the analytics in these cases.

This is one of the unfortunate areas of the SDK that are difficult for external developers to contribute to effectively :(

PR #81, opened 2015-01-27
Include Data subspec in Podfile

ccwasden:
Data subspec wasn't coming through, so I had no access to BTData


burnto:
Hi @ccwasden - thanks for this PR. 

Many developer have no need for `BTData`, so we intentionally omitted it from `default_subspecs`. 

If you need to use `BTData` in your project, you can optionally include it just by adding it to your `Podfile` alongside `Braintree`:

```ruby
pod 'Braintree'
pod 'Braintree/Data'
```

I'm closing this, but please reply if you feel further clarification or discussion is warranted. Thanks! :star2:

ccwasden:
That's perfect I don't have a problem with it, but just FYI your docs on
this page might benefit from that explanation
https://developers.braintreepayments.com/ios+ruby/guides/fraud-tools

burnto:
@ccwasden excellent point. I'm making a change to those docs now.

PR #84, opened 2015-02-09
Added BTClientMetadata to public headers in order to make Braintree work with swift

Reflejo:
This small change makes Braintree compatible with swift and the new cocoapods' framework support.

### DIFF
--- a/Braintree.podspec
+++ b/Braintree.podspec
@@ -42,7 +42,7 @@ Pod::Spec.new do |s|
 
   s.subspec "API" do |s|
     s.source_files  = "Braintree/API/**/*.{h,m}"
-    s.public_header_files = "Braintree/API/@Public/*.h"
+    s.public_header_files = "Braintree/API/@Public/*.h", "Braintree/API/Client/BTClientMetadata.h"
     s.weak_frameworks = "PassKit"
     s.frameworks = "AddressBook"
   end


mickeyreiss:
Thanks for the PR. Do you have any insight as to why only this private header is needed for swift support but no others? Is there any other way to add compatibility without marking this file as part of the public API?

Reflejo:
@mickeyreiss `BTClient_Metadata` is importing `BTClientMetadata.h` as a quoted header and exposing a "metadata" property on the `BTClient` class publicly. The problem is that `BTClientMetadata` is not in the scope of Braintree's public headers. Sadly you can't include quoted headers inside your public headers that are not by itself public. Three options arise here:

a) Make `BTClientMetadata.h` public
b) Make `BTClient_Metadata.h` private (not sure if that is an option).
c) Import `BTClientMetadata.h` using a system import as in: `#import <Braintree/BTClientMetadata.h>` instead of as a quoted header.

mickeyreiss:
Interesting. I don't see `BTClient_Metadata` as public anywhere. It is intended to be private.

The `BTClient` `metadata` property is declared in `BTClient_Metadata.h`, which is only imported by `.m` files.

Are interface extensions treated as "public" by the swift compiler even if they are not located in (or imported into) a public header?

Reflejo:
@mickeyreiss `BTClient_Metadata` is public, it's actually here: `./Braintree/API/@Public/BTClient_Metadata.h` which is included as public in the podspecs: `s.public_header_files = "Braintree/API/@Public/*.h"`

mickeyreiss:
@Reflejo Ahh - that explains it. I believe that is a mistake, and I think the best solution would be to move `BTClient_Metadata.h` out of the `@Public` directory.

mickeyreiss:
I'm working on another PR to make that change right now.

Reflejo:
Closed in favor of #85

mickeyreiss:
Thanks for helping us iron this one out, @Reflejo!

PR #85, opened 2015-02-09
Conslidate BTClient_Metadata.h with BTClient_Internal.h

mickeyreiss:
`BTClient_Metadata.h` was intended to be a private header but it has
been in `@Public` incidentally since it was introduced. This commit
corrects that by moving it to the BTClient extension in
`BTClient_Internal.h`.

This is an alternate fix for #84.

cc/ @Reflejo


Reflejo:
:+1:

mickeyreiss:
Merging since there hasn't been any more feedback on this diff.

PR #87, opened 2015-02-11
Added status property to BTPaymentProvider

Reflejo:
This provides information about the status of the transaction by the time that requestsDismissalOfViewController: is called. See #73

### DIFF
--- a/Braintree/Payments/@Public/BTPaymentProvider.h
+++ b/Braintree/Payments/@Public/BTPaymentProvider.h
@@ -20,7 +20,7 @@ typedef NS_ENUM(NSInteger, BTPaymentProviderType) {
 
 /// Status of the last (or ongoing) transaction
 typedef NS_ENUM(NSInteger, BTPaymentProviderStatus) {
-    /// First status: The payment method was not created yet
+    /// The payment method was not created yet (default status)
     BTPaymentProviderStatusUninitialized = 0,
 
     /// Payment method created (either PayPal, Venmo or ApplePay)
@@ -34,7 +34,7 @@ typedef NS_ENUM(NSInteger, BTPaymentProviderStatus) {
     BTPaymentProviderStatusError,
 
     /// Payment method creation is complete with success
-    BTPaymentProviderStatusSuccees,
+    BTPaymentProviderStatusSuccess,
 
     /// The payment method creator has cancelled
     BTPaymentProviderStatusCancelled
--- a/Braintree/Payments/@Public/BTPaymentProvider.h
+++ b/Braintree/Payments/@Public/BTPaymentProvider.h
@@ -18,7 +18,7 @@ typedef NS_ENUM(NSInteger, BTPaymentProviderType) {
     BTPaymentProviderTypeApplePay
 };
 
-/// Status of the last (or ongoing) transaction
+/// Status of the last (or ongoing) payment method creation
 typedef NS_ENUM(NSInteger, BTPaymentProviderStatus) {
     /// The payment method was not created yet (default status)
     BTPaymentProviderStatusUninitialized = 0,
@@ -118,7 +118,7 @@ typedef NS_OPTIONS(NSInteger, BTPaymentMethodCreationOptions) {
 - (BOOL)canCreatePaymentMethodWithProviderType:(BTPaymentProviderType)type;
 
 
-/// Status of the last (or ongoing) transaction 
+/// Status of the last (or ongoing) payment method creation 
 @property (nonatomic, assign) BTPaymentProviderStatus status;
 
 #if BT_ENABLE_APPLE_PAY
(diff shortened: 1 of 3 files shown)


Braintree/Payments/@Public/BTPaymentProvider.h
```diff
@@ -18,6 +18,29 @@ typedef NS_ENUM(NSInteger, BTPaymentProviderType) {
     BTPaymentProviderTypeApplePay
 };
 
+/// Status of the last (or ongoing) transaction
```
mickeyreiss: > transaction

We should probably favor the term "payment method creation" to avoid confusion with Braintree's typical notion of [_transaction_ creation](https://developers.braintreepayments.com/ios+ruby/sdk/server/transaction-processing/create#minimal-example).

mickeyreiss:
@Reflejo thanks. This code looks great and is inline with our general vision for the SDK.

Is this ready for merge as far as you are concerned?

One thing that's missing before we can merge is test coverage, probably in `BTPaymentsSpec`. Let me know if you'd like to to do that, or if you'd prefer for me to pick up that aspect.

Reflejo:
@mickeyreiss I can add the tests (maybe tonight), do you want new cases or hook up existing methods with stuff like:

```objc
expect([provider status]).to.equal(BTPaymentProviderStatusInitialized);
```

mickeyreiss:
I think either one would work. If I were doing it, I'd probably add new cases.

Reflejo:
Just added some test cases. @mickeyreiss  Could you check it it's what you had in mind?

Reflejo:
@mickeyreiss Have you check the analytics stuff? I can rollback my `[self informDelegateDidFailWithError:error];` changes and refactor this a bit if firing analytics for those errors is a problem.

mickeyreiss:
Those tests look good. Thanks for taking the time to do that.
***
Yes, I looked into it, and we should avoid sending the analytics in those cases. Maybe we could add a boolean to `informDelegateDidFailWithError` so that the rest of the logic can stay unified.

Reflejo:
@mickeyreiss Just added the flag.

mickeyreiss:
:+1:

Braintree/Payments/@Public/BTPaymentProvider.h
```diff
@@ -18,6 +18,29 @@ typedef NS_ENUM(NSInteger, BTPaymentProviderType) {
...
+    BTPaymentProviderStatusProcessing,
+
+    /// Payment method failed with error
+    BTPaymentProviderStatusError,
+
+    /// Payment method creation is complete with success
+    BTPaymentProviderStatusSuccess,
+
+    /// The payment method creator has cancelled
+    BTPaymentProviderStatusCancelled
```
burnto: Super minor, but we apparently favor "Canceled" with one 'l' in our codebase. e.g.

https://github.com/braintree/braintree_ios/blob/a1522fecf8bd14f50919594cb7b8d20e8a827d9f/Braintree/Venmo/BTVenmoAppSwitchReturnURL.h#L9

Apparently this is an American vs British thing.
http://www.quickanddirtytips.com/education/grammar/canceled-or-cancelled

Reflejo: Fixed

Braintree/Payments/@Public/BTPaymentProvider.h
```diff
@@ -18,6 +18,29 @@ typedef NS_ENUM(NSInteger, BTPaymentProviderType) {
...
+
+    /// The payment method creator, having obtained user payment details and/or user
+    /// authorization, is now processing the results
+    BTPaymentProviderStatusProcessing,
+
+    /// Payment method failed with error
+    BTPaymentProviderStatusError,
+
+    /// Payment method creation is complete with success
+    BTPaymentProviderStatusSuccess,
```
Reflejo: Maybe to be consistent with [this other enum](https://github.com/braintree/braintree_ios/blob/a1522fecf8bd14f50919594cb7b8d20e8a827d9f/Braintree/Venmo/BTVenmoAppSwitchReturnURL.h#L9), this should be `BTPaymentProviderStatusSucceeded` and `BTPaymentProviderStatusFailed`?. (also rename every *Status* to *State*?):

BTPaymentProviderStateUninitialized
BTPaymentProviderStateInitialized
BTPaymentProviderStateFailed
BTPaymentProviderStateSucceeded
BTPaymentProviderStateCanceled

.cc @burnto

mickeyreiss: `BTVenmoAppSwitchReturnURL` models [`x-callback-url`](http://x-callback-url.com/), so it is not necessarily parallel to payment method creation.

Regardless, I'm +1 for _Status_ -> _State_ and +1 for _Error_ -> _Failed_.

`¯\_(ツ)_/¯` What do you think, @burnto?

burnto:
@Reflejo this great. Thanks for your contribution.

Issue #60, opened 2014-10-28
Apple Pay entitlement is required even when not used


sergey-sportsetter:
Using Braintree 3.4.0, v.zero drop-in UI

Upon submission to Apple, got the message:
>App contains references to [Apple Pay] in 'SportSetter', but is missing the required entitlement '.com.apple.developer.in-app-payments'

Previously, this entitlement was not necessary

indisee:
have same issue with iTC — upload build and get mail from apple. Do not use any of ApplePay set up or usage code, xcode's capabilities and developer.apple.com do not have ApplePay enabled. 
Is there some workaround to prevent this warning?
Braintree (3.4.2)

jemmons:
+1

xinghengwang:
Is this issue getting fixed soon? 
What is the work around?

burnto:
Yes we are working on a fix and will update here shortly.

burnto:
Hi all, apologies for the delay. We have a solution for this that we're doing some final verification on. Expect a commit and an update here soon.

mickeyreiss:
@burnto's fix is now on master as of https://github.com/braintree/braintree_ios/commit/e142286234522a05f183b563bbad0c60c2537f88.

burnto:
Starting in e142286 (which will be part of a tagged release in the next few days), Apple Pay is only part of the Braintree iOS build when the optional `Braintree/Apple-Pay` subspec is specified in your `Podfile`. 

The basic tests are [pretty](https://github.com/braintree/braintree_ios/blob/e142286234522a05f183b563bbad0c60c2537f88/Rakefile#L149-L151) [simple](https://github.com/braintree/braintree_ios/blob/e142286234522a05f183b563bbad0c60c2537f88/Rakefile#L160-L162), if a little coarse, and just use `nm` to check for symbols that look suspiciously like Apple's. Of course the real test is whether your app submission results in a warning, so once you get a chance to update, please let us know if you encounter further issues.

Thanks all!

lupugabriel:
Hi everybody... i added to my podfile:

pod "Braintree"
pod "Braintree/Enable-Apple-Pay"

and i have this error: 

[!] Unable to satisfy the following requirements:
- `Braintree/Enable-Apple-Pay` required by `Podfile`

Can you help?

burnto:
[...]

PR #88, opened 2015-02-11
Added missing `BT_ENABLE_APPLE_PAY` check

Reflejo:
Without this, when `BT_ENABLE_APPLE_PAY` is undefined the compiler complains about `BTClientTokenApplePayStatusValueTransformer` not being defined.

### DIFF
--- a/Braintree/API/@Public/BTClientToken.h
+++ b/Braintree/API/@Public/BTClientToken.h
@@ -96,6 +96,8 @@ extern NSString *const BTClientTokenPayPalNonLiveDefaultValueMerchantUserAgreeme
 
 - (NSString *)btVenmo_status;
 
+#if BT_ENABLE_APPLE_PAY
+
 #pragma mark Apple Pay
 
 @property (nonatomic, readonly, strong) NSDictionary *applePayConfiguration DEPRECATED_MSG_ATTRIBUTE("");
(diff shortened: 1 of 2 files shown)


burnto:
Thanks @Reflejo - this is great. I'd like to reduce the scope of the check to within the implementation. See #89. What do you think?

Reflejo:
My motto is "fail early, fail hard". But I understand why you'd want the implementation to still work even when `BT_ENABLE_APPLE_PAY` is not defined and make it behave more like as a "kill switch" for apple pay. The only problem with the "work as normal even when apple pay is not enabled by the compilation flag" approach is that there is no warning or error if one mistakenly forgets to set `-D BT_ENABLE_PAYPAL=1`.

I, for once, don't like the flag checks everywhere because makes the code more complex and (as it depends on a global flag) hard to follow. Also: there is no BT_ENABLE_PAYPAL or BT_ENABLE_VENMO which makes it less consistent but I'm sure you guys have a good reason for adding this on apple pay.

Reflejo:
Having said that, I'm :+1: with the other approach and you can close this PR

burnto:
Thanks @Reflejo. We don't like the `BT_ENABLE_APPLE_PAY` preprocessor macro checks everywhere either. We added this macro and this build configuration option to address #60. This approach was the only option given the way we integrated and exposed Apple Pay in v.zero. We hope to better decouple the various payment methods in the next major version bump.

burnto:
Going with #89 

Thanks @Reflejo! :star2: :sunglasses: :star2:

PR #94, opened 2015-02-18
[Bugfix] SSL Pinning should pass through unexpected authentication methods

mickeyreiss:
#### Background

Today, there was a configuration change in sandbox that caused our HTTP client to break. In particular, the sandbox began to request [SSL client certificates](https://docs.oracle.com/cd/E19424-01/820-4811/aakhe/index.html). Our [status page](https://status.braintreepayments.com/incidents/ypbc8csb6hpk) has more details about the incident.

Unfortunately, our HTTP client, `BTHTTP`, uses [SSL pinning code](https://github.com/braintree/braintree_ios/blob/3.5.0/Braintree/API/Networking/BTHTTP.m#L378-L389) that did not handle a challenge with this type of authentication method. Specifically, we never called the callback in 
our implementation of `URLSession:didReceiveChallenge:completionHandler:` when the authentication method type was anything other than `NSURLAuthenticationMethodServerTrust`. (Note the missing `else` [here](https://github.com/braintree/braintree_ios/blob/3.5.0/Braintree/API/Networking/BTHTTP.m#L389).)

#### Resolution

The primary issue has now resolved with deployment to our sandbox API servers, reverting the SSL client certificate request.

In addition, this PR fixes our SSL pinning code to avoid the "infinite spinning" in case of future changes to our server-side SSL implementation.

One thing that is missing here is a test case to reproduce the specific issue, but I do not know how to add client verification to the [test HTTPS server](https://github.com/braintree/braintree_ios/blob/3.5.0/Specs/Braintree-API-Integration-Specs/SSL/https_server.rb).

### DIFF
--- a/Braintree/API/Networking/BTHTTP.m
+++ b/Braintree/API/Networking/BTHTTP.m
@@ -373,8 +373,8 @@
 }
 
 
-- (void)URLSession:(__unused NSURLSession *)session didReceiveChallenge:(NSURLAuthenticationChallenge *)challenge completionHandler:(void (^)(NSURLSessionAuthChallengeDisposition, NSURLCredential *))completionHandler{
-    if ([[[challenge protectionSpace] authenticationMethod] isEqualToString: NSURLAuthenticationMethodServerTrust]) {
+- (void)URLSession:(__unused NSURLSession *)session didReceiveChallenge:(NSURLAuthenticationChallenge *)challenge completionHandler:(void (^)(NSURLSessionAuthChallengeDisposition, NSURLCredential *))completionHandler {
+    if ([[[challenge protectionSpace] authenticationMethod] isEqualToString:NSURLAuthenticationMethodServerTrust]) {
         SecTrustRef serverTrust = [[challenge protectionSpace] serverTrust];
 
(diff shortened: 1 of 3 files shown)


mickeyreiss:
I'd recommend looking at this diff without whitespace at  https://github.com/braintree/braintree_ios/pull/94/files?w=1

Braintree/API/Networking/BTHTTP.m
```diff
@@ -384,8 +382,10 @@ - (void)URLSession:(__unused NSURLSession *)session didReceiveChallenge:(NSURLAu
             NSURLCredential *credential = [NSURLCredential credentialForTrust:serverTrust];
             completionHandler(NSURLSessionAuthChallengeUseCredential, credential);
         } else {
-            completionHandler(NSURLSessionAuthChallengeRejectProtectionSpace, nil);
+            completionHandler(NSURLSessionAuthChallengeRejectProtectionSpace, NULL);
         }
+    } else {
+        completionHandler(NSURLSessionAuthChallengePerformDefaultHandling, NULL);
```
mickeyreiss: This is the key line that was missing.

burnto:
:+1: looks good to me.

mickeyreiss:
Thanks for the :eyes: !

PR #98, opened 2015-02-20
Fix/remove prefix dep

sebbean:
fixed deprecated getBytes: call
added foundation header to remove prefix dependency

### DIFF
--- a/braintree/BTEncryption/BTAES.m
+++ b/braintree/BTEncryption/BTAES.m
@@ -14,13 +14,13 @@
 +(NSData*) encrypt:(NSData *) data withKey:(NSData *) key Iv:(NSData *) iv {
   char keyPtr[kCCKeySizeAES256 + 1];
   bzero( keyPtr, sizeof( keyPtr ) );
-  [key getBytes:keyPtr];
+  [key getBytes:keyPtr length:sizeof(key)];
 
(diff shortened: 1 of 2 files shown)


braintree/BTEncryption/BTAES.m
```diff
@@ -14,13 +14,13 @@ +(NSData*) encrypt:(NSData*) data withKey:(NSData*) key {
 +(NSData*) encrypt:(NSData *) data withKey:(NSData *) key Iv:(NSData *) iv {
   char keyPtr[kCCKeySizeAES256 + 1];
   bzero( keyPtr, sizeof( keyPtr ) );
-  [key getBytes:keyPtr];
+  [key getBytes:keyPtr length:sizeof(key)];
```
mickeyreiss: @sebbean This could be a nit since I haven't worked with this code recently, but shouldn't we use `sizeof(keyPtr)` rather than `sizeof(key)`? Otherwise there is a risk of a buffer overrun when the `key` data is bigger than `keyPtr`.

mickeyreiss:
[Merged manually](https://github.com/braintree/braintree_ios/compare/2.3.0...2.3.1) and released in [2.3.1](https://github.com/braintree/braintree_ios/releases/2.3.1).

Issue #101, opened 2015-02-26
Word wrapping for summaryTitle and summaryDescription


nirinchev:
Currently BTDropInViewController's summaryTitle and summaryDescription don't get wrapped if the text overflows and the title overlays the price, while the description is cut off at the edge of the screen. I believe that this should either be handled gracefully by wrapping to the next line or should be better documented.

mickeyreiss:
This sounds like a valid edge case that we are not handling. Feel free to submit a pull request if you need this urgently. The code for these labels is in [BTUISummaryView](https://github.com/braintree/braintree_ios/blob/d89d633ee7033d73d487dbcf202a7cf3bc617e4d/Braintree/UI/Views/Payments%20Components/BTUISummaryView.m#L29-L49).

nirinchev:
Thanks, I already saw where modifications should be made, but I'm reluctant to submit a PR, since I'm not Obj C developer (I use the library via Xamarin bindings) and I'm afraid I'm not familiar with the testing frameworks native developers use. Still if no one picks it up soon, I'll try to do it myself :)

burnto:
Thanks @nirinchev. We'll update here if/when someone works on this, but until then consider if fair game.

mickeyreiss:
@nirinchev If you'd feel more comfortable sending us a patch privately, that would be fine too!

PR #107, opened 2015-03-14
Fix wrapping of texts in SummaryView

nirinchev:
This should fix https://github.com/braintree/braintree_ios/issues/101.

Please note that, since I'm not an Obj C developer, I haven't written any tests to validate correct behaviour. If anyone is more comfortable with the testing frameworks braintree uses, feel free to pick it up and add tests.

Here are two screenshots demonstrating the issue before and after the fix:
**Before fix**
alt text
**After fix**
alt text


mickeyreiss:
Hi @nirinchev! This looks great. Thanks for the code :octocat:

Issue #120, opened 2015-05-07
Maestro card number limit


marekswiecznik:
In BTUICardFormView, I can't enter 6304000000000000 (listed here https://developers.braintreepayments.com/ios+ruby/reference/general/testing), form accepts only 12 digits for MasterCard. From what I have found online, Maestro card number can be 12 or 19 (international debit cards) long.

burnto:
Thanks @marekswiecznik - we'll fix this.

burnto:
Thanks again @marekswiecznik. I've addressed this in #122. Any feedback?

marekswiecznik:
Thank you! Works great. Should I close this issue now or wait till next release?

burnto:
Great to hear, @marekswiecznik. 

Sorry I didn't respond before dropping off the grid for a week. :palm_tree:

PR #122, opened 2015-05-08
Fix Maestro card number limit

burnto:
Don't auto-advance if card type might allow for longer card numbers.

This addresses #120


intelliot:
LGTM

burnto:
Thanks @intelliot

PR #131, opened 2015-06-10
Fix setupViews being called twice in BTPaymentButton

simonyangme:
Also gets rid of "negative or zero sizes are not supported in the flow layout" log message when BTPaymentButton is shown.


mickeyreiss:
It's that simple?!

Thanks for this. We'll merge this after some internal testing and include the change in an upcoming release.

burnto:
Thanks @simonyangme!

PR #137, opened 2015-06-19
Change `@import` to `#import` (Close #124)

### LINKED ISSUE #124: Replace @import with #import to support projects where modules are disabled
Hi,

We are working on a project where modules are disabled (because it contains objc++ files). This forbids @import usage (and therefore generates error with Intercom). Could you replace these imports with regular #import. The compiler will automatically convert regular imports into module imports for Apple frameworks.
```objective-c
@import Foundation;
@import UIKit;
```
becomes

```objective-c
#import <Foundation/Foundation.h>
#import <UIKit/UIKit.h>
```
Thanks.

intelliot:
Coinbase-Integration-Specs:
- Enable ARC
- Set Base SDK to Latest iOS (iOS 8.3) [Oddly, default was OS X]


mickeyreiss:
> Oddly, default was OS X 

Odd indeed...

Should we add a sanity check to the `Rakefile` to prevent ourselves from committing `@import`s in the future?

PR #147, opened 2015-07-23
Replace deprecated APIs

intelliot:
Code review:
- [ ] @burnto 
- [x] @skunkworks 

```
CFURLCreateStringByAddingPercentEscapes() ->
  -stringByAddingPercentEncodingWithAllowedCharacters:

-stringByReplacingPercentEscapesUsingEncoding: ->
  -stringByRemovingPercentEncoding

UIAlertView -> UIAlertController
```

### DIFF
--- a/Braintree/API/Networking/BTURLUtils.m
+++ b/Braintree/API/Networking/BTURLUtils.m
@@ -48,7 +48,13 @@
 }
 
 + (NSString *)stringByURLEncodingAllCharactersInString:(NSString *)aString {
-    return [aString stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLQueryAllowedCharacterSet]];
+    // See Section 2.2. http://www.ietf.org/rfc/rfc2396.txt
+    NSString *reservedCharacters = @";/?:@&=+$,";
+
+    NSMutableCharacterSet *URLQueryPartAllowedCharacterSet = [[NSCharacterSet URLQueryAllowedCharacterSet] mutableCopy];
+    [URLQueryPartAllowedCharacterSet removeCharactersInString:reservedCharacters];
+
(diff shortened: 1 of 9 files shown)


skunkworks:
We made a subset of these changes already in feature/remodel, so thanks for doing this. Only feedback I have is to check in the drop-in code for >= iOS 8 before using UIAlertController (i.e. `if ([UIAlertController class]) {...}`). We'll need to fall back to using action sheet or alert view if it's not available and just ignore the deprecation warnings.

Braintree/API/Networking/BTURLUtils.m
```diff
@@ -48,12 +48,7 @@ + (NSString *)queryStringWithDictionary:(NSDictionary *)dict {
 }
 
 + (NSString *)stringByURLEncodingAllCharactersInString:(NSString *)aString {
-    NSString *encodedString = (__bridge_transfer NSString * ) CFURLCreateStringByAddingPercentEscapes(NULL,
-                                                                                                      (__bridge CFStringRef)aString,
-                                                                                                      NULL,
-                                                                                                      (CFStringRef)@"&()<>@,;:\\\"/[]?=+$|^~`{}",
-                                                                                                      kCFStringEncodingUTF8);
-    return encodedString;
+    return [aString stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLQueryAllowedCharacterSet]];
```
burnto: Since this method is used to encode keys and values, we need to also be sure to also encode the reserved characters (`;/?:@&=+$,`).

burnto: See also https://www.ietf.org/rfc/rfc2396.txt section 2.2

skunkworks:
:+1:

Issue #182, opened 2015-10-14
No way to handle switching back to the app from Safari/PayPal app via the iOS task manager or dismissing of SFSafariViewController via Done button


corristo:
I think there should be a way to be notified about these things, to be able to notify backend about payment cancellation, for example.

I'm using 4.0 beta branch.

corristo:
Well, I've found the following in doc comments:

```
/// @note During the app switch authorization, the user may switch back to your app manually. In this case, the caller
/// will not receive a cancellation via the completionBlock. Rather, it is the caller's responsibility to observe
/// `UIApplicationDidBecomeActiveNotification` and `UIApplicationWillResignActiveNotification` using `NSNotificationCenter`
/// if necessary.
```

This explains app switching, but SFSafariViewController is still an issue.

skunkworks:
Hi @corristo, I've merged in a pull request that should take care of this. Thanks for reporting the issue!

corristo:
@skunkworks :+1:

PR #185, opened 2015-10-19
Improved support for PayPal flow cancellation

skunkworks:
Addresses this issue: https://github.com/braintree/braintree_ios/issues/182

Calls the tokenization callback for cancellation when:
* `SFSafariViewController` "Done" button is tapped
* When returning to the app via iOS Task Manager or using the Home button to navigate back to the app

### DIFF
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -156,18 +156,30 @@
 		4163BF7E1BCEDE6A0047CBA3 /* BTVenmoAppSwitchReturnURLSpec.m in Sources */ = {isa = PBXBuildFile; fileRef = A71F7DDF1B616BA0005DA1B0 /* BTVenmoAppSwitchReturnURLSpec.m */; };
 		4163BF7F1BCEDE6A0047CBA3 /* BTVenmoAppSwitchURLSpec.m in Sources */ = {isa = PBXBuildFile; fileRef = A71F7DE01B616BA0005DA1B0 /* BTVenmoAppSwitchURLSpec.m */; };
 		4163BF801BCEDE6A0047CBA3 /* BTVenmoDriver_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A71F7DE61B6180A3005DA1B0 /* BTVenmoDriver_Tests.swift */; };
-		4163BF811BCEDE6A0047CBA3 /* BTHTTPTestProtocol.m in Sources */ = {isa = PBXBuildFile; fileRef = A7D4630C1B4B16C100A09C46 /* BTHTTPTestProtocol.m */; };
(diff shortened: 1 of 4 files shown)


skunkworks:
@intelliot @mickeyreiss I'd like to get your thoughts on this approach of using NSNotificationCenter to monitor when the user manually cancels the PayPal authorization flow. I think this will greatly simplify how UI is handled, updated, and reset throughout the app switch lifecycle.

intelliot:
My concern is that returning to the app via iOS Task Manager (or using the Home button to navigate back to the app) does not actually mean the payment was necessarily canceled: If the buyer switches back to the PayPal Wallet app (or Safari) and authorizes the payment, it should still work.

In the SFSafariViewController case, I imagine it should be possible for developers to handle delegate events themselves:

```
- (void)paymentDriver:(id)driver requestsPresentationOfViewController:(UIViewController *)viewController {
    if ([viewController isKindOfClass:[SFSafariViewController class]]) {
        [(SFSafariViewController *)viewController setDelegate:self];
    ]
    // ...
}
```

skunkworks:
I'd argue that if someone goes back to the app that the tokenization (a.k.a. payment and app switch) has been cancelled, no matter what the state is of the web app or PayPal app. This is a simple and straightforward way to define it, and it reduces complexity for both us and for merchants.

What should happen the user swaps back to the merchant app, then goes back to the PayPal app and completes the authorization? What should happen if the merchant app crashes, or if the user logs out of the app in between?

The motivation for this is to reduce the number of code paths that merchants have to manage to determine whether a PayPal payment was successfully authorized, failed, or cancelled. As-is, there are three code paths *just for handling the case where the user cancels*:
1. User cancels from within the web app, the tokenization callback block is invoked
2. User taps the Done button in SFSafariViewController, merchant sets up a delegate to receive this message
3. User manually swaps back to the app via Task Manager or Home button, merchant subscribes to `UIApplicationDidBecomeActive` notification and receives this notification

This means that if the UI has to be reset after a payment is cancelled, they'll have to reset their UI in three different places.

intelliot:
> What should happen the user swaps back to the merchant app, then goes back to the PayPal app and completes the authorization?
Ideally, the authorization should work. It should behave the same as if the user didn't do the swap in between.

> What should happen if the merchant app crashes
Ideally, the authorization should work. Apps can be purged from memory at any time.

> or if the user logs out of the app in between?
Assuming "the app" is the merchant app: ideally, the authorization should be ignored because the checkout is no longer valid.

imho, we should guide merchant integrations so that they don't have UI that has to be reset if PayPal authorization is canceled. For PayPal, the only cancelation handling that I expect is for analytics: to enable merchants to know how many times users canceled from within the web app.

skunkworks:
Closed in favor of https://github.com/braintree/braintree_ios/pull/187

PR #187, opened 2015-10-22
4.0 safari cancellation redux

skunkworks:
Addresses this issue: https://github.com/braintree/braintree_ios/issues/182

Calls the tokenization callback for cancellation when:
* `SFSafariViewController` "Done" button is tapped


intelliot:
:green_apple:

PR #199, opened 2015-11-25
Prevent calling requestsDismissalOfViewController

Reflejo:
This fix prevents calling requestsDismissalOfViewController when there is nothing to dismiss. The same way `requestsPresentationOfViewController` is not being called when there is nothing to present.

Before this, the library was crashing on runtime when using swift and iOS < 9 as this is passing `nil` to a function signature that is marked as `non_null`.

### DIFF
--- a/BraintreePayPal/BTPayPalDriver.m
+++ b/BraintreePayPal/BTPayPalDriver.m
@@ -317,7 +317,10 @@ typedef NS_ENUM(NSUInteger, BTPayPalPaymentType) {
 - (void)setAppSwitchReturnBlock:(void (^)(BTPayPalAccountNonce *tokenizedAccount, NSError *error))completionBlock
                  forPaymentType:(BTPayPalPaymentType)paymentType {
     appSwitchReturnBlock = ^(NSURL *url) {
-        [self informDelegatePresentingViewControllerNeedsDismissal];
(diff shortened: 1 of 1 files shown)


burnto:
Cool, thanks for this. 

I wonder if we could reasonably just move this check to the `inform` method. @skunkworks and @intelliot can make the call.

Reflejo:
@burnto That was my first approach. Wasn't sure what to do with

```objc
else {
        [[BTLogger sharedLogger] critical:@"Unable to dismiss View Controller to end PayPal flow. BTPayPalDriver needs a viewControllerPresentingDelegate<BTViewControllerPresentingDelegate> to be set."];
}
```

but yeah, seems safer to do it there

intelliot:
This PR looks good to me. In the `requestsPresentationOfViewController` case, `informDelegatePresentingViewControllerRequestPresent` does not get called when there won't be a view controller to present. This makes `informDelegatePresentingViewControllerNeedsDismissal` work the same way.

Thanks, @Reflejo !

Reflejo:
Any chance to bump the podspec version with this fix? It was pretty bad on non ios9

Reflejo:
Ah just saw it, never mind. Thanks!

Issue #202, opened 2015-12-16
The link for "manual integration instructions" in Readme links to GHE


prazgaitis:
The link should probably go here instead: 
https://github.com/braintree/braintree_ios/blob/3.x/Docs/Manual%20Integration.md

burnto:
Good catch. Thanks @prazgaitis

PR #203, opened 2015-12-16
Fix manual integration instructions link in README

burnto:
This addresses #202 and clarifies that the manual integrations doc applies to v3, and that v4 instructions are coming soon.

@skunkworks @intelliot @demerino :eyes: ?


skunkworks:
thanks for fixing my bug :shaved_ice:

PR #212, opened 2016-02-04
Add Korean UI strings

sydneyitguy:


intelliot:
Although I don't know Korean, I did a sanity check (Google Translate) and it looks :+1: from that perspective :)

demerino:
We're in the process of updating our i18n strings. We appreciate the PR and we'll ensure that Korean is included in the near future.

PR #217, opened 2016-02-09
Report exceptions caused by Braintree iOS SDK

skunkworks:
**Problem**: The Braintree iOS SDK does not have any visibility into app crashes caused by our SDK, which offers a poor experience for users as well as frustration and lost revenue for merchants.

**Proposal**: Offer *opt-in* exception reporting in our iOS SDK such that when your app crashes, it will send a crash event to Braintree.

**How it works**: At the end of your app delegate `application:didFinishLaunchingWithOptions:`, call `[BTReporting enable]`.

One limitation with iOS apps is that only one crash handler can be attached to an app. Our solution is to wrap any existing crash handler that your app has integrated (e.g. Crashlytics, HockeyApp).

When an exception occurs, the Braintree crash handler searches the stack trace to see if the crash involves Braintree-related code. If so, it sets a flag that the app crashed and sends a crash notification the next time that your app initializes the Braintree SDK.

It will then forward the crash to the wrapped crash handler, if one exists.

**Limitations**: The Braintree iOS SDK will only report when Objective-C exceptions occur. Other errors such as signal aborts due to memory access violations will not be reported.

The reason we've chosen not to write a signal handler for now is because most crashes caused by the SDK have been Objective-C exceptions, and we think this is sufficient for a first cut of this feature. In addition, signal handlers are [fraught with peril](https://www.mikeash.com/pyblog/friday-qa-2011-04-01-signal-handling.html).


skunkworks:
We have decided for now to defer this feature. Sorry.

PR #227, opened 2016-02-21
Add missing _Nullable specification

sberrevoets:
This missing specifier is preventing the project to build in Xcode 7.3 (beta 3). I realize 7.3 is not officially released and things could change between now and the final release, but it's probably good to add this regardless of Xcode version.

### DIFF
--- a/BraintreePayPal/PayPalOneTouch/Analytics/PPFPTITracker.h
+++ b/BraintreePayPal/PayPalOneTouch/Analytics/PPFPTITracker.h
@@ -33,7 +33,7 @@
 
 
 /// The delegate which actually sends the data
-@property (nonatomic, weak, readwrite) id<PPFPTINetworkAdapterDelegate> networkAdapterDelegate;
+@property (nonatomic, weak, readwrite, nullable) id<PPFPTINetworkAdapterDelegate> networkAdapterDelegate;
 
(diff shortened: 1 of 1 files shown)


BraintreePayPal/PayPalOneTouch/Analytics/PPFPTITracker.h
```diff
@@ -33,7 +33,7 @@
 
 
 /// The delegate which actually sends the data
-@property (nonatomic, weak, readwrite) id<PPFPTINetworkAdapterDelegate> networkAdapterDelegate;
+@property (nonatomic, weak, readwrite) _Nullable id<PPFPTINetworkAdapterDelegate> networkAdapterDelegate;
```
skunkworks: Thanks for the fix! Do you mind if we move the nullability annotation into the parentheses? It's just a style suggestion.

```
+@property (nonatomic, weak, readwrite, nullable) id<PPFPTINetworkAdapterDelegate> networkAdapterDelegate;
```

sberrevoets: Yes, fixed. I don't work with Objective-C that much anymore, so I just picked what Xcode suggested. Thanks!

skunkworks:
Thanks again for opening up this PR

PR #238, opened 2016-04-28
Fix the flow issue when choosing paypal from inner adding payment met…

flexme:
Currently if I present a `BTDropInViewController` from my own view controller, then tap "change payment method", tap "+" on the navigation bar to add a payment method, then select Paypal. After Paypal flow is done, my own view controller will be notified directly through `dropInViewController:didSucceedWithTokenization` delegate method, which is wrong (At least It's very different from other flows).

I think the correct flow here is to return to the "Payment Method Selection" view, where user can see the new added Paypal method, and can tap 'Pay' button there.

### DIFF
--- a/BraintreeUI/Drop-In/BTDropInViewController.m
+++ b/BraintreeUI/Drop-In/BTDropInViewController.m
@@ -685,10 +685,15 @@
         
         [errorAlert show];
     } else if (paymentMethodNonce) {
-        NSMutableArray *newPaymentMethods = [NSMutableArray arrayWithArray:self.paymentMethodNonces];
-        [newPaymentMethods insertObject:paymentMethodNonce atIndex:0];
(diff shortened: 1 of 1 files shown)


skunkworks:
@flexme Thanks for opening this PR. I'm having some issues duplicating the behavior. Which version of the SDK are you using?

Here's the behavior I see: the first time you add a payment method, it will call `dropInViewController:didSucceedWithTokenization`. When you add subsequent payment methods (*Change payment method*), it will return back to the the Drop-in view controller, at which point you can tap the "Pay" button.

demerino:
Take a look at `4.4.0`+. We've added a new Drop-In component and exposed our UI components (along with styling). More info [here](https://github.com/braintree/braintree_ios/blob/master/Docs/Drop-In-Update.md).

PR #239, opened 2016-04-29
Update theme on child components if set/changed

theill:
If you setup a theme on `BTUICardFormView` it's not correctly propagated to children for fields and float labels. This patch does that.


skunkworks:
👍 Thanks for opening this PR!

PR #241, opened 2016-05-24
Include missing sources

AlexDenisov:
Currently static library (libBraintree.a) has missing two symbols:

 - `_OBJC_CLASS_$_BTUICardPhoneNumberField`
 - `_OBJC_CLASS_$_BTUIUnionPayVectorArtView`

The patch fixes the problem by adding corresponding classes into the
target.


skunkworks:
Thanks for the fix!

PR #258, opened 2016-09-02
 Updates to 'paymentMethodNonces' declaration.

BenMaer:
* `paymentMethodNonces` property is now declared using generics to indicate that the array accepts instances of only `BTPaymentMethodNonce` and its subclasses.

 * `paymentMethodNonces` property is declared with `copy` instead of `strong`, and the setter method updated to make a copy of the passed in array, so we don't accidentally end up with a mutable array, which can't be managed.


BenMaer:
Apologies, I accidentally merged some other changes into this branch. Hopefully you can consider this small change again :)

demerino:
Thanks for the PR @BenMaer 

We're actually retiring our old `BraintreeUI` Drop-In in favor of our new Drop-In: https://github.com/braintree/braintree-ios-drop-in

So we're only doing necessary maintenance on it at this time. But we would love for you to checkout the new Drop-In and let us know what you think. Feel free to create issues on either repo with feedback and suggestions.

PR #281, opened 2016-11-10
Do not call viewControllerPresentingDelegate for dismissal if breakout via Done button

danielbowden:
On iOS 10 when the PayPal flow is presented in a SFSafariViewController a user can break out by pressing the builtin 'Done' button. (as addressed in a2af94cf5d01b06e0cf81834ad3cb3f543ac984c back in Oct 2015 and then again here 077ee24aafacfc6d019141eb823cbb39bdf1c330 and here c7ee6748a1ce4c2b528dcb1cd52642b36602763d).

Currently if the PayPal view is presented on an already presented view controller, the act of hitting the 'Done' button will dismiss PayPal AND dismiss the first party view controller that was underneath it.

All this PR does is move the existing check for the `SFSafariViewControllerFinishedURL` to before the `informDelegatePresentingViewControllerNeedsDismissal` method. This prevents the delegate from being called if the `SFSafariViewController` is being cancelled via the 'Done' button and therefore prevents the double dismissal.

The existing behaviour for dismissal when either tapping the X in the top right hand corner of the PayPal UI or successfully authenticating still both occur as expected.

Thanks

### DIFF
--- a/BraintreePayPal/BTPayPalDriver.m
+++ b/BraintreePayPal/BTPayPalDriver.m
@@ -328,17 +328,17 @@ typedef NS_ENUM(NSUInteger, BTPayPalPaymentType) {
 - (void)setAppSwitchReturnBlock:(void (^)(BTPayPalAccountNonce *tokenizedAccount, NSError *error))completionBlock
                  forPaymentType:(BTPayPalPaymentType)paymentType {
     appSwitchReturnBlock = ^(NSURL *url) {
-        if (self.safariViewController) {
-            [self informDelegatePresentingViewControllerNeedsDismissal];
-        } else {
-            [self informDelegateWillProcessAppSwitchReturn];
-        }
-        
         // Before parsing the return URL, check whether the user cancelled by breaking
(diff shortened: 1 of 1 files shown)


danielbowden:
CI build fail looks unrelated. Can you kick it again?

skunkworks:
Thanks for opening this PR! I've restarted the Travis build. We'll take a closer look at the PR.

danielbowden:
Thanks @skunkworks

demerino:
What does your `paymentDriver:requestsDismissalOfViewController` look like? It sounds like you might want to use something like this:

```
- (void)paymentDriver:(__unused id)driver requestsDismissalOfViewController:(UIViewController *)viewController {
    [viewController dismissViewControllerAnimated:YES completion:nil];
}
```

Which is a bit different than our docs. If that works, let us know so we can make the appropriate changes to our docs.

danielbowden:
Thanks @demerin. 

My `paymentDriver:requestsDismissalOfViewController` delegate method was having the host view controller call `dismissViewControllerAnimated:completion:`

Changing it to use the `viewController` parameter from the delegate method instead looks like it now works as expected.

Issue #285, opened 2016-11-16
[Warning] Incorrect @param causing compiler warning in BTAPIClient.h


adomanico:
Seeing a compiler warning regarding 'completionBlock' rather then 'completion in the param comment

/<directory>/Carthage/Build/iOS/BraintreeCore.framework/Headers/BTAPIClient.h:67:12: Parameter 'completionBlock' not found in the function declaration

```swift
/// @param completionBlock Callback that returns an array of payment method nonces.
- (void)fetchPaymentMethodNonces:(BOOL)defaultFirst
                      completion:(void(^)(NSArray <BTPaymentMethodNonce *> * _Nullable paymentMethodNonces, NSError * _Nullable error))completion;
```

Both fetchPaymentMethodNonces functions have this issue

adomanico:
Deployment target is 10.0.

Not seeing this warning on a 9.0 deployment target

demerino:
Thanks @adomanico 

We'll review the issue and PR.

PR #286, opened 2016-11-16
[Fix] @param warning for incorrect argument name

adomanico:
Fixes https://github.com/braintree/braintree_ios/issues/285


demerino:
👍 
Thanks for pointing this out. Looks like our Carthage tests didn't catch this (targeting 9+).

Issue #290, opened 2016-11-17
Core not compatible with app extensions


bsmith11:
Currently, the core framework is not compatible with app extensions because it uses APIs that are unavailable to app extensions. Namely, `[UIApplication sharedApplication]`.

demerino:
Thanks for pointing out this use case.  We'll review your PR.

skunkworks:
We've merged this into a separate branch for now and will update you when this makes it into `master`. Thanks!

demerino:
`4.7.3` has been released and should include the changes you need. Let us know if you have any issues.

PR #291, opened 2016-11-17
[fix] Add app extension compatibility

bsmith11:
#290 

This is the implementation that was used in PINCache: https://github.com/pinterest/PINCache/pull/72

It has since been removed, but as far as I can tell, only because the feature requiring `[UIApplication sharedApplication]` was removed as well:
https://github.com/pinterest/PINCache/pull/99

### DIFF
--- a/BraintreeCore/BTAnalyticsMetadata.m
+++ b/BraintreeCore/BTAnalyticsMetadata.m
@@ -175,8 +175,13 @@
     if ([UIApplication class] == nil) {
         return nil;
     }
-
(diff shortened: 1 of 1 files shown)


demerino:
Thanks for putting this PR together - we'll take a closer look in the next few days. 

Because these changes return `false` when `isAppExtension` - we'll also want to add something to the analytics metadata to mark these events as being from an `app extensions` so they don't pollute our data.

skunkworks:
Thanks for opening this PR @bsmith11 . Your changes seem reasonable, but I think we have a little bit of work on our end to make analytics from app extensions more coherent. Our current plan is to merge this to a separate branch and work off of that before we merge to master.

PR #292, opened 2016-11-17
Add support for App Extensions

skunkworks:
See also https://github.com/braintree/braintree_ios/pull/291

### Summary

When running in an iOS app extension, we don't have access to `[UIApplication sharedApplication]`, which is used in BraintreeCore's BTAnalyticsMetadata.

We've also hijacked the `deviceScreenOrientation` analytics property to return `AppExtension` when running inside an app extension, which may be useful later.


demerino:
👍

bluk:
:shipit:

PR #299, opened 2016-12-25
Declare as static variables

chuganzy:
As titled.

### DIFF
--- a/BraintreeUI/Views/Custom Views/BTUIFormField.m
+++ b/BraintreeUI/Views/Custom Views/BTUIFormField.m
@@ -5,11 +5,11 @@
 
 @import QuartzCore;
 
-const CGFloat formFieldTopMargin = 7;
-const CGFloat formFieldLabelHeight = 15;
-const CGFloat formFieldVerticalSpace = 1;
-const CGFloat formFieldTextFieldHeight = 20;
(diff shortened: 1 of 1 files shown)


demerino:
Thanks for submitting the PR! Can to add any background information on why this change is required?

I also wanted to mention that additional work on `BraintreeUI` (beyond bug fixes) has been suspended in favor of moving towards our `new Drop-In`: https://github.com/braintree/braintree-ios-drop-in

chuganzy:
> Can to add any background information on why this change is required?

These variables are not exported in header file but linkable which means we cannot define variables with same name. For example, if we define `formFieldTopMargin` with same way, it causes a compile error. We shouldn't make these variables likable so added `static` keyword.
If we want to export them, we should rename them to more unique ones.

> I also wanted to mention that additional work on BraintreeUI (beyond bug fixes) has been suspended in favor of moving towards our new Drop-In: https://github.com/braintree/braintree-ios-drop-in

Oh, I didn't know that!

demerino:
Got it. Thanks for the info.

It is unofficially deprecated (we haven't explicitly deprecated it in the code yet). We've updated our official docs to use the New Drop-In: https://developers.braintreepayments.com/guides/drop-in/ios/v4

skunkworks:
👍

PR #303, opened 2017-01-01
Make CI passable

chuganzy:
As titled.

It seems that `pod install` fails if BraintreeDropIn is already installed under ./Pods directory and its latest commit hash differs from the hash written in .lock file. So as a workaround, delete ./Pods directory and retry installation when `pod install` failed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
 
 install:
   - bundle install
-  - bundle exec pod repo update
+  - bundle exec pod repo update --silent
   - bundle exec pod install
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
 install:
   - bundle install
   - bundle exec pod repo update --silent
-  - bundle exec pod install
+  - bundle exec pod install || { rm -rf ./Pods; bundle exec pod install; }
 
 script:
   - echo $SIMULATOR_ID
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -11,7 +11,7 @@ before_install:
 
 install:
   - bundle install
-  - bundle exec pod repo update
+  - bundle exec pod repo update --silent
```
chuganzy: I think the log is too long and wasteful so added `--silent` flag, but if you consider it's required, I'll revert https://github.com/braintree/braintree_ios/pull/303/commits/8ec8cab9feb9b441a7c4b336a92309cb8a8c55db commit.

.travis.yml
```diff
@@ -11,8 +11,8 @@ before_install:
 
 install:
   - bundle install
-  - bundle exec pod repo update
-  - bundle exec pod install
+  - bundle exec pod repo update --silent
+  - bundle exec pod install || { rm -rf ./Pods; bundle exec pod install; }
```
chuganzy: `pod install` seems to fail if ./Pods is newer under Git on this project.

.travis.yml
```diff
@@ -11,8 +11,8 @@ before_install:
 
 install:
   - bundle install
-  - bundle exec pod repo update
-  - bundle exec pod install
+  - bundle exec pod repo update --silent
```
chuganzy: The log is too long and wasteful so added this option, but if you consider it is necessary, I will revert https://github.com/braintree/braintree_ios/pull/303/commits/8ec8cab9feb9b441a7c4b336a92309cb8a8c55db.

skunkworks:
Thanks for the help! :1st_place_medal:

PR #313, opened 2017-02-14
Fixing compilation under Xcode 8.3 beta.

charlieMonroe:
Xcode 8.3 beta has new errors when importing files under with the wrong case. In this case, there was "Meta*D*ata", while the file is named "Meta*d*ata".


charlieMonroe:
Sorry, missed the issue and the fact this is fixed in master.

PR #323, opened 2017-03-28
Resolve build errors in XCode 8.3  caused by the use incorrect case o…

alamantia:
Resolve build errors in XCode 8.3  caused by the use incorrect case of include files

### DIFF
--- a/Braintree/API/Utility/BTAnalyticsMetadata.m
+++ b/Braintree/API/Utility/BTAnalyticsMetadata.m
@@ -1,4 +1,4 @@
-#import "BTAnalyticsMetaData.h"
+#import "BTAnalyticsMetadata.h"
 #import "BTClient.h"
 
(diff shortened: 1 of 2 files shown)


demerino:
@alamantia Thanks for taking the time to submit this PR. We will take a look and see if there are any other similar issues. We've had a few of these come up recently.

vitorll:
I was about to open a PR with that 👍

demerino:
Thanks for your patience on this. We're actually updating `master` today and we'll update `braintree:3.x` afterwards. (I just noticed that this PR wasn't for master).

demerino:
`3.9.9` is now available with this fix.

Thanks for your contribution on this: https://github.com/braintree/braintree_ios/blob/3.x/CHANGELOG.md#399-2017-04-03

PR #328, opened 2017-04-07
Fix documentation issue

intelliot:
'@return' command used in a comment that is attached to a method returning void


intelliot:
This will be addressed in the next release

PR #337, opened 2017-07-18
Set port correctly in base url

peterstuart:
`components.port` is being set to an `NSString` when it expects an `NSNumber`. For some reason this works on iOS 10, but it causes a crash on iOS 11 (building with Xcode 9 beta 3).

### DIFF
--- a/BraintreeCore/BTAPIClient.m
+++ b/BraintreeCore/BTAPIClient.m
@@ -130,10 +130,11 @@ NSString *const BTAPIClientErrorDomain = @"com.braintreepayments.BTAPIClientErro
     NSURLComponents *components = [[NSURLComponents alloc] init];
     components.scheme = [BTAPIClient schemeForEnvironmentString:environment];
     NSString *host = [BTAPIClient hostForEnvironmentString:environment];
-    NSArray *hostComponents = [host componentsSeparatedByString:@":"];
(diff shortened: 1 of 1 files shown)


demerino:
@peterstuart 
Thanks for the PR. We will verify with the most recent Xcode 9 Beta.

MikeSilvis:
Thanks @peterstuart for fixing this.

@demerino We were currently running an older version of Braintree, and this bit us in production when we released today (using Xcode 8 to archive still) due to the changes in Bitcode it seems. For these kind of crashes, is this something that gets emailed out to existing Braintree customers?

Thanks,
Mike

demerino:
@MikeSilvis Apologies for the production issue. We did send an email shortly after the release of [4.8.5](https://github.com/braintree/braintree_ios/releases/tag/4.8.5) to merchants that would be impacted.

In addition, we also announce _every_ release to our Google Group as noted in the README: https://github.com/braintree/braintree_ios#releases

But if you didn't receive the email, then you should contact support@braintreepayments.com so we can investigate why you were not notified if you were using a Tokenization Key (in production) at the time of the release of `4.8.5` (~2 months ago) so we can fix/adjust our messaging for similar issues in the future.

MikeSilvis:
No problem, thanks! I'll go ahead and reach out to support.

PR #344, opened 2017-08-23
Change cookie policy to accept always

apascual:
Some banks like SEB in Sweden have troubles with the default cookies policy **NSHTTPCookieAcceptPolicyOnlyFromMainDocumentDomain**. This patch fixes those issues.

### DIFF
--- a/Braintree3DSecure/BTThreeDSecureAuthenticationViewController.m
+++ b/Braintree3DSecure/BTThreeDSecureAuthenticationViewController.m
@@ -76,6 +76,25 @@
 #pragma mark UIWebViewDelegate
 
 - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)request navigationType:(UIWebViewNavigationType)navigationType {
+
+    // BANK-ID
+    NSString *customScheme = @"bankid";
+    if([[[request URL] scheme] isEqualToString:customScheme]) {
+        UIApplication *application = [UIApplication sharedApplication];
+        if ([application canOpenURL:[NSURL URLWithString:[NSString stringWithFormat:@"%@://", customScheme]]]) {
+
+            NSString *bankIDUrl = [[request URL] absoluteString];
+            NSRange range = [bankIDUrl rangeOfString:@"redirect="];
+            if(range.length > 0 ) {
+                bankIDUrl = [bankIDUrl substringToIndex:range.location];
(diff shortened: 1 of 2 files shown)


demerino:
@apascual We've been working no a new solution to this problem where we use a _browser switch_ instead of a _web view_. This feature went out and it would be great if you can confirm that attempting to use a SEB card now works. You can use the new 3DS Payment Flow demo to test it: https://github.com/braintree/braintree_ios/blob/master/Demo/Features/3D%20Secure/BraintreeDemoThreeDSecurePaymentFlowViewController.m#L78-L124

You can access the this feature from the Demo app by changing the integration (under settings).

Let us know if you have any questions or comments. We're in the process of updating our documentation and Drop-in integration for the new 3DS Payment Flow.

apascual:
@demerino Thanks for the update, I'll give it a try and give back to you.

Just for a better understanding, please confirm if this is the new 3DSecure flow:
1) User taps pay button on App
2) The app opens the system's default browser with 3DSecure auth link
3) After verification (or failure), the browser redirects back to the app (using URL Scheme? or how?)
4) App receives the result on 3DSecure and proceeds

demerino:
Yup, this is just about right.

- The switch is to `SFSafariViewController` on iOS 9+ (falls back to Safari <iOS 9)
- A URL Scheme is required. The same URL scheme required for other Braintree integrations that use browser/app switching: https://developers.braintreepayments.com/guides/paypal/client-side/ios/v4#setup-for-app-switch

We appreciate you taking the time to check it out.

crookedneighbor:
@apascual Have you been able to test the new release?

demerino:
@apascual 
Yeah, if you look at the `Settings`:
1. Enable `Production` environment
2. Disable `Use Tokenization Key`
3. Scroll down on the settings page to ensure 3DS is always enabled.

crookedneighbor:
@apascual Any update?

demerino:
@apascual We've verified that the new integration works with SEB cards when using *SMS authentication*. There is a known issue where the BankID app switch isn't working correctly from the SEB side of things which they are working on.

Let us know if you have any questions or feedback. The official docs should be updated soon for this new integration.

Closing the PR - we're in the process of deprecating the webview integration.

apascual:
@demerino Yeah, sorry, I tried to test the SEB integration but I don't have access anymore to that card 😔

PR #348, opened 2017-09-07
fixing apple payment method as default

rksaraf:
When parsing Apple Pay nonces from the vault the isDefault was never set.

### DIFF
--- a/BraintreeApplePay/BTApplePayCardNonce.m
+++ b/BraintreeApplePay/BTApplePayCardNonce.m
@@ -2,22 +2,8 @@
 
 @implementation BTApplePayCardNonce
 
-- (instancetype)initWithNonce:(NSString *)nonce
-         localizedDescription:(NSString *)description
-                         type:(NSString *)type
-                    isDefault:(BOOL)isDefault
-                         json:(BTJSON *)json {
-    self = [super initWithNonce:nonce localizedDescription:description type:type isDefault:isDefault];
-    if (self) {
-        _binData = [[BTBinData alloc] initWithJSON:json[@"binData"]];
-    }
-    return self;
-}
-
-#pragma mark - Deprecated methods
-
 - (instancetype)initWithNonce:(NSString *)nonce localizedDescription:(NSString *)description type:(NSString *)type json:(BTJSON *)json {
-    self = [super initWithNonce:nonce localizedDescription:description type:type];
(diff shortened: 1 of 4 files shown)


bluk:
LGTM. The only thing is that we might want to keep the older init method and deprecate it just because it was part of the public API.

rksaraf:
@bluk - Thanks. I just updated it.

rksaraf:
@bluk - sorry. misplaced a ';'. Just fixed that.

rksaraf:
@bluk just checking in on the status of this

bluk:
@rksaraf Sorry for the late response. Discussed this with a couple team members. Currently, there is no support for setting Apple Pay as the default for vaulting in the configuration. How do you intend to use this change?

rksaraf:
@bluk when our app adds an Apple Pay payment method (or any payment method) we set that as the default payment method in the vault via our server. When it initially was not getting updated in the iOS app we noticed that the response from reloading the vault in the iOS app (using the Drop-In UI) was there was a flag that had the payment method for Apple Pay as the default but the SDK was not parsing that, so this change just adds that parsing in.

In terms of use we try and keep the last payment method saved and displayed in our own UI so the user does not have to select a payment method each time they make a purchase. This works fine for all other payment methods so we wanted to keep that consistent for Apple Pay to allow the user to just quickly purchase and authorize via Apple Pay UI rather than having to select Apple Pay again in the Drop-In UI and then authorize in Apple Pay again. If the user wishes to change their payment method they simply need to click on the cell that has the current (default) payment method listed and that launches the Drop-In UI sheet to allow them to select one one their vaulted payment methods or add a new one.

rksaraf:
@bluk any updates?

demerino:
@rksaraf I think this makes sense. Do you mind adding/updating unit tests for this. Perhaps around https://github.com/braintree/braintree_ios/blob/master/UnitTests/BTApplePay_Tests.swift#L231-L279

demerino:
Also, how do you feel about keeping the original `init` method and just updating it:
```
self = [super initWithNonce:nonce localizedDescription:description type:type isDefault:[json[@"default"] isTrue]];
```

rksaraf:
@demerino - I can do that, don't really have a preference - it's your codebase :)

I'll also look at updating the tests. I'll post here when that's done (might not get to it until later this week).

demerino:
Great, we'll take a look and try to get this into the next release. Thanks for taking the time to contribute, we appreciate it!

rksaraf:
<3

PR #379, opened 2018-01-11
Drop-In.strings(Arabic)

abdallahnh:


bluk:
We don't plan to support Arabic in this version of Drop-in. Arabic is supported in the latest version of [Drop-in](https://github.com/braintree/braintree-ios-drop-in). We would love for you to check out the new version, and give us feedback.

bluk:
Thanks for the PR, but will close as discussed above. Please let us know if there's any feedback for the latest version of Drop-in that can help your integration.

PR #388, opened 2018-02-02
Fix Xcode warning

keith:
There's a new warning with Xcode for retaining self that requires you to
reference self in that case.


demerino:
@keith We'll look into this.

Which version of Xcode are you using?

keith:
This is a new warning with Xcode 9.3

keith:
Or sorry, maybe it's just that the default changed. And now it's on by default in CocoaPods HEAD https://github.com/CocoaPods/Xcodeproj/pull/541/files#diff-2fbe57544f33c804869fe7b12f1728eaR356

demerino:
Thanks for the info. We are looking into fixing this in the next release.

demerino:
This should be fixed in `4.11.0`. Thanks for letting us know about the issue.

keith:
Thanks!

PR #395, opened 2018-03-23
 Ensure animations are consistent for PayPal SFSafariViewController flow

nudge:
Our designer noticed that the presentation of `SFSafariViewController` does not perform the correct presentation animation. For example on iOS 10 devices, in the `BTViewControllerPresentingDelegate` implementation, calling `parentViewController.present(viewController)` causes the `SFSafariViewController` to be pushed instead of presented.

This pull request ensures the animation is consistent with the presentation style.


demerino:
👍 Nice catch

demerino:
Included in `4.14.0`.

PR #412, opened 2018-08-14
Chore: Support Update Card Tokenization

dmiluski:
Problem:
While leveeraging BTCardClient tokenization process, I found Venmo Android was able to leverage the tokenization process to updating a card, but iOS client was not. This leads to any update card process initialized on iOS to fail.

Upon further inspection, I found Android's builder process never setting the card number, while iOS implementation was setting value to "" given the requirement of BTCard.h for a non-null parameter.

Approach:
Remove number from a requirement for creating a BTCard for use in tokenization. Create a separate init function which states its use is for updating card rather than making current init number optional.

Change:
+ Update update card variant of BTCard initializer
+ Add test

### DIFF
--- a/BraintreeCard/BTCard.m
+++ b/BraintreeCard/BTCard.m
@@ -77,6 +77,17 @@ NSString *const BTCardGraphQLTokenizationMutation = @""
     return self;
 }
 
+- (instancetype)initWithExpirationMonth:(NSString *)expirationMonth
+                         expirationYear:(NSString *)expirationYear
+                                    cvv:(nullable NSString *)cvv {
+    if (self = [self initWithParameters:@{}]) {
+        _expirationMonth = expirationMonth;
+        _expirationYear = expirationYear;
+        _cvv = cvv;
+    }
+    return self;
+}
+
 #pragma mark -
 
 - (NSDictionary *)parameters {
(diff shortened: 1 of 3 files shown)


demerino:
You should be able to use the default initializer and set properties as needed:

```
BTCard *card = [[BTCard alloc] init];
card.expirationYear = @"...";
card.expirationMonth = @"...";
card.cvv = @"...";
```

dmiluski:
Totally understand that recommendation, and I can move forward with that on our side? But are we the only client that performs update operation which requires tokenization? Would this help anyone else?

dmiluski:
Ended up implementing a little something locally:

```
extension BTCard {
    // Work-a-Round until Braintree PR is merged:
    // https://github.com/braintree/braintree_ios/pull/412
    //
    // Helper Method for creating BTCard without Number (Useful when updating)
    convenience init(number: String?, expMonth: String, expYear: String, cvv: String?, postalCode: String?) {
        self.init(number: number ?? "", expirationMonth: expMonth, expirationYear: expYear, cvv: cvv)
        self.number = number
        self.postalCode = postalCode
    }
}
```

Still running into a little trouble. Not sure why this isn't working where it worked when edited in this repo code. Perhaps I missed something. Will follow up tomorrow.

PR #422, opened 2018-10-04
#420 use "#ifdef DEBUG" instead of "if"

### LINKED ISSUE #420: Replace `#if DEBUG` in headers by `#ifdef DEBUG`
### General information

* SDK/Library version: Any (4.17.0)
* Environment: Any
* iOS Version and Device: Any
* Integration type and version: CocoaPods 1.5.3

### Issue description

Please consider replacing `#if DEBUG` in headers by more common `#ifdef DEBUG`

#### The difference

-  `#ifdef DEBUG` evaluates to true (code we will be in) when `DEBUG` is defined
    - It is common to define `DEBUG=1` in debug builds, don't define anything (no `DEBUG=0`) in release builds
- `#if DEBUG` evaluates to true if `DEBUG` is defined as `1`

By "common" I mean common in iOS development. For example Xcode's Obj-C project templates are using this concept.

screen shot 2018-09-27 at 10 59 39

#### How does it work now?

Construct `#if DEBUG` is used to provide some methods in headers in DEBUG mode only.

- If `DEBUG=1` is defined, the code will _be_ compiled in
- If `DEBUG=0` is defined, the code will _not be_ compiled in
    - If _no_ `DEBUG` is defined at all, it will by default be evaluated as `0`, turning the condition to `#if 0` -> code will not be compiled in

#### What is the problem?

In short: strict compiler flags.

The fallback of non-defined value to `0` is actually the problem. With stricter compiler settings using `-Wundef` flag ([docs](https://www.cleancss.com/explain-command/gcc/4801)), this behaviour actually raises a warning.

When used together with `-Werror` (treat warnings as errors, [docs](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html), warning is actually treated as error.

As a result, using Braintree causes *non-debug builds to fail* for project with this strict compiler settings.

##### Wrap import by pragma ignore
In Obj-C it is technically possible to wrap all imports of Braintree SDK with pragma ignore (`#pragma clang diagnostic ignored ...`). However this is not even possible when using Braintree as a module in Swift, where we don't make the import explicitly ourselves.

##### Treat -Wundef still just as error
It is possible to opt-out `-Wunder` from `-Werror` by `-Wno-error=undef` flag to still keep `-Wunder` just as error. However this is not ideal as it disables this behavior also for the project, not just for the Braintree SDK.

lukaskukacka:
This resolves #420

### DIFF
--- a/BraintreePayPal/PayPalOneTouch/Configuration/PPOTConfiguration.h
+++ b/BraintreePayPal/PayPalOneTouch/Configuration/PPOTConfiguration.h
@@ -59,7 +59,7 @@ typedef void (^PPOTConfigurationCompletionBlock)(PPOTConfiguration *currentConfi
 */
 + (PPOTConfiguration *)configurationWithDictionary:(NSDictionary *)dictionary;
 
-#if DEBUG
+#ifdef DEBUG
 + (void)useHardcodedConfiguration:(BOOL)useHardcodedConfiguration;
 #endif
 
--- a/BraintreePayPal/PayPalOneTouch/Configuration/PPOTConfiguration.m
+++ b/BraintreePayPal/PayPalOneTouch/Configuration/PPOTConfiguration.m
@@ -350,7 +350,7 @@ typedef void (^PPOTConfigurationFileDownloadCompletionBlock)(NSData *fileData);
 
 #pragma mark - debug-only stuff
 
-#if DEBUG
+#ifdef DEBUG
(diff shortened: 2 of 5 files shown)


lukaskukacka:
Could someone merge this PR please? I don't have write access, so I cannot merge it myself.

crookedneighbor:
I'd like one other Braintree dev to look over it and approve it first. I'll make sure someone looks at it today.

PR #426, opened 2018-10-23
Use dispatch groups, early return

supermarin:
Semaphores will lock the calling thread, and in the current implementation,
it will happen every single time this method is invoked.
We're observing a small percentage of app freezes caused by this method.

The new implementation early returns if the config was already initialized.
Another way of implementing it is with dispatch_once(), but that wouldn't
allow clearing out the config and re-fetching (not implemented in this PR to
minimize changes).

### DIFF
--- a/BraintreeCore/BTAPIClient.m
+++ b/BraintreeCore/BTAPIClient.m
@@ -252,15 +252,20 @@ NSString *const BTAPIClientErrorDomain = @"com.braintreepayments.BTAPIClientErro
     // Note: Configuration queue is SERIAL. This helps ensure that each request for configuration
     //       is processed independently. Thus, the check for cached configuration and the fetch is an
     //       atomic operation with respect to other calls to this method.
-    //
-    // Note: Uses dispatch_semaphore to block the configuration queue when the configuration fetch
-    //       request is waiting to return. In this context, it is OK to block, as the configuration
-    //       queue is a background queue to guarantee atomic access to the remote configuration resource.
+    static BTConfiguration *configuration;
+
+
     dispatch_async(self.configurationQueue, ^{
+        if (configuration) {
+            dispatch_async(dispatch_get_main_queue(), ^{
+                completionBlock(configuration, nil);
+            });
+            return;
+        }
+
+        dispatch_group_t group = dispatch_group_create();
+        dispatch_group_enter(group);
         __block NSError *fetchError;
-        
-        dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
-        __block BTConfiguration *configuration;
         NSString *configPath = self.tokenizationKey ? @"v1/configuration" : [self.clientToken.configURL absoluteString];
         [self.configurationHTTP GET:configPath parameters:@{ @"configVersion": @"3" } completion:^(BTJSON * _Nullable body, NSHTTPURLResponse * _Nullable response, NSError * _Nullable error) {
             if (error) {
@@ -297,16 +302,13 @@ NSString *const BTAPIClientErrorDomain = @"com.braintreepayments.BTAPIClientErro
                     }
                 }
             }
-            
-            // Important: Unlock semaphore in all cases
-            dispatch_semaphore_signal(semaphore);
-        }];
-        
-        dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
 
-        dispatch_async(dispatch_get_main_queue(), ^{
-            completionBlock(configuration, fetchError);
-        });
+            dispatch_async(dispatch_get_main_queue(), ^{
+                completionBlock(configuration, fetchError);
+            });
+            dispatch_group_leave(group);
+        }];
+        dispatch_group_wait(group, DISPATCH_TIME_FOREVER);
     });
 }
 


supermarin:
You can test the functionality isolated in this snippet:
``` objc
//
//  ViewController.m
//  dispatch
//
//  Created by Marin Usalj on 10/22/18.
//  Copyright © 2018 Marin Usalj. All rights reserved.
//

#import "ViewController.h"

@interface ViewController ()
@end
@implementation ViewController

static dispatch_queue_t serialq;
static dispatch_queue_t concurrentq;

- (void)getConfigurationWithCompletion:(void (^)(NSString *))completionBlock {
    static NSString *configuration;

    dispatch_async(serialq, ^{
        if (configuration) {
            NSLog(@"returning cached config.");
            dispatch_async(dispatch_get_main_queue(), ^{
                completionBlock(configuration);
            });
            return;
        }

        NSLog(@"grouping...");
        dispatch_group_t group = dispatch_group_create();
        dispatch_group_enter(group);
        [self GET:^(NSString *result) {
            configuration = result;
            dispatch_async(dispatch_get_main_queue(), ^{
                completionBlock(configuration);
            });
            dispatch_group_leave(group);
        }];
        dispatch_group_wait(group, DISPATCH_TIME_FOREVER);
    });
}

- (void)GET:(void (^)(NSString *))completionBlock {
    dispatch_async(concurrentq, ^{
        NSLog(@"requesting network...");
        sleep(1);
        NSLog(@"network request done...");
        completionBlock(@"WAT");
    });
}

- (void)viewDidLoad {
    [super viewDidLoad];
    // Do any additional setup after loading the view, typically from a nib.
    concurrentq = dispatch_queue_create("work q", DISPATCH_QUEUE_CONCURRENT);
    serialq = dispatch_queue_create("exec q", DISPATCH_QUEUE_SERIAL);
}

- (void)viewDidAppear:(BOOL)animated {
    [super viewDidAppear:animated];

    for (int i = 1; i <= 5; i ++) {
        NSLog(@"Getting configuration for the %d time.", i);
        [self getConfigurationWithCompletion:^(NSString *conf) {
            NSLog(@"Got configuration %d: %@", i, conf);
        }];
    }
}

@end

```

demerino:
Thanks for taking the time to create this PR, we'll take a look.

BraintreeCore/BTAPIClient.m
```diff
@@ -297,16 +302,13 @@ - (void)fetchOrReturnRemoteConfiguration:(void (^)(BTConfiguration *, NSError *)
...
 
-        dispatch_async(dispatch_get_main_queue(), ^{
-            completionBlock(configuration, fetchError);
-        });
+            dispatch_async(dispatch_get_main_queue(), ^{
+                completionBlock(configuration, fetchError);
+            });
+            dispatch_group_leave(group);
+        }];
+        dispatch_group_wait(group, DISPATCH_TIME_FOREVER);
```
kastiglione: I think this should be a small wait time. Maybe the next request gets through even though the first one isn't.

supermarin: Mind elaborating on this @kastiglione?

If I'm understanding this correctly, the network library itself should retry / wait for the maximum wait time and call the callback with fetchError. In that case, this operation will pop from the dispatch queue and the next one will miss the cache & re-request again.

kastiglione: It's a question of how long is the network wait time? My suggestion is that this should control how long to wait before trying the a new queued fetch (if one exists).

kastiglione: For a concrete example:

1. First request is made
2. Second request is queued

Currently, second request will wait for however long the network library waits (including maybe retries).

I'm suggesting this function can own how long to wait for, instead of letting the networking decide. If it the first one is slow for some reason, a second one might get through. Just a thought.

supermarin: Gotcha, yep it makes sense what you're suggesting.

It seems like timeout is 60 sec (both default and what they pass in).
https://github.com/braintree/braintree_ios/blob/9a192145eb8c5ae03ca7903a70773fcbeeb63d1c/BraintreePayPal/PayPalOneTouch/Configuration/PPOTConfiguration.m#L37

I think the handler should be rewritten in that case, since requests in flight would be racing on `configuration` setting, and the  callers would just get a completion with nil value (not a proper error).

There's a third option of passing the timeout to the network lib, that way you get both

supermarin:
@demerino any updates on this?

billwerges:
Hi @supermarin,

We are going to be releasing an update this week that utilizes a static configuration object and returns early if we have already fetched the config. We would appreciate any feedback on whether that solves the issue you're seeing with the app freezes your seeing.

supermarin:
@billwerges That sounds like it might fix it, if you've also removed semaphores (or any kind of multithread locking while > 1 request is still in flight). Would you mind pointing me to the implementation you're about to release?

crookedneighbor:
@supermarin any updates on @billwerges's question?

supermarin:
Nope, I haven't got pinged again about the issue. It's possible it was resolved in some of the updates.

crookedneighbor:
I'm going to close this for now then, but if it becomes an issue again, feel free to comment here.

PR #430, opened 2018-11-27
Fix a static analyzer memory error in BTAPIClient

rjstelling:
`graphQLURLForEnvironment:` could return `nil` but was marked as returning a `nonnull` object.


rjstelling:
This has now regressed. The fixes in this PR have vanished from the master branch.

PR #461, opened 2019-06-07
 Adding collectPayPalDeviceInfo to PPDataCollector

manibrundha:
Adding a function to PPDataCollector to get PayPalDeviceInfo

### DIFF
--- a/BraintreePayPal/PayPalDataCollector/PPDataCollector.m
+++ b/BraintreePayPal/PayPalDataCollector/PPDataCollector.m
@@ -22,10 +22,17 @@
 
 @implementation PPDataCollector
 
-+ (NSString *)generateClientMetadataID:(NSString *)clientMetadataID disableBeacon:(BOOL)disableBeacon data:(NSDictionary *)data {
+
++ (PPRMOCMagnesSDKResult *)generateMagnesResultWithClientMetadataID:(NSString *)clientMetadataID disableBeacon:(BOOL)disableBeacon data:(NSDictionary *)data {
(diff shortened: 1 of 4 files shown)


BraintreePayPal/PayPalDataCollector/PPDataCollector.m
```diff
@@ -62,4 +69,21 @@ + (nonnull NSString *)collectPayPalDeviceData {
...
+    if (payloadDict == nil) {
+        return @"";
+    }
+    NSError *error;
+
+    NSData *data = [NSJSONSerialization dataWithJSONObject:payloadDict options:0 error:&error];
+    if (!data) {
+        NSLog(@"ERROR: Failed to get PayPal DeviceInfo , error = %@", error);
+    }
+    return [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
```
billwerges: NSString's `initWithData:` method requires a non-null `data` value. This should probably return early if `NSJSONSerialization` fails and `data` is null, similar to the early return when `payloadDict` is null.

BraintreePayPal/PayPalDataCollector/Public/PPDataCollector.h
```diff
@@ -43,4 +43,14 @@
...
 
+/**
+Collects Device Info for PayPal.
+
+ This should be used when the user is paying with PayPal.
+ @param pairingID a pairing ID to associate with this clientMetadataID must be 10-32 chars long or null
+
+ @return a deviceInfo string that should be passed into server-side calls.
+ */
++ (nonnull NSString *)collectPayPalDeviceInfoWithClientMetadataID:(nullable NSString *)pairingID;
```
scannillo: Can I get more context on what a `pairingID` is?

demerino: `pairingID` is another name for the `clientMetadataId`. It is the ID related for risk data.

PR #466, opened 2019-07-12
Correct importing of BTConfiguration+ThreeDSecure

ejensen:
This PR fixes the following warning cased by importing the `BTConfiguration+ThreeDSecure` implementation file rather than the header:
> instance method ‘cardinalAuthenticationJWT’ in category from BTConfiguration+ThreeDSecure.o conflicts with same method from another category


demerino:
Thanks for the PR, we'll get this fix out early next week.

Issue #485, opened 2019-09-18
Carthage Submodule Integration Results in Dirty Submodule


dmiluski:
### General information

* SDK/Library version: "4.28.0"
* Environment: N/A
* iOS Version and Device: (Occurs in SDK integration)
* Integration type and version: Carthage 0.33.0

### Issue description

When integrating braintree_ios version `4.28.0` we noticed our submodule directories being dirtied on  boostrap.

Given Cartfile.resolved containing
```
github "braintree/braintree_ios" "4.28.0"
```

And installing dependencies via command via submodules
```
#!/usr/bin/env sh

set -eu

PLATFORMS="ios"

carthage bootstrap --no-build --use-ssh --use-submodules --platform "$PLATFORMS" "$@"
```

The integrated submodule is immediately dirty Project Directory:
<img width="542" alt="Screen Shot 2019-09-18 at 8 50 30 AM" src="https://user-images.githubusercontent.com/5083390/65164275-741e4f80-d9f1-11e9-8293-6ea118c315d1.png">

Then Submodule Directory:
<img width="489" alt="Screen Shot 2019-09-18 at 8 52 04 AM" src="https://user-images.githubusercontent.com/5083390/65164382-a0d26700-d9f1-11e9-8440-6d05a55ccd4a.png">

## Why ❓ 
When Carthage dependencies have dependencies themselves, Carthage resolves those versions and pulls in the tree of dependencies in the top level of the project and SYMLinks to the dependency directory.

## Requested Fix
Ignore this file in BT .gitignore

```
# Carthage
#
# Add this line if you want to avoid checking in source code from Carthage dependencies.
Carthage/Checkouts
Carthage/Build
```

scannillo:
Hey @dmiluski. 
> we noticed our submodule directories being dirtied

I don't know what you mean by this.

CardinalMobile is the provider we use for ThreeDSecure verifications. It is not included in our SDK by mistake and therefore we don't want to add it to our SDKs `.gitignore`. You can instead add it to your app's `.gitignore` if you don't want to utilize ThreeDSecure.

PR #486, opened 2019-09-18
Issue 485: Carthage Submodule Integration Dirty

dmiluski:
#### Issue:
https://github.com/braintree/braintree_ios/issues/485

#### Problem:
When integrating braintree_ios sdk via Carthage (as submodules), consumer Applications repos are dirtied due to a SYMLink between the cloned carthage checkout directory, and the Applications root Carthage/Checkout directory.

This had not been an issue with our previous integration of Release 4.24.0, which didn't have the Cardinal dependency yet to reveal this issue.

#### Why
When Carthage integrates dependencies, it inspects each dependency as that dependencies dependency, and eventually constructs a Cartfile.resolved include all dependencies in the graph.

As part of this resolution, all dependencies of dependencies are cloned to the top level which may then be shared with dependencies (if multiple dependencies reuse the same item). This is done via a SYMLink.

In our case, this exibited the behavior of a new file being created in the cloned braintree submodule:
`<App_ProjectDir>/Carthage/Checkouts/braintree_ios/Carthage/Checkouts/CardinalMobile`

This new SYMLINK left our project dirty. Given the SYMLink exists in the braintree_ios folder structure, the folder which needs this .gitignore is the braintree_ios repo.

[Default Swift Github gitignore](https://github.com/github/gitignore/blob/master/Swift.gitignore)
By default, when creating new github repo's a .gitignore may be provided by Github, often with default items such as this as well as CocoaPods. In this case, this is mirroring what the CocoaPods Pods path is doing just above this change.

### Outcome
CardinalMobile will still be included when an application integrated braintree_ios based on Carthage dependency resolution by the public Cartfile in this repo. The difference is, after this change, consuming applications will be able to still consume as a submodule, but not have a dirty repo.

### Reproduction Example:
https://github.com/dmiluski/braintree-ios-issue-demo

### Precedent already used at Braintree
https://github.com/braintree/braintree-ios-drop-in/blob/master/.gitignore#L38

This was likely overlooked in this repo as it did not include Carthage dependencies of its own prior to `CardinalMobile`

### Checklist

- [x] Include Carthage Directory .gitignore to play friendly with Carthage integrations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,12 @@ DerivedData
 # CocoaPods
 Pods
 
+# Carthage
+#
+# Add this line if you want to avoid checking in source code from Carthage dependencies.
+Carthage/Checkouts
+Carthage/Build
+
 # This file is needed for signing the ipa with our enterprise cert;
 # you can get it from the iOS Developer Center.
 EverybodyVenmo.mobileprovision
@@ -32,4 +38,3 @@ EverybodyVenmo.mobileprovision
 .hockeyapp
 
 *.pid
-


scannillo:
Hi. See our response to this solution [here](https://github.com/braintree/braintree_ios/issues/485#issuecomment-532755238). We cannot accept this PR.

dmiluski:
Hi @scannillo , I updated this PR with a reference to a reproduction example repo, as well as cited the similar approach utilized in braintree-ios-drop-in.

This is an issue for us, and I'd like to chat a bit more to see if I can help this move forward as this will affect other developers if not addressed.

PR #554, opened 2020-11-30
Convert CI to GH Actions

scannillo:
### Summary of changes

* Remove travis file
* Add GH actions file
* Make PPRiskMagnes.framework an embedded framework in PayPalDataCollector, which allows the integration tests to run

### Checklist

- ~Added a changelog entry~

- @scannillo
- @sestevens


sshropshire (approved):

.github/workflows/testing.yml
```diff
@@ -0,0 +1,48 @@
...
+jobs:
+  unit_test_job:
+    name: Unit Tests
+    runs-on: macOS-latest
+    steps:
+      # Checks out a copy of your repository
+      - name: Checkout repository
+        uses: actions/checkout@v1
+      - name: Install gems
+        run: bundle install
```
scannillo, line 12: Can we do `bundle install cocoapods` since we're only using the `cocoapods` gem in the tasks below?

scannillo: Opted to use GH actions version of CocoaPods. Please comment if you prefer we use the version in our Gemfile.

PR #566, opened 2020-12-22
Add paymentTypeCountryCode to BTLocalPaymentRequest

sestevens:
### Summary of changes

- Add `paymentTypeCountryCode` to `BTLocalPaymentRequest`

### Checklist

- [x] Added a changelog entry

- @sestevens

### DIFF
--- a/Demo/UI Tests/PayPal UI Tests/PayPal_BillingAgreement_UITests.swift
+++ b/Demo/UI Tests/PayPal UI Tests/PayPal_BillingAgreement_UITests.swift
@@ -48,7 +48,7 @@ class PayPal_BillingAgreement_UITests: XCTestCase {
     func testPayPal_billingAgreement_cancelsSuccessfully_whenTappingCancelButtonOnPayPalSite() {
         let webviewElementsQuery = app.webViews.element.otherElements
 
-        self.waitForElementToAppear(webviewElementsQuery.links["Cancel Sandbox Purchase"])
+        self.waitForElementToAppear(webviewElementsQuery.links["Cancel Sandbox Purchase"], timeout: 20)
 
         webviewElementsQuery.links["Cancel Sandbox Purchase"].forceTapElement()
 
--- a/Sources/BraintreePaymentFlow/Public/BraintreePaymentFlow/BTLocalPaymentRequest.h
+++ b/Sources/BraintreePaymentFlow/Public/BraintreePaymentFlow/BTLocalPaymentRequest.h
@@ -23,6 +23,15 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, nullable, copy) NSString *paymentType;
 
 /**
+ The country code of the local payment.
+
+ This value must be one of the supported country codes for a given local payment type listed at the link below. For local payments supported in multiple countries, this value may determine which banks are presented to the customer.
+
+ https://developers.braintreepayments.com/guides/local-payment-methods/client-side-custom/ios/v4#invoke-payment-flow
+ */
+@property (nonatomic, nullable, copy) NSString *paymentTypeCountryCode;
+
+/**
  Optional: A non-default merchant account to use for tokenization.
  */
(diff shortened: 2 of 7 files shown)


Demo/UI Tests/PayPal UI Tests/PayPal_BillingAgreement_UITests.swift
```diff
@@ -48,7 +48,7 @@ class PayPal_BillingAgreement_UITests: XCTestCase {
     func testPayPal_billingAgreement_cancelsSuccessfully_whenTappingCancelButtonOnPayPalSite() {
         let webviewElementsQuery = app.webViews.element.otherElements
 
-        self.waitForElementToAppear(webviewElementsQuery.links["Cancel Sandbox Purchase"])
+        self.waitForElementToAppear(webviewElementsQuery.links["Cancel Sandbox Purchase"], timeout: 20)
```
sestevens, line 51: This change isn't directly related to this PR, but I saw the test failing in CI, so I increased the timeout to get CI passing. (Same thing on the file below.)

scannillo: 🙏

Sources/BraintreePaymentFlow/Public/BraintreePaymentFlow/BTLocalPaymentRequest.h
```diff
@@ -22,6 +22,15 @@ NS_ASSUME_NONNULL_BEGIN
...
 @property (nonatomic, nullable, copy) NSString *paymentType;
 
+/**
+ The country code of the local payment.
+
+ This value must be one of the supported country codes for a given local payment type listed at the link below. For local payments supported in multiple countries, this value may determine which banks are presented to the customer.
+
+ https://developers.braintreepayments.com/guides/local-payment-methods/client-side-custom/ios/v4#invoke-payment-flow
+ */
+@property (nonatomic, nullable, copy) NSString *paymentTypeCountryCode;
```
scannillo, line 32: Would it be more clear to just call it `countryCode`, since that is what the value is? `paymentTypeCountryCode` is throwing me off a little.

scannillo: > For local payments supported in multiple countries, this value may determine which banks are presented to the customer.

I'm a little confused by this docstring. Is it saying that for LPMs that support multiple countries, this value represents the one that was actually shown to the customer?

sestevens: Both the name of the property and the docstring match what was implemented in the JS SDK. Since there's a property called `paymentType` (declared just above this one), calling it `paymentTypeCountryCode` makes it clear that it's the country code for the specific local payment type (as opposed to the merchant or customer's country, I suppose).

I interpret the docstring as saying that this property may determine which banks _will_ be presented to the customer. (Not which ones already have been.)

sshropshire (approved):

Sources/BraintreePaymentFlow/Public/BraintreePaymentFlow/BTLocalPaymentRequest.h
scannillo: Yeh, I just saw that we used the same phrasing as in JS. 

The plural on `banks` is throwing me off, since I assume this value can only represent 1 countryCode.

sestevens: I take it to mean that there could be more than one bank available in a given country.

scannillo: AH - yes. Thank you

scannillo (approved):

PR #567, opened 2020-12-22
Add GH Actions workflow to test package managers

sestevens:
### Summary of changes

- Add GitHub actions workflow to test integration with supported package managers

### Checklist

- ~Added a changelog entry~

- @sestevens

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,8 +1,8 @@
-name: Check package managers
+name: Build
 on: [pull_request]
 jobs:
-  check_cocoapods:
-    name: Check CocoaPods
+  cocoapods:
+    name: CocoaPods
     runs-on: macOS-latest
     steps:
       - name: Check out repository
@@ -13,8 +13,8 @@ jobs:
         run: pod install
       - name: Run pod lib lint
         run: pod lib lint --allow-warnings
-  check_carthage:
-    name: Check Carthage
+  carthage:
+    name: Carthage
     runs-on: macOS-latest
     steps:
       - name: Check out repository
--- a/.github/workflows/check-package-managers.yml
+++ b/.github/workflows/check-package-managers.yml
@@ -1,4 +1,4 @@
-name: test-package-managers
+name: Check package managers
 on: [pull_request]
 jobs:
   check_cocoapods:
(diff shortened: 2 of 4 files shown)


.github/workflows/check-package-managers.yml
```diff
@@ -0,0 +1,27 @@
+name: Check package managers
+on: [pull_request]
```
scannillo, line 2: I wonder if instead of having this run on every PR (& PR commit), we just have it run on every merge/push to master? Just so it doesn't slow down too much the test runs on each commit.

sestevens: Yeah, I wanted to test out how long this would take, and right now the package manager checks are actually faster than the test workflow. Since they're running in parallel, running the package manager checks on PRs won't impact the total time.

That being said, I'd like to add a check for SPM, so we'll have to see how fast that runs. :)

sestevens:
I'm marking this as "Ready for Review." I'm still planning to add a check for SPM at some point, but I figure it would be helpful to check for CocoaPods and Carthage in the meantime.

.github/workflows/check-package-managers.yml
```diff
@@ -0,0 +1,27 @@
+name: Check package managers
```
scannillo, line 1: I think we can update the names of our workflows and jobs so they read a little better in the bottom of the PR.

Maybe something like ...
```
Tests / Unit
Tests / UI
Tests / Integration
Build Demo / Carthage
Build Demo / CocoaPods
```

sestevens: I'm down for renaming the jobs! But the Carthage check doesn't actually build a demo app. It's just running `carthage build --no-skip-current`, which I think just makes sure the frameworks build.

sestevens: Maybe we could do:
```
Tests / Unit
Tests / UI
Tests / Integration
Build / Carthage
Build / CocoaPods
```

scannillo: I like it 👍

sestevens: Updated!

scannillo (approved):
🏁

.github/workflows/build.yml
```diff
@@ -0,0 +1,27 @@
...
+  carthage:
+    name: Carthage
+    runs-on: macOS-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v1
+      - name: Use Xcode 12.3
+        run: sudo xcode-select -switch /Applications/Xcode_12.3.app
+      - name: Remove SPMTest
+        run: rm -rf SampleApps/SPMTest && git add . && git commit -m 'Remove SPMTest app to avoid Carthage timeout'
```
sshropshire, line 25: question: are these git changes persisted anywhere? Or is it just a local commit.

sshropshire (approved):

.github/workflows/build.yml
sestevens: Should just be a local commit that gets wiped away after the job finishes!

hollabaq86 (approved):

PR #569, opened 2020-12-29
3DS2 UI Customization Updates

sestevens:
### Summary of changes

Fix 3DS2 UI customization and adds support for Swift Package Manager

- Add wrapper classes around corresponding Cardinal 3DS2 UI customization classes. The actual Cardinal classes will be created at runtime, which allows us to support SPM.

### Checklist

- [x] Added a changelog entry

- @sestevens 
- @scannillo


Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureV2ButtonCustomization.h
```diff
@@ -0,0 +1,28 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * The ButtonCustomization class provides methods for the 3DS Requestor App to pass button customization parameters to the 3DS SDK.
+ */
+@interface BTThreeDSecureV2ButtonCustomization : BTThreeDSecureV2BaseCustomization
+
+/**
+ * @property backgroundColor Colour code in Hex format. For example, the colour code can be “#999999”.
```
sarahkoop, line 17:
```suggestion
 * @property backgroundColor Color code in Hex format. For example, the color code can be “#999999”.
```

Not sure if these came from Cardinal docs and we want them to match, but could change this to match the variable name.

Sources/BraintreeThreeDSecure/BTProxyCardinal.h
```diff
@@ -25,6 +31,7 @@ typedef NS_ENUM(NSUInteger, BTProxyCardinalResponseActionCode) {
 
 @protocol BTProxyCardinalSessionConfiguration <NSObject>
 @property (nonatomic, assign) BTProxyCardinalSessionEnvironment deploymentEnvironment;
+@property (nonatomic, strong) id uiCustomization;
```
sshropshire, line 34: random question: would it make sense for `uiCustomization` be a `BTThreeDSecureV2UICustomization` instead of `id` here?

sshropshire (approved):
🔥

Sources/BraintreeThreeDSecure/V2UICustomization/BTThreeDSecureV2ButtonCustomization.m
```diff
@@ -0,0 +1,40 @@
...
+- (instancetype)init {
+    self = [super init];
+    if (self) {
+        self.cardinalValue = [NSClassFromString(@"ButtonCustomization") new];
+    }
+
+    return self;
+}
+
+- (void)setBackgroundColor:(NSString *)backgroundColor {
```
sarahkoop, line 21: Should there be any validation of the values passed into the color fields to match the format required by Cardinal? (the docs say hex format is required)

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureV2ButtonCustomization.h
sestevens: Good call!

Sources/BraintreeThreeDSecure/BTProxyCardinal.h
sestevens: No, it needs to be `id`, unfortunately. At runtime it will be a Cardinal class, but to work around not having an xcframework for Cardinal, we can't use Cardinal classes at compile time, which is why we're using `id` here.

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureV2ButtonCustomization.h
sestevens: I went through and updated a bunch of these doc strings to be more clear and correct. I don't think they need to match Cardinal's docs exactly.

Sources/BraintreeThreeDSecure/V2UICustomization/BTThreeDSecureV2ButtonCustomization.m
sestevens: My initial thought is that we should keep these classes as simple as possible and just pass values along to Cardinal. But I'm curious if anyone else has opinions on this?

sestevens: Also, I just checked to see if Cardinal logs any errors or anything around this, and it doesn't look like they do.

sarahkoop: Ok then keeping it simple seems reasonable to me, unless anyone else opposes.

sarahkoop (approved):
Looks good to me 🌟

scannillo:
Do we want to add a UI test that setting customization via `BTThreeDSecureV2UICustomization` properly reflects in the 3DS2 challenge screen?

scannillo:
> Do we want to add a UI test that setting customization via BTThreeDSecureV2UICustomization properly reflects in the 3DS2 challenge screen?

[We can't test colors using XCTest](https://stackoverflow.com/a/38676658).

scannillo:
Tests are passing reliable locally, and CI test failures are part of a caching issue we are currently working to resolve. Merging PR!

PR #573, opened 2021-01-05
Add UI tests for American Express

sestevens:
### Summary of changes

- Convert American Express integration tests into UI tests. 

We weren't able to figure out why, but our Amex integration tests were failing when run on fresh simulators, which meant they were failing pretty consistently in CI. We converted them into UI tests, and we're hoping this makes them more stable 🤞 

### Checklist

- ~Added a changelog entry~

- @sestevens 
- @scannillo


demerino (approved):
Weird that the integration tests fail

Demo/UI Tests/American Express UI Tests/AmericanExpress_UITests.swift
```diff
@@ -0,0 +1,35 @@
...
+        app.launchArguments.append("-ClientToken")
+        app.launchArguments.append("-Integration:BraintreeDemoAmexViewController")
+        app.launch()
+
+        waitForElementToAppear(app.buttons["Valid card"])
+    }
+
+    func testValidCard_receivesRewardsBalance() {
+        app.buttons["Valid card"].tap()
+        waitForElementToAppear(app.buttons["45256433 Points, 316795.03 USD"], timeout: 10)
```
sarahkoop, line 21: Do these `waitForElementToAppear` statements need to be wrapped in some sort of assertion?  It's unclear if the tests will actually fail or just return false if the wait times out.

sestevens: The test will fail if the element isn't found in the given amount of time. [`waitForElementToAppear`](https://github.com/braintree/braintree_ios/blob/bf20534a349502c74603eeba8b61e42ba4801b47/Demo/UI%20Tests/Helpers/BTUITest.swift#L4) calls XCTest's `waitForExpectations` method, which fails the test if the expectation isn't fulfilled.

sarahkoop: Ah ok thanks for the explanation!

sarahkoop (approved):
🌟

PR #579, opened 2021-01-08
Make GET and POST methods on BTAPIClient Public

sestevens:
### Summary of changes

- Braintree Drop-in uses a couple of methods on BTAPIClient that are currently internal only. BT Drop-in was redeclaring these methods in an internal header in order to access them. We're making these methods public in order to avoid a "ambiguous reference" error in Drop-in. (See https://github.com/braintree/braintree-ios-drop-in/pull/256.)

We aren't adding a changelog entry or doc comments for these methods, since we don't intend them to be used by merchants.

### Checklist

- ~Added a changelog entry~

- @scannillo
- @sestevens

### DIFF
--- a/Sources/BraintreeCore/Public/BraintreeCore/BTAPIClient.h
+++ b/Sources/BraintreeCore/Public/BraintreeCore/BTAPIClient.h
@@ -33,6 +33,20 @@ typedef NS_ENUM(NSInteger, BTAPIClientErrorType) {
 };
 
 /**
+ :nodoc:
+*/
+typedef NS_ENUM(NSInteger, BTAPIClientHTTPType) {
+    /// Use the Gateway
+    BTAPIClientHTTPTypeGateway = 0,
+
+    /// Use the Braintree API
+    BTAPIClientHTTPTypeBraintreeAPI,
+
+    /// Use the GraphQL API
+    BTAPIClientHTTPTypeGraphQLAPI,
+};
+
+/**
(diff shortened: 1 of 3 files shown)


Sources/BraintreeCore/Public/BraintreeCore/BTAPIClient.h
```diff
@@ -91,6 +105,7 @@ typedef NS_ENUM(NSInteger, BTAPIClientErrorType) {
                       completion:(void(^)(NSArray <BTPaymentMethodNonce *> * _Nullable paymentMethodNonces, NSError * _Nullable error))completion;
 
 /**
+ :nodoc:
```
sarahkoop, line 108: Not sure I understand why `:nodoc:` is being added above the existing docs for these methods. Are they also only supposed to be used internally?

sshropshire (approved):

Sources/BraintreeCore/Public/BraintreeCore/BTAPIClient.h
sestevens: Yeah, that was our thought. These methods default to using the REST API instead of GraphQL, and the methods that we're exposing let you choose which API to use. But really, all of these methods are meant to be used by other Braintree modules, not by merchants directly.

sarahkoop: Got it 👍

sarahkoop (approved):
🌟

Issue #576, opened 2021-01-06
Unable to archive project which use BraintreeDataCollector or PayPalDataCollector


JonathanDowning:
**Integration Details (please complete the following information):**
 - SDK/Library Version: master branch/commit hash `acf23311388bc6be10f9a1b7740c77fa3a42c1f5`
 - Environment: N/A
 - iOS Version: N/A
 - Device: N/A
 - Integration type & version: Swift Package Manager

**Describe the bug**
Xcode cannot archive `BraintreeDataCollector` or `PayPalDataCollector` (or targets which depend on these).

These targets depend on `xcframeworks`, I suspect this is not a coincidence. 

Xcode emits the following error for each target:

`BraintreeDataCollector`:
> 'kDataCollector.h' file not found

`PayPalDataCollector`:
> No such module 'PPRiskMagnes'

**To Reproduce**
1. In Terminal, from the project directory: `git checkout acf23311388bc6be10f9a1b7740c77fa3a42c1f5` (or pull the current master branch)
2. `open Package.swift`
3. Select either `PayPalDataCollector` or `BraintreeDataCollector` from the schemes dropdown
4. Click Product > Archive
5. See error

**Expected behavior**
The project should archive successfully

**Screenshots**
<img width="1611" alt="Screen Shot 2021-01-06 at 16 15 59" src="https://user-images.githubusercontent.com/2230175/103821250-cbf42600-503b-11eb-94c5-b906b26a378e.png">
<img width="1611" alt="Screen Shot 2021-01-06 at 16 16 14" src="https://user-images.githubusercontent.com/2230175/103821252-cc8cbc80-503b-11eb-812b-dee4737bb2dd.png">

scannillo:
Hi @JonathanDowning - thank you for opening this issue and providing detail. I am able to replicate the issue. We will keep you posted as we work on a fix.

scannillo:
Hey @JonathanDowning - after some investigating, multiple sources can confirm this is an Xcode bug. A report has been filed for it [here](https://bugs.swift.org/browse/SR-13803).

Related issues:
- [Swift forums](https://forums.swift.org/t/binary-dependency-fails-on-archive/40780/7)
- [Apple dev forums](https://developer.apple.com/forums/thread/662247)

We'll keep you posted when we release a workaround.

JonathanDowning:
[...]

PR #582, opened 2021-01-13
Spm binary dep bug

scannillo:
### The Issue

- Addresses GH issue #576 - where a merchant is unable to archive an app that uses either the `BraintreeDataCollector` or `PayPalDataCollector` modules via SPM.
    - This stems from a known Xcode bug for SPM libraries that use binary dependencies. A bug report was filed for it [here](https://bugs.swift.org/browse/SR-13803).
- Related issues: [Swift forums](https://forums.swift.org/t/binary-dependency-fails-on-archive/40780/7) & [Apple dev forums](https://developer.apple.com/forums/thread/662247)

### This PR
The current workaround is to have the `KountDataCollector` & `PPRiskMagnes` targets be exposed as product libraries and require merchants to also include them via SPM.

- Update Package.swift to expose `KountDataCollector` & `PPRiskMagnes`
- Update README instructions
- Update SPMTest app to include `KountDataCollector` and `PPRiskMagnes` via SPM

### Checklist

- [X] Added a changelog entry
- [X] Archive via Xcode
- [X] Archive via command line (`xcodebuild archive -project SampleApps/SPMTest/SPMTest.xcodeproj/ -scheme SPMTest`)

### Authors
@scannillo


Package.swift
```diff
@@ -47,9 +47,17 @@ let package = Package(
             name: "BraintreeVenmo",
             targets: ["BraintreeVenmo"]
         ),
+        .library(
+            name: "KountDataCollector",
```
scannillo, line 51: What do folks think about renaming the SPM library names for these to be something along the lines of:
- `KountDataCollector` -> `BraintreeDataCollector_BinaryDependency` or
- `PPRiskMagnes` -> `PayPalDataCollector_Dependency`      ?

sestevens: I'd lean towards leaving the names as they are, just for our own sanity. But I could be persuaded otherwise, if you think the rename would be helpful.

README.md
```diff
@@ -46,16 +46,28 @@ The following Braintree libraries offer official Swift Package Manager support:
...
 
-** - To use the the `BraintreeThreeDSecure` library via SPM, you must [manually include](#manually-including-a-framework) the `CardinalMobile.framework` located in the `Frameworks` dir.
+#### BraintreeDataCollector
 
-#### Manually including a framework
+To use the `BraintreeDataCollector` library, you must also include the `KountDataCollector` library via SPM.
+
+#### PayPalDataCollector
+
+The `BraintreePaymentFlow`, `BraintreeThreeDSecure`, `BraintreePayPal`, and `BraintreeVenmo` libraries all require both the `PayPalDataCollector` and `PPRiskMagnes` libraries be included via SPM.
```
sestevens, line 64: This is making me wonder if we need a full breakdown of all the frameworks that are required for each payment method / flow. We'll definitely have that in the dev doc guides for each payment method, once those are updated for v5. I'm not sure if we also need it in the README?

sestevens (approved):

Package.swift
scannillo: I guess I was only proposing we change the name of the SPM Library that merchants will tick the box for, and keep the names for the targets, internal use (etc). 

I don't feel too strongly but wanted to throw the idea out there in case we think this might make integration less confusing for merchants.

sestevens: Right - I just meant that it might be confusing for us (or other people on the team in the future) to keep track of two sets of names.

README.md
scannillo: This is a good question 🤔 . It definitely is a bit confusing.

Would we want to include instructions per package manager for each module, or maybe to a dependency-tree-type diagram agnostic to each package manager?

hollabaq86: I like the idea of a visual diagram or table to help folks determine which packages they need.

Package.swift
sarahkoop: For what it's worth, in Android we rename the Magnes library to `paypal-data-collector` but leave Kount as `kount-data-collector`. I think the rename did make development with Magnes a little more confusing, but we continued to rename it for consistency. So, I would probably also lean towards leaving the names as they are, but maybe we should try to align the names of both across SDKs?

sarahkoop (approved):

Package.swift
scannillo: Thanks for the feedback - we will keep the names as is!

README.md
scannillo: Since CocoaPods & SPM resolve the Braintree dependencies for a merchant (ex: if you include `BraintreeApplePay` it will handle including it's dependency `BraintreeCore`), we think a visual diagram could be more confusing to merchants.

For our Carthage users, we outline the required frameworks in the [dev docs](https://developers.braintreepayments.com/guides/drop-in/setup-and-integration/ios/v4#carthage). There's a few payment methods that need updated Carthage instructions, but we'll tackle them with our big v5 docs overhaul.

PR #583, opened 2021-01-13
Add `accountType` to `BTThreeDSecureRequest`

sestevens:
### Summary of changes

- Add `accountType` to `BTThreeDSecureRequest`

### Checklist

- [x] Added a changelog entry

- @sestevens


Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
```diff
@@ -20,16 +20,30 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /**
- Error codes that describe errors that occur during 3D Secure.
+ 3D Secure version
```
sestevens, line 23: I think this was a copy-paste error at some point, so I fixed up this doc string.

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
```diff
@@ -20,16 +20,30 @@
...
     /// 3DS 2.0
     BTThreeDSecureVersion2
 };
 
+/**
+ The account type
+ */
+typedef NS_ENUM(NSInteger, BTThreeDSecureAccountType) {
+    /// Unspecified
+    BTThreeDSecureAccountTypeUnspecified,
```
sestevens, line 38: FYI - the only reason we need an "unspecified" option is because Objective-C enums cannot be nil.

sarahkoop (approved):
Looks good to me 🌟

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
```diff
@@ -45,6 +59,13 @@ typedef NS_ENUM(NSInteger, BTThreeDSecureVersion){
  */
 @property (nonatomic, copy) NSDecimalNumber *amount;
 
+/**
+ Optional. The account type selected by the cardholder
+
+ @note Some cards can be processed using either a credit or debit account and cardholders have the option to choose which account to use.
+ */
+@property (nonatomic, assign) BTThreeDSecureAccountType accountType;
```
scannillo, line 67: I think if we're going to make this an enum, we should also consider doing the same for the `shippingMethod` property on `BTThreeDSecureRequest`.

Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
```diff
@@ -52,6 +52,10 @@ - (void)performThreeDSecureLookup:(BTThreeDSecureRequest *)request
             requestParameters[@"dfReferenceId"] = request.dfReferenceId;
         }
 
+        if (request.accountTypeAsString) {
```
scannillo, line 55: Do we need the `accountTypeAsString` property? Is it easier to read having the logic be here like this:   

```
if (request.accountType) {
    case BTThreeDSecureAccountTypeCredit:
         requestParameters[@"accountType"] = @"credit";
    case BTThreeDSecureAccountTypeDebit:
         requestParameters[@"accountType"] = @"debit";
    default:
         requestParameters[@"accountType"] = nil;
}
```

sshropshire (approved):

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
sestevens: Yeah, I agree. Let's do that as a separate PR, though.

Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
sestevens: Yeah, I can see why that might seem simpler. Here's my rationale for the `accountTypeAsString` property.

1) It makes the testing story simpler. If we test `accountTypeAsString` independently, we don't need to add extra tests for `performThreeDSecureLookup` for each path through the switch statement.

2) This type of if-statement:

```objc
if (request.parameter) {
    requestParameters[@"param"] = request.parameter;
}
```

is the pattern we follow with all of the other properties, so throwing a switch statement in there might actually make things less readable.

3) This particular enum only has three cases, but if we add an enum for `shippingMethod`, it'll have a lot of cases. We'll want to handle both enums in the same way, and having two switch statements in the `performThreeDSecureLookup` method doesn't seem super readable to me. I think having `asString` methods for both is the better way to go.

scannillo (approved):

PR #588, opened 2021-01-22
Remove BTTokenizationService

sestevens:
### Summary of changes

`BTTokenizationService` was used in Drop-in to tokenize Venmo accounts if BraintreeVenmo was found at runtime. We are now requiring the BraintreeVenmo framework in Drop-in, so we can use `BTVenmoDriver` directly.

`BTCardClient` and `BTCardClient+UnionPay` were also registering with `BTTokenizationService`, but they are no longer being used through `BTTokenizationService` in the base SDK or Drop-in.

**NOTE:** `BTTokenizationService` was being used in the `BraintreeDemoCustomMultiPayViewController` demo integration. We already have separate demo view controllers for Card, Venmo and PayPal, so we think it's OK to remove this.

### Checklist

- [x] Added a changelog entry

- @sestevens 
- @scannillo


Demo/Application/Features/Venmo - Custom Button/BraintreeDemoCustomVenmoButtonViewController.m
```diff
@@ -14,7 +14,6 @@ - (void)viewDidLoad {
     [super viewDidLoad];
     self.venmoDriver = [[BTVenmoDriver alloc] initWithAPIClient:self.apiClient];
     self.title = NSLocalizedString(@"Custom Venmo Button", nil);
-    self.paymentButton.hidden = ![self.venmoDriver isiOSAppAvailableForAppSwitch];
```
sestevens, line 17: In order to maintain a UI test that checks for a message saying that the Venmo app is not installed, we needed to make the Venmo button always visible.

sshropshire (approved):

sarahkoop (approved):

PR #590, opened 2021-01-26
Add offerPayLater to BTPayPalRequest

demerino:
### Summary of changes
Add `offerPayLater` to `BTPayPalRequest`. This will default the customer to the Pay Later tab if they qualify during the Checkout experience.

### Checklist

- [x] Added a changelog entry

### Authors
- @demerino

### DIFF
--- a/Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalRequest.h
+++ b/Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalRequest.h
@@ -169,7 +169,7 @@ typedef NS_ENUM(NSInteger, BTPayPalRequestUserAction) {
 @property (nonatomic) BOOL offerCredit;
 
 /**
- Optional: Offers PayPal Pay Later if the customer qualifies. Defaults to false.
+ Optional: Offers PayPal Pay Later if the customer qualifies. Defaults to false. Only available with PayPal Checkout.
  */
 @property (nonatomic) BOOL offerPayLater;
 
--- a/Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalRequest.h
+++ b/Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalRequest.h
@@ -169,6 +169,11 @@ typedef NS_ENUM(NSInteger, BTPayPalRequestUserAction) {
 @property (nonatomic) BOOL offerCredit;
 
 /**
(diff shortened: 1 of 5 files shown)


demerino:
In another PR, I might remove `offerCredit` and rework the PayPal Credit demo into a Pay Later demo. I'll investigate if this is appropriate.

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalRequest.h
```diff
@@ -168,6 +168,11 @@ typedef NS_ENUM(NSInteger, BTPayPalRequestUserAction) {
  */
 @property (nonatomic) BOOL offerCredit;
 
+/**
+ Optional: Offers PayPal Pay Later if the customer qualifies. Defaults to false.
```
sestevens, line 172: Do we want to mention that this is only applicable to checkout flows (not billing agreement)?

sestevens: Or is it applicable to both but we only send the analytics event for checkout? Maybe I'm just confused by this code above:
```objc
if (paymentType == BTPayPalPaymentTypeCheckout && self.payPalRequest.offerPayLater) {
        NSString *eventName = [NSString stringWithFormat:@"ios.%@.webswitch.paylater.offered.%@", [self.class eventStringForPaymentType:paymentType], success ? @"started" : @"failed"];

        [self.apiClient sendAnalyticsEvent:eventName];
}
```

UnitTests/BraintreePayPalTests/BTPayPalDriver_Checkout_Tests.swift
```diff
@@ -552,6 +552,30 @@ class BTPayPalDriver_Checkout_Tests: XCTestCase {
         XCTAssertTrue(postedAnalyticsEvents.contains("ios.paypal-single-payment.webswitch.credit.offered.started"))
     }
```
scannillo, line 554: We should probably cover the case for defaulting to NO if `offer_pay_later` isn't set. 

Looks like we do that for `offer_paypal_credit` in `testCheckout_whenRemoteConfigurationFetchSucceeds_postsPaymentResourceWithShippingAddress`.

scannillo (approved):
⭐️

sestevens (approved):

PR #591, opened 2021-01-27
Update sdkVersion param for server-side 3DS lookup to include platform

sestevens:
### Summary of changes

- Update sdkVersion param for server-side 3DS lookup to include platform as well as version. This fixes a discrepancy observed between client-side and server-side lookups.
- We also discovered that client-side lookup requests were missing the requested version param, so we're adding it now.

### Checklist

- ~Added a changelog entry~

- @sestevens 
- @scannillo


sarahkoop (approved):

sestevens:
We may also want to start sending the requested 3DS version in client-side lookup requests. Waiting to hear back on this...

sshropshire (approved):
🌊

PR #593, opened 2021-01-27
Make BTPaymentMethodNonceParser private

sestevens:
### Summary of changes

- We'd like to make some changes to this class in the future, and potentially remove the need for payment clients/drivers to register with `BTPaymentMethodNonceParser` in their `load` methods. In order to do that without a breaking change, we're making this class private.

### Checklist

- [x] Added a changelog entry

- @sestevens 
- @scannillo


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Breaking Changes
   * Make `shippingMethod` property on `BTThreeDSecureRequest` an enum instead of a string
   * Remove `BTTokenizationService`
+  * Make `BTPaymentMethodNonceParser` private
```
sarahkoop, line 7: What do you all think of listing classes made private as "removed" in the CHANGELOG? That's what we've been doing in Android, since from an integration perspective the class does not exist anymore.

sarahkoop (approved):

CHANGELOG.md
scannillo: Yeh - we talked about this for a bit. We were thinking it could potentially confuse our future-selves/ SDK team members if we say "remove" instead of "make private".

demerino (approved):

PR #599, opened 2021-02-02
Checkout with vault

demerino:
### Summary of changes
- Add `requestBillingAgreement` to `BTPayPalRequest`
- Update usage of `billingAgreementDescription` on `BTPayPalRequest`. Opted to reuse this property when using `requestBillingAgreement` and updated the description, rather than making another property that does the same thing.

### Checklist

- [x] Added a changelog entry

### Authors
- @demerino

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalDriver.m
+++ b/Sources/BraintreePayPal/BTPayPalDriver.m
@@ -158,6 +158,8 @@ NSString * _Nonnull const PayPalEnvironmentMock = @"mock";
 
         parameters[@"offer_paypal_credit"] = @(request.offerCredit);
 
+        parameters[@"offer_pay_later"] = @(request.offerPayLater);
+
         experienceProfile[@"no_shipping"] = @(!request.isShippingAddressRequired);
 
         experienceProfile[@"brand_name"] = request.displayName ?: [configuration.json[@"paypal"][@"displayName"] asString];
@@ -618,6 +620,12 @@ NSString * _Nonnull const PayPalEnvironmentMock = @"mock";
 
     if ((paymentType == BTPayPalPaymentTypeCheckout || paymentType == BTPayPalPaymentTypeBillingAgreement) && self.payPalRequest.offerCredit) {
         NSString *eventName = [NSString stringWithFormat:@"ios.%@.webswitch.credit.offered.%@", [self.class eventStringForPaymentType:paymentType], success ? @"started" : @"failed"];
+
+        [self.apiClient sendAnalyticsEvent:eventName];
+    }
+
+    if (paymentType == BTPayPalPaymentTypeCheckout && self.payPalRequest.offerPayLater) {
+        NSString *eventName = [NSString stringWithFormat:@"ios.%@.webswitch.paylater.offered.%@", [self.class eventStringForPaymentType:paymentType], success ? @"started" : @"failed"];
 
         [self.apiClient sendAnalyticsEvent:eventName];
     }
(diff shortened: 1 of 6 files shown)


scannillo:
This seems a bit confusing since we have a `requestBillingAgreement` public method on `BTPayPalDriver.h` [here](https://github.com/braintree/braintree_ios/blob/4c47ffd41cf1c595a88efe6986b1ecfed3602d9b/Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalDriver.h#L102-L114).

Can you clarify when we'd expect merchants to use one over the other?

demerino:
This is to enable the Checkout With Vault flow. Maybe I should change `requestBillingAgreement` -> `checkoutWithVault`.

Sources/BraintreePayPal/BTPayPalDriver.m
```diff
@@ -152,6 +152,14 @@ - (void)requestPayPalCheckout:(BTPayPalRequest *)request
             if (request.amount != nil) {
                 parameters[@"amount"] = request.amount;
             }
+            if (request.requestBillingAgreement) {
+                parameters[@"request_billing_agreement"] = @(request.requestBillingAgreement);
+                if (request.billingAgreementDescription.length > 0) {
+                    NSMutableDictionary *billingAgreementDetails = [NSMutableDictionary dictionary];
+                    billingAgreementDetails[@"description"] = request.billingAgreementDescription;
+                    parameters[@"billing_agreement_details"] = billingAgreementDetails;
```
billwerges, line 160: This block could be simplified down to
`parameters[@"billing_agreement_details"] = @{@"description": request.billingAgreementDescription};`

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalRequest.h
```diff
@@ -190,6 +190,11 @@ typedef NS_ENUM(NSInteger, BTPayPalRequestUserAction) {
  */
 @property (nonatomic, nullable, strong) UIWindow *activeWindow;
 
+/**
+ Optional: If `true`, the customer will be prompted to consent to a billing agreement during PayPal Checkout flows.
+ */
+@property (nonatomic) BOOL requestBillingAgreement;
```
sestevens, line 196: It looks like there are a [few](https://github.com/braintree/braintree_ios/blob/0c3860281da41f3b95eca3556ae65d3f0f7eb901/Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalRequest.h#L142) [other](https://github.com/braintree/braintree_ios/blob/0c3860281da41f3b95eca3556ae65d3f0f7eb901/Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalRequest.h#L147) [properties](https://github.com/braintree/braintree_ios/blob/0c3860281da41f3b95eca3556ae65d3f0f7eb901/Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalRequest.h#L174) that only apply to one of the two PayPal flows. I don't know if we'd want to do this refactor now, but I wonder if the ideal state would be to have a `BTPayPalRequest` base class, and two subclasses: `BTPayPalCheckoutRequest` and `BTPayPalBillingAgreementRequest`. That way, it would be a bit more clear which properties apply to only one flow and which apply to both.

demerino: 🤔 I like that

sestevens:
We're going to add checkout and vault subclasses for `BTPayPalRequest`. Let's hold off on merging this until that work is done.

sestevens:
Closing in favor of #619

PR #603, opened 2021-02-03
Add environment property to BTConfiguration

sestevens:
### Summary of changes

- Add `environment` property to `BTConfiguration` so that we don't have to manually parse the config json to find this field

### Checklist

- [x] Added a changelog entry

- @sestevens 
- @scannillo


sshropshire (approved):
🌊

sarahkoop (approved):

PR #595, opened 2021-02-01
[do not merge] v5 GA audit

demerino:
### Summary of changes
A discussion for additional cleanup.

### Authors
- @demerino

PR #606, opened 2021-02-04
App switch clean up

scannillo:
### Summary of changes

- Addresses feedback in #595
- Removes `BTAppSwitchDelegate`
- Renames:
    - `BTAppSwitch` -> `BTAppContextSwitcher`
    -  `handleAppSwitchReturnURL()` -> `handleReturnURL()`
    -  `canHandleAppSwitchReturnURL()` -> `canHandleReturnURL()`
    -  `BTAppSwitchHandler` -> `BTAppContextSwitchDriver`
- Clarify docstring comments

### Checklist

- [x] Added a changelog entry
- [x] Update Migration guide

- @scannillo 
- @sestevens


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
   * Make `shippingMethod` property on `BTThreeDSecureRequest` an enum instead of a string
   * Remove `BTTokenizationService`
   * Make `BTPaymentMethodNonceParser` private
+  * Remove `BTAppSwitchDelegate`
```
sestevens, line 8: This delegate was something merchants could use to be notified when the SDK was about to open an `SFSafariViewController` (for 3DS and local payments) or switch to the Venmo app. We don't think this is necessary at this point. 

For the `SFSafariViewController`, the merchant is responsible for presenting it in their `BTViewControllerPresentingDelegate` methods, so they already know that it's being presented.

And for switching to the Venmo app, it's more reliable to use app delegate methods to detect when their app is transitioning from foreground to background and vice versa.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
...
+  * Rename `BTAppSwitch` to `BTAppContextSwitcher`
```
sestevens, line 9: We tried to come up with a name that would indicate that this class handles app switches as well as browser switches (i.e., returning from `SFSafariViewController`). If anyone has other suggestions, we'd love to hear them!

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
...
+    * Rename `handleAppSwitchReturnURL()` to `handleReturnURL()`
+    * Rename `canHandleAppSwitchReturnURL()` to `canHandleReturnURL()`
+    * Remove `unregisterAppSwitchHandler()`
```
sestevens, line 12: The `unregister` method was never used.

Sources/BraintreeCore/Public/BraintreeCore/BTAppContextSwitcher.h
```diff
@@ -0,0 +1,85 @@
...
+*/
++ (void)setReturnURLScheme:(NSString *)returnURLScheme;
+
+/**
+ Handles a return from app context switch
+
+ @param url The URL that was opened to return to your app
+ @return `YES` if the app switch successfully handled the URL, or `NO` if the attempt to handle the URL failed.
+*/
++ (BOOL)handleOpenURL:(NSURL *)url; // TODO: implement proper swift name
```
sestevens, line 68: Whoops, I need to handle this TODO...

Sources/BraintreeCore/Public/BraintreeCore/BTAppContextSwitcher.h
```diff
@@ -0,0 +1,85 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+#pragma mark - BTAppContextSwitchDriver protocol
+
+/**
+ A protocol for handling the return from gathering payment information from a browser or another app.
+ @note The app context may switch to a SFSafariViewController or to a native app, such as Venmo.
+*/
+@protocol BTAppContextSwitchDriver
```
sestevens, line 11: We went with "driver" rather than "handler" in this name since the classes that implement this protocol are both drivers: `BTVenmoDriver` and `BTPaymentFlowDriver`.

Sources/BraintreeVenmo/BTVenmoDriver.m
```diff
@@ -208,21 +202,19 @@ - (void)vaultVenmoAccountNonce:(NSString *)nonce {
 #pragma mark - App switch
 
 - (BOOL)isiOSAppAvailableForAppSwitch {
-    BOOL isAtLeastIos9 = ([[self.device systemVersion] intValue] >= 9);
```
sestevens, line 211: No need to check for iOS 9 since we only support iOS 12+ now.

UnitTests/BraintreeCoreTests/BTAppContextSwitch_Tests.swift
```diff
@@ -0,0 +1,111 @@
+import XCTest
+
+class BTAppContextSwitch_Tests: XCTestCase {
```
sestevens, line 3: The diff shows this as a new file, but we actually just renamed `BTAppSwitch_Tests.swift`.

UnitTests/BraintreeTestShared/MockViewControllerPresentingDelegate.swift
```diff
@@ -0,0 +1,21 @@
+import XCTest
+import BraintreeCore
+
+@objc public class MockViewControllerPresentingDelegate : NSObject, BTViewControllerPresentingDelegate {
```
sestevens, line 4: This originally lived in `MockDelegates.swift`, which we renamed to `MockViewControllerPresentingDelegate.swift`. After deleting the `MockAppSwitchDelegate`, there was only one class left in the file, so it made sense for the file name to match the class.

Sources/BraintreeCore/Public/BraintreeCore/BTAppContextSwitcher.h
```diff
@@ -0,0 +1,85 @@
...
+*/
++ (void)handleReturnURL:(NSURL *)url NS_SWIFT_NAME(handleReturnURL(_:));
+
+@end
+
+#pragma mark - BTAppContextSwitcher
+
+/**
+ Handles return URLs when returning from app context switch and routes the return URL to the correct app context switch driver class.
+ @note `returnURLScheme` must contain your app's registered URL Type that starts with the app's bundleID. When your app returns from app switch, the app delegate should call `handleOpenURL:` or `handleOpenURLContext:`
```
sarahkoop, line 36: Nitpick we use `bundleID` here and `bundle ID` in line 43.

V5_MIGRATION.md
```diff
@@ -105,18 +105,38 @@ BTAppSwitch.setReturnURLScheme("com.your-company.your-app.payments")
...
 If your app supports multi-tasking, you must set the `BTPayPalRequest.activeWindow` property to ensure that the PayPal flow launches from the correct window.
 
-## App Switch
+## App Context Switching
 
-v5 removes the `options` and `sourceApplication` params on methods in `BTAppSwitch`. 
+v5 renames the `BTAppSwitch` class to `BTAppContextSwitcher` to clarify that it is used for flows that requiring switching contexts, either by opening an `SFSafariViewController` or by opening a different app (specifically, Venmo).
 
-If you're using `UISceneDelegate`, you don't need to make any code changes. 
+`BTAppSwitchDelegate` was removed in v5. If you were using these delegate methods to determine when control switched between your app and the Venmo app, we recommend using app or scene delegate methods instead. If you were `BTAppSwitchDelegate` to determine when an `SFSafariViewController` was presented or dismissed, we recommend using the `BTViewControllerPresentingDelegate` methods instead.
```
sarahkoop, line 112:
```suggestion
`BTAppSwitchDelegate` was removed in v5. If you were using these delegate methods to determine when control switched between your app and the Venmo app, we recommend using app or scene delegate methods instead. If you were using `BTAppSwitchDelegate` to determine when an `SFSafariViewController` was presented or dismissed, we recommend using the `BTViewControllerPresentingDelegate` methods instead.
```

V5_MIGRATION.md
```diff
@@ -105,18 +105,38 @@ BTAppSwitch.setReturnURLScheme("com.your-company.your-app.payments")
...
+  return true
+}
+```
 
+If you're using `UISceneDelegate`, use the following code to pass a return URL to `BTAppContextSwitcher`:
+
+```swift
+func scene(_ scene: UIScene, openURLContexts URLContexts: Set<UIOpenURLContext>) {
+  URLContexts.forEach { context in
+    if context.url.scheme?.localizedCaseInsensitiveCompare("com.my-app.your-app.payments") == .orderedSame {
```
sarahkoop, line 127:
```suggestion
    if context.url.scheme?.localizedCaseInsensitiveCompare("com.your-company.your-app.payments") == .orderedSame {
```

sarahkoop (approved):
Looks good to me 🌟

CHANGELOG.md
hollabaq86: `BTSwitchAllTheThings` 🤣  tbh I think what you have is the best. Just spent five minutes seriously considering other names and I don't like any of my ideas

hollabaq86 (approved):

scannillo:
Not sure why the unit tests are failing for something that was fixed. Maybe it's doing some caching of files that weren't changed? We can ignore this failure right now.

PR #609, opened 2021-02-06
Add BTPayPalCheckoutRequest

sestevens:
### Summary of changes

- Add `BTPayPalCheckoutRequest`, which subclasses `BTPayPalRequest` and contains all properties that apply only to checkout (aka one-time payment) flows

_Follow-up PRs will include:_
- Adding `BTPayPalVaultRequest`
- Deprecating `requestBillingAgreement` and `requestOneTimePayment` methods
- Adding `tokenizePayPalAccount` method
- Updating the migration guide

### Checklist

- [x] Added a changelog entry

- @sestevens


Sources/BraintreePayPal/BTPayPalCheckoutRequest.m
```diff
@@ -0,0 +1,55 @@
...
+
+    if (self = [super init]) {
+        _amount = amount;
+        _offerPayLater = NO;
+        _intent = BTPayPalRequestIntentAuthorize;
+    }
+    return self;
+}
+
+- (NSDictionary<NSString *, NSObject *> *)parametersWithConfiguration:(BTConfiguration *)configuration isBillingAgreement:(BOOL)isBillingAgreement {
```
sestevens, line 19: The `isBillingAgreement` param will go away once we add the `BTPayPalVaultRequest` subclass.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.m
```diff
@@ -0,0 +1,55 @@
+#import "BTPayPalCheckoutRequest_Internal.h"
+#import "BTPayPalRequest_Internal.h"
+
+@implementation BTPayPalCheckoutRequest
+
+- (instancetype)initWithAmount:(NSString *)amount {
```
sarahkoop, line 6: This may be outside the scope of this PR, but what do you think about removing `amount` from the constructor and setting it in the same way as the other parameters?

sarahkoop, line 6: This may be outside the scope of this PR, but what do you think about removing `amount` from the constructor and setting it in the same way as the other parameters?

sestevens: Hmm.... I think the advantage of having it as part of the initializer is that you can't forget to set it. It's the only required field for this class, so it makes sense to me for the code to enforce that requirement.

That being said, it is a bit weird that this one field is different from all the rest.

sarahkoop: Just wanted to get thoughts on it because in Android we took it out of the constructor in favor of matching the way other fields are set.  Since it's the only required field it also makes sense to enforce setting it.

sarahkoop (approved):

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalCheckoutRequest.h
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ Base initializer - do not use.
+ */
+- (instancetype)init __attribute__((unavailable("Please use initWithAmount:")));
+
+/**
+ Initialize a PayPal request with an amount for a one-time payment.
+
+ @param amount Used for a one-time payment. Amount must be greater than or equal to zero, may optionally contain exactly 2 decimal places separated by '.', optional thousands separator ',', and is limited to 7 digits before the decimal point.
```
hollabaq86, line 40: I wonder if we include an example amount for these descriptions?

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalCheckoutRequest.h
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ Used for a one-time payment.
+
+ Amount must be greater than or equal to zero, may optionally contain exactly 2 decimal places separated by '.', optional thousands separator ',', and is limited to 7 digits before the decimal point.
+*/
+@property (nonatomic, readonly, strong) NSString *amount;
+
+/**
+ Optional: A valid ISO currency code to use for the transaction. Defaults to merchant currency code if not set.
```
hollabaq86, line 53: do we know which ISO standard this follows? Is it 2 or three characters? We should include this detail with an example

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalCheckoutRequest.h
```diff
@@ -0,0 +1,70 @@
...
+ @note This is only used for one-time payments.
```
hollabaq86, line 54:
```suggestion
 @note This can only be used for one-time payments/checkout flows.
```

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalCheckoutRequest.h
```diff
@@ -0,0 +1,70 @@
...
+*/
+@property (nonatomic, nullable, copy) NSString *currencyCode;
+
+/**
+ Optional: Payment intent. Only applies when using checkout flow. Defaults to `BTPayPalRequestIntentAuthorize`.
```
hollabaq86, line 59:
```suggestion
 Optional: Payment intent. Only applies when using one-time payment/checkout flow. Defaults to `BTPayPalRequestIntentAuthorize`.
```

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalCheckoutRequest.h
```diff
@@ -0,0 +1,70 @@
...
+*/
+@property (nonatomic) BTPayPalRequestIntent intent;
+
+/**
+ Optional: Offers PayPal Pay Later if the customer qualifies. Defaults to false. Only available with PayPal Checkout.
```
hollabaq86, line 64:
```suggestion
 Optional: Offers PayPal Pay Later if the customer qualifies. Defaults to false. Only available with one-time payment flows/PayPal Checkout.
```

hollabaq86: we should make sure all these descriptions use the same snippet to describe PP checkout/ OTP flows hehe

hollabaq86 (approved):

Sources/BraintreePayPal/BTPayPalCheckoutRequest.m
hollabaq86: 🤔  it would be nice if iOS and Android were in parity

sshropshire: Yeah we removed the amount parameter on Android because Billing Agreements don't require an amount. It's definitely required for one-time / checkout requests, but does that mean the constructor for `BTPayPalVaultRequest` would have zero parameters?

sshropshire (approved):
🌊

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalRequest.h
```diff
@@ -63,21 +44,6 @@ typedef NS_ENUM(NSInteger, BTPayPalRequestUserAction) {
 */
 @interface BTPayPalRequest : NSObject
```
sshropshire, line 45: question: Is it possible for merchants to instantiate the base class?

Sources/BraintreePayPal/BTPayPalCheckoutRequest.m
sestevens: Yep!

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalRequest.h
sestevens: Yeah. I'm not sure how to make the initializer available to the SDK without it also being available to merchants. It's something I can play around with, though.

sshropshire: Yeah it was a thought. Obj-C doesn't have an abstract class concept though this makes sense.

sestevens: Yeah, an abstract class would be perfect here

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalCheckoutRequest.h
sestevens: Looks like we have more info in the [Braintree docs](https://developers.braintreepayments.com/guides/paypal/checkout-with-paypal/ios/v4#currency-presentment), which link out to the [PayPal docs](https://developer.paypal.com/docs/api/reference/currency-codes/). I can include the PayPal docs link here and specify that it's three-character ISO-4217.

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalCheckoutRequest.h
sestevens: Sounds good! I'll look through all the doc comments and make them consistent.

sestevens: I'm going to open a separate PR with these updates to the doc strings: #615

sestevens:
The failing unit tests are passing locally, sigh. I'm going to go ahead and merge this. Let's do more investigation into the CI failures soon.

PR #610, opened 2021-02-08
SPM Analytics Metadata

scannillo:
### Summary
Add Analytics metadata for if Swift Package Manager is used.

### Note
It would be nice to have an `iosPackageManager` field that returns a string `CocoaPods`, `Swift Package Manager`, or `Carthage`. However, there is no way to differentiate between a Carthage integration & a manual integration. 

We could leave as-is, or instead have an `iosPackageManager` field with the options of:
- `CocoaPods`
- `Swift Package Manager`
- `Carthage or Manual integration`

### Checklist

- ~Added a changelog entry~

### DIFF
--- a/Sources/BraintreeCore/BTAnalyticsMetadata.m
+++ b/Sources/BraintreeCore/BTAnalyticsMetadata.m
@@ -2,14 +2,6 @@
 #import "Braintree-Version.h"
 #import "BTKeychain.h"
 
-#if __has_include(<Braintree/BraintreePaymentFlow.h>) // CocoaPods
-#import <Braintree/BTLogger_Internal.h>
-#elif SWIFT_PACKAGE // SPM
-#import "../BraintreeCore/BTLogger_Internal.h"
-#else // Carthage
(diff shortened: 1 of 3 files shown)


sestevens:
I think we should have one field with three possible values: `cocoapods`, `spm`, `carthage/manual`. Just my 2 cents!

scannillo:
> @sestevens  I think we should have one field with three possible values: cocoapods, spm, carthage/manual. Just my 2 cents!

I am down for that. 

Do we think somewhere/someone at BT is using `iosIsCocoaPods` and that our removing it would break anything? The only scenario I can think of is people's SUMO searches.

sestevens:
Hmm... I'd be surprised if removing that field broke anything.

scannillo:
Update made 👍 . We couldn't find a way to have a unit test test for each package string returned. We did manually test in our SPM sample app and saw the correct value.

sestevens (approved):
Looks good to me!

sshropshire (approved):

PR #619, opened 2021-02-10
PayPal Checkout with Vault (Take 2)

sestevens:
### Summary of changes

- Add `requestBillingAgreement` to `BTPayPalCheckoutRequest`
- Move `billingAgreementDescription` from `BTPayPalVaultRequest` to `BTPayPalRequest`

### Checklist

- [x] Added a changelog entry

- @sestevens

### DIFF
--- a/Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalCheckoutRequest.h
+++ b/Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalCheckoutRequest.h
@@ -65,6 +65,11 @@ typedef NS_ENUM(NSInteger, BTPayPalRequestIntent) {
  */
 @property (nonatomic) BOOL offerPayLater;
 
+/**
+ Optional: If set to true, this enables the Checkout with Vault flow, where the customer will be prompted to consent to a billing agreement during checkout.
+ */
+@property (nonatomic) BOOL requestBillingAgreement;
+
 @end
 
 NS_ASSUME_NONNULL_END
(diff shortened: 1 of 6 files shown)


Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalCheckoutRequest.h
```diff
@@ -65,6 +65,11 @@ typedef NS_ENUM(NSInteger, BTPayPalRequestIntent) {
  */
 @property (nonatomic) BOOL offerPayLater;
 
+/**
+ Optional: If set to true, this enables the Checkout with Vault flow, where the customer will be prompted to consent to a billing agreement during checkout.
+ */
+@property (nonatomic) BOOL requestBillingAgreement;
```
sestevens, line 71: There was some discussion in the previous PR about whether to call this something else, like `checkoutWithVault`. I went with `requestBillingAgreement` since that's what we use on JS, but I'm curious what other people think makes the most sense.

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalVaultRequest.h
```diff
@@ -8,11 +8,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface BTPayPalVaultRequest : BTPayPalRequest
 
-/**
- Optional: Display a custom description to the user for a billing agreement.
-*/
-@property (nonatomic, nullable, copy) NSString *billingAgreementDescription;
```
sestevens, line 14: It's a little weird that this class doesn't have any public properties now that `billingAgreementDescription` has been moved to `BTPayPalRequest`. But I think it still makes sense to keep the separate class since it makes our public API a little nicer and we might have things to add here in the future.

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalCheckoutRequest.h
sshropshire: I like that it lines up with JS. An alternative could be `checkoutWithVaultEnabled`, but SDK parity feels like the best call.

sshropshire (approved):

demerino (approved):

PR #621, opened 2021-02-10
Update Cardinal to 2.2.5

sestevens:
### Summary of changes

- Update Cardinal frameworks in the SDK and SPMTest sample app to 2.2.5
- Update Cartfile (_note: We're in the process of updating the hosted version of Cardinal, so this won't work quite yet_)

### Checklist

- [x] Added a changelog entry

- @hollabaq86 
- @sestevens


sarahkoop (approved):

scannillo (approved):
Nice! Also might be useful if we update our knowledge repo to point out the places that the iOS repo requires updating Cardinal since now it is more than just in the SDK 🙃

sestevens:
Good call! We just updated it. 

We might also want to update the Carthage job on GH Actions to run the CarthageTest app, instead of just running `carthage build`. I was expecting the job to fail until Cardinal v2.2.5 was hosted, but it's passing since `carthage build` doesn't actually pull down our SDK's dependencies.

demerino (approved):

PR #622, opened 2021-02-11
Add release.yml

sestevens:
### Summary of changes

- Add GitHub Actions release workflow

### Checklist

- [ ] Add token for CocoaPods ([like this](https://medium.com/swlh/automated-cocoapod-releases-with-github-actions-8526dd4535c7))
- ~Added a changelog entry~

- @sestevens

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,80 @@
+name: Release
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to release"
+        required: true
+jobs:
+  release:
+    name: Release
+    runs-on: macOS-latest
+    steps:
+      - name: Check out repository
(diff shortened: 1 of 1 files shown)


scannillo:
Are we planning to add running the tests and building each demo app to this process?

scannillo:
Also - did we find out if it is possible to include a release to GHE as well?

sestevens:
> Are we planning to add running the tests and building each demo app to this process?

That's a good question. I think it should be a separate workflow. One idea I have is to update build.yml and test.yml so that those workflows can be run manually. Then we can run them before running the release workflow. Does that sound good?

> Also - did we find out if it is possible to include a release to GHE as well?

I haven't investigated that yet. I think we can do that manually for now and update the script later.

PR #623, opened 2021-02-11
Update v5 changelog to include links to betas

sestevens:
### Summary of changes

- Add links to v5 betas at the end of the v5 CHANGELOG entry so that it's clear that everything in the betas is included in v5.

### Checklist

- ~Added a changelog entry~

- @sestevens 
- @scannillo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@
   * Add `offerPayLater` and `requestBillingAgreement` to `BTPayPalCheckoutRequest`
 * Update CardinalMobile.framework to v2.2.5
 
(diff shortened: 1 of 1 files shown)


sarahkoop (approved):
Looks good. I know there is now a step in the release process to extract CHANGELOG entries into the release notes, I think this note should/will be included there as well, but will the links work correctly in the release notes?

sestevens:
> Looks good. I know there is now a step in the release process to extract CHANGELOG entries into the release notes, I think this note should/will be included there as well, but will the links work correctly in the release notes?

Ah, good question. I'll manually add this note to the release notes with the full links.

PR #629, opened 2021-02-12
Update unit tests to use MockAPIClient

sestevens:
### Summary of changes

I noticed that a lot of the unit tests that randomly fail (especially in CI) are using `BTAPIClient` with a mock `BTHTTP` when they could be using `MockAPIClient`. Since `BTAPIClient` kicks off a configuration request in the initializer, I have a hunch that there could be a race condition that causes tests to sporadically fail.

This PR updates unit tests to use `MockAPIClient` where possible. It doesn't address the sporadic failures in the unit tests for `BTAPIClient` itself.

### Checklist

- ~Added a changelog entry~

- @sestevens

### DIFF
--- a/UnitTests/BraintreeVenmoTests/BTVenmoDriver_Tests.swift
+++ b/UnitTests/BraintreeVenmoTests/BTVenmoDriver_Tests.swift
@@ -486,26 +486,7 @@ class BTVenmoDriver_Tests: XCTestCase {
 
     // Note: testing of handleReturnURL is done implicitly while testing authorizeAccountWithCompletion
 
-    // MARK: - Drop-in
-
-    /// Helper
-    func client(_ configurationDictionary: Dictionary<String, String>) -> BTAPIClient {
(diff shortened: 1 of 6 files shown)


UnitTests/BraintreeVenmoTests/BTVenmoDriver_Tests.swift
```diff
@@ -486,26 +486,7 @@ class BTVenmoDriver_Tests: XCTestCase {
 
     // Note: testing of handleReturnURL is done implicitly while testing authorizeAccountWithCompletion
 
-    // MARK: - Drop-in
-
-    /// Helper
-    func client(_ configurationDictionary: Dictionary<String, String>) -> BTAPIClient {
```
sestevens, line 492: These helper methods were never used.

sarahkoop (approved):

scannillo (approved):

PR #630, opened 2021-02-12
Update GH Actions Test and Build Workflows

sestevens:
### Summary of changes

- Use actions/checkout@v2 instead of v1
- Use Xcode 12.4 instead of 12.3
- Update Carthage check to build the CarthageTest demo app. This is a more thorough check since it pulls down our dependencies (CardinalMobile and PPRiskMagnes).
- Remove `--allow-warnings` flag from `pod lib lint` so that it fails if there are warnings
- Allow Test and Build workflows to be run manually as well as on pull requests. This way we can run them manually prior to releasing.

### Checklist

- ~Added a changelog entry~

- @sestevens

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,38 +1,49 @@
 name: Build
-on: [pull_request]
+on: [pull_request, workflow_dispatch]
 jobs:
   cocoapods:
     name: CocoaPods
     runs-on: macOS-latest
     steps:
       - name: Check out repository
-        uses: actions/checkout@v1
-      - name: Use Xcode 12.3
-        run: sudo xcode-select -switch /Applications/Xcode_12.3.app
+        uses: actions/checkout@v2
+      - name: Use Xcode 12.4
+        run: sudo xcode-select -switch /Applications/Xcode_12.4.app
       - name: Install CocoaPod dependencies
         run: pod install
       - name: Run pod lib lint
-        run: pod lib lint --allow-warnings
+        run: pod lib lint
   carthage:
     name: Carthage
     runs-on: macOS-latest
     steps:
       - name: Check out repository
-        uses: actions/checkout@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -1,38 +1,49 @@
...
-      - name: Run carthage build
-        run: carthage build --no-skip-current
+        run: |
+          git checkout $GITHUB_HEAD_REF
+          rm -rf SampleApps/SPMTest
+          git add SampleApps/SPMTest
+          git commit -m 'Remove SPMTest app to avoid Carthage timeout'
+      - name: Use current branch
+        run: |
+          echo 'git "file://'$PWD'" "'$GITHUB_HEAD_REF'"' > SampleApps/CarthageTest/Cartfile
```
scannillo, line 34: Does this need to be two lines?

.github/workflows/build.yml
```diff
@@ -1,38 +1,49 @@
...
+      - name: Run carthage update
+        run: cd SampleApps/CarthageTest && carthage update
```
scannillo, line 36: This could be a separate PR - but maybe we can require our repo uses [Carthage 0.37.0](https://github.com/Carthage/Carthage/releases/tag/0.37.0), and we can get rid of that `carthage.sh` script (which is only used in the Rakefile at this point).

.github/workflows/build.yml
sestevens: I can make it one

sestevens:
```suggestion
        run: echo 'git "file://'$PWD'" "'$GITHUB_HEAD_REF'"' > SampleApps/CarthageTest/Cartfile
```

scannillo (approved):

sshropshire (approved):

PR #637, opened 2021-02-23
Fix Carthage timeout 

scannillo:
### Summary of changes

- Updated Rakefile & GH Actions `release.yml` to remove the SPMTest app before tagging the new SDK version.
    - Removing the SPM test app is necessary, or else Carthage users trying to use the new `--use-xcframeworks` flag will get a timeout.
    - After we've removed the SPMTest app, tagged, and pushed, we recommit the SPMTest app to continue to use for internal development purposes.
    - See this [_Setting an Environment Variable_ ](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable) section of GH docs.
    
### Authors
@scannillo


.github/workflows/release.yml
```diff
@@ -29,6 +29,14 @@ jobs:
           plutil -replace CFBundleVersion -string ${{ github.event.inputs.version }} -- 'Demo/Application/Supporting Files/Braintree-Demo-Info.plist'
           plutil -replace CFBundleVersion -string ${{ github.event.inputs.version }} -- 'Sources/BraintreeCore/Info.plist'
 
+      # Remove the SPMTest app to prevent Carthage timeout; store SPMTest app in temp dir
```
sestevens, line 32: Maybe we should do this before the `Update version` step. `Update version` and `Tag and push new version` kind of go together, since once is updating a bunch of files and the other is committing those changes.

.github/workflows/release.yml
```diff
@@ -29,6 +29,14 @@ jobs:
...
+      - name: Remove SPMTest demo app
+        run: |
+          echo "TEMP_DIR=$(mktemp -d)" >> $GITHUB_ENV
+          mv SampleApps/SPMTest/ ${{ env.TEMP_DIR }}/
+          git add SampleApps/SPMTest
+          git commit -m 'Remove SPMTest app to avoid Carthage timeout'
```
sestevens, line 38: Would it make sense for the commit message to indicate that this is part of the release process? Maybe it could be:

```
Pre-release: Remove SPMTest app to avoid Carthage timeout
```

.github/workflows/release.yml
```diff
@@ -39,6 +47,14 @@ jobs:
           git tag ${{ github.event.inputs.version }} -a -m 'Release ${{ github.event.inputs.version }}'
           git push origin tag ${{ github.event.inputs.version }}
 
+      # Restore the SPMTest app for development purposes
+      - name: Restore SPMTest demo app
+        run: |
+          mv ${{ env.TEMP_DIR }}/SPMTest/ SampleApps/
+          git add SampleApps/SPMTest
+          git commit -m 'Restore SPMTest app for development'
```
sestevens, line 55: Similarly, this could be `Post-release: Restory SPMTest app for development`.

sshropshire (approved):
🌊

Rakefile
```diff
@@ -109,11 +109,28 @@ end
...
+  # Restore SPMTest app to prevent Carthage timeout
+  task :restore_spm_test_app do
+    run! "mv temp/SPMTest/ SampleApps/"
+    run! "rm -rf temp/"
+    run! "git add SampleApps/SPMTest"
+    run! "git commit -m 'Restore SPMTest app for development'"
+  end
+
+  task :build_demo do
+    remove_spm_test_app
```
sshropshire, line 133: random: Does `git reset --hard HEAD^` at the end of this task prevent us from having to call `restore_spm_test_app`?

scannillo: Exactly! For just building the demo_app, we don't want to commit the change and instead just RESET it.

.github/workflows/release.yml
scannillo: DONE

.github/workflows/release.yml
scannillo: DONE

.github/workflows/release.yml
scannillo: DONE

sestevens (approved):

PR #639, opened 2021-02-26
Allow multiple versions of the reference docs

sestevens:
### Summary of changes

- Update Rakefile and release.yml steps for publishing reference docs to allow multiple version to exist at once

I went ahead and published docs for [4.37.0](https://braintree.github.io/braintree_ios/4.37.0/), so they now live alongside the docs for [5.0.0](https://braintree.github.io/braintree_ios/5.0.0/). Ideally, there would be a drop-down allowing you to toggle between versions, but currently you have to change the version in the URL.

[Related PR in iOS Drop-in](https://github.com/braintree/braintree-ios-drop-in/pull/273)

### Checklist

- ~Added a changelog entry~

- @sestevens

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,8 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Use Xcode 12.4
         run: sudo xcode-select -switch /Applications/Xcode_12.4.app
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -12,6 +12,8 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
sestevens, line 16: This allows us to checkout the `gh-pages` branch later in the workflow.

sshropshire (approved):

scannillo (approved):

PR #649, opened 2021-03-18
Add CardinalMobile.xcframework version 2.2.5-1

sestevens:
### Summary of changes

- Add `CardinalMobile.xcframework` and update `Package.swift` and `Braintree.podspec` to use it
- Update SPM instructions and v5 migration guide
- Update SPMTest app to use `CardinalMobile.xcframework`

I tested this by updating our SPMTest app to go through a basic 3DS2 flow. I was also able to successfully archive and upload the SPMTest app to Test Flight.

**NOTE:** We can also remove the code in `BraintreeThreeDSecure` that checks for CardinalMobile at runtime, which is what enabled our SPM work-around. But I think we should do that in a separate PR.

### Checklist

- [x] Added a changelog entry

- @sestevens


V5_MIGRATION.md
```diff
@@ -64,34 +64,22 @@ In v4, 3D Secure classes were housed in the `BraintreePaymentFlow` module. In v5
 
 #### Integration
 
-<details><summary>CocoaPods</summary>
-<p>
+##### CocoaPods
```
sestevens, line 67: We added the collapsable sections for each package manager because the SPM instructions were initially very long. Now that we only have a sentence of two for each, I think we can get rid of the collapsable sections.

sshropshire (approved):

SWIFT_PACKAGE_MANAGER.md
```diff
@@ -44,7 +46,7 @@ Make sure to select your app's target in the _Provide build settings from_ drop-
 
 ### BraintreeThreeDSecure
 
-Currently, to use the `BraintreeThreeDSecure` library via SPM, you must manually include `CardinalMobile.framework`. Once we receive an xcframework version of CardinalMobile, these steps will no longer be required.
+~Currently, to use the `BraintreeThreeDSecure` library via SPM, you must manually include `CardinalMobile.framework`. Once we receive an xcframework version of CardinalMobile, these steps will no longer be required.~ These steps are only required for versions 5.0.0 to 5.2.0. For versions _greater than_ 5.2.0, you can include CardinalMobile via Swift Package Manager.
```
sarahkoop, line 49: I feel like crossing this statement out is confusing since above we are directing people to this section if they are using version 5.0.0-5.2.0

sarahkoop (approved):

SWIFT_PACKAGE_MANAGER.md
sestevens: Hmmm... I can rework this section.

sestevens, line 49:
```suggestion
If you are using versions 5.0.0 to 5.2.0, in order to use the `BraintreeThreeDSecure` library with SPM, you must manually include `CardinalMobile.framework`. For versions _greater than_ 5.2.0, you can include CardinalMobile via Swift Package Manager.
```

sestevens: @sarahkoop Updated! Could you take a look and let me know if my changes make sense?

sarahkoop: I think this is very clear now 👍

sestevens: Awesome, thanks for the feedback!

scannillo:
👏

PR #655, opened 2021-04-02
Update PPRiskMagnesOC to 4.0.12

sestevens:
### Summary of changes

- Update PPRiskMagnesOC to 4.0.12. This version of Magnes resolves a duplicate symbols errors that a merchant reported with Magnes and another SDK.

Updating this in 4.x because the merchant who reported the issue is using Drop-in, and the latest GA release of Drop-in uses version 4.x of Braintree iOS. We won't need to update this in v5 because v5 uses the Swift version of Magnes, not the Obj-C version.

### Checklist

- [x] Added a changelog entry

- @sestevens


sshropshire (approved):

scannillo (approved):
Looks good!

I think it is worth updating in v5 though eventually, it will make it easier to track.

sestevens:
> Looks good!
> 
> I think it is worth updating in v5 though eventually, it will make it easier to track.

We don't use the Obj-C version of Magnes in v5, so we don't need to update anything there 🙂

scannillo:
> We don't use the Obj-C version of Magnes in v5, so we don't need to update anything there 🙂

You're right! Thank you

PR #667, opened 2021-04-22
Remove carthage.sh script for Xcode 12 workaround

sestevens:
### Summary of changes

- Removing Carthage workaround script that's no longer necessary

### Checklist

- ~Added a changelog entry~

- @sestevens


sarahkoop (approved):
Looks good to me - Do we want to add a note in the README or anywhere that Xcode 12.5 is required for Carthage?

sestevens:
> Looks good to me - Do we want to add a note in the README or anywhere that Xcode 12.5 is required for Carthage?

Oh, it's not necessarily required! Merchants could continue using older versions of Carthage and Xcode along with the workaround script (which we got from the Carthage repo) -- we just don't need to use it in our Rakefile anymore.

sshropshire (approved):

scannillo (approved):

PR #669, opened 2021-04-23
Update PPDataCollector doc string

sestevens:
### Summary of changes

- We got some feedback about a misleading doc string in `PPDataCollector`. The device data collected can be used for cards, not just PayPal and Venmo transactions, so this PR updates that comment.

### Checklist

- ~Added a changelog entry~

- @sestevens

### DIFF
--- a/Sources/PayPalDataCollector/Public/PayPalDataCollector/PPDataCollector.swift
+++ b/Sources/PayPalDataCollector/Public/PayPalDataCollector/PPDataCollector.swift
@@ -22,7 +22,7 @@ import Security
     /**
      Collects device data.
 
-     - Returns: A device data string that should be passed into server-side calls, such as `Transaction.sale`. This JSON serialized string contains a PayPal fraud ID.
+     - Returns: A JSON string containing a device data identifier that should be passed into server-side calls, such as `Transaction.sale`.
     */
     @objc public class func collectPayPalDeviceData() -> String {
         return "{\"correlation_id\":\"\(PPDataCollector.generateClientMetadataID())\"}"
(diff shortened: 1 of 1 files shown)


Sources/PayPalDataCollector/Public/PayPalDataCollector/PPDataCollector.swift
```diff
@@ -17,10 +20,9 @@ import Security
     }
 
     /**
-     Collects device data for PayPal.
+     Collects device data.
 
-     This should be used when the user is paying with PayPal or Venmo only.
-     - Returns: A deviceData string that should be passed into server-side calls, such as `Transaction.sale`, for PayPal transactions. This JSON serialized string contains a PayPal fraud ID.
+     - Returns: A device data string that should be passed into server-side calls, such as `Transaction.sale`. This JSON serialized string contains a PayPal fraud ID.
```
scannillo, line 25: `PayPal fraud ID` is this language we use in the docs? I don't think we use this anywhere else. 

Maybe we stick with `correlationID`? Writing this because it could confuse a merchant to think this ID is only applicable for PayPal checkouts.

sestevens: I didn't actually see any specific name for this in our docs. Maybe we could just reword this to something like:

```suggestion
     - Returns: A JSON serialized string containing device data that should be passed into server-side calls, such as `Transaction.sale`.
```

sestevens: Or maybe instead of "containing device data" it should be "containing a device data identifier", since it doesn't actually contain raw device data.

sshropshire (approved):

sarahkoop (approved):

PR #670, opened 2021-05-03
Make Venmo Context flow opt-in

scannillo:
### Summary of changes

- Launch the new VenmoContext flow only if the `paymentMethodUsage` property on `BTVenmoRequest` is explicitly set
    - Adds a default enum case to `BTVenmoPaymentMethodUsage` for `None`
- Reorganize methods in `BTVenmoDriver.m` to align with pragma mark descriptions

### Question
- I have an open question regarding how we should communicate this new "feature" to merchants, using this opt-in pattern. See PR comment.

### Checklist

- ~Added a changelog entry~
- [JIRA](https://engineering.paypalcorp.com/jira/browse/DTBTSDK-942)

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeVenmo/BTVenmoDriver.m
+++ b/Sources/BraintreeVenmo/BTVenmoDriver.m
@@ -205,7 +205,7 @@ static BTVenmoDriver *appSwitchedDriver;
 
 #pragma mark - App switch
 
-- (void)performAppSwitch:(NSURL *)appSwitchURL  completion:(void (^)(BTVenmoAccountNonce * _Nullable venmoAccount, NSError * _Nullable error))completionBlock {
+- (void)performAppSwitch:(NSURL *)appSwitchURL completion:(void (^)(BTVenmoAccountNonce * _Nullable venmoAccount, NSError * _Nullable error))completionBlock {
     if (!appSwitchURL) {
         NSError *error = [NSError errorWithDomain:BTVenmoDriverErrorDomain
                                              code:BTVenmoDriverErrorTypeInvalidRequestURL
@@ -378,4 +378,3 @@ static BTVenmoDriver *appSwitchedDriver;
 }
 
 @end
-
--- a/Sources/BraintreeVenmo/BTVenmoDriver.m
+++ b/Sources/BraintreeVenmo/BTVenmoDriver.m
@@ -132,64 +132,52 @@ static BTVenmoDriver *appSwitchedDriver;
             return;
         }
 
+        self.venmoRequest = venmoRequest;
+
         NSString *merchantProfileID = venmoRequest.profileID ?: configuration.venmoMerchantID;
+        NSString *bundleDisplayName = [self.bundle objectForInfoDictionaryKey:@"CFBundleDisplayName"];
 
-        NSDictionary *params = @{
-            @"query": @"mutation CreateVenmoPaymentContext($input: CreateVenmoPaymentContextInput!) { createVenmoPaymentContext(input: $input) { venmoPaymentContext { id } } }",
-            @"variables": @{
-                    @"input": @{
-                            @"paymentMethodUsage": venmoRequest.paymentMethodUsageAsString,
-                            @"merchantProfileId": merchantProfileID,
-                            @"customerClient": @"MOBILE_APP",
-                            @"intent": @"CONTINUE"
-                    }
-            }
-        };
+        BTMutableClientMetadata *metadata = [self.apiClient.metadata mutableCopy];
+        metadata.source = BTClientMetadataSourceVenmoApp;
 
-        [self.apiClient POST:@"" parameters:params httpType:BTAPIClientHTTPTypeGraphQLAPI completion:^(BTJSON *body, __unused NSHTTPURLResponse *response, __unused NSError *error) {
+        if (venmoRequest.paymentMethodUsage != BTVenmoPaymentMethodUsageNone) { // Venmo Context flow
+            NSDictionary *params = @{
+                @"query": @"mutation CreateVenmoPaymentContext($input: CreateVenmoPaymentContextInput!) { createVenmoPaymentContext(input: $input) { venmoPaymentContext { id } } }",
+                @"variables": @{
+                        @"input": @{
+                                @"paymentMethodUsage": venmoRequest.paymentMethodUsageAsString,
+                                @"merchantProfileId": merchantProfileID,
+                                @"customerClient": @"MOBILE_APP",
+                                @"intent": @"CONTINUE"
+                        }
+                }
+            };
 
-            BTMutableClientMetadata *metadata = [self.apiClient.metadata mutableCopy];
-            metadata.source = BTClientMetadataSourceVenmoApp;
-            NSString *bundleDisplayName = [self.bundle objectForInfoDictionaryKey:@"CFBundleDisplayName"];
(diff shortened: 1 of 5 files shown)


Sources/BraintreeVenmo/Public/BraintreeVenmo/BTVenmoRequest.h
```diff
@@ -34,7 +36,7 @@ typedef NS_ENUM(NSInteger, BTVenmoPaymentMethodUsage) {
 /**
  * If set to `.multiUse`, the resulting payment method will be authorized for future payments and can be vaulted. If set to `.singleUse`, the resulting payment method will be authorized for a one-time payment and cannot be vaulted.
  *
- * Defaults to `.multiUse`.
+ * Defaults to `.none`.
  */
```
scannillo, line 40: What distinctions are there between the old Venmo flow and the two flows offered via Venmo Context (multi-use & single-use)?  How can we make that clear in this docstring?

I'm a little confused on this, since I know the "old" Venmo flow had a concept of defaulting to `multi-use`.

Sources/BraintreeVenmo/BTVenmoDriver.m
```diff
@@ -217,10 +205,42 @@ - (void)vaultVenmoAccountNonce:(NSString *)nonce {
 
 #pragma mark - App switch
 
+- (void)performAppSwitch:(NSURL *)appSwitchURL  completion:(void (^)(BTVenmoAccountNonce * _Nullable venmoAccount, NSError * _Nullable error))completionBlock {
```
sestevens, line 208:
```suggestion
- (void)performAppSwitch:(NSURL *)appSwitchURL completion:(void (^)(BTVenmoAccountNonce * _Nullable venmoAccount, NSError * _Nullable error))completionBlock {
```

Sources/BraintreeVenmo/BTVenmoDriver_Internal.h
```diff
@@ -38,4 +38,9 @@
  */
 @property (nonatomic, assign) BOOL shouldVault;
 
+/**
+ Exposed for testing the payment method usage associated with this request
```
sestevens, line 42: I'm a little confused about why this need to be exposed for testing. Shouldn't tests already have access to the venmo request since it's passed into the `tokenize` method?

scannillo: You are right! We don't need it for testing.

Sources/BraintreeVenmo/Public/BraintreeVenmo/BTVenmoRequest.h
sestevens: Maybe we can call it `.unspecified` instead of `.none`, since (as you mentioned), it will default to multi-use. The doc string could be something like:

```
If set to `.multiUse`, the resulting payment method will be authorized for future payments and can be vaulted. If set to `.singleUse`, the resulting payment method will be authorized for a one-time payment and cannot be vaulted. If not specified, the resulting payment method will be authorized for future payment and can be vaulted; however, the updated Venmo UI will not be used.
```

Sources/BraintreeVenmo/Public/BraintreeVenmo/BTVenmoRequest.h
```diff
@@ -6,10 +6,12 @@ NS_ASSUME_NONNULL_BEGIN
  Usage for the tokenized Venmo account: either multi-use or single use
```
sestevens, line 6:
```suggestion
 Usage for the resulting payment method
```
I'm not sure if we actually need to say what the cases are here. But if we do, we should probably list the new case as well.

sestevens:
We should update this line:
https://github.com/braintree/braintree_ios/blob/f7fa072659897c77723ee0805d9fd0faf583d6c2/Sources/BraintreeVenmo/Public/BraintreeVenmo/BTVenmoRequest.h#L30

to something like:
```
If this property is set to false, you can still vault the Venmo account on your server, provided that `paymentMethodUsage` is not set to `.singleUse`.
```

since vaulting is possible if `paymentMethodUsage` is `.multiUse` or `.none` -- but not if it's `.singleUse`.

Sources/BraintreeVenmo/Public/BraintreeVenmo/BTVenmoRequest.h
scannillo: Do you know what Venmo UI updates does a merchant get with the VenmoContext flow?

demerino: There is new language and a `half sheet` style UI (think apple pay). Maybe we say something like...

>Leaving this value as `none` will result in the legacy Venmo flow. It is recommended to select `multiUse` or `singleUse` for the best customer experience.

demerino (approved):

Sources/BraintreeVenmo/Public/BraintreeVenmo/BTVenmoRequest.h
scannillo: Good feedback, and thank you for the clarifications! See updated docstring comment.

Sources/BraintreeVenmo/BTVenmoDriver_Internal.h
```diff
@@ -38,4 +38,5 @@
  */
 @property (nonatomic, assign) BOOL shouldVault;
 
+
```
sestevens, line 41:
```suggestion
```

Sources/BraintreeVenmo/Public/BraintreeVenmo/BTVenmoRequest.h
```diff
@@ -25,17 +27,20 @@ typedef NS_ENUM(NSInteger, BTVenmoPaymentMethodUsage) {
...
  *
  * Defaults to false.
  */
 @property (nonatomic) BOOL vault;
 
 /**
- * If set to `.multiUse`, the resulting payment method will be authorized for future payments and can be vaulted. If set to `.singleUse`, the resulting payment method will be authorized for a one-time payment and cannot be vaulted.
+ * If set to `.multiUse`, the Venmo payment will be authorized for future payments and can be vaulted.
+ * If set to `.singleUse`, the Venmo payment will be authorized for a one-time payment and cannot be vaulted.
+ * If set to `.unspecified`, the legacy Venmo UI flow will launch. It is recommended to use `.multiUse` or `.singleUse` for the best customer experience.
```
sestevens, line 39: Nice 👏

Sources/BraintreeVenmo/Public/BraintreeVenmo/BTVenmoRequest.h
```diff
@@ -25,17 +27,20 @@ typedef NS_ENUM(NSInteger, BTVenmoPaymentMethodUsage) {
...
  *
- * Defaults to `.multiUse`.
+ * Defaults to `.unspecified`.
  */
+
```
sestevens, line 43:
```suggestion
```

sestevens: Not sure if this extra space will mess up viewing the doc string in Xcode or the generated Jazzy docs.

sestevens (approved):
Looks good! I just noticed a couple extra blank lines that I think can be removed.

PR #672, opened 2021-05-06
Fix bug where `userAction` on `BTPayPalCheckoutRequest` was ignored

sestevens:
### Summary of changes

- Fix a bug where `userAction` on `BTPayPalCheckoutRequest` was ignored, which meant that the button in the PayPal web page displayed the wrong text.

The bug was that we had a local variable called `approvalURL` as well as a property, `self.approvalURL`. We added the `userAction` query param to `self.approvalURL`, but then were actually using `approvalURL` (the local variable, which didn't have the `userAction` query param) to open the web page.

The tests didn't catch this because they are checking `self.approvalURL`, which was constructed correctly (but then not actually used). Instead of setting `self.approvalURL` early on, I moved it so that it is set right before we pass the URL to `ASWebAuthenticationSession`. That way, we're testing the URL that we're actually using.

### Checklist

- [x] Added a changelog entry

- @sestevens

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## unreleased
 * Update Kount SDK to v4.1.5
+* Fix bug where `userAction` on `BTPayPalCheckoutRequest` was ignored
 
 ## 4.37.1 (2021-04-06)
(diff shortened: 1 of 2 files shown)


scannillo (approved):

demerino (commented):
👍

demerino (approved):

PR #678, opened 2021-05-17
Fix docs for `amount` on `BTPayPalCheckoutRequest`

sestevens:
### Summary of changes

- Remove incorrect docs indicating that commas are allowed in the amount on BTPayPalCheckoutRequest

### Checklist

- ~Added a changelog entry~

- @sestevens


sshropshire (approved):

sarahkoop (approved):
Looks good to me. Does this only apply to iOS? I noticed our Android doc string also says thousands separator is allowed.

sestevens:
> Looks good to me. Does this only apply to iOS? I noticed our Android doc string also says thousands separator is allowed.

Android needs to be updated to! There's a ticket in the sprint for it.

Issue #671, opened 2021-05-05
'Braintree/Braintree-Swift.h' file not found in React Native project using CocoaPods


jbatra-umeey:
**Integration Details**

 - Braintree SDK Version: 5.3.0
 - Environment: : Sandbox
 - iOS Version: : iOS 13.4.1
 - Device: [e.g. iPhone 12] : iPhone 7
 - Integration type & version: Cocoapods 1.10.1

**Describe the bug**

When updating to 5.3.0 we get an error during the compilation of the header file 'Braintree/Braintree-Swift.h' . The current version we are using is 4.37.1
In continuation to issue https://github.com/braintree/braintree_ios/issues/651

Podfile :

```
require_relative '../node_modules/react-native/scripts/react_native_pods'
require_relative '../node_modules/@react-native-community/cli-platform-ios/native_modules'

platform :ios, '12.0'
use_frameworks!
inhibit_all_warnings!
target 'demoReactNativeApp' do
  config = use_native_modules!

pod 'Braintree'
pod 'Braintree/Venmo'

  use_react_native!(
    :path => config[:reactNativePath],
    # to enable hermes on iOS, change `false` to `true` and then install pods
    :hermes_enabled => false
  )

  # Enables Flipper.
  #
  # Note that if you have use_frameworks! enabled, Flipper will not work and
  # you should disable the next line.
  use_flipper!()

  post_install do |installer|
    react_native_post_install(installer)
  end

end
```

**To Reproduce**
There are no explicit steps to reproduce, other than try to update the version

**Expected behavior**
Success build with the latest version

**Screenshots**
<img width="1802" alt="Screenshot 2021-05-05 at 1 34 46 PM" src="https://user-images.githubusercontent.com/75033678/117113860-b48a3700-ada8-11eb-9fef-573c41deb00d.png">

In the project directory, I can see this file present. Any changes needed in the main project's build settings?

<img width="1352" alt="Screenshot 2021-05-05 at 4 05 34 PM" src="https://user-images.githubusercontent.com/75033678/117129237-0ee0c300-adbc-11eb-8212-6680d31a1609.png">

As a workaround can we skip `PayPalDataCollector ` from cocoa pods?

sestevens:
> As a workaround can we skip PayPalDataCollector from cocoa pods?

[...]

PR #680, opened 2021-05-18
Bugfix for React Native

scannillo:
### Summary of changes

- See [this GH issue](https://github.com/braintree/braintree_ios/issues/671) calling out issues using ReactNative via CocoaPods and our SDK.
- This PR:
    - Updates the import syntax for importing Braintree modules written in Swift into Objective-C code.
    - Adds a React Native sample app for our CI builds to test against on each PR.
        - _Note: since we don't officially support React Native, we can remove this anytime. But, we figured it was a nice-to-have since ReactNative is a popular framework._
        - _Note: Everything in `SampleApps/ReactNativeTest` is from mostly template ReactNative app. It wasn't straightforward to remove the Android code, so we can do that in another PR._
        - **We welcome discussion on including this app in our repo/CI**
    - Updates DEVELOPMENT.md

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo 
@sestevens


.github/workflows/build.yml
```diff
@@ -49,3 +49,19 @@ jobs:
...
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Use Xcode 12.4
+        run: sudo xcode-select -switch /Applications/Xcode_12.4.app
+      - name: Use current branch
+        run: sed -i '' "s/branch => .*/branch => \'"$GITHUB_HEAD_REF"\'/" SampleApps/ReactNativeTest/ios/Podfile
+      - name: Install npm dependencies
+        run: cd SampleApps/ReactNativeTest && npm install
+      - name: Install CocoaPod dependencies
```
sestevens, line 64:
```suggestion
      - name: Install CocoaPods dependencies
```

sarahkoop:
The import syntax fix looks good to me! With respect to including the React Native app in this repo and CI - Would we consider it a PR/release blocker if CI fails because of the React Native build? Also, would we plan to consistently update the React Native version to the latest? 

I think it would be helpful to keep it in this repo and CI if we want to consistently make sure our SDK is compatible with the latest React Native version, but that also seems closer to official support.  Otherwise, it might make more sense to move it to a separate repo to use for testing if/when any issues are opened.

sarahkoop (approved):

scannillo:
> The import syntax fix looks good to me! With respect to including the React Native app in this repo and CI - Would we consider it a PR/release blocker if CI fails because of the React Native build? Also, would we plan to consistently update the React Native version to the latest?

> I think it would be helpful to keep it in this repo and CI if we want to consistently make sure our SDK is compatible with the latest React Native version, but that also seems closer to official support. Otherwise, it might make more sense to move it to a separate repo to use for testing if/when any issues are opened.

@sarahkoop - this is great feedback. I don't think would treat the ReactNative build as a blocker. Also if it gets flakey or inconsistent we can easily remove it from the CI build.

In terms of updating the ReactNative app to the latest Braintree SDK, the Podfile (`pod 'Braintree', :git => 'https://github.com/braintree/braintree_ios', :branch => 'master'`) will automatically do that itself each time `pod install` runs in CI.

Another idea - we can have the sample app live in this repo, and just not run it in CI until an issue comes in like you were saying. I'm open! We could also try it in CI for a bit and if we hate it, remove it.

sarahkoop:
Trying it in CI for a bit and removing if it causes problems sounds good to me!

SampleApps/ReactNativeTest/ios/EmptySwiftFile.swift
```diff
@@ -0,0 +1,8 @@
+//
+//  EmptySwiftFile.swift
+//  ReactNativeTest
+//
+//  Created by Cannillo, Sammy on 5/18/21.
+//
+
+import Foundation
```
sestevens, line 8: Do we need both `Empty.swift` and `EmptySwiftFile.swift` or can we remove one of them?

SampleApps/ReactNativeTest/ios/ReactNativeTest/AppDelegate.m
```diff
@@ -0,0 +1,112 @@
...
+    [self initializeReactNativeApp];
+  #else
+    EXUpdatesAppController *controller = [EXUpdatesAppController sharedInstance];
+    controller.delegate = self;
+    [controller startAndShowLaunchScreen:self.window];
+  #endif
+
+  [super application:application didFinishLaunchingWithOptions:launchOptions];
+
+  // [BTAppSwitch setReturnURLScheme:@"com.your-company-name.your-app-name.payments"];
```
sestevens, line 61: Is this supposed to be here?

sestevens:
I'm a little nervous about adding the React Native sample app to our repo since the repo already takes a really long time to clone. But I'm OK with trying it for now. Maybe eventually we can move all of these sample apps into their own repo. It would be nice to remove the Android code, though, since there seems to be a lot of it.

crookedneighbor:
I definitely think the react native sample app should be in a separate repo (and maybe include a link to the repo in the readme).

scannillo:
> I'm a little nervous about adding the React Native sample app to our repo since the repo already takes a really long time to clone.

This is totally valid. It's a lot of extra code.

> Maybe eventually we can move all of these sample apps into their own repo. It would be nice to remove the Android code, though, since there seems to be a lot of it.

I did try removing the Android code, but it wasn't immediately straightforward. We could follow that up in another PR.

> I definitely think the react native sample app should be in a separate repo (and maybe include a link to the repo in the readme).

This is also fair. We can throw it into it's own repo just to have it to potentially test against? I don't know how often we'll need it. Also - a sample react-native repo might be a good way for us to gauge interest. If we get a ton of people stumbling on it asking for official support, could be a sign we should move toward that. It also could give them the wrong idea that we do offer support for it, if we have a repo for it. 🤔

SampleApps/ReactNativeTest/ios/ReactNativeTest/AppDelegate.m
scannillo: Nah! Don't need it!

scannillo:
As a summary - sounds like folks are hesitant to include this sample app in our repo. Totally valid. 

I will update this PR to remove the sample app. We can then move it into it's own sample repo.

sestevens (approved):

PR #683, opened 2021-05-19
Enable presenting PayPal native checkout UI

sestevens:
### Summary of changes

- Enable merchants to present the PayPal native checkout UI. This PR doesn't handle the response from the native checkout SDK -- that will be covered in a separate PR.

### Checklist

- ~Added a changelog entry~

- @sestevens 
- @scannillo


Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeAccountNonce.swift
```diff
@@ -0,0 +1,5 @@
+import BraintreePayPal
+
+@objc public class BTPayPalNativeAccountNonce: BTPayPalAccountNonce {
```
sestevens, line 3: Oops, this needs a doc comment.

Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeVaultRequest.swift
```diff
@@ -0,0 +1,27 @@
+import BraintreePayPal
+
+/**
+ Options for the PayPal Vault flow.
+ */
+@objc public class BTPayPalNativeVaultRequest: BTPayPalVaultRequest, BTPayPalNativeRequest {
```
sarahkoop, line 6: Is this class needed? I thought this flow was only for checkout requests.

Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeClient.swift
```diff
@@ -0,0 +1,70 @@
...
+    @objc(tokenizePayPalAccountWithPayPalRequest:completion:)
+    public func tokenizePayPalAccount(with nativeRequest: BTPayPalRequest, completion: @escaping (BTPayPalNativeAccountNonce?, NSError?) -> Void) {
+        guard let request = nativeRequest as? (BTPayPalRequest & BTPayPalNativeRequest) else {
+            completion(nil, BTPayPalNativeError.invalidRequest as NSError)
+            return
+        }
+
+        let orderCreationClient = BTPayPalNativeOrderCreationClient(with: apiClient)
+        orderCreationClient.createOrder(with: request) { result in
+            switch result {
```
scannillo, line 40: I like this using the Swift Result type ([link](https://www.hackingwithswift.com/articles/161/how-to-use-result-in-swift) for people who want to read) 👏

Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeError.swift
```diff
@@ -0,0 +1,46 @@
+import Foundation
+
+/**
+ Error returned from the native PayPal flow
+ */
+@objc public enum BTPayPalNativeError: Int, LocalizedError {
```
scannillo, line 6: nice!

Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeVaultRequest.swift
scannillo: +1 To this question. I was under the assumption vault & all flows should work with the Native XO, but I know Susan saw some errors when testing Billing Agreement. So, maybe Native XO isn't compatible with vault yet?

sestevens: I think we need to create a demo app so that we can test this. But what we heard from the Native Checkout team is that you can request a billing agreement and it will fall back to a web flow.

scannillo (approved):

Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeError.swift
```diff
@@ -0,0 +1,46 @@
...
+    case fetchConfigurationFailed
+
+    /// PayPal is not enabled for this merchant in the Braintree Control Panel
+    case payPalNotEnabled
+
+    /// Could not find PayPal client ID in the Braintree configuration
+    case payPalClientIDNotFound
+
+    /// Invalid environment identifier found in the Braintree configuration
+    case invalidEnvironment
```
sarahkoop, line 20: Curious what we are considering an invalid environment here? In Android we are currently using the logic if `production` then send to `Live` environment, and everything else is `Sandbox`

sarahkoop (approved):
🌟

Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeError.swift
sestevens: Yeah, we could do the same here. We _should_ only ever get back production or sandbox, so this error shouldn't ever occur. 🤷‍♀️ I don't really have a strong preference either way, but I'm happy to change this if you think defaulting to sandbox is better!

sarahkoop: I don't feel strongly but will just make sure that we are doing the same on Android and iOS. How would a merchant get configured for an environment other than production or sandbox? It might be a confusing error for them to try to resolve. But maybe it is best of error out if they aren't configured correctly.

sestevens: > How would a merchant get configured for an environment other than production or sandbox?

This should never happen -- but if it did happen, the merchant would need to reach out to Support and we'd need to investigate. Some of these error messages are really intended to help us debug more than to help merchants debug. So maybe it's better to keep it?

PR #685, opened 2021-05-20
Return PayPal account nonce from BraintreePayPalNative module

sestevens:
### Summary of changes

- Tokenize PayPal account returned from PayPalCheckout SDK and return it to the merchant

### Checklist

- ~Added a changelog entry~

- @sestevens


Sources/BraintreeCore/BTAPIClient_Internal.h
```diff
@@ -32,11 +32,6 @@ typedef NS_ENUM(NSInteger, BTAPIClientAuthorizationType) {
 @property (nonatomic, strong) BTAPIHTTP *braintreeAPI;
 @property (nonatomic, strong) BTGraphQLHTTP *graphQL;
 
-/**
- Client metadata that is used for tracking the client session
-*/
-@property (nonatomic, readonly, strong) BTClientMetadata *metadata;
```
sestevens, line 38: This needed to be moved to the public header to be accessible to Swift.

Sources/BraintreeCore/Public/BraintreeCore/BTAPIClient.h
```diff
@@ -69,6 +69,11 @@ typedef NS_ENUM(NSInteger, BTAPIClientHTTPType) {
 - (instancetype)copyWithSource:(BTClientMetadataSourceType)source
                    integration:(BTClientMetadataIntegrationType)integration;
 
+/**
+ Client metadata that is used for tracking the client session
```
sestevens, line 73:
```suggestion
 :nodoc:
 Client metadata that is used for tracking the client session
```

Sources/BraintreePayPal/BTPayPalAccountNonce.m
```diff
@@ -16,6 +18,75 @@ @interface BTPayPalAccountNonce ()
 
 @implementation BTPayPalAccountNonce
 
+- (nullable instancetype)initWithJSON:(BTJSON *)json {
```
sestevens, line 21: I thought it made sense to have the JSON parsing contained to an initializer, instead of putting it somewhere in the `BraintreePayPalNative` module. At some point, we could update `BTPayPalDriver` to use this initializer too, instead of doing the parsing in the driver.

scannillo: +1 I like that idea of also updating BTPayPalDriver. I wouldn't be opposed to doing that in this PR. If not - making a jira or adding a TODO for it here. (I feel this is something we can easily forget)

Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeTokenizationRequest.swift
```diff
@@ -0,0 +1,44 @@
...
+            "_meta": [
+                "source": clientMetadata.sourceString,
+                "integration": clientMetadata.integrationString,
+                "sessionId": clientMetadata.sessionID,
+            ]
+        ]
+
+        if let checkoutRequest = request as? BTPayPalNativeCheckoutRequest {
+            account["options"] = ["validate" : false]
+            account["intent"] = checkoutRequest.intentAsString
```
scannillo, line 39: Confirming - it is intended we aren't passing intent for VaultRequests?

sestevens: Yep! Intent is only relevant to checkout, not vault.

Sources/BraintreeCore/Public/BraintreeCore/BTPaymentMethodNonce.h
```diff
@@ -21,7 +21,7 @@ NS_ASSUME_NONNULL_BEGIN
  @param nonce       A transactable payment method nonce.
  @return A Payment Method Nonce, or `nil` if nonce is nil.
 */
-- (nullable instancetype)initWithNonce:(NSString *)nonce;
+- (nullable instancetype)initWithNonce:(NSString * _Nullable)nonce;
```
sarahkoop, line 24: I believe nullable is the default in obj-c, but want to confirm - does adding these nullability annotations affect existing merchant integrations?

Sources/BraintreePayPal/BTPayPalAccountNonce.m
```diff
@@ -16,6 +18,75 @@ @interface BTPayPalAccountNonce ()
...
+    BTJSON *payPalAccount = json[@"paypalAccounts"].asArray.firstObject;
+    NSString *nonce = payPalAccount[@"nonce"].asString;
+    BOOL isDefault = payPalAccount[@"default"].isTrue;
+
+    if (self = [super initWithNonce:nonce type:@"PayPal" isDefault:isDefault]) {
+        BTJSON *details = payPalAccount[@"details"];
+        _email = details[@"email"].asString ?: details[@"payerInfo"][@"email"].asString;
+        _clientMetadataID = details[@"correlationId"].asString;
```
sarahkoop, line 29: Is the correlationId the same as clientMetadataID? If so, should the clientMetadataID be passed for correlationID in the TODO below?

sarahkoop (approved):

Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeTokenizationClient.swift
```diff
@@ -0,0 +1,36 @@
...
+    func tokenize(returnURL: String,
+                  request: BTPayPalRequest,
+                  completion: @escaping (Result<BTPayPalNativeAccountNonce, BTPayPalNativeError>) -> Void) {
+
+        let tokenizationRequest = BTPayPalNativeTokenizationRequest(returnURL: returnURL,
+                                                                    request: request,
+                                                                    correlationID: "", // TODO: get this value from PayPalCheckout SDK
+                                                                    clientMetadata: apiClient.metadata)
+
+        apiClient .post("/v1/payment_methods/paypal_accounts", parameters: tokenizationRequest.parameters()) { body, _, error in
```
sshropshire, line 22:
```suggestion
        apiClient.post("/v1/payment_methods/paypal_accounts", parameters: tokenizationRequest.parameters()) { body, _, error in
```

sshropshire (approved):

Sources/BraintreeCore/Public/BraintreeCore/BTPaymentMethodNonce.h
sestevens: We have `NS_ASSUME_NONNULL_BEGIN` at the top of this file, so non-nullable is the default here. But allowing the initializer to accept `null` as well as non-null string shouldn't be a breaking change for merchants.

Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeTokenizationClient.swift
sestevens:
```suggestion
        apiClient.post("/v1/payment_methods/paypal_accounts", parameters: tokenizationRequest.parameters()) { body, _, error in
```

sestevens: I'm unresolving this because it doesn't look like the change was actually committed?

Sources/BraintreePayPal/BTPayPalAccountNonce.m
sestevens: Yes, it should be the same value. But we need to get it from the Native Checkout SDK after the customer completes the UI flow and pass it to the Braintree Gateway when we tokenize. Here, we're parsing it from the gateway tokenization response.

PR #701, opened 2021-06-24
Hide Cancel Modal

sshropshire:
### Summary of changes

- Hide cancel modal

### Checklist

### Authors

- @sestevens 
- @sshropshire

### DIFF
--- a/Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeClient.swift
+++ b/Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeClient.swift
@@ -48,6 +48,7 @@ import PayPalCheckout
                                                                        environment: order.environment)
 
                 PayPalCheckout.Checkout.set(config: payPalNativeConfig)
+                PayPalCheckout.Checkout.showsExitAlert = false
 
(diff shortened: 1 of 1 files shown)


Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeClient.swift
```diff
@@ -48,6 +48,7 @@ import PayPalCheckout
                                                                        environment: order.environment)
 
                 PayPalCheckout.Checkout.set(config: payPalNativeConfig)
+                PayPalCheckout.Checkout.showsExitAlert = false
```
sarahkoop, line 51: I thought the option to hide the modal was supposed to be part of the merchant PayPalRequest, are we just always hiding it instead?

sestevens: I guess that would be a question for the team. I was under the impression that we always want to hide it.

scannillo (approved):

Sources/BraintreePayPalNative/Public/BraintreePayPalNative/BTPayPalNativeClient.swift
scannillo: We always want to hide it!

PR #704, opened 2021-07-02
Remove BraintreeUI from Demo App

scannillo:
### Summary of changes

- `BraintreeUI` has been deprecated forever and no longer exists in our Braintree iOS product. We kept using it in the Demo app to avoid having to refactor our Demo app's UI.
- This PR removes `BraintreeUI` from the Demo app.
- This PR creates a basic `BTCardFormView` to be used throughout the Demo.
  - It is used on the 3DS and UnionPay screens.

**NOTE:**
This new card form is super basic. It doesn't do text validation (ex: you need to type in the `/` when entering the expiration date,  doesn't add spaces for easy visualization when typing a card number, doesn't prevent too-long card numbers, etc. I think this is OK and if we really hate it we can do the extra work to add back in those validations. Thoughts?

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo & lady gaga


scannillo:
TODO: Card form doesn't fit properly on small phones (iPhone 5s & iPod Touch 7).

sarahkoop (approved):
This looks good to me! It might be nice in the future to add autofill buttons for the 3DS1 and 3DS2 test cards that we typically use so the form validation doesn't matter as much

scannillo:
> This looks good to me! It might be nice in the future to add autofill buttons for the 3DS1 and 3DS2 test cards that we typically use so the form validation doesn't matter as much

I like this idea. I'll throw a ticket into our JIRA.

sshropshire (approved):

PR #706, opened 2021-07-12
Fix build of package when parallelize build is enabled in Xcode scheme

danielctull:
### Summary of changes

Adds PPRiskMagnes as a dependency of PayPalDataCollector. Without this, projects using Braintree fail to build when the "parallelize build" is true as Xcode does not have a way to determine this dependency.

I checked the provided Xcode project and this dependency definition matches what is defined for the two framework targets. 

### Checklist

- [x] Added a changelog entry 

### Authors

- @danielctull

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+ * Add `PPRiskMagnes` dependency to `PayPalDataCollector` to allow application builds to be parallelized.
+ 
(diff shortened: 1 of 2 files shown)


scannillo:
👋  Hello @danielctull - thanks for opening a PR. Using XCode 12, with the "Parallelize Build" option enabled in the Build Scheme, I don't run into any issues for a project pulling in Braintree via SPM. So - I'm not able to replicate.

Can you provide more specific steps to replicate?

scannillo:
👋  Closing due to inactivity.

PR #710, opened 2021-07-13
Update Gems

sarahkoop:
### Summary of changes

- Address PR #709 
- Run `bundle update`: Update `activesupport`, `addressable`, `concurrent-ruby`, `ffi`, `i18n`, `jazzy`, `manitest`, `rake`, `rexml`, `xcodeproj`, and update bundler version

### Checklist

### Authors

- @sarahkoop


Gemfile.lock
```diff
@@ -114,4 +116,4 @@ DEPENDENCIES
   rake
 
 BUNDLED WITH
-   2.2.9
+   2.2.23
```
sarahkoop, line 119: Upgrading to the latest bundler version as well - since this is a patch version I don't think there are any major changes here.

scannillo (approved):

PR #713, opened 2021-07-19
Xcode13-beta3 Fix

scannillo:
### Background

[GitHub Issue](https://github.com/braintree/braintree-ios-drop-in/issues/343)

[XCode 13-beta3](https://developer.apple.com/documentation/xcode-release-notes/xcode-13-beta-release-notes) release notes call out the following change:
 > Linking Swift packages from application extension targets or watchOS applications no longer emits unresolvable warnings about linking to libraries not safe for use in application extensions. This means that code referencing APIs annotated as unavailable for use in app extensions must now themselves be annotated as unavailable for use in application extensions, in order to allow that code to be used in both apps and app extensions. (66928265)

### Changes

In places that we use Apple APIs which aren't usable in App Extensions (`UIApplication.sharedApplication`) we need to explicitly mark these methods as not usable for App Extensions.

### Resources
- [Twitter thread](https://twitter.com/twannl/status/1415643862761152513)

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* Swift Package Manager
+  * Adds `NS_EXTENSION_UNAVAILABLE` annotations to methods unavailable for use in app extensions. Fixes (Drop-In issue #343)[https://github.com/braintree/braintree-ios-drop-in/issues/343] for Xcode 13-beta3.
+
 ## 5.4.2 (2021-06-24)
 * Swift Package Manager
   * Remove product libraries for `KountDataCollector`, `PPRiskMagnes`, and `CardinalMobile` (requires Xcode 12.5+)
--- a/Sources/BraintreeCore/Public/BraintreeCore/BTPreferredPaymentMethods.h
+++ b/Sources/BraintreeCore/Public/BraintreeCore/BTPreferredPaymentMethods.h
@@ -24,7 +24,7 @@ NS_ASSUME_NONNULL_BEGIN
 
  @param apiClient An API client
 */
-- (instancetype)initWithAPIClient:(BTAPIClient *)apiClient NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithAPIClient:(BTAPIClient *)apiClient NS_DESIGNATED_INITIALIZER NS_EXTENSION_UNAVAILABLE("Uses APIs (i.e UIApplication.sharedApplication) not available for use in App Extensions.");
 
 /**
  :nodoc:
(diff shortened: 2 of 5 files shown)


sshropshire (approved):

sarahkoop (approved):

scannillo:
@sarahkoop & @sshropshire - looking at this again, I am hoping to get your opinion on something. So, I just found [this PR](https://github.com/braintree/braintree_ios/pull/292/files) from a few years ago. It is basically using the workaround of `[UIApplication performSelector:@selector(sharedApplication)];` instead of accessing `UIApplication.sharedApplication` directly, which quiets the compiler.

I opted for the `NS_EXTENSION_UNAVAILABLE` annotations, since it is more explicit / clear that these methods won't work in an App Extension. This will make these methods unusable in an App Extension.

My PR adds `NS_EXTENSION_UNAVAILABLE` into BraintreePayPal, BraintreeVenmo, and PreferredPayments. We don't "officially" claim App Extension support and I don't really see why a merchant would want to use these modules in an App Extension ... but I can't say they don't for sure.

My question is if you think the workaround in that old PR is worth also doing instead of the `UNAVAILABLE` annotations so that people can compile these modules within App Extensions, or if it's better to be explicit. Or ... is adding the annotation _technically_ a breaking change? (We don't really claim official support for "App Extensions" - but want your thoughts). Thanks!

sarahkoop:
I'm not super familiar with App Extensions... do we think this is something an iOS developer / merchant would assume is supported in an iOS SDK unless explicitly stated that we don't support?  I'd lean towards the workaround to prevent breaking anything if any merchants are using those modules in App Extensions. And maybe next major version stating explicitly whether we support them or not going forward. But if you think App Extension support is not assumed I could see the case for this simpler annotation fix.

scannillo:
> I'm not super familiar with App Extensions... do we think this is something an iOS developer / merchant would assume is supported in an iOS SDK unless explicitly stated that we don't support? I'd lean towards the workaround to prevent breaking anything if any merchants are using those modules in App Extensions. And maybe next major version stating explicitly whether we support them or not going forward. But if you think App Extension support is not assumed I could see the case for this simpler annotation fix.

Sorry, I deleted my prior comment since I realized I had misspoke. So, even before adding these annotations, these methods wouldn't compile for an App Extension, since they're using APIs unavailable to App Extensions. So in that case - I think adding these annotations would make no difference to a merchant trying to use these modules in an App Extension. Given that - these changes feel OK to me.

I like your idea of taking a more formal stance on our support for App Extensions in future versions. 

Let me know if that makes sense @sarahkoop.

gobera:
@scannillo What's the next step with this? Are you releasing the fix anytime now from the foregoing?

scannillo:
Hey @gobera - yup, we should be able to get this released by the end of the week.

scannillo:
@gobera  - I am curious, are you blocked on this by braintree_ios or braintree-ios-drop-in?

PR #715, opened 2021-07-21
Add `BTThreeDSecureCardAddChallenge`

sarahkoop:
### Summary of changes

- Add `cardAddChallengeRequested` to `BTThreeDSecureRequest`
- Create `BTThreeDSecureCardAddChallenge` enum to allow for true, false, or nil options for this param 

### Checklist

- [x] Added a changelog entry

- @kate-paypal 
- @sarahkoop

### DIFF
--- a/Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
+++ b/Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
@@ -70,10 +70,10 @@ NSString * const BTThreeDSecureFlowValidationErrorsKey = @"com.braintreepayments
             requestParameters[@"dataOnlyRequested"] = @(request.dataOnlyRequested);
         }
         
-        if (request.addCardChallengeRequested == BTThreeDSecureAddCardChallengeRequested) {
-            requestParameters[@"addCard"] = @(true);
-        } else if (request.addCardChallengeRequested == BTThreeDSecureAddCardChallengeNotRequested) {
-            requestParameters[@"addCard"] = @(false);
+        if (request.cardAddChallengeRequested == BTThreeDSecureCardAddChallengeRequested) {
+            requestParameters[@"cardAdd"] = @(true);
+        } else if (request.cardAddChallengeRequested == BTThreeDSecureCardAddChallengeNotRequested) {
+            requestParameters[@"cardAdd"] = @(false);
         }
 
         NSMutableDictionary *additionalInformation = [NSMutableDictionary dictionary];
--- a/Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
+++ b/Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
@@ -69,6 +69,12 @@ NSString * const BTThreeDSecureFlowValidationErrorsKey = @"com.braintreepayments
         if (request.dataOnlyRequested) {
             requestParameters[@"dataOnlyRequested"] = @(request.dataOnlyRequested);
         }
+        
+        if (request.addCardChallengeRequested == BTThreeDSecureAddCardChallengeRequested) {
+            requestParameters[@"addCard"] = @(true);
+        } else if (request.addCardChallengeRequested == BTThreeDSecureAddCardChallengeNotRequested) {
+            requestParameters[@"addCard"] = @(false);
+        }
 
         NSMutableDictionary *additionalInformation = [NSMutableDictionary dictionary];
         if (request.billingAddress) {
--- a/Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
+++ b/Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
@@ -70,9 +70,9 @@ NSString * const BTThreeDSecureFlowValidationErrorsKey = @"com.braintreepayments
             requestParameters[@"dataOnlyRequested"] = @(request.dataOnlyRequested);
         }
         
-        if (request.cardAddChallengeRequested == BTThreeDSecureCardAddChallengeRequested) {
+        if (request.cardAddChallenge == BTThreeDSecureCardAddChallengeRequested) {
             requestParameters[@"cardAdd"] = @(YES);
-        } else if (request.cardAddChallengeRequested == BTThreeDSecureCardAddChallengeNotRequested) {
+        } else if (request.cardAddChallenge == BTThreeDSecureCardAddChallengeNotRequested) {
             requestParameters[@"cardAdd"] = @(NO);
         }
 
--- a/Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
+++ b/Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
@@ -73,7 +73,7 @@ NSString * const BTThreeDSecureFlowValidationErrorsKey = @"com.braintreepayments
         if (request.cardAddChallengeRequested == BTThreeDSecureCardAddChallengeRequested) {
             requestParameters[@"cardAdd"] = @(YES);
         } else if (request.cardAddChallengeRequested == BTThreeDSecureCardAddChallengeNotRequested) {
-            requestParameters[@"cardAdd"] = @(false);
+            requestParameters[@"cardAdd"] = @(NO);
         }
 
         NSMutableDictionary *additionalInformation = [NSMutableDictionary dictionary];
--- a/Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
+++ b/Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
@@ -71,7 +71,7 @@ NSString * const BTThreeDSecureFlowValidationErrorsKey = @"com.braintreepayments
         }
         
         if (request.cardAddChallengeRequested == BTThreeDSecureCardAddChallengeRequested) {
-            requestParameters[@"cardAdd"] = @(true);
+            requestParameters[@"cardAdd"] = @(YES);
         } else if (request.cardAddChallengeRequested == BTThreeDSecureCardAddChallengeNotRequested) {
             requestParameters[@"cardAdd"] = @(false);
         }
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
@@ -53,7 +53,7 @@
     self = [super init];
     if (self) {
         _versionRequested = BTThreeDSecureVersion2;
-        _addCardChallengeRequested = BTThreeDSecureAddCardChallengeUnspecified;
+        _cardAddChallengeRequested = BTThreeDSecureCardAddChallengeUnspecified;
     }
 
     return self;
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
@@ -53,6 +53,7 @@
     self = [super init];
     if (self) {
         _versionRequested = BTThreeDSecureVersion2;
+        _addCardChallengeRequested = BTThreeDSecureAddCardChallengeUnspecified;
     }
 
     return self;
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
@@ -53,7 +53,6 @@
     self = [super init];
     if (self) {
         _versionRequested = BTThreeDSecureVersion2;
-        _cardAddChallengeRequested = BTThreeDSecureCardAddChallengeUnspecified;
     }
 
     return self;
--- a/Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
+++ b/Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
@@ -71,7 +71,7 @@ typedef NS_ENUM(NSInteger, BTThreeDSecureShippingMethod) {
 };
 
 /**
- The add card challenge request
+ The card add challenge request
  */
 typedef NS_ENUM(NSInteger, BTThreeDSecureCardAddChallenge) {
     /// Unspecified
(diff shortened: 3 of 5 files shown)


Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
```diff
@@ -140,6 +154,15 @@ typedef NS_ENUM(NSInteger, BTThreeDSecureShippingMethod) {
  */
 @property (nonatomic) BOOL dataOnlyRequested;
 
+/**
+ Optional. An authentication created using this property should only be used for adding a payment method to the merchant's vault and not for creating transactions.
+ Defaults to BTThreeDSecureAddCardChallengeUnspecified.
+ If set to BTThreeDSecureAddCardChallengeRequested, the authentication challenge will be requested from the issuer to confirm adding new card to the merchant's vault.
+ If set to BTThreeDSecureAddCardChallengeUnspecified and amount is 0, the authentication challenge will be presented to the user.
+ If set to BTThreeDSecureAddCardChallengeNotRequested, when the amount is 0, the authentication challenge will not be presented to the user
```
scannillo, line 162: Love this - super clear!

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
```diff
@@ -140,6 +154,15 @@ typedef NS_ENUM(NSInteger, BTThreeDSecureShippingMethod) {
...
+ */
+@property (nonatomic) BTThreeDSecureCardAddChallenge cardAddChallengeRequested;
```
scannillo, line 164: How come on iOS we are using an enum but Android a bool? Should we keep them the same across platforms?

crookedneighbor: Definitely agree that we should make this consistent across the sdks.

kate-paypal: The enum was used because we weren't certain of how to mimic the behavior of a nullable bool in objective c.

scannillo: AH you're right - we can't really do nullable bools in Objective-C. We could do some sort of custom setter on the Bool and then have a flag to detect if the value was set by the merchant or not ... but that's definitely kind of hacky.

I think the most "Objective-C" friendly route here is the enum. I think it is OK if we have to diverge from parity here to do what makes most sense for the language. _If we were in Swift, this would be as easy as `Bool?`._

But honestly, since the false case is different from the not-set case ... there technically are "3" options to this property and maybe an enum actually makes the most sense across SDKs? Thoughts?

scannillo, line 164:
```suggestion
@property (nonatomic, assign) BTThreeDSecureCardAddChallenge cardAddChallengeRequested;
```

sarahkoop: The Android change with nullable `Boolean` as the parameter was released this week, but we could definitely deprecate that method in favor of an enum.  I do think a `Boolean` for a param with `null`, `true`, and `false` options does make most sense from a Java language perspective, but if we want full parity on this feature we can make the change there too.

scannillo: Gotcha - yeh. Maybe we can second pass about whether or not we want to deprecate the existing Bools or just leave them as-is?

Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
```diff
@@ -69,6 +69,12 @@ - (void)performThreeDSecureLookup:(BTThreeDSecureRequest *)request
         if (request.dataOnlyRequested) {
             requestParameters[@"dataOnlyRequested"] = @(request.dataOnlyRequested);
         }
+        
+        if (request.cardAddChallengeRequested == BTThreeDSecureCardAddChallengeRequested) {
+            requestParameters[@"cardAdd"] = @(true);
```
scannillo, line 74:
```suggestion
            requestParameters[@"cardAdd"] = @(YES);
```

Sources/BraintreeThreeDSecure/BTPaymentFlowDriver+ThreeDSecure.m
```diff
@@ -69,6 +69,12 @@ - (void)performThreeDSecureLookup:(BTThreeDSecureRequest *)request
...
+        } else if (request.cardAddChallengeRequested == BTThreeDSecureCardAddChallengeNotRequested) {
+            requestParameters[@"cardAdd"] = @(false);
```
scannillo, line 76:
```suggestion
            requestParameters[@"cardAdd"] = @(NO);
```

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
```diff
@@ -53,6 +53,7 @@ - (instancetype)init {
     self = [super init];
     if (self) {
         _versionRequested = BTThreeDSecureVersion2;
+        _cardAddChallengeRequested = BTThreeDSecureCardAddChallengeUnspecified;
```
scannillo, line 56:
```suggestion
```

We can omit this since ObjC will default the enum to the first value option in the ENUM list.

UnitTests/BraintreeThreeDSecureTests/BTPaymentFlowDriver+ThreeDSecure_Tests.swift
```diff
@@ -76,6 +78,41 @@ class BTPaymentFlowDriver_ThreeDSecure_Tests: XCTestCase {
...
+        driver.performThreeDSecureLookup(threeDSecureRequest) { (lookup, error) in
+            XCTAssertFalse(self.mockAPIClient.lastPOSTParameters!["cardAdd"] as! Bool)
+
+            expectation.fulfill()
+        }
+
+        waitForExpectations(timeout: 3, handler: nil)
+    }
+
+    func testPerformThreeDSecureLookup_whenAddCardChallengeUnspecified_doesNotSendAddCardParameter() {
```
scannillo, line 99:
```suggestion
    func testPerformThreeDSecureLookup_whenCardAddChallengeRequestedNotSet_doesNotSendAddCardParameter() {
```

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
```diff
@@ -140,6 +154,15 @@ typedef NS_ENUM(NSInteger, BTThreeDSecureShippingMethod) {
...
 @property (nonatomic) BOOL dataOnlyRequested;
 
+/**
+ Optional. An authentication created using this property should only be used for adding a payment method to the merchant's vault and not for creating transactions.
+ Defaults to BTThreeDSecureAddCardChallengeUnspecified.
+ If set to BTThreeDSecureAddCardChallengeRequested, the authentication challenge will be requested from the issuer to confirm adding new card to the merchant's vault.
+ If set to BTThreeDSecureAddCardChallengeUnspecified and amount is 0, the authentication challenge will be presented to the user.
+ If set to BTThreeDSecureAddCardChallengeNotRequested, when the amount is 0, the authentication challenge will not be presented to the user
+ */
+@property (nonatomic, assign) BTThreeDSecureCardAddChallenge cardAddChallengeRequested;
```
scannillo, line 164: Since this is an enum now, maybe the name `cardAddChallengeRequested` isn't the most clear. 

So instead of merchant code looking like:
`request.cardAddChallengeRequested = .requested`

maybe we can do something like
`request.cardAddChallengeType = .requested` or `request.cardAddChallenge = .requested` or `request.cardAddChallengeRequirement = . requested` ? Not sure on which would be best, but we can brainstorm. Thoughts?

scannillo: Also, for the `notRequested` option the docstring says that the challenge will _not_ be presented, in any case. Am I understanding that right?

If that's the case - I feel like the name `notRequested` might not communicate that. 🤔

sarahkoop: After reviewing the updated JS doc strings and GW PR on this I'm thinking we might want to reword the doc strings here. I don't know that we can say for sure that a challenge will be presented to the user, just that an add card authentication will be requested from the issuer.

The breakdown is: 
true -> card add challenge will be requested
false -> card add challenge will not be requested
null and amount 0 -> card add challenge will be requested
null and amount greater than 0 -> card add challenge will not be requested

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
sarahkoop: Cool good to know 👍

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
scannillo: `null and amount 0 -> card add authentication will be requested` Blade confirmed was behavior before this PR, as well so we are good on that not being a breaking change.

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
```diff
@@ -70,6 +70,20 @@ typedef NS_ENUM(NSInteger, BTThreeDSecureShippingMethod) {
     BTThreeDSecureShippingMethodShipToStore
 };
 
+/**
+ The add card challenge request
```
scannillo, line 74: Nitpick - let's keep it consistent throughout for "add card" or "card add".

sarahkoop: Good call - done.

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
sarahkoop: Yeah I think we are good on the breaking change part 👍 . I've updated the param to be set as `request.cardAddChallenge`, and updated the doc string to try and provide more clarity on the different scenarios.

Since this param is actually determining whether or not the card add authentication will be requested from the issuer (rather than what specifically will be displayed to the user), I think `.requested` and `.notRequested` still make the most sense here, but open to other ideas.

scannillo: Awesome - these clarifications are 👌  super helpful

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
```diff
@@ -140,6 +154,16 @@ typedef NS_ENUM(NSInteger, BTThreeDSecureShippingMethod) {
...
 @property (nonatomic) BOOL dataOnlyRequested;
 
+/**
+ Optional. An authentication created using this property should only be used for adding a payment method to the merchant's vault and not for creating transactions.
+ Defaults to BTThreeDSecureAddCardChallengeUnspecified.
+ If set to BTThreeDSecureAddCardChallengeRequested, the authentication challenge will be requested from the issuer to confirm adding new card to the merchant's vault.
+ If set to BTThreeDSecureAddCardChallengeNotRequested the authentication challenge will not be requested from the issuer.
+ If set to BTThreeDSecureAddCardChallengeUnspecified, when the amount is 0, the authentication challenge will be requested from the issuer.
+ If set to BTThreeDSecureAddCardChallengeUnspecified, when the amount is greater than 0, the authentication challenge will not be requested from the issuer.
+ */
```
scannillo, line 164:
```suggestion
/**
 Optional. An authentication created using this property should only be used for adding a payment method to the merchant's vault and not for creating transactions.
 
 Defaults to BTThreeDSecureAddCardChallengeUnspecified.
 
 If set to BTThreeDSecureAddCardChallengeRequested, the authentication challenge will be requested from the issuer to confirm adding new card to the merchant's vault.
 If set to BTThreeDSecureAddCardChallengeNotRequested the authentication challenge will not be requested from the issuer.
 If set to BTThreeDSecureAddCardChallengeUnspecified, when the amount is 0, the authentication challenge will be requested from the issuer.
 If set to BTThreeDSecureAddCardChallengeUnspecified, when the amount is greater than 0, the authentication challenge will not be requested from the issuer.
 */
```

scannillo: Nitpick / take it or leave it

UnitTests/BraintreeThreeDSecureTests/BTPaymentFlowDriver+ThreeDSecure_Tests.swift
```diff
@@ -76,6 +78,41 @@ class BTPaymentFlowDriver_ThreeDSecure_Tests: XCTestCase {
         waitForExpectations(timeout: 3, handler: nil)
     }
 
+    func testPerformThreeDSecureLookup_whenAddCardChallengeNotRequested_sendsAddCardFalse() {
```
scannillo, line 81:
```suggestion
    func testPerformThreeDSecureLookup_whenCardAddChallengeNotRequested_sendsAddCardFalse() {
```

UnitTests/BraintreeThreeDSecureTests/BTPaymentFlowDriver+ThreeDSecure_Tests.swift
```diff
@@ -76,6 +78,41 @@ class BTPaymentFlowDriver_ThreeDSecure_Tests: XCTestCase {
...
+        driver.performThreeDSecureLookup(threeDSecureRequest) { (lookup, error) in
+            XCTAssertFalse(self.mockAPIClient.lastPOSTParameters!["cardAdd"] as! Bool)
+
+            expectation.fulfill()
+        }
+
+        waitForExpectations(timeout: 3, handler: nil)
+    }
+
+    func testPerformThreeDSecureLookup_whenCardAddChallengeRequestedNotSet_doesNotSendAddCardParameter() {
```
scannillo, line 99:
```suggestion
    func testPerformThreeDSecureLookup_whenCardAddChallengeRequestedNotSet_doesNotSendCardAddParameter() {
```

UnitTests/BraintreeThreeDSecureTests/BTPaymentFlowDriver+ThreeDSecure_Tests.swift
scannillo: Do we also need a test for if the cardAddChallenge was set to `.requested`? 🤔

sarahkoop: The `.requested` case is covered in `testPerformThreeDSecureLookup_sendsAllParameters`

scannillo: Ah gotcha - good call

scannillo (approved):
🚀

sshropshire (approved):

sarahkoop:
Flakey UI test for PayPal cancellation flow is failing - merging this unrelated change.

PR #723, opened 2021-08-09
V6 - Bump to iOS 13+

scannillo:
### Summary of changes

* Bump minimum supported deployment target to iOS 13+
  * Bump each XCode target
  * Bump package managers (Podspec & Package.swift)
* Require Xcode 13
* Remove no-longer-needed API annotations
* API changes
  * Remove `BTAppContextSwitcher.handleOpenURL` in favor of `handleOpenURLContext`
  * In iOS 13, [Apple has moved some of the AppDelegate responsibilities to the SceneDelegate](https://www.donnywals.com/understanding-the-ios-13-scene-delegate/). So now, merchants should only need to call `handleOpenURL` from their `SceneDelegate`, with uses a `URLContext`.
* Add migration guide
  * We want to update the migration per-PR change, to make it more manageable than changing it all at once at the end of the next-major-version work.
* Use Swift 5.5 build tools for CocoaPods & SPM.

### Checklist

- [X] Added a changelog entry

### Authors
@sarahkoop @scannillo


sshropshire (approved):
🌊

Sources/BraintreeCore/Public/BraintreeCore/BTAppContextSwitcher.h
```diff
@@ -59,21 +59,13 @@ NS_ASSUME_NONNULL_BEGIN
...
 + (void)setReturnURLScheme:(NSString *)returnURLScheme;
 
-/**
- Handles a return from app context switch
-
- @param url The URL that was opened to return to your app
- @return `YES` if the app switch successfully handled the URL, or `NO` if the attempt to handle the URL failed.
-*/
-+ (BOOL)handleOpenURL:(NSURL *)url NS_SWIFT_NAME(handleOpenURL(_:));
-
```
scannillo, line 69: Thinking through this again - I actually am not sure we should delete this method. I initially was thinking through the lens of "if a merchant is **creating** an app using XCode 13 and targeting at the minimum iOS 13+, it will be auto-configured to use scenes. Removing the scene support is not encouraged."

However I forgot to think of this case "A merchant who **created** their app in a previous Xcode version (before scenes), and hasn't updated to add scene support. Example - a merchant started their app in Xcode 11, iOS 12+. Now they're updating their app to run on Xcode 13 and support iOS 13+." In doing this, Xcode doesn't enforce adding scene support. Here's an example of when we [updating our Demo app](https://github.com/braintree/braintree_ios/commit/e61f74f82ee84c47d7477ddd76b6c251f6af56c1) to include scene support.

TLDR: I don't think all merchants will necessarily be adding scene support to older projects. We should keep this method around but clarify that this one is meant for use if your app "Does not support scenes" and the other method is meant for if your app "does support scenes".

Apologies for the back and forth on this one, Sarah! Let me know if this thinking makes sense.

Sources/BraintreeCore/Public/BraintreeCore/BTAppContextSwitcher.h
```diff
@@ -5,25 +5,25 @@ NS_ASSUME_NONNULL_BEGIN
...
  @return `YES` when the SDK can process the return URL
 */
 + (BOOL)canHandleReturnURL:(NSURL *)url NS_SWIFT_NAME(canHandleReturnURL(_:));
 
 /**
- Complete payment flow after returning from app or browser switch.
+ :nodoc: Complete payment flow after returning from app or browser switch.
 
- @param url The URL you receive in `scene:openURLContexts:` (or `application:openURL:options:` if iOS 12)
+ @param url The URL you receive in `scene:openURLContexts:`  (or `application:openURL:options:` if not using SceneDelegate)
```
scannillo, line 26:
```suggestion
 @param url The URL you receive in `scene:openURLContexts:` (or `application:openURL:options:` if not using SceneDelegate)
```

Sources/BraintreeCore/Public/BraintreeCore/BTAppContextSwitcher.h
```diff
@@ -60,20 +60,24 @@ NS_ASSUME_NONNULL_BEGIN
 + (void)setReturnURLScheme:(NSString *)returnURLScheme;
 
 /**
- Handles a return from app context switch
-
+ Handles a return from app context switch.
+
```
scannillo, line 64:
```suggestion

```

scannillo:
✅  this looks ready to me! We can wait for CI to pass

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* Breaking Changes
+    * Bump minimum supported deployment target to iOS 13+
+    * Require Xcode 13
+    * Bump Swift Tools Version to 5.5 for CocoaPods & SPM
```
scannillo, line 7:
```suggestion
* Breaking Changes
  * Bump minimum supported deployment target to iOS 13+
  * Require Xcode 13
    * Bump Swift Tools Version to 5.5 for CocoaPods & SPM
```

V6_MIGRATION.md
```diff
@@ -0,0 +1,14 @@
...
+
+_Documentation for v6 will be published to https://developer.paypal.com/braintree/docs once it is available for general release._
+
+## Table of Contents
+
+1. [Supported Versions](#supported-versions)
+
+## Supported Versions
+
+v6 supports a minimum deployment target of iOS 13+. It requires the use of Xcode 13+ and Swift 5.1+. If your application contains Objective-C code, the `Enable Modules` build setting must be set to `YES`.
```
scannillo, line 13:
```suggestion
v6 supports a minimum deployment target of iOS 13+. It requires the use of Xcode 13+. If your application contains Objective-C code, the `Enable Modules` build setting must be set to `YES`.
```

PR #727, opened 2021-08-16
iOS v6 - Update Venmo Payment Context Flow

scannillo:
### Summary

> NEXT_MAJOR_VERSION - make paymentMethodUsage a required param in the initializer & remove UNSPECIFIED case from enum

This PR addresses the above comment to achieve parity with Android v4.

### Changes

- Remove `BTVenmoPaymentMethodUsageUnspecified` case.
- Require merchant to initialize `BTVenmoRequest` with a `paymentMethodUsage` (single or multi use)
  - Always use the Venmo PaymentContext flow.
- Update UnitTests accordingly.

### Checklist

- [X] Added a CHANGELOG entry
- [X] Add MIGRATION guide entry

### Authors
@scannillo @sarahkoop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,13 @@
 # Braintree iOS SDK Release Notes
 
-## unreleased
+## unreleased (next-major-version)
 * Breaking Changes
   * Bump minimum supported deployment target to iOS 13+
   * Require Xcode 13
     * Bump Swift Tools Version to 5.5 for CocoaPods & SPM
+  * BraintreeVenmo
+    * Remove `.unspecified` case from `BTVenmoPaymentMethodUsage` enum
+    * Require `paymentMethodUsage` param in `BTVenmoRequest` initializer
 
 ## unreleased
 * Re-organize `/Frameworks` binaries into nested `/FatFrameworks` and `/XCFrameworks` directories.
--- a/Demo/MockVenmo/AppSwitcher.swift
+++ b/Demo/MockVenmo/AppSwitcher.swift
@@ -45,5 +45,4 @@ class AppSwitcher {
             .value
             .flatMap({ URL(string: $0) })
     }
-
 }
--- a/Demo/MockVenmo/AppSwitcher.swift
+++ b/Demo/MockVenmo/AppSwitcher.swift
@@ -45,4 +45,10 @@ class AppSwitcher {
             .value
             .flatMap({ URL(string: $0) })
     }
+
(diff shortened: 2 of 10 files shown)


Demo/MockVenmo/AppSwitcher.swift
```diff
@@ -45,4 +45,10 @@ class AppSwitcher {
             .value
             .flatMap({ URL(string: $0) })
     }
+
+    // TODO: We can update this MockVenmo returnURL to include the `resource_id` to more closely mimic the VenmoContext Flow.
+    // OPTION 1: We add another "SUCCESS-CONTEXT" type button which triggers the context flow,
+    // this way we can easily test both.
+    // OPTION 2: We just update the existing "SUCCESS" button URL.
+
```
scannillo, line 53: I am currently investigating with the PwVenmo team if this is possible (see #service-pw-venmo). My gut says that they won't have a static test `resource_id` sandbox value for us to mimic the `returnURL` with. But, it is worth investigation!

scannillo: I don't think this is going to happen. I will remove this.

CHANGELOG.md
```diff
@@ -1,10 +1,13 @@
 # Braintree iOS SDK Release Notes
 
-## unreleased
+## unreleased (next-major-version)
 * Breaking Changes
   * Bump minimum supported deployment target to iOS 13+
   * Require Xcode 13
     * Bump Swift Tools Version to 5.5 for CocoaPods & SPM
+  * BraintreeVenmo
+    * Remove `.unspecified` case from `BTVenmoPaymentMethodUsage` enum
```
sarahkoop, line 9: Do we want to note the enum number change here too?

sshropshire (approved):

CHANGELOG.md
scannillo: Hm - I don't think we need to call it out. Obj-C devs should know about how enums use the first value as the default (or zero) case. If you think it is confusing though, we can call it out explicitily!

sarahkoop: Ah no I forgot we already discussed that 👍

Demo/MockVenmo/AppSwitcher.swift
```diff
@@ -45,4 +45,5 @@ class AppSwitcher {
             .value
             .flatMap({ URL(string: $0) })
     }
+
```
scannillo, line 48:
```suggestion
```

PR #734, opened 2021-09-01
[iOS v6] Remove Fat Frameworks

scannillo:
### Summary of changes

This PR gives me so much joy 🥲

So, for v6 we can require Carthage 0.38.0+, which offers full support for xcframeworks. Xcframeworks are the encouraged binary type by Apple at this point, so this is the direction we should go.

#### Changes:

- Remove Fat Frameworks in favor of only using xcframeworks.

---

**NOTE:**
Another thought we can punt on is not supporting Carthage at all for v6. This [thread](https://github.com/braintree/braintree_ios/discussions/705) started to discuss. Sumo numbers currently show 10% still on Carthage for iOS v5 in sand, so I lean toward keeping it. We can chat about this in a stand.

**REMINDER:**
[Our hosted-assets ](https://github.braintreeps.com/braintree/hosted-client-libraries/tree/master/mobile/ios/carthage-frameworks)should remain, so that old SDK versions using Carthage can properly download the binaries. However, going forward for new SDK versions, we only need to include xcframeworks 😄 .

_If we offer 0 Carthage support in v6, then we don't need hosted-assets for future SDK versions._

---

### Checklist

- [ ] Added a changelog entry

### Authors
@scannillo


sarahkoop:
I know reorganizing the frameworks folder isn't a small task - but do we want to keep the XCFrameworks in a subfolder still or move them back to the top level `Frameworks` directory?

scannillo:
> I know reorganizing the frameworks folder isn't a small task - but do we want to keep the XCFrameworks in a subfolder still or move them back to the top level Frameworks directory?

Lol ... _it definitely isn't_. This crossed my mind too. The reason I went with leaving it as-is, was that ... the potential that one of these 3rd party doesn't give us an xcframework for a certain version release and we have to add back in a fat framework. We also can make this change at a later point/non breaking change.

But I went back and forth on it. Thoughts?

sarahkoop (approved):
Leaving them in a subfolder makes sense to me - this looks good 🌟

sshropshire (approved):

PR #741, opened 2021-09-24
PwPP: Add correlation id to BTPayPalRequest

demerino:
### Summary of changes

- Add `correlation_id` to BTPayPalRequest.

Allows for the linking of STC and RDA _before_ the PwPP flow starts.

### Checklist

- [ ] Added a changelog entry

### Authors
- @demerino

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalRequest.m
+++ b/Sources/BraintreePayPal/BTPayPalRequest.m
@@ -53,6 +53,10 @@ NSString *const BTPayPalCallbackURLScheme = @"sdk.ios.braintree";
         parameters[@"merchant_account_id"] = self.merchantAccountID;
     }
 
+    if (self.correlationId) {
+        parameters[@"correlation_id"] = self.correlationId;
+    }
+
     if (self.shippingAddressOverride) {
         experienceProfile[@"address_override"] = @(!self.isShippingAddressEditable);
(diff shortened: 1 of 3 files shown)


Sources/BraintreePayPal/BTPayPalRequest.m
```diff
@@ -53,6 +53,10 @@ - (NSString *)landingPageTypeAsString {
         parameters[@"merchant_account_id"] = self.merchantAccountID;
     }
 
+    if (self.correlationId) {
```
crookedneighbor, line 56: Should this be `clientMetadataId` instead? That's what a merchant will get from the GraphQL call.

scannillo (approved):
Looks good! +1 to Blade's comment on checking out which naming convention is more merchant-friendly.

Sources/BraintreePayPal/BTPayPalRequest.m
demerino: I think correlation id is much clearer and aligns with what the value actually is: an ID that correlates a bunch of data (in this case, STC data and RDA data). I am not sure why graphQL chose `clientMetadataId` :(

Even the value that the gateway takes is called `correlation_id`

sshropshire (approved):
🌊

PR #744, opened 2021-10-05
[NEXT_MAJOR_VERSION] Release Carthage xcframeworks only

scannillo:
### Summary

- This piggybacks off [this PR](https://github.com/braintree/braintree_ios/pull/734).
- The latest Carthage version (0.38.0) adds support for xcframework binaries, and it actually works really well. We will be requiring this version of Carthage (and up) for iOS v6. So, let's only upload the binaries for this framework type (it is the encouraged/recommended type by both Apple & Carthage).
- This PR makes it so that the only pre-uploaded binaries that we provide are in the xcframework format.

### Checklist

- ~Added a changelog entry~

I didn't think we needed to call this out, since we already call out that we're bumping to 0.38.0+ in our previous PR. Open to thoughts if folks disagree, though!

### Authors
@scannillo


sarahkoop (approved):
Looks good to me - I agree no CHANGELOG needed.

PR #752, opened 2021-11-01
Fix Carthage Release Steps for .Frameworks

scannillo:
### Summary of changes

In our latest iOS release attempt we ran into this [issue](https://github.com/braintree/braintree_ios/runs/4071524081?check_suite_focus=true): `Building universal frameworks with common architectures is not possible. The device and simulator slices for "<Framework>" both build for: arm64`. 

This looks to be a known issue when creating `.framework` binaries using XCode 13. [See Carthage issue here](https://github.com/Carthage/Carthage/issues/3146#issuecomment-927781844).

According to that GH issue, the fix for this seems to be going back to the old ways of using a special `carthage.sh` script.

### Changes

* Re-add `carthage.sh` script, for XCode 13
* Use `carthage.sh` script when creaing `.framework` binaries for release
* I noticed our Rakefile was missing a task for building `.framework` binaries for Carthage, so I added it
    * _We only really use the Rakefile at this point as a fallback for when automated iOS releases fail._
* Add additional task to `Carthage` CI build workflow to build frameworks _without_ the `--use-xframeworks` option. Previously, our CI on PRs wasn't able to catch this since our PR CI was only testing with the `--use-xframeworks` option.

### Notes

In the next-major-version, we are removing `.framework` support entirely 👯 . (See [PR](https://github.com/braintree/braintree_ios/pull/744)).

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


sshropshire (approved):

.github/workflows/build.yml
```diff
@@ -83,7 +83,11 @@ jobs:
         run: cd SampleApps/CarthageTest && carthage update --use-xcframeworks
       - name: Build CarthageTest
         run: xcodebuild -project 'SampleApps/CarthageTest/CarthageTest.xcodeproj' -scheme 'CarthageTest' clean build CODE_SIGNING_ALLOWED=NO
-  spm_xcode_beta:
+      - name: Run carthage build without --use-xcframeworks
+        run: |
+          ./carthage.sh build --no-skip-current
+          ./carthage.sh archive "BraintreeAmericanExpress" "BraintreeApplePay" "BraintreeCard" "BraintreeCore" "BraintreeDataCollector" "BraintreePaymentFlow" "BraintreePayPal" "BraintreeThreeDSecure" "BraintreeUnionPay" "BraintreeVenmo" "PayPalDataCollector" --output Braintree.framework.zip
```
scannillo, line 89: I'm not sure it's super valuable to test the archive on the PR CI. I think we can remove that and just test that the frameworks build.

sarahkoop (approved):

.github/workflows/build.yml
scannillo: Opting to leave as-is for now.

carthage.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# carthage.sh
+# Usage example: ./carthage.sh build --platform iOS
+
+set -euo pipefail
+
+xcconfig=$(mktemp /tmp/static.xcconfig.XXXXXX)
+trap 'rm -f "$xcconfig"' INT TERM HUP EXIT
+
+# For Xcode 12 make sure EXCLUDED_ARCHS is set to arm architectures otherwise
```
scannillo, line 11:
```suggestion
# For Xcode 13 make sure EXCLUDED_ARCHS is set to arm architectures otherwise
```

PR #753, opened 2021-11-04
Add note to Rakefile

scannillo:
### Summary of changes

- Add basic note to Rakefile clarifying why we use this rake task (it came up recently with Wayfair).


sshropshire (approved):

Rakefile
```diff
@@ -177,6 +177,8 @@ namespace :spm do
     run! 'git checkout SampleApps/SPMTest'
   end
 
+
+  # Note: This task is for edge case LE merchants who insist on integrating directly with xcframeworks, though we encourage integrating directly through a package manager instead.
```
sshropshire, line 181:
```suggestion
  # Note: This task is for edge case merchants who insist on integrating directly with xcframeworks, though we encourage integrating directly through a package manager instead.
```
Random nit, since this could apply to any merchant that wants to integrate directly.

PR #778, opened 2022-01-12
Add `BTCardClientErrorTypeCardAlreadyExists`

sarahkoop:
### Summary of changes

- Add explicit error case for error code returned from Gateway when a card already exists in the vault.

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais 
- @sarahkoop

### DIFF
--- a/Sources/BraintreeCard/BTCardClient.m
+++ b/Sources/BraintreeCard/BTCardClient.m
@@ -235,20 +235,20 @@ NSString *const BTCardClientGraphQLTokenizeFeature = @"tokenize_credit_cards";
 
 - (NSError *)cardTokenizationErrorForError:(NSError *)error {
     NSError *callbackError = error;
-    NSDictionary<NSErrorUserInfoKey,id> *errorMap = error.userInfo;
-    BTJSON *errorFields = [errorMap objectForKey:BTHTTPJSONResponseBodyKey];
-    
-    BTJSON *errorCode = errorFields[@"fieldErrors"][0][@"fieldErrors"][0][@"code"];
-    
-    // Gateway error code for card already exists
-    if ([errorCode.asString  isEqual: @"81724"]) {
-        callbackError = [NSError errorWithDomain:BTCardClientErrorDomain
-                                            code:BTCardClientErrorTypeCardAlreadyExists
-                                        userInfo:[self.class validationErrorUserInfo:error.userInfo]];
-    } else {
-        callbackError = [NSError errorWithDomain:BTCardClientErrorDomain
-                                            code:BTCardClientErrorTypeCustomerInputInvalid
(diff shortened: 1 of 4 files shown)


Sources/BraintreeCard/Public/BraintreeCard/BTCardClient.h
```diff
@@ -26,6 +26,9 @@ typedef NS_ENUM(NSInteger, BTCardClientErrorType) {
 
     /// Customer provided invalid input
     BTCardClientErrorTypeCustomerInputInvalid,
+    
+    /// Card already exists as a saved payment method
+    BTCardClientErrorTypeCardAlreadyExists,
```
sarahkoop, line 31: Pending discussion of whether or not this should be considered a breaking change

Sources/BraintreeCard/BTCardClient.m
```diff
@@ -237,4 +233,24 @@ - (BOOL)isGraphQLEnabledForCardTokenization:(BTConfiguration *)configuration {
     return graphQLFeatures && [graphQLFeatures containsObject:BTCardClientGraphQLTokenizeFeature];
 }
 
+- (NSError *)cardTokenizationErrorForError:(NSError *)error {
```
crookedneighbor, line 236: I find this name a little confusing. Can you explain the `ErrorForError` bit?

Sources/BraintreeCard/Public/BraintreeCard/BTCardClient.h
crookedneighbor: I don't think this is a breaking change. As long as the merchant is handling a catch all error of "something went wrong" or just passing back the error message from the error directly, it should be fine.

Sources/BraintreeCard/BTCardClient.m
sarahkoop: We updated this helper method name to clarify. Let us know if it needs more refining 👍

Sources/BraintreeCard/BTCardClient.m
```diff
@@ -237,4 +244,20 @@ - (BOOL)isGraphQLEnabledForCardTokenization:(BTConfiguration *)configuration {
     return graphQLFeatures && [graphQLFeatures containsObject:BTCardClientGraphQLTokenizeFeature];
 }
 
+- (NSError *)constructCallbackErrorResponseForErrorCode:(BTJSON *)errorCode error:(NSError *)error {
```
sshropshire, line 247:
```suggestion
- (NSError *)constructCallbackErrorForErrorCode:(String *)errorCode error:(NSError *)error {
```
Super nit.

sshropshire (approved):

cgdibble (approved):

PR #789, opened 2022-02-23
Cache Configuration Updates

jaxdesmarais:
### Summary of changes

- This updates our HTTP configuration requests to now cache the configuration for subsequent requests
- Adds a timeout to invalidate the cached configuration after 5 minutes

### Checklist

- [x] Added a changelog entry

- @sarahkoop @jaxdesmarais


Sources/BraintreeCore/BTHTTP.m
```diff
@@ -110,7 +112,11 @@ - (void)GET:(NSString *)aPath completion:(void(^)(BTJSON *body, NSHTTPURLRespons
 }
 
 - (void)GET:(NSString *)aPath parameters:(NSDictionary *)parameters completion:(void(^)(BTJSON *body, NSHTTPURLResponse *response, NSError *error))completionBlock {
-    [self httpRequest:@"GET" path:aPath parameters:parameters completion:completionBlock];
+    if ([aPath containsString:@"configuration"]) {
```
demerino, line 115: What about adding a `GET cache:TRUE....` method instead of using this branching logic? Then future GETS could leverage it.

sarahkoop: Yes good idea - updated this to have a `shouldCache` param and made the internal method names more generic 👍

Sources/BraintreeCore/BTAPIClient.m
```diff
@@ -137,31 +137,6 @@ + (BTAPIClientAuthorizationType)authorizationTypeForAuthorization:(NSString *)au
     }
 }
 
-- (instancetype)copyWithSource:(BTClientMetadataSourceType)source
```
sarahkoop, line 140: Using this method prevents the `BTAPIClient` from sharing a cache, we are removing this method so that feature clients can be instantiated with a single `BTAPIClient` and share a cache

sshropshire: Is this a breaking change?

Sources/BraintreeCore/BTHTTP.m
```diff
@@ -139,15 +149,56 @@ - (void)DELETE:(NSString *)aPath parameters:(NSDictionary *)parameters completio
...
+            cachedResponse = nil;
+        }
+
+        if (cachedResponse != nil) {
+            [self handleRequestCompletion:cachedResponse.data response:cachedResponse.response error:nil completionBlock:completionBlock];
+        } else {
+            NSURLSessionTask *task = [self.session dataTaskWithRequest:request completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
+                if (data != nil && response != nil) {
+                    NSCachedURLResponse *cachedURLResponse = [[NSCachedURLResponse alloc]initWithResponse:response data:data];
+                    [[NSURLCache sharedURLCache] storeCachedResponse:cachedURLResponse forRequest:request];
```
sshropshire, line 172: This could end up caching a failed response, like if it comes back as HTTP `5xx`.

Sources/BraintreeCore/BTAPIClient.m
Epreuve: Shouldn't be?  iirc `-` is private vs `+` public?

UnitTests/BraintreePayPalTests/BTPayPalDriver_Tests.swift
```diff
@@ -626,7 +626,6 @@ class BTPayPalDriver_Tests: XCTestCase {
         let payPalDriver = BTPayPalDriver(apiClient: apiClient)
 
         XCTAssertEqual(payPalDriver.apiClient?.metadata.integration, BTClientMetadataIntegrationType.custom)
-        XCTAssertEqual(payPalDriver.apiClient?.metadata.source, BTClientMetadataSourceType.payPalBrowser)
```
Epreuve, line 629: Should we rename this test or remove it?

Sources/BraintreeCore/BTAPIClient.m
sarahkoop: Ah yes good catch - put back the method and tests and added a next major version note 👍

UnitTests/BraintreeVenmoTests/BTVenmoDriver_Tests.swift
```diff
@@ -613,7 +613,6 @@ class BTVenmoDriver_Tests: XCTestCase {
         let venmoDriver = BTVenmoDriver(apiClient: apiClient)
         
         XCTAssertEqual(venmoDriver.apiClient.metadata.integration, BTClientMetadataIntegrationType.custom)
-        XCTAssertEqual(venmoDriver.apiClient.metadata.source, BTClientMetadataSourceType.venmoApp)
```
Epreuve, line 616: Similarly, rename or remove?  Is there no default source now, and if so, will this bork analytics going forward?

jaxdesmarais: https://github.com/braintree/braintree_ios/pull/789#discussion_r814809470

Sources/BraintreeCore/BTHTTP.m
jaxdesmarais: Good catch - [added a check here for this](https://github.com/braintree/braintree_ios/blob/4d649671887c2ea4da6486be37ed2a594b3925ba/Sources/BraintreeCore/BTHTTP.m#L342-L346)

demerino (approved):
This looks good! And we did a round of manual testing?

sarahkoop:
> This looks good! And we did a round of manual testing?

Yes we manually tested with debug points to confirm the cache was being used and invalidated after 5 minutes 👍

cgdibble (approved):

Sources/BraintreeCore/BTHTTP.m
```diff
@@ -295,6 +338,12 @@ - (void)handleRequestCompletion:(NSData *)data response:(NSURLResponse *)respons
         }
         return;
     }
+    
+    // We should only cache the response if we do not have an error
+    if (request != nil && shouldCache) {
```
sshropshire, line 343: We might need to check the HTTP response code. I'm not sure all the possible ones the gateway can produce, but  anything in the `2xx` range is probably good to cache.

sshropshire: Oh ok I didn't see the checks in the diff I had to expand the code fold 🙌

PR #792, opened 2022-03-02
Add SEPA Debit Model Classes

sarahkoop:
### Summary of changes

- Implement `BTSEPADebitRequest` and `BTSEPADebitNonce` 
- Add BTSEPADebitClient with methods to be implemented in future PR

### Checklist

- @jaxdesmarais 
- @sarahkoop


sshropshire (approved):

Sources/BraintreeSEPADebit/BTSEPADebitNonce.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+/// A  payment method nonce representing a SEPA Debit payment.
+@objcMembers public class BTSEPADebitNonce {
```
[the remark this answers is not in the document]
kaylagalway: @sarahkoop @jaxdesmarais and I all chatted and we will create a research ticket to look a little more into what we can do about this

UnitTests/BraintreeSEPADirectDebitTests/BTSEPADirectDebitClient_Tests.swift
```diff
@@ -0,0 +1,7 @@
+import XCTest
```
kaylagalway, line 1: Do we need to add this test class before we've added any XCTestCases?

Braintree.xcodeproj/project.pbxproj
```diff
@@ -6504,9 +6605,104 @@
...
+				GCC_C_LANGUAGE_STANDARD = gnu11;
+				GCC_DYNAMIC_NO_PIC = NO;
+				GCC_OPTIMIZATION_LEVEL = 0;
+				GCC_PREPROCESSOR_DEFINITIONS = (
+					"DEBUG=1",
+					"$(inherited)",
+				);
+				GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR;
+				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
+				INFOPLIST_FILE = "BraintreePaymentFlowTests copy-Info.plist";
```
kaylagalway, line 6640: This will need to be deleted as well

kaylagalway (commented):
BraintreePaymentFlowTests has a duplicate that just needs to be removed

Braintree.xcodeproj/project.pbxproj
```diff
@@ -6504,9 +6605,104 @@
...
+				DEVELOPMENT_TEAM = 43253H4X22;
+				ENABLE_NS_ASSERTIONS = NO;
+				FRAMEWORK_SEARCH_PATHS = (
+					"$(inherited)",
+					"$(PROJECT_DIR)/Frameworks",
+				);
+				GCC_C_LANGUAGE_STANDARD = gnu11;
+				GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR;
+				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
+				INFOPLIST_FILE = "BraintreePaymentFlowTests copy-Info.plist";
```
kaylagalway, line 6685: This one as well

Sources/BraintreeSEPADebit/BTSEPADebitNonce.swift
```diff
@@ -0,0 +1,28 @@
...
+    /// The IBAN last four characters.
+    public let ibanLastFour: String?
+    
+    /// The customer ID.
+    public let customerID: String?
+    
+    /// The `BTSEPADebitMandateType`.
+    public let mandateType: BTSEPADebitMandateType?
+        
+    init(json: BTJSON) {
```
kaylagalway: I realized this should probably be an optional init right? `init?` I looked at our other modules and the `BTPaymentMethodNonce` objective-c init and it seems like it is nullable.

UnitTests/BraintreeSEPADirectDebitTests/BTSEPADirectDebitClient_Tests.swift
jaxdesmarais: I think this is fine for now as this is all going into a feature branch and we will add tests as we add implementation pieces and will for sure add client tests.

Sources/BraintreeSEPADebit/BTSEPADebitNonce.swift
jaxdesmarais: 🤔 I think we can to align with the Objective-C inits today, but am going to add a next major version note to remove non-optional internal inits in the future. I am not tied to one way or the other and I think we'd need to make this optional anyways if we can eventually inherit from `BTPaymentMethodNonce` because the super init will be optional which will force us to make this optional as well.

kaylagalway: I think the `BTPaymentMethodNonce` relies on the `nonce` value not being nil, otherwise it returns `nil` from it's nullable init. We can probably follow the same pattern, make the `nonce` non-optional on `BTSepaDebitNonce` and unwrap it in this `init?`

kaylagalway (approved):
Thanks for the updates!

PR #795, opened 2022-03-10
Add SEPA Direct Debit to Demo App

jaxdesmarais:
### Summary of changes

- Add SEPA Direct Debit to Demo App

Note: there is not functionality as currently the tokenize method is not implimented

### Checklist

- @jaxdesmarais @sarahkoop


Demo/Application/Base/BraintreeDemoContainmentViewController.m
```diff
@@ -253,6 +267,18 @@ - (void)containIntegrationViewController:(UIViewController *)viewController {
     return block;
 }
 
+- (void (^)(NSString *paymentMethodNonceString))nonceStringCompletionBlock {
+    static id block;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
```
sshropshire, line 273: q: Does this prevent retries?

sshropshire: May not be important since it's a demo app but having to rebuild the app to retry could make troubleshooting more difficult.

sshropshire (approved):

Demo/Application/Base/BraintreeDemoContainmentViewController.m
```diff
@@ -76,15 +85,19 @@ - (void)updateStatus:(NSString *)status {
 - (void)tappedStatus {
     NSLog(@"Tapped status!");
 
-    if (self.latestTokenizedPayment) {
-        NSString *nonce = self.latestTokenizedPayment.nonce;
+    if (self.latestTokenizedPayment || self.latestTokenizedPaymentString) {
```
crookedneighbor, line 88: Can I get some context on this change? Why do we have to ways of retrieving the nonce now here?

jaxdesmarais: Yeah, so [this comment](https://github.com/braintree/braintree_ios/pull/792#discussion_r818851499) goes into a bit more details, but it has to do with some Objective-C intricacies with Swift. We have a research story for this to see if it's possible, but has not been an obvious solution yet. Since this is going into a feature branch (and not master), we plan to revisit it before merging in the feature branch. Since `BTPaymentMethodNonce` is public we also will need to find a way to force what we want to happen while allowing us to do it without breaking changes and also avoiding a mixed module (as we cannot just add a bridging header to this module because that would break our Swift Package Manager integration since it doesn't allow Obj-C + Swift to be mixed in a single module).

Demo/Application/Base/BraintreeDemoContainmentViewController.m
```diff
@@ -64,6 +65,14 @@ - (void)setLatestTokenizedPayment:(id)latestPaymentMethodOrNonce {
     }
 }
 
+- (void)setLatestTokenizedPaymentString:(NSString*)latestNonceString {
+    _latestTokenizedPaymentString = latestNonceString;
```
kaylagalway, line 69: any reason you chose to use `_` to access the variable instead of `self` ? Seems like a typical setter through `self` would be okay here, unless I'm missing a BT requirement/piece of funcationality

Demo/Application/Base/BraintreeDemoContainmentViewController.m
kaylagalway: Would it be possible to add some comments in this function to give merchants insight into what is happening here when they check out our demo app? I agree that the reason for using `dispatch` is unclear, will probably be unclear to them as well?

kaylagalway (approved):
Had one question, but looks good otherwise - agree with Steven's comments about `dispatch` and I wonder if we could add some comments to the code of that function.

Demo/Application/Base/BraintreeDemoContainmentViewController.m
jaxdesmarais: Nope, not a specific requirement. Used the `_` instead of `self` to match the above `setLatestTokenizedPayment` method that is used for the `BTPaymentMethodNonce` type. Since we don't have access to that here (currently) we just matched the method implementation. Still hoping we can resolve that issue and remove this function if we are able to get the Swift class to play nice with `BTPaymentMethodNonce`. The demo app as a whole seems to use both, so we can certainly update to use self if we feel strongly.

Demo/Application/Base/BraintreeDemoContainmentViewController.m
jaxdesmarais: We can add docs, though I am not sure I have the context to add clear enough documentation to this as dispatch is used rather extensively throughout the SDK currently. Feels a bit weird to add those docs here but nowhere else where we are also using the legacy API in the demo app? It was used by us to match the implementation where we use a nonce as mixing 2 different ways of returning the nonce felt weird.

sshropshire: We do use dispatch a lot but `dispatch_once` is the confusing part for me here. I'm sure it made sense at one point in time, but a good question from the uninitiated would be "why is this completion block allowed to execute only once?"

jaxdesmarais: My understanding of `dispatch_once` is "perform something once and only once", which is precisely what we're doing and want to do here? We want to get the nonce and ensure the nonce is set as a variable that’s accessible before we execute the next step, which in our case is tapping to make a transaction. We would run into errors if we attempted to retrieve the nonce before this step had completed execution, correct?

This method can be called multiple times (as it is today across all of our ViewControllers). We can run the demo app and make subsistent card transactions (for example) and that calls the completion block and updates the status each time.

Happy to make changes to this, I do think our demo app contains a lot of anti-patterns that I don’t think we’d want merchants to follow. 🙂 Also just trying to wrap my head around this myself!

sshropshire: It may be out of scope for this PR but yeah overall that `static` dispatch token isn't bound to any single instance of `BraintreeDemoContainmentViewController`, it's almost equivalent to a `fileprivate var` in Swift.

jaxdesmarais: Yeah, I think it's probably out of scope here, but totally agree with asking why on it! I think it brings up a great point about how our demo app modernization with a switch to Swift is probably equally as important as the rest of the SDK since it doesn't represent what we would actually want merchant to do in some cases.

PR #796, opened 2022-03-16
Implement Create Mandate Request

jaxdesmarais:
### Summary of changes

- This is a partial implementation of the `SEPADebitClient` around the create mandate portion of the request. Future PRs will cover TODOs and additional functionality.

### Checklist

- @sarahkoop @jaxdesmarais


UnitTests/BraintreeSEPADirectDebitTests/BTSEPADirectDebitClient_Tests.swift
```diff
@@ -1,7 +1,78 @@
...
+                    if error != nil, let error = error as NSError? {
+                        XCTAssertEqual(error.domain, "CannedError")
+                        XCTAssertEqual(error.code, 0)
+                        XCTAssertEqual(error.localizedDescription, "This is a create mandate fake error")
+                    } else if nonce != nil {
+                        XCTFail("This request should fail.")
+                    }
+                }
+        } else {
+            XCTFail("This should not get here as we use iOS 13.")
```
crookedneighbor, line 53: Is this bolierplate necessary? Can the ios 13 check go in a dedicated environment test or something? Or is there a particular reason to guard it here?

UnitTests/BraintreeSEPADirectDebitTests/SEPADirectDebitAPI_Tests.swift
```diff
@@ -0,0 +1,91 @@
...
+        sepaDirectDebitRequest.iban = "FR891751244434203564412313"
+        sepaDirectDebitRequest.customerID = "A0E243A0A200491D929D"
+        sepaDirectDebitRequest.mandateType = .oneOff
+        sepaDirectDebitRequest.billingAddress = billingAddress
+        sepaDirectDebitRequest.merchantAccountID = "eur_pwpp_multi_account_merchant_account"
+    }
+
+    func testBuildHttpRequest_withAllParams() {
+        let expectedHTTPBody = """
+{\"sepa_debit\":{\"customer_id\":\"A0E243A0A200491D929D\",\"mandate_type\":\"ONE_OFF\",\"account_holder_name\":\"John Doe\",\"iban\":\"FR891751244434203564412313\",\"billing_address\":{\"admin_area_2\":\"Annaberg-buchholz\",\"country_code\":\"FR\",\"address_line_2\":\"#170\",\"address_line_1\":\"Kantstraße 70\",\"admin_area_1\":\"Freistaat Sachsen\",\"postal_code\":\"09456\"}},\"merchant_account_id\":\"eur_pwpp_multi_account_merchant_account\",\"cancel_url\":\"https:\\/\\/example.com\",\"return_url\":\"https:\\/\\/example.com\"}
```
sshropshire, line 27: nit: indentation

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
```diff
@@ -16,7 +37,7 @@ import BraintreeCore
     /// Required. The customer ID.
     public var customerID: String?
     
-    /// Required. The `BTSEPADebitMandateType`.
+    /// Optional. The `BTSEPADebitMandateType`. If not set, defaults to `.oneOff`
     public var mandateType: BTSEPADirectDebitMandateType?
```
sshropshire, line 41: nit: I noticed on Android it's non-null with `ONE_OFF` as the default value. Would it make sense to do the same here and avoid the optional type?

sshropshire (approved):
Couple nits but overall LGTM 🔥

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
jaxdesmarais: Yeah, so this is a bit weird. With all of the requests on the SDK, we currently default all of the values to `nil`  and make them optional so that we can do this syntax:
```
let sepaDirectDebitRequest = BTSEPADirectDebitRequest()
sepaDirectDebitRequest.accountHolderName = "John Doe"
sepaDirectDebitRequest. = ...
```
If we don't make them optional, they would be required when you create the object like:
```
let sepaDirectDebitRequest = BTSEPADirectDebitRequest(
    accountHolderName: "John Doe",
    ...
)
```
The latter is the more "Swift-y" way to do it, but it feels odd to make this change for 1 request type in Swift when the others can all use the `.` syntax. It also felt weird to make 1 part of the request non-optional and required when creating the Swift request when the rest of it could use the `.` syntax. We should explore which pattern we want to move forward with as we convert other modules.

We are still defaulting to `.onOff` in the init if not provided by the merchant.

UnitTests/BraintreeSEPADirectDebitTests/SEPADirectDebitAPI_Tests.swift
jaxdesmarais: The test fails with any syntax other than this because of how the object is compiled. Let me dig into it and see if there is a different way to go about this because I agree, this is _ugly_

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
sshropshire:
```suggestion
    public var mandateType: BTSEPADirectDebitMandateType = .oneOff
```
Yeah we did the same thing in Android [SEPADirectDebitRequest.java](https://github.com/braintree/braintree_android/blob/40e579ec02ede738aeab774ce31e187335f2d104/SEPADirectDebit/src/main/java/com/braintreepayments/api/SEPADirectDebitRequest.java#L14) we could do it like this in Swift.

jaxdesmarais: Yeah, so in Swift those defaults are provided in the init instead, [which we currently do in this PR](https://github.com/braintree/braintree_ios/blob/4a6cf68f9bb398a9d05784c72cf5fa2fd58c44cf/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift#L66). If we moved to this, even if we set a default, the merchant would have to set the mandate like so even if using the `.` syntax:
```
let sepaDirectDebitRequest = BTSEPADirectDebitRequest(mandateType: .oneOff)
sepaDirectDebitRequest.accountHolderName = "John Doe"
sepaDirectDebitRequest. = ...
```

Sources/BraintreeSEPADirectDebit/CreateMandateResult.swift
```diff
@@ -0,0 +1,48 @@
+import Foundation
+
+struct CreateMandateResult: Decodable {
```
kaylagalway, line 3: Should we add docs to this struct?

kaylagalway (approved):
lgtm - only left one comment!

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
sshropshire:
```suggestion
    public var mandateType: BTSEPADirectDebitMandateType
```
Oh word yeah my nit was actually with the type being optional. Having it in `init` is smooth, can we do this 👆 instead?

sshropshire: I see cases for both here though but I've always felt like the less optionals the better. On the other hand, it does align nicely with the fact that the merchant isn't required to set this property.

jaxdesmarais: Ah gotcha gotcha. Yeah, I agree. I could go either way with this, honestly. I lean slightly more to defaulting all to optional now to match what we do in obj-c, but would love to move away from optionals as we decide on a pattern moving forward with Swift conversions. Would be interesting to start up a GH discussion to see how many merchants are actually using the `.` syntax since I feel like that could influence our decisions moving forward.

PR #799, opened 2022-03-22
SEPA Direct Debit Mandate Web Flow

jaxdesmarais:
### Summary of changes

- Implement ASWebAuthenticationSession to handle web flow - note there is some duplication of code here as we need to support iOS 12 still. In the next major version we should be able to clean up a lot of this.
- TODOs will be handled in a future PR

### Checklist

- @sarahkoop 
- @jaxdesmarais


Braintree.xcodeproj/project.pbxproj
```diff
@@ -368,7 +369,7 @@
 		BEFE6DA627E27070002BE9C8 /* MockSEPADirectDebitAPI.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEFE6DA527E27070002BE9C8 /* MockSEPADirectDebitAPI.swift */; };
 		BEFE6DA827E289DF002BE9C8 /* MockWebAuthenticationSession.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEFE6DA727E289DF002BE9C8 /* MockWebAuthenticationSession.swift */; };
 		BEFE6DAA27E28A0B002BE9C8 /* MockViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEFE6DA927E28A0B002BE9C8 /* MockViewController.swift */; };
-		BEFE6DAC27E28DFF002BE9C8 /* WebAuthenticationSession.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEFE6DAB27E28DFF002BE9C8 /* WebAuthenticationSession.swift */; };
+		BEFE6DAD27E399AB002BE9C8 /* WebAuthenticationSession.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEFE6DAB27E28DFF002BE9C8 /* WebAuthenticationSession.swift */; };
```
jaxdesmarais, line 372: This changed as the target for this file was incorrect

Demo/Application/Base/BraintreeDemoContainmentViewController.m
```diff
@@ -74,9 +74,11 @@ - (void)setLatestTokenizedPaymentString:(NSString*)latestNonceString {
 }
 
 - (void)updateStatus:(NSString *)status {
-    [(UIButton *)self.statusItem.customView setTitle:NSLocalizedString(status, nil) forState:UIControlStateNormal];
-    [(UIButton *)self.statusItem.customView setTitleColor:UIColor.blackColor forState:UIControlStateNormal];
-    NSLog(@"%@", ((UIButton *)self.statusItem.customView).titleLabel.text);
+    dispatch_async(dispatch_get_main_queue(), ^{
```
jaxdesmarais, line 77: We were not handling the status update for any of our views on the main thread - this now prevents the warnings we were seeing in Xcode.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -36,16 +43,29 @@ import BraintreeCore
             if error != nil {
                 completion(nil, error)
                 return
-            } else if result != nil {
-                // TODO: future PR start ASWebAuthenticationSession with result.approvalURL
+            } else if result != nil, let result = result {
```
kaylagalway, line 46: Double nil check here I think? Could probably change to:
```
if let result = result
```

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -55,8 +75,21 @@ import BraintreeCore
...
+                        switch success {
+                        case true:
+                            // TODO: call BTSEPADirectDebitClient.tokenize
+                            return
+                        case false:
+                            completion(nil, error)
+                            return
+                        }
+                    }
+                }
```
kaylagalway, line 92: We should probably have a final `else` here that passes a generic error in the completion, even if `error` is nil.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -69,4 +102,56 @@ import BraintreeCore
             completion(result, error)
         }
     }
+    
+    func startAuthenticationSessionWithoutContext(
```
kaylagalway, line 106: I was thinking we could add some documentation to these internal sepa functions since the flow won't be documented externally. Might help folks in the future when they are trying to figure out the order of this multi-step SEPA process.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -36,16 +43,29 @@ import BraintreeCore
...
+                        case true:
+                            // TODO: call BTSEPADirectDebitClient.tokenize
+                            return
+                        case false:
+                            completion(nil, error)
+                            return
+                        }
+                    }
+                }
             }
```
kaylagalway, line 61: We should add another else here. If `error` is nil and `result` is nil, we should pass a generic error in the completion.

kaylagalway (commented):
Added a few comments/requests during a pair review

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -36,16 +43,29 @@ import BraintreeCore
...
         }
     }
     
     /// Initiates an `ASWebAuthenticationSession` to display a mandate to the user. Upon successful mandate creation, tokenizes the payment method and returns a result
     /// - Parameters:
     ///   - request: a BTSEPADebitRequest
-    /// - Note: This function should only be used for iOS 12 support.
+    /// - Note: This function should only be used for iOS 12 support. This function cannot be invoked on a device running iOS 13 or higher.
```
sshropshire, line 68: I think there's a way to annotate this method as only being available up to and including iOS 12.

jaxdesmarais: I don't _think_ the `@available` tags allow for this. I know they are introducing an `@unavailable` tag in Swift 5.6 that would allow us to do that but don't think it exists with what we have currently.

Sources/BraintreeSEPADirectDebit/CreateMandateResult.swift
```diff
@@ -39,7 +39,9 @@ struct CreateMandateResult: Decodable {
         let bodyContainer = try messageContainer.nestedContainer(keyedBy: CodingKeys.self, forKey: .body)
         let sepaDebitContainer = try bodyContainer.nestedContainer(keyedBy: CodingKeys.self, forKey: .sepaDebitAccount)
         
-        approvalURL = try sepaDebitContainer.decode(String.self, forKey: .approvalURL)
+        // Defaulting the approval URL to the string "null" if the API returns nil for this field because the
+        // mandate has already been approved.
+        approvalURL = try sepaDebitContainer.decodeIfPresent(String.self, forKey: .approvalURL) ?? "null"
```
kaylagalway, line 44: Ah I missed this the first time - but we should probably create the constant for `"null"` in this class. Then we should reference that constant when we check it's value later. That way if for any reason this `"null"` string is ever changed to something else, the things that check it would be automatically changed to check the correct value as well.

kaylagalway (commented):
One last request - apologies for missing it the first time

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
kaylagalway: Same comment for here, could use guards for both the `error != nil` and the `result` check. 
https://github.com/braintree/braintree_ios/pull/799/files#r833424928

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
kaylagalway: If you put @available above this function declaration, it will provide an error for the merchant trying to use it. I think that would probably be the ideal solve.

Sources/BraintreeSEPADirectDebit/CreateMandateResult.swift
sshropshire: Could we make `approvalURL` optional here instead of having `"null"`? Or maybe even an empty `""` would work well.

kaylagalway (approved):
Thanks for the updates!

Sources/BraintreeSEPADirectDebit/CreateMandateResult.swift
jaxdesmarais: I went back and fourth on this, on Android, the literal string returned  from the API is "null" which is what we do the check for on Android. I think we could make it optional and then do a nil check, but then it would be hard to handle if something went wrong and the result of the URL was nil vs now what we do is get "null" from the API, which Swift converts to nil, then we convert back to "null" in the decoding step. It could potentially leave us in a situation when the URL is unexpectedly actually nil but we handle it as if the mandate was already approved, which would be incorrect.

sshropshire: Nice yeah I agree with Kayla here then. If we put it in a constant and maybe add a comment with this information right above it to keep the historical context. It'll help us cut down on WTFs/min lol.

PR #801, opened 2022-03-25
Convert PayPal Demo ViewControllers to Swift

jaxdesmarais:
### Summary of changes

We've been wanting to convert the demo app to Swift, so I figured no time but the present on this open dev Friday. 🚀  Once we have all the features converted we can tackle the base views.

- Convert `BraintreeDemoPayPalCheckoutViewController`, `BraintreeDemoPayPalPayLaterViewController`, and `BraintreeDemoPayPalVaultViewController` to Swift in the Demo app.
- Kept all functionality/language/etc the same as when it was implemented in Objective-C.

### Checklist

- @jaxdesmarais


sshropshire (approved):

sarahkoop (approved):
This is awesome! 

My only consideration is that for the Obj-C to Swift conversion, it was nice to have a pre-built Obj-C integration to test and confirm that the Swift modules were still compatible with Obj-C integrations. The benefits of having the demo in Swift are probably still greater, just wanted to note that so we can think about other ways to test those conversions.

jaxdesmarais:
> This is awesome!
> 
> My only consideration is that for the Obj-C to Swift conversion, it was nice to have a pre-built Obj-C integration to test and confirm that the Swift modules were still compatible with Obj-C integrations. The benefits of having the demo in Swift are probably still greater, just wanted to note that so we can think about other ways to test those conversions.

Totally agree. I think we will for sure need to be more intentional about checking the generated interface for modules in Swift (I think we have that included in our migration guide for obj-c > swift). With NXO we didn't have anything in objective-c so this was part of our general cadence for anything exposed publicly, but will certainly be something we want to be conscious of as we do the conversions.

PR #803, opened 2022-03-28
Add Analytics to SEPA Feature

jaxdesmarais:
### Summary of changes

- Send analytic events on each API call, success, and failure path throughout the SEPA flow

### Checklist

- @jaxdesmarais


Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -39,34 +39,46 @@ import BraintreeCore
         context: ASWebAuthenticationPresentationContextProviding,
         completion:  @escaping (BTSEPADirectDebitNonce?, Error?) -> Void
     ) {
+        apiClient.sendAnalyticsEvent("ios.sepa-direct-debit.selected.started")
         createMandate(request: request) { createMandateResult, error in
+            self.apiClient.sendAnalyticsEvent("ios.sepa-direct-debit.create-mandate.requested")
```
sarahkoop, line 44: On Android we ended up moving the `requested` event to just before the method invocation (for all API calls). I think it makes more sense to send the analytics event before the API call returns, in case there is a problem with the API call.

sshropshire (approved):

PR #804, opened 2022-03-30
Fix crash caused by failed http requests

cltnschlosser:
### Summary of changes

`Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'data parameter is nil'` could happen for failed http requests (no response body).

### Checklist

- [x] Added a changelog entry

- @cltnschlosser

### DIFF
--- a/Sources/BraintreeCore/BTGraphQLHTTP.m
+++ b/Sources/BraintreeCore/BTGraphQLHTTP.m
@@ -67,15 +67,7 @@ static NSString *BraintreeVersion = @"2018-03-06";
         return;
     }
 
-    if (data == nil || data.length == 0) {
-        NSError *error = [[NSError alloc] initWithDomain:BTHTTPErrorDomain
-                                                    code:BTHTTPErrorCodeUnknown
-                                                userInfo:@{NSLocalizedDescriptionKey: @"An unexpected error occurred with the HTTP request."}];
-        [self callCompletionBlock:completionBlock body:nil response:nil error:error];
-        return;
-    }
-
-    NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
+    NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data?:[NSData data] options:0 error:NULL];
     BTJSON *body = [[BTJSON alloc] initWithValue:json];
 
     // Success case
--- a/Sources/BraintreeCore/BTGraphQLHTTP.m
+++ b/Sources/BraintreeCore/BTGraphQLHTTP.m
@@ -67,8 +67,7 @@ static NSString *BraintreeVersion = @"2018-03-06";
         return;
     }
 
-    NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
-    BTJSON *body = [[BTJSON alloc] initWithValue:json];
+    BTJSON *body = (data.length == 0) ? [BTJSON new] : [[BTJSON alloc] initWithData:data];
 
     // Success case
     if ([body asDictionary] && ![body[@"errors"] asArray]) {
--- a/Sources/BraintreeCore/BTGraphQLHTTP.m
+++ b/Sources/BraintreeCore/BTGraphQLHTTP.m
@@ -67,7 +67,15 @@ static NSString *BraintreeVersion = @"2018-03-06";
         return;
     }
 
-    NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data?:[NSData data] options:0 error:NULL];
+    if (data == nil || data.length == 0) {
+        NSError *error = [[NSError alloc] initWithDomain:BTHTTPErrorDomain
+                                                    code:BTHTTPErrorCodeUnknown
+                                                userInfo:@{NSLocalizedDescriptionKey: @"An unexpected error occurred with the HTTP request."}];
+        [self callCompletionBlock:completionBlock body:nil response:nil error:error];
+        return;
+    }
+
+    NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
     BTJSON *body = [[BTJSON alloc] initWithValue:json];
 
     // Success case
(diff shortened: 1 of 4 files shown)


cltnschlosser:
OHHTTPStubs overwrites nil data with `[NSData data]` (Zero length data), and the code falls into, which seems fine.
```
        statusCode = 500;
        errorCode = BTHTTPErrorCodeServerError;
        errorBody[@"error"] = @{@"message": @"An unexpected error occurred"};
```

So I can't accurately test this code with the way you propose (For the record: I like what you proposed). From NSURLSession data is Nullable, and if I just put `data = nil` in the code it does crash, so I'll find some other way to do this that I can prevent the crash and still accurately test.

cltnschlosser:
Pushed up some changes (including a test). It's basically what I had before, but I think that is okay.

jaxdesmarais (commented):
It looks like a compromise of the solution that you proposed and the one we proposed that will allow for easier testing (due to the OCMock limitations) along with handling both nil data and 0 byte data would be the following:
```
    BTJSON *body;
    
    if (data == nil || data.length == 0) {
        NSError *error = [[NSError alloc] initWithDomain:BTHTTPErrorDomain
                                                            code:BTHTTPErrorCodeUnknown
                            userInfo:@{NSLocalizedDescriptionKey: @"An unexpected error occurred with the HTTP request."}];
        [self callCompletionBlock:completionBlock body:nil response:nil error:error];
        return;
    } else {
        NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
        body = [[BTJSON alloc] initWithValue:json];
    }
```

Then we should be able to use a similar test to what you proposed:
```
- (void)testHttpError_returnsError {
    [HTTPStubs stubRequestsPassingTest:^BOOL(__unused NSURLRequest *request) {
        return YES;
    } withStubResponse:^HTTPStubsResponse *(__unused NSURLRequest *request) {
        // Ideally would test with both nil and empty data, but OHTTPStubs doesn't support nil data.
        return [HTTPStubsResponse responseWithData:[NSData data] statusCode:500 headers:@{}];
    }];

    XCTestExpectation *expectation = [self expectationWithDescription:@"callback invoked"];
    [http POST:@"" completion:^(BTJSON *body, __unused NSHTTPURLResponse *response, NSError *error) {
        XCTAssertNil(body);
        XCTAssertEqualObjects(error.userInfo[NSLocalizedDescriptionKey], @"An unexpected error occurred with the HTTP request.");
        XCTAssertEqualObjects(error.domain, BTHTTPErrorDomain);
        XCTAssertEqual(error.code, BTHTTPErrorCodeUnknown);
        [expectation fulfill];
    }];

    [self waitForExpectationsWithTimeout:2 handler:nil];
}
```

cltnschlosser:
Fine with me :) Updated.

Looks like CI doesn't work for forks. IDK if you just want to run locally or copy my branch. Tested locally and it should be good to go (all green).

Sources/BraintreeCore/BTGraphQLHTTP.m
```diff
@@ -63,6 +63,14 @@ - (void)handleRequestCompletion:(NSData *)data
...
     if (error) {
+        [self callCompletionBlock:completionBlock body:nil response:(NSHTTPURLResponse *)response error:error];
+        return;
+    }
+
+    if (data == nil || data.length == 0) {
+        NSError *error = [[NSError alloc] initWithDomain:BTHTTPErrorDomain
+                                                    code:BTHTTPErrorCodeUnknown
+                                                userInfo:@{NSLocalizedDescriptionKey: @"An unexpected error occurred with the HTTP request."}];
         [self callCompletionBlock:completionBlock body:nil response:nil error:error];
```
cltnschlosser, line 74: @jaxdesmarais Should I include response here? It's included for other errors below.

jaxdesmarais:
Hello @cltnschlosser -

We are confirming with an internal team that there are not cases when an empty data object would be valid to ensure we should be handling both a nil and empty data the same. Once we have heard back, we can open a PR outside of the fork to ensure the CI is all 🟢  and will tag you in that PR (and close this one). Thanks again for opening this PR.

cltnschlosser:
Okay, thanks!

If it helps the internal discussion the previous logic would consider empty data an error, it would fall under this block:
```objc
} else {
    statusCode = 500;
    errorCode = BTHTTPErrorCodeServerError;
    errorBody[@"error"] = @{@"message": @"An unexpected error occurred"};
}
```

jaxdesmarais:
Hello @cltnschlosser -

We've confirmed with an internal team that we should be handling the expected responses and errors from GraphQL with how the code is already implemented and no changes to this should be required. Are you able to provide details to us on how you were able to reproduce the issue that sparked opening this PR so we can investigate further?

cltnschlosser:
If the network request fails outright, `data` can be nil and it will cause the app to crash. If you want it to work exactly like it does now, but not crash, I can make a commit for that. Super simple, just need to treat nil as empty.

Sources/BraintreeCore/BTGraphQLHTTP.m
```diff
@@ -63,11 +63,11 @@ - (void)handleRequestCompletion:(NSData *)data
 {
     // Network error
     if (error) {
-        [self callCompletionBlock:completionBlock body:nil response:nil error:error];
+        [self callCompletionBlock:completionBlock body:nil response:(NSHTTPURLResponse *)response error:error];
         return;
     }
 
-    NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
+    NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data?:[NSData data] options:0 error:NULL];
```
cltnschlosser, line 70: @jaxdesmarais To reproduce the crash (since OHHTTPStubs doesn't let you send nil data), you can add `data = nil` at the beginning of this function and then this line will crash with the message in the PR description.

jaxdesmarais: Was more curious about the payment flow you were going through to reproduce this crash but understand that this may have just been a general bug fix. We will get a PR up shortly for this.

cltnschlosser: Oh sorry. Yeah I don't know, I'm just working off crash logs. It's somewhat rare, so it seems to be some sort of error state. I don't think it's something you could reliably reproduce (Maybe with a mock server or something that returns 500s)

PR #805, opened 2022-04-01
Add isSandbox bool to PPDataCollector Functions

jaxdesmarais:
### Summary of changes

- Allow passing isSandbox bool for data collection in `clientMetadataID` and `collectPayPalDeviceData` functions

### Checklist

- [x] Added a changelog entry

### Authors

- @sarahkoop @jaxdesmarais

### DIFF
--- a/Sources/BraintreeDataCollector/BTDataCollector.m
+++ b/Sources/BraintreeDataCollector/BTDataCollector.m
@@ -135,8 +135,12 @@ static Class PayPalDataCollectorClass;
 + (NSString *)generatePayPalClientMetadataID:(BOOL)isSandbox {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wundeclared-selector"
-    if (PayPalDataCollectorClass && isSandbox ? [PayPalDataCollectorClass respondsToSelector:@selector(sandboxGenerateClientMetadataID)] : [PayPalDataCollectorClass respondsToSelector:@selector(generateClientMetadataID)]) {
-        return isSandbox ? [PayPalDataCollectorClass performSelector:@selector(sandboxGenerateClientMetadataID)] : [PayPalDataCollectorClass performSelector:@selector(generateClientMetadataID)];
+    if (PayPalDataCollectorClass) {
+        if (isSandbox && [PayPalDataCollectorClass respondsToSelector:@selector(sandboxGenerateClientMetadataID)]) {
+            return [PayPalDataCollectorClass performSelector:@selector(sandboxGenerateClientMetadataID)];
+        } else if (!isSandbox && [PayPalDataCollectorClass respondsToSelector:@selector(generateClientMetadataID)]) {
+            return [PayPalDataCollectorClass performSelector:@selector(generateClientMetadataID)];
+        }
     }
 #pragma clang diagnostic pop
     
--- a/Sources/BraintreeDataCollector/BTDataCollector.m
+++ b/Sources/BraintreeDataCollector/BTDataCollector.m
@@ -135,10 +135,8 @@ static Class PayPalDataCollectorClass;
 + (NSString *)generatePayPalClientMetadataID:(BOOL)isSandbox {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wundeclared-selector"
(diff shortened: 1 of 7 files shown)


Sources/BraintreeDataCollector/BTDataCollector.m
```diff
@@ -130,11 +132,11 @@ - (void)collectDeviceData:(void (^)(NSString * _Nonnull))completion {
...
 #pragma mark - Helper methods
 
-+ (NSString *)generatePayPalClientMetadataID {
++ (NSString *)generatePayPalClientMetadataID:(BOOL)isSandbox {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wundeclared-selector"
-    if (PayPalDataCollectorClass && [PayPalDataCollectorClass respondsToSelector:@selector(generateClientMetadataID)]) {
-        return [PayPalDataCollectorClass performSelector:@selector(generateClientMetadataID)];
+    if (PayPalDataCollectorClass && isSandbox ? [PayPalDataCollectorClass respondsToSelector:@selector(sandboxGenerateClientMetadataID)] : [PayPalDataCollectorClass respondsToSelector:@selector(generateClientMetadataID)]) {
+        return isSandbox ? [PayPalDataCollectorClass performSelector:@selector(sandboxGenerateClientMetadataID)] : [PayPalDataCollectorClass performSelector:@selector(generateClientMetadataID)];
```
jaxdesmarais, line 139: Because we are using `@selector` methods to call into the PPDataCollectorClass under the hood, we need to have 2 different functions that take no parameters (since selectors cannot send params). 1 func that takes no params and sets the environment as sandbox and 1 that takes no params and sets the environment as live.

Sources/PayPalDataCollector/Public/PayPalDataCollector/PPDataCollector.swift
```diff
@@ -5,6 +5,8 @@ import Security
  Enables you to collect data about a customer's device and correlate it with a session identifier on your server.
  */
 @objc public class PPDataCollector: NSObject {
+    
+    static var mangnesEnvironment: MagnesSDK.Environment = .LIVE
```
jaxdesmarais, line 9: Adding this static variable so that we can more easily test our logic

Sources/PayPalDataCollector/Public/PayPalDataCollector/PPDataCollector.swift
```diff
@@ -15,21 +17,45 @@ import Security
      - Parameter pairingID: A pairing ID to associate with this clientMetadataID must be 10-32 chars long or null
      - Returns: A client metadata ID to send as a header
      */
+    // NEXT_MAJOR_VERSION: make this not a class function
```
jaxdesmarais, line 20: class/static functions are a bit weird in this file, guessing it's needed because of how we call into this class is BTDataCollector under the hood? This would also be a breaking change but is weird to test and generally not the best practice, so lots of next major version notes to find a better way to do this in next major version.

Sources/PayPalDataCollector/Public/PayPalDataCollector/PPDataCollector.swift
```diff
@@ -41,10 +67,22 @@ import Security
         return result?.getPayPalClientMetaDataId() ?? ""
     }
 
+    // NEXT_MAJOR_VERSION: make this not a class function
+    @objc class func generateClientMetadataID(isSandbox: Bool) -> String {
+        PPDataCollector.generateClientMetadataID("", disableBeacon: false, isSandbox: isSandbox, data: nil)
+    }
```
jaxdesmarais, line 73: Duplication of this function below because `@selector` can't pass in parameters. We should be able to clean this up in next major version so we don't need 3 functions doing basically the same thing.

Sources/BraintreeDataCollector/BTDataCollector.m
sshropshire: nit: could we go with a nested if here to reduce the line length? `if (isSandbox)` instead of the ternary operator would tighten it up a bit.

sshropshire (approved):

cgdibble (approved):

PR #807, opened 2022-04-12
Update Configuration Threading Logic

sarahkoop:
### Summary of changes

- Update demo app to vault Venmo and use Payment Context
- Update demo app to not embed & sign Magnes framework (static framework should not be embedded)
- Update configuration cache request policy
- Moved logic for dispatch queue of configuration calls
- Remove unnecessary flushing of analytics when app enters background
- Throttle requests for configuration to reduce load on servers and network connection lost errors
- Add analytics events for network connection lost errors in Venmo flow

**Note:** The scope of impact for this is older devices running older iOS and Venmo app versions

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais 
- @sarahkoop

### DIFF
--- a/Demo/UI Tests/Venmo UI Tests/Venmo_UITests.swift
+++ b/Demo/UI Tests/Venmo UI Tests/Venmo_UITests.swift
@@ -22,11 +22,11 @@ class Venmo_UITests: XCTestCase {
         demoApp.buttons["Venmo (custom button)"].tap()
     }
     
-    func testTokenizeVenmo_whenSignInSuccessfulWithPaymentContext_returnsNonce() {
+    func testTokenizeVenmo_whenSignInSuccessfulWithPaymentContext_returnsToApp() {
         waitForElementToBeHittable(mockVenmo.buttons["SUCCESS WITH PAYMENT CONTEXT"])
         mockVenmo.buttons["SUCCESS WITH PAYMENT CONTEXT"].tap()
 
-        XCTAssertTrue(demoApp.buttons["Got a nonce. Tap to make a transaction."].waitForExistence(timeout: 15))
+        XCTAssertTrue(demoApp.buttons["Failed to store Venmo Account in vault"].waitForExistence(timeout: 15))
     }
     
     func testTokenizeVenmo_whenSignInSuccessfulWithoutPaymentContext_returnsNonce() {
--- a/Sources/BraintreeCore/BTAPIClient.m
+++ b/Sources/BraintreeCore/BTAPIClient.m
@@ -90,6 +90,7 @@ NSString *const BTAPIClientErrorDomain = @"com.braintreepayments.BTAPIClientErro
             configurationCache = [[NSURLCache alloc] initWithMemoryCapacity:1 * 1024 * 1024 diskCapacity:0 diskPath:nil];
         });
         configuration.URLCache = configurationCache;
+        // Use the caching logic defined in the protocol implementation, if any, for a particular URL load request.
         configuration.requestCachePolicy = NSURLRequestUseProtocolCachePolicy;
         _configurationHTTP.session = [NSURLSession sessionWithConfiguration:configuration];
 
@@ -273,20 +274,12 @@ NSString *const BTAPIClientErrorDomain = @"com.braintreepayments.BTAPIClientErro
 #pragma mark - Remote Configuration
 
 - (void)fetchOrReturnRemoteConfiguration:(void (^)(BTConfiguration *, NSError *))completionBlock {
-    // Guarantee that multiple calls to this method will successfully obtain configuration exactly once.
+    // Fetches or returns the configuration and caches the response in the GET BTHTTP call if successful
     //
     // Rules:
     //   - If cachedConfiguration is present, return it without a request
-    //   - If cachedConfiguration is not present, fetch it and cache the succesful response
-    //     - If fetching fails, return error and the next queued will try to fetch again
-    //
-    // Note: Configuration queue is SERIAL. This helps ensure that each request for configuration
-    //       is processed independently. Thus, the check for cached configuration and the fetch is an
-    //       atomic operation with respect to other calls to this method.
-    //
-    // Note: Uses dispatch_semaphore to block the configuration queue when the configuration fetch
-    //       request is waiting to return. In this context, it is OK to block, as the configuration
-    //       queue is a background queue to guarantee atomic access to the remote configuration resource.
+    //   - If cachedConfiguration is not present, fetch it and cache the successful response
+    //   - If fetching fails, return error
     __block NSError *fetchError;
     __block BTConfiguration *configuration;
     NSString *configPath = @"v1/configuration"; // Default for tokenizationKey
--- a/Sources/BraintreeCore/BTAPIClient.m
+++ b/Sources/BraintreeCore/BTAPIClient.m
@@ -287,61 +287,49 @@ NSString *const BTAPIClientErrorDomain = @"com.braintreepayments.BTAPIClientErro
     // Note: Uses dispatch_semaphore to block the configuration queue when the configuration fetch
     //       request is waiting to return. In this context, it is OK to block, as the configuration
     //       queue is a background queue to guarantee atomic access to the remote configuration resource.
-    dispatch_async(self.configurationQueue, ^{
-        __block NSError *fetchError;
-
-        dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
-        __block BTConfiguration *configuration;
-        NSString *configPath = @"v1/configuration"; // Default for tokenizationKey
-        if (self.clientToken) {
-            configPath = [self.clientToken.configURL absoluteString];
-        }
-        [self.configurationHTTP GET:configPath parameters:@{ @"configVersion": @"3" } shouldCache:YES completion:^(BTJSON * _Nullable body, NSHTTPURLResponse * _Nullable response, NSError * _Nullable error) {
-            if (error) {
-                fetchError = error;
-            } else if (response.statusCode != 200) {
-                NSError *configurationDomainError =
-                [NSError errorWithDomain:BTAPIClientErrorDomain
-                                    code:BTAPIClientErrorTypeConfigurationUnavailable
-                                userInfo:@{
-                                           NSLocalizedFailureReasonErrorKey: @"Unable to fetch remote configuration from Braintree API at this time."
-                                           }];
-                fetchError = configurationDomainError;
-            } else {
-                configuration = [[BTConfiguration alloc] initWithJSON:body];
-                if (!self.braintreeAPI) {
-                    NSURL *apiURL = [configuration.json[@"braintreeApi"][@"url"] asURL];
-                    NSString *accessToken = [configuration.json[@"braintreeApi"][@"accessToken"] asString];
-                    self.braintreeAPI = [[BTAPIHTTP alloc] initWithBaseURL:apiURL accessToken:accessToken];
-                }
-                if (!self.http) {
-                    NSURL *baseURL = [configuration.json[@"clientApiUrl"] asURL];
-                    if (self.clientToken) {
-                        self.http = [[BTHTTP alloc] initWithBaseURL:baseURL authorizationFingerprint:self.clientToken.authorizationFingerprint];
-                    } else if (self.tokenizationKey) {
-                        self.http = [[BTHTTP alloc] initWithBaseURL:baseURL tokenizationKey:self.tokenizationKey];
-                    }
-                }
-                if (!self.graphQL) {
-                    NSURL *graphQLBaseURL = [BTAPIClient graphQLURLForEnvironment:configuration.environment];
-                    if (self.clientToken) {
-                        self.graphQL = [[BTGraphQLHTTP alloc] initWithBaseURL:graphQLBaseURL authorizationFingerprint:self.clientToken.authorizationFingerprint];
-                    } else if (self.tokenizationKey) {
-                        self.graphQL = [[BTGraphQLHTTP alloc] initWithBaseURL:graphQLBaseURL tokenizationKey:self.tokenizationKey];
-                    }
+    __block NSError *fetchError;
+    __block BTConfiguration *configuration;
+    NSString *configPath = @"v1/configuration"; // Default for tokenizationKey
+    if (self.clientToken) {
+        configPath = [self.clientToken.configURL absoluteString];
+    }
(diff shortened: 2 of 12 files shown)


Demo/UI Tests/Venmo UI Tests/Venmo_UITests.swift
```diff
@@ -22,11 +22,11 @@ class Venmo_UITests: XCTestCase {
         demoApp.buttons["Venmo (custom button)"].tap()
     }
     
-    func testTokenizeVenmo_whenSignInSuccessfulWithPaymentContext_returnsNonce() {
+    func testTokenizeVenmo_whenSignInSuccessfulWithPaymentContext_returnsToApp() {
         waitForElementToBeHittable(mockVenmo.buttons["SUCCESS WITH PAYMENT CONTEXT"])
         mockVenmo.buttons["SUCCESS WITH PAYMENT CONTEXT"].tap()
 
-        XCTAssertTrue(demoApp.buttons["Got a nonce. Tap to make a transaction."].waitForExistence(timeout: 15))
+        XCTAssertTrue(demoApp.buttons["Failed to store Venmo Account in vault"].waitForExistence(timeout: 15))
```
sarahkoop, line 29: This UI test if only used to test that the SDK successfully switches to the mock Venmo app and returns. Since we updated the demo app to include vaulting, this flow now ultimately fails, because the mock Venmo app does not return a valid nonce for vaulting. We decided to just update this error message, so that we can keep vaulting in the demo app flow for future testing, and since this test still covers its intended scope.

Sources/BraintreeCore/BTAPIClient.m
```diff
@@ -90,7 +90,8 @@ - (nullable instancetype)initWithAuthorization:(NSString *)authorization sendAna
             configurationCache = [[NSURLCache alloc] initWithMemoryCapacity:1 * 1024 * 1024 diskCapacity:0 diskPath:nil];
         });
         configuration.URLCache = configurationCache;
-        configuration.requestCachePolicy = NSURLRequestReturnCacheDataElseLoad;
+        // Use the caching logic defined in the protocol implementation, if any, for a particular URL load request.
+        configuration.requestCachePolicy = NSURLRequestUseProtocolCachePolicy;
```
sshropshire, line 94: random q: does config endpoint enforce caching with HTTP headers?

Sources/BraintreeCore/BTAPIClient.m
```diff
@@ -273,75 +274,55 @@ - (void)fetchPaymentMethodNonces:(BOOL)defaultFirst completion:(void (^)(NSArray
...
-                    NSURL *baseURL = [configuration.json[@"clientApiUrl"] asURL];
-                    if (self.clientToken) {
-                        self.http = [[BTHTTP alloc] initWithBaseURL:baseURL authorizationFingerprint:self.clientToken.authorizationFingerprint];
-                    } else if (self.tokenizationKey) {
-                        self.http = [[BTHTTP alloc] initWithBaseURL:baseURL tokenizationKey:self.tokenizationKey];
-                    }
+    //   - If cachedConfiguration is not present, fetch it and cache the successful response
+    //   - If fetching fails, return error
+    __block NSError *fetchError;
+    __block BTConfiguration *configuration;
```
sshropshire, line 284: Would it be possible to move these into the closure to avoid `__block` scope?

Sources/BraintreeCore/BTAPIClient.m
```diff
@@ -273,75 +274,55 @@ - (void)fetchPaymentMethodNonces:(BOOL)defaultFirst completion:(void (^)(NSArray
...
-    //       is processed independently. Thus, the check for cached configuration and the fetch is an
-    //       atomic operation with respect to other calls to this method.
-    //
-    // Note: Uses dispatch_semaphore to block the configuration queue when the configuration fetch
-    //       request is waiting to return. In this context, it is OK to block, as the configuration
-    //       queue is a background queue to guarantee atomic access to the remote configuration resource.
-    dispatch_async(self.configurationQueue, ^{
-        __block NSError *fetchError;
-
-        dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
```
sshropshire, line 293: It's nice to see the semaphore code removed 🙌 . Was this the root cause?

Sources/BraintreeCore/BTAnalyticsService.m
```diff
@@ -241,27 +240,6 @@ - (void)flush:(void (^)(NSError *))completionBlock {
     }];
 }
 
-#pragma mark - Private methods
-
-- (void)appWillResign:(NSNotification *)notification {
-    UIApplication *application = notification.object;
```
sshropshire, line 247: qq: Are analytics all processed in the foreground now?

Sources/BraintreeCore/BTHTTP.m
```diff
@@ -156,15 +156,19 @@ - (void)httpRequestWithCaching:(NSString *)method path:(NSString *)aPath paramet
...
+        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 0.1 * NSEC_PER_SEC), dispatch_get_main_queue(), ^{
+            if (cachedResponse != nil) {
+                [self handleRequestCompletion:cachedResponse.data request:nil shouldCache:NO response:cachedResponse.response error:nil completionBlock:completionBlock];
+            } else {
+                NSURLSessionTask *task = [self.session dataTaskWithRequest:request completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
+                    [self handleRequestCompletion:data request:request shouldCache:YES response:response error:error completionBlock:completionBlock];
+                }];
+                [task resume];
+            }
+        });
```
sshropshire, line 171: Is the `httpRequestWithCaching:` only used for `BTConfiguration` requests?

Sources/BraintreeVenmo/BTVenmoDriver.m
```diff
@@ -159,6 +159,9 @@ - (void)tokenizeVenmoAccountWithVenmoRequest:(BTVenmoRequest *)venmoRequest comp
 
             [self.apiClient POST:@"" parameters:params httpType:BTAPIClientHTTPTypeGraphQLAPI completion:^(BTJSON *body, __unused NSHTTPURLResponse *response, NSError *err) {
                 if (err) {
+                    if ([err.localizedDescription isEqualToString:@"The network connection was lost."]) {
```
sshropshire, line 162: Is there another check we can make to see if it's a _network connection error_ that doesn't rely on a `"network connection was lost."` string?

Sources/BraintreeCore/BTAPIClient.m
jaxdesmarais: Do you mean like passing a `Cache-Control` header? If so the preferred method for `NSURLSession` is to use the `requestCachePolicy` on the configuration as there is some weirdness with headers not being respected (for example `NSURLSession` also ignores `Keep-Alive` headers).

Sources/BraintreeCore/BTAPIClient.m
jaxdesmarais: ✅

Sources/BraintreeCore/BTAPIClient.m
jaxdesmarais: Having this is an semaphore was at least part of the root cause.

Sources/BraintreeCore/BTAnalyticsService.m
jaxdesmarais: We didn't change the thread where analytics are being processed which is done in the `sendAnalyticsEvent` function. Here we were unnecessarily flushing analytics on the background thread and updating this results in all of the analytics events being sent as expected, so was unnecessarily adding strain.

Sources/BraintreeCore/BTHTTP.m
jaxdesmarais: Yep! We only want to cache configuration requests (not anything else)

Sources/BraintreeVenmo/BTVenmoDriver.m
jaxdesmarais: Yep, looks like we can check the `error.code` instead. lldb was losing it's mind when we were trying to do that yesterday, but it looks like it's fine now and Xcode is happy letting us do that. 🤦

Sources/BraintreeCore/BTAPIClient.m
sshropshire: Right yeah I guess that's a question for the gateway actually. If config has caching behavior built in we could leverage that on Android too.

Sources/BraintreeCore/BTAnalyticsService.m
sshropshire: Yeah background tasks do still run when the app is closed by the user, it may be good to do this in the future. On Android we're using `WorkManager` to do timed uploads every 30 seconds, I'm wondering if iOS has something similar.

Sources/BraintreeCore/BTHTTP.m
sshropshire: Is the overall solution to prevent multiple BT API requests within a 1 second timeframe?

Sources/BraintreeVenmo/BTVenmoDriver.m
sshropshire: Nice yeah that'd be dope in case that string changes.

Sources/BraintreeCore/BTAnalyticsService.m
jaxdesmarais: Yeah, previously we were getting a _ton_ of flush analytics errors which this resolves. I think we should certainly look into something similar in the future. When we re-write the core module in Swift I think there is certainly a lot of optimization to be done as we move to newer APIs with those changes in the future.

Sources/BraintreeVenmo/BTVenmoDriver.m
jaxdesmarais: Yup! Updated! Those this error hasn't changed since I think Swift dropped so probably a lot of things would blow up if they ever changed that string (but now we won't be one of them).

Sources/BraintreeCore/BTAPIClient.m
jaxdesmarais: Yeah certainly something to look into for both SDKs

Sources/BraintreeCore/BTHTTP.m
jaxdesmarais: Yeah, essentially we are throttling this request by 0.1 seconds

sshropshire: Oh aight Analytics HTTP requests happen as soon as they're triggered on iOS that is rough.

sshropshire: Don't mean to nit I genuinely have questions lol. Did we choose 0.1 because it worked best in testing?

sarahkoop: Yeah we tried 0.05 and 0.075 but we still saw an increased number of Network connection lost errors. 0.1 seems to be the least amount of time that still improved the errors.

demerino (approved):
Is it accurate to say that testing this solution is heavily dependent on using an older device? What is the monitoring strategy to identify these dropped requests? This looks like a good pass at the fix, any reservations I have are around observability and testing moving forward to catch future issues.

sarahkoop:
> Is it accurate to say that testing this solution is heavily dependent on using an older device? What is the monitoring strategy to identify these dropped requests? This looks like a good pass at the fix, any reservations I have are around observability and testing moving forward to catch future issues.

Yes, testing heavily depended on an older Venmo app version on an older device. We believe the older Venmo app version was more crucial to testing because there was a big change in loading time between Venmo app versions (newer versions of the app have a loading screen and the app switch process takes much longer than older versions, where we switch back to the merchant app almost immediately and continue with API calls).  We didn't have a newer iOS versioned device with an older Venmo app, so couldn't confirm if the iOS version also plays a large role. 

We are planning to also meet with the observability team to discuss client side monitoring in general and hopefully improve our process for catching these issues sooner.

sshropshire (approved):
This may be a good reference PR for starting the Swift refactor. A serial dispatch queue could help in the future to halt outbound requests until the config is available.

Sources/BraintreeCore/BTHTTP.m
```diff
@@ -156,15 +156,19 @@ - (void)httpRequestWithCaching:(NSString *)method path:(NSString *)aPath paramet
...
-        } else {
-            NSURLSessionTask *task = [self.session dataTaskWithRequest:request completionHandler:^(NSData *data, NSURLResponse *response, NSError *error) {
-                [self handleRequestCompletion:data request:request shouldCache:YES response:response error:error completionBlock:completionBlock];
-            }];
-            [task resume];
-        }
+        
+        // The increase in speed of API calls with cached configuration caused an increase in "network connection lost" errors.
+        // Adding this delay allows us to throttle the network requests slightly to reduce load on the servers and decrease connection lost errors.
+        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 0.1 * NSEC_PER_SEC), dispatch_get_main_queue(), ^{
```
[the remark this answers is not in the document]
sarahkoop: Hmm.. we probably shouldn't change this value without the same amount of testing we did initially (we also tested smaller increments and this seemed to be the shortest amount of time that improved the issue). It's also not used elsewhere so I think leaving it here is ok, but happy to change it if others disagree

scannillo:
👋 Just catching up on some PRs while I was away - this is definitely an interesting one. I'm curious @jaxdesmarais - do you have background info on how this issue arose? Was it a specific merchant request? What exactly did they report?

scannillo:
👋 @jaxdesmarais - thanks for that thorough explanation, this helps a lot!

I'm curious - so in the original PR #789, we are manually wiping and then and restoring URL responses to the cache, using the `NSURLCache.sharedURLCache` shared instance. I also see [this cache instance](https://github.com/braintree/braintree_ios/blob/221f48098f30f172089d92fc5ff3a7d8f6648934/Sources/BraintreeCore/BTAPIClient.m#L87). Is this leftover/not being used?

jaxdesmarais:
Yeah, the tl;dr is that we were accessing the cache in BTAPIClient, but we were never actually adding the config to the cache under the hood in BTHTTP.

This resulted in every time we attempted to access the cache here, nothing was present so we would then call `fetchOrReturnRemoteConfiguration` again, which called into `BTHTTP` to access the cache, but since nothing was actually cached just resulting in us always calling the API for a new config. PR #789 made it so that we would actually cache the config during the underlying method calls from BTAPIClient > BTHTTP to cache the config if there were no errors as well as access and use that cache if present.

We can likely clean this up even further during the Swift conversion though as this cache instance I don't believe is ever used.

PR #809, opened 2022-04-21
Handle Nil Data Returned From NSURLSessionTask

jaxdesmarais:
### Summary of changes

- Update for PR #804 
- We do not currently handle nil data in our request completion for GraphQL - in cases where we get nil data and no error, we will return an error as nil data and nil error would be unexpected behavior of [`dataTaskWithRequest`](https://developer.apple.com/documentation/foundation/nsurlsession/1407613-datataskwithrequest)

### Checklist

- [x] Added a changelog entry

- @sarahkoop @jaxdesmarais @cltnschlosser

### DIFF
--- a/Sources/BraintreeCore/BTGraphQLHTTP.m
+++ b/Sources/BraintreeCore/BTGraphQLHTTP.m
@@ -67,18 +67,15 @@ static NSString *BraintreeVersion = @"2018-03-06";
         return;
     }
 
-    BTJSON *body;
-    
     if (data == nil) {
         NSError *error = [[NSError alloc] initWithDomain:BTHTTPErrorDomain
                                                             code:BTHTTPErrorCodeUnknown
                             userInfo:@{NSLocalizedDescriptionKey: @"An unexpected error occurred with the HTTP request."}];
         [self callCompletionBlock:completionBlock body:nil response:(NSHTTPURLResponse *)response error:error];
         return;
-    } else {
-        NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
-        body = [[BTJSON alloc] initWithValue:json];
     }
+    NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
+    BTJSON *body = [[BTJSON alloc] initWithValue:json];
 
     // Success case
     if ([body asDictionary] && ![body[@"errors"] asArray]) {


Sources/BraintreeCore/BTGraphQLHTTP.h
```diff
@@ -23,6 +23,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @interface BTGraphQLHTTP : BTHTTP
 
+- (void)handleRequestCompletion:(nullable NSData *)data
+                       response:(nullable NSURLResponse *)response
+                          error:(nullable NSError *)error
+                completionBlock:(void (^)(BTJSON * _Nonnull, NSHTTPURLResponse * _Nonnull, NSError * _Nonnull))completionBlock;
+
```
jaxdesmarais, line 30: There is not an easy way to test this new behavior as `OHHTTPStubs` doesn't let you send nil data and automatically converts it to an empty data object. Adding this to the internal header allows us to unit test this update. I explored overriding the `HTTPStubsResponse` but that also requires eventually overriding pieces of `Foundation` due to how `HTTPStubsResponse` works under the hood.

Sources/BraintreeCore/BTGraphQLHTTP.m
```diff
@@ -63,12 +63,22 @@ - (void)handleRequestCompletion:(NSData *)data
 {
     // Network error
     if (error) {
-        [self callCompletionBlock:completionBlock body:nil response:nil error:error];
+        [self callCompletionBlock:completionBlock body:nil response:(NSHTTPURLResponse *)response error:error];
```
jaxdesmarais, line 66: We are including the response in other error cases but not here, this will bring our error completion here in line with our other error cases.

Sources/BraintreeCore/BTGraphQLHTTP.m
```diff
@@ -63,12 +63,22 @@ - (void)handleRequestCompletion:(NSData *)data
...
         return;
     }
```
cltnschlosser, line 68: On line 187, this block is duplicated before calling `handleRequestCompletion` so it can be removed from there.

Sources/BraintreeCore/BTGraphQLHTTP.m
```diff
@@ -63,12 +63,22 @@ - (void)handleRequestCompletion:(NSData *)data
...
+    if (data == nil) {
+        NSError *error = [[NSError alloc] initWithDomain:BTHTTPErrorDomain
+                                                            code:BTHTTPErrorCodeUnknown
+                            userInfo:@{NSLocalizedDescriptionKey: @"An unexpected error occurred with the HTTP request."}];
+        [self callCompletionBlock:completionBlock body:nil response:(NSHTTPURLResponse *)response error:error];
+        return;
+    } else {
+        NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
+        body = [[BTJSON alloc] initWithValue:json];
+    }
```
cltnschlosser, line 81: Since the if returns, you can flatten this and treat the `if (data == nil)` line a swift guard
```swift
guard let data = data else {
    // handle error
    return
}
```
```suggestion
    if (data == nil) {
        NSError *error = [[NSError alloc] initWithDomain:BTHTTPErrorDomain
                                                            code:BTHTTPErrorCodeUnknown
                            userInfo:@{NSLocalizedDescriptionKey: @"An unexpected error occurred with the HTTP request."}];
        [self callCompletionBlock:completionBlock body:nil response:(NSHTTPURLResponse *)response error:error];
        return;
    }
    NSDictionary *json = [NSJSONSerialization JSONObjectWithData:data options:0 error:NULL];
    BTJSON *body = [[BTJSON alloc] initWithValue:json];
```

Purely stylistic, but in my opinion it's a nice benefit to not have to predefined `BTJSON *body;`

Sources/BraintreeCore/BTGraphQLHTTP.h
cltnschlosser: Nice!

cltnschlosser (approved):
None of my comments are required, just some code cleanup and style nits.

Sources/BraintreeCore/BTGraphQLHTTP.m
jaxdesmarais: Good catch, updated

sshropshire (approved):

andrespelaezp (approved):

PR #811, opened 2022-05-09
Add additional error analytics events

mattwylder:
### Summary of changes

- Added Network Connection Lost analytics event to all post requests

### Checklist

- [x] Added a changelog entry

- @sarahkoop 
- @jaxdesmarais 
- @mattwylder

### DIFF
--- a/Sources/BraintreeApplePay/BTApplePayClient.m
+++ b/Sources/BraintreeApplePay/BTApplePayClient.m
@@ -24,6 +24,7 @@
 #endif
 
 NSString *const BTApplePayErrorDomain = @"com.braintreepayments.BTApplePayErrorDomain";
+NSInteger const NetworkConnectionLostCode = -1005;
 
 @implementation BTApplePayClient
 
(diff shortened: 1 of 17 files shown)


Sources/BraintreeApplePay/BTApplePayClient.m
```diff
@@ -24,6 +24,7 @@
 #endif
 
 NSString *const BTApplePayErrorDomain = @"com.braintreepayments.BTApplePayErrorDomain";
+NSInteger const NetworkConnectionLostCode = -1005;
```
sshropshire, line 27: 🧶 : is there a central place we could put this constant to make it more DRY?

sshropshire (approved):
🌊  small nit but overall LGTM!

Sources/BraintreeApplePay/BTApplePayClient.m
jaxdesmarais: Yep, we can define this as a macro in `BTAPIClient` for now which is the preferred way to do this with how it's being used. This will also avoid the linker warnings we are getting that are making the SPM and Cocoapods builds fail.

andrespelaezp (approved):
LGTM

PR #813, opened 2022-05-17
Create NXO Data Models

sshropshire:
### Summary of changes

- Add `tokenize()` method to `BTPayPalNativeCheckoutClient`
- Create `BTPayPalNativeCheckoutNonce`
- Create `BTPayPalNativeCheckoutRequest`

### Checklist

### Authors

- @mattwylder
- @sarahkoop 
- @sshropshire


Podfile.lock
```diff
@@ -45,4 +45,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 75163f16229528991a9364c7c1a44cd57a30cac6
 
-COCOAPODS: 1.11.2
+COCOAPODS: 1.11.3
```
jaxdesmarais, line 48: Is bumping the cocoapods version needed for this PR?

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutRequest.swift
```diff
@@ -0,0 +1,13 @@
...
+
+@objcMembers
+public class BTPayPalNativeCheckoutRequest: NSObject {
+    public let returnURL: String
+    public let amount: String
+    
+    public init(returnURL: String, amount: String) {
+        self.returnURL = returnURL
+        self.amount = amount
+        super.init()
```
jaxdesmarais, line 11: Is this `super.init` just a placeholder for now or is it needed for something that I am missing currently?

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
```diff
@@ -3,6 +3,17 @@ import Foundation
...
+    @objc(initWithAPIClient:)
+    public init(apiClient: BTAPIClient) {
+        self.apiClient = apiClient
+    }
+    
+    public func tokenize(request: BTPayPalNativeCheckoutRequest, completion: (BTPayPalNativeCheckoutNonce?, Error?) -> Void) {
+        // TODO: start native checkout
+    }
 }
+
```
jaxdesmarais, line 19: nit: it looks like we don't have a newline space in the other files at the end but do here. I don't lean strongly one way or the other as long as we are consistent. For SEPA it looks like we did include a newline space at the end of files. 🤷

jaxdesmarais (approved):
Just a couple of very small comments but otherwise looks good

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutRequest.swift
sshropshire: I think we need to call it because we're subclassing `NSObject`.

sshropshire: Just checked with Sarah actually we should be good to remove this.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
jaxdesmarais: Yeah, I'm fine with either tbh as long as it's consistent. I think Swiftlint handles this for Unified, but would be nice to stay consistent so if we _do_ switch to a linter post Swift conversion or add a style guide we've already got some standards in place.

andrespelaezp (approved):

PR #816, opened 2022-05-27
WIP - [NEXT_MAJOR_VERSION] Kount Module

jaxdesmarais:
🚨🚨🚨 Currently a draft/wip - need to see what's blowing up with CI or if this is good 🚨🚨🚨

### Summary of changes

To convert `BraintreeDataCollector` to Swift we need for the entire module to be Swift. Unfortunately, Kount only offers Objective-C, which won't work for us because you cannot mix Obj-C and Swift for Swift Package Manager 🙃  The Kount solution is to use a bridging header in a Swift codebase, which doesn't play nicely with Frameworks, which we need since that is the whole point of an SDK. But also, still Obj-C so won't work with SPM

The solution in this PR is to:
- Extract the Kount code into a Module that `BraintreeDataCollector` depends on (basically create our own wrapper for the Kount SDK
- This allows us to import the new `KountDataCollector` into a future Swift `BraintreeDataCollector` as expected

Testing in the Demo app, we still get the data we expect back from Kount with this solution.

### Checklist

- [ ] Added a changelog entry

- @jaxdesmarais


Package.swift
```diff
@@ -113,6 +116,12 @@ let package = Package(
             dependencies: ["BraintreeCore"],
             publicHeadersPath: "Public"
         ),
+        .target(
+            name: "KountDataCollectorWrapper",
```
jaxdesmarais, line 120: For SPM we cannot call the target `KountDataCollector` as the [binary target and xcframework name need to match](https://stackoverflow.com/questions/71539013/spm-artifact-not-found-for-target-aaa-xcode-13-3-only). This does mean for SPM only, a merchant would need to do `import KountDataCollectorWrapper` in their `ViewController` while the other package managers could do `import KountDataCollector`. Should all be updated to call the wrapper module `KountDataCollectorWrapper`? Is there a different/better name for this module that wraps the Kount SDK?

Sources/KountDataCollector/BTKountDataCollector.m
```diff
@@ -0,0 +1,7 @@
+#if __has_include(<Braintree/KountDataCollector.h>)
+#import <Braintree/KountDataCollector.h>
+#else
+#import <KountDataCollector/KountDataCollector.h>
+#endif
+
+int _keep;
```
jaxdesmarais, line 7: For SPM we need a dummy `.m` file as we cannot have a module with only `.h` files ([reference 1](https://forums.swift.org/t/swift-package-and-xcframework-target-for-c-library-where-to-include-the-header/51163) and [reference 2](https://stackoverflow.com/questions/68502255/how-to-define-obj-c-target-with-a-header-file-so-other-targets-can-implement-it)). 

There are a few ways to go about this we can either have something like we are using here `int _keep` or we could likely do something like `void __dummy() {}`. Essentially the file needs to have something other than an import statement in it. Was thinking we should document this somehow - maybe with just a note describing this and the 2 links from above? I know we don't typically add doc strings for internal things, but this seems like a good thing to document. Thoughts?

sarahkoop: +1 on a doc string here for internal explanation

SampleApps/SPMTest/SPMTest/ViewController.swift
```diff
@@ -9,6 +9,7 @@ import BraintreeThreeDSecure
 import BraintreeUnionPay
 import BraintreeVenmo
 import PayPalDataCollector
+import KountDataCollectorWrapper
```
sarahkoop, line 12: Looking at the SPM vs Carthage import statements, I think it would be good for them to be aligned

Package.swift
jaxdesmarais: Sick, will update to `BraintreeKountDataCollector`

mattwylder (approved):

sshropshire (approved):

SampleApps/CarthageTest/Cartfile
```diff
@@ -1 +1 @@
-github "braintree/braintree_ios"
\ No newline at end of file
+github "braintree/braintree_ios" "kount-module"
```
sshropshire, line 1: random: does this make it a breaking change?

jaxdesmarais: This is the sample app and we have needed to point to the new branch for all of the new modules we make. Once `next-major-version` is merged into master/main we should be able to update this in that PR or in a fast follow. But merchants would not need to point to our branch or next major version once the entire feature branch is merged in, if that's the question.

sshropshire: Got it nah I realize now that `kount-module` is a GitHub branch 🤦 . I thought this was a list of modules lol.

jaxdesmarais: Ah no worries, understandably confusing. Package managers :shakes_fist:

PR #817, opened 2022-05-31
Update Cardinal to 2.2.5 3

jaxdesmarais:
### Summary of changes

- Update Cardinal frameworks in the SDK to 2.2.5-3
- Note: we still do not have version 2.2.5-3 that works with x84_64 slice for .frameworks so this update is only for SPM, Cocoapods, and Carthage users using .xcframeworks.

### Checklist

- [x] Added a changelog entry

- @sarahkoop @jaxdesmarais


jaxdesmarais:
Seems to be a weird carthage issue going on here - hoping it's an issue with this branch sitting for a while. Going to close and make these updates on a clean branch and 🤞 .

PR #825, opened 2022-06-14
Convert BTKeychain, BTLogger, and BTURLUtils to Swift

jaxdesmarais:
This PR is to be merged into the next-major-version branch

### Summary of changes

- Convert `BTKeychain`, `BTLogger`, and `BTURLUtils` to Swift
- `BTLogger` has been reduced to a minimum, with the intention of eventually removing it entirely. Formerly a wrapper for `NSLog` with log level logic and pre-packaged string formatting, it is now essentially an enum with associated string values. `BTLogger` no longer calls NSLog, so all invocations have been replaced with an `NSLog` call containing the new `BTLogger` strings. Because of its limited use throughout the codebase we felt this was reasonable. We also feel the new setup better expresses `BTLogger`'s behavior. Test assertions have been removed which we believe reasonably reflect the change in `BTLogger`'s behavior. A closure was also removed from the Objective-C version which was only used for tests.
- BTKeychain has more or less been literally translated from Objective-C to Swift. It uses `CoreFoundation` which is not the most readable in Swift or Objective-C.
- All changes have been tested in the demo app to ensure everything is still working as expected

### Checklist

- @mattwylder @jaxdesmarais


andrespelaezp (approved):

UnitTests/BraintreeVenmoTests/BTVenmoClient_Tests.swift
```diff
@@ -90,20 +91,9 @@ class BTVenmoClient_Tests: XCTestCase {
         waitForExpectations(timeout: 2, handler: nil)
     }
 
-    func testTokenizeVenmoAccount_whenReturnURLSchemeIsNil_logsCriticalMessageAndCallsBackWithError() {
+    func testTokenizeVenmoAccount_whenReturnURLSchemeIsNil_andCallsBackWithError() {
         let venmoClient = BTVenmoClient(apiClient: mockAPIClient)
         BTAppContextSwitcher.sharedInstance().returnURLScheme = ""
-
-        
-        var criticalMessageLogged = false
```
kaylagalway, line 98: Do we need to add new tests to replace these removed tests? Or is testing this logic handled somewhere else now

UnitTests/BraintreeCoreTests/BTLogger_Internal_Tests.m
```diff
@@ -1,58 +0,0 @@
-#import "BTLogger_Internal.h"
-@import BraintreeCore;
-@import XCTest;
-
-@interface BTLogger_Internal_Tests : XCTestCase
```
kaylagalway, line 5: Same question for these tests

Sources/BraintreeCoreSwift/BTKeychain.swift
```diff
@@ -0,0 +1,76 @@
+import Foundation
+import Security
+
+@objcMembers public class BTKeychain: NSObject {
```
kaylagalway, line 4: Same [documentation question](https://github.com/braintree/braintree_ios/pull/825/files#r914133820) for this file. May be irrelevant if merchants don't touch these.

Sources/BraintreeCoreSwift/BTKeychain.swift
```diff
@@ -0,0 +1,76 @@
...
+        guard let dataString = String(data: data, encoding: .utf8) else { return "" }
+        return dataString
+    }
+    
+    static func keychainForKey(_ key: String) -> String {
+        String(format: "com.braintreepayments.Braintree-API.", key)
+    }
+    
+    static func setData(_ data: Data, forKey key: String) -> Bool {
+        var success: Bool = true
```
kaylagalway, line 22: Do we need this success `Bool`? Could we just return true or false or an error in each place that we set it's value?

kaylagalway (commented):
Looks good! I only have some lack of context questions - going to sync up with @mattwylder and talk through the PR

Sources/BraintreeCore/BTAPIClient.m
```diff
@@ -36,7 +55,7 @@ - (nullable instancetype)initWithAuthorization:(NSString *)authorization {
 - (nullable instancetype)initWithAuthorization:(NSString *)authorization sendAnalyticsEvent:(BOOL)sendAnalyticsEvent {
     if(![authorization isKindOfClass:[NSString class]]) {
         NSString *reason = @"BTClient could not initialize because the provided authorization was invalid";
-        [[BTLogger sharedLogger] error:reason];
+        NSLog(@"%@ Missing analytics session metadata - will not send event  %@", [BTLogLevelDescription stringFor:BTLogLevelError], reason);
```
sshropshire, line 58: q: Do we still use `BTLogger` or is it being removed in this PR?

Sources/BraintreeCoreSwift/BTKeychain.swift
sshropshire: This comes up a lot on Android. We don't have a consistent convention for single return statements in place. This could be a good start to the discussion though, there's some good [stack discussion](https://stackoverflow.com/q/36707) on the topic too.

UnitTests/BraintreeVenmoTests/BTVenmoClient_Tests.swift
mattwylder: Since the functionality of `BTLogger` is different now, I think it's ok to remove these tests. It may be worthwhile to create a new test file for "stupid" tests of `BTLogger` itself. @kaylagalway and I have started doing this for the other files!

sshropshire (approved):
LGTM 🚀  Removing `BTLogger` in favor of `NSLog` makes things a lot simpler.

Sources/BraintreeCoreSwift/BTKeychain.swift
kaylagalway: Yeah definitely two sides to this, just read through that stack discussion for a bit and I can see both sides on it. In the end I'm good either way! Maybe there is even a way to optimize this logic with enums so that the number of returns or bool state setting becomes less of a question/less bug prone.

sshropshire: Definitely yeah there's for sure an opportunity to refactor to make things more readable. Having a single return variable in a method this long could easily lead to bugs.

PR #833, opened 2022-07-07
Convert `BTAppContextSwitcher` and `BTClientMetadata` to Swift

mattwylder:
### Summary of changes

- Converted `BTAppContextSwitcher` to Swift with some API changes. 
   - Static wrapper methods have been removed because, in our opinion, they hid that `BTAppContextSwitcher` is a singleton. All invocations have been changed to directly call `sharedInstance` instead of the equivalent method.
   - The original Objective-C used a publisher-subscriber pattern, where `Class` types were held in a `Set`, in which static functions were invoked. We replaced the `Set` with an array `[BTAppContextSwitchClient.Type]` which behaves similarly. It appears the original use of `Type` and `static` is meant to prevent the use of multiple instances of the same type. In a later major version we would like to turn this into a more traditional publisher-subscriber pattern containing and directly using instances of `BTAppContextSwitch` rather than abstract `Type`.
- Converted `BTClientMetadata` and `BTMutableClientMetadata` to Swift
  - The Swift mutable version is only used to maintain Objective-C compatibility.
- Converted `BTClientMetadataSourceType` and `BTClientMetadataIntegrationType` to Swift. Because of an inconsistency with which the Swift to Objective-C conversion handled their names, we renamed them `BTClientMetadataSource` and `BTClientMetadataIntegration` respectively. 
- Preprocessor Swift module import statements have been added to multiple files.

### Checklist

- [x] Added a changelog entry

- @mattwylder 
- @jaxdesmarais


Sources/BraintreeCore/Public/BraintreeCore/BTAPIClient.h
```diff
@@ -1,18 +1,18 @@
...
 
 #define NETWORK_CONNECTION_LOST_CODE -1005
 
 @class BTConfiguration;
 @class BTJSON;
-
 @class BTPaymentMethodNonce;
 
+typedef NS_ENUM(NSInteger, BTClientMetadataSource);
+typedef NS_ENUM(NSInteger, BTClientMetadataIntegration);
```
mattwylder, line 14: Forward declarations to avoid header import issues with Cocoapods.

Sources/BraintreePaymentFlow/BTPaymentFlowClient.m
```diff
@@ -39,7 +39,7 @@
 #import <BraintreeCoreSwift/BraintreeCoreSwift-Swift.h>
 #endif
 
-@interface BTPaymentFlowClient () <SFSafariViewControllerDelegate>
+@interface BTPaymentFlowClient () <SFSafariViewControllerDelegate, BTAppContextSwitchClient>
```
mattwylder, line 42: Moved this protocol conformance from the header file to the `.m` to make Cocoapods happy. Cocoapods won't let you import an outside framework in a `.h` file

UnitTests/BraintreeCoreTests/BTAPIClient_Tests.m
```diff
@@ -3,22 +3,7 @@
 @import BraintreeCore;
 @import BraintreeTestShared;
 @import XCTest;
-
-@interface StubBTClientMetadata : BTClientMetadata
```
mattwylder, line 7: unused, removed.

Sources/BraintreeCore/BTAPIClient.m
```diff
@@ -136,8 +136,8 @@ + (BTAPIClientAuthorizationType)authorizationTypeForAuthorization:(NSString *)au
 }
 
 // NEXT_MAJOR_VERSION: remove this unused method
-- (instancetype)copyWithSource:(BTClientMetadataSourceType)source
-                   integration:(BTClientMetadataIntegrationType)integration
+- (instancetype)copyWithSource:(BTClientMetadataSource)source
+                   integration:(BTClientMetadataIntegration)integration
```
sshropshire, line 140: 👏

sshropshire (approved):
🌊

kaylagalway (approved):
Looks good! Nice job working around all the package managers

PR #832, opened 2022-07-06
Enable shipping callback in NXO module

stephaniechiu:
### Summary of changes

This PR allows merchants to optionally create shipping callbacks by using the `onShippingChange` closure.

### Checklist

- [ ] Added a changelog entry

- stephaniechiu

PR #836, opened 2022-07-12
Shipping callback in NXO module

stephaniechiu:
### Summary of changes

This PR allows merchants to optionally create shipping callbacks by using the `onShippingChange` closure that's from the NXO SDK.

If a merchant wants to have shipping options, once they set the amount using `BTPayPalNativeCheckoutRequest(amount: "4.30")`, they will be able to call the `onShippingChange` closure and perform actions using the input parameters `change` and `action`.

The `onShippingChange` parameter was created in `BTPayPalNativeCheckoutRequest` like so `public var onShippingChange: ((PayPalCheckout.ShippingChange, PayPalCheckout.ShippingChangeAction) -> Void)?`

In `ShippingChangeAction`, a merchant can use `patch(request: PatchRequest)` `approve()`, `reject()`

Original PR - https://github.com/braintree/braintree_ios/pull/832

### Checklist

- [ ] Added a changelog entry

- stephaniechiu


Demo/Application/Features/PayPal - Native Checkout/BraintreeDemoPayPalNativeCheckoutViewController.swift
```diff
@@ -1,10 +1,28 @@
 import Foundation
 import UIKit
 import BraintreePayPalNativeCheckout
+import PayPalCheckout
```
jaxdesmarais, line 4: Agree with @jonathajones in previous PR that we do not want to require merchants to import `PayPalCheckout` to use specific features. Instead we should either create Braintree types that can be translated to the NXO types as or create a wrapper for the shipping callback types that can be passed through to the callback without the merchant having to pull in `PayPalCheckout` directly.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutRequest.swift
```diff
@@ -2,12 +2,33 @@
 import BraintreePayPal
 #endif
 
-/// Options for the PayPal Checkout and PayPal Checkout with Vault flows.
+import PayPalCheckout
+
+/**
+ Options for the PayPal Checkout and PayPal Checkout with Vault flows.
+ */
```
jaxdesmarais, line 9: Nit: update to /// doc strings.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutRequest.swift
```diff
@@ -2,12 +2,33 @@
...
-    let hermesPath: String = "v1/paypal_hermes/create_payment_resource"
+    /**
+     Initializes a PayPal Checkout request.
+
+     - Parameter amount: Used for a one-time payment. Amount must be greater than or equal to zero, may optionally contain exactly 2 decimal places separated by '.' and is limited to 7 digits before the decimal point.
+
+     - Parameter shippingCallback: Optional. Used for shipping changes and shipping change actions.
+
+     - Returns: A PayPal Checkout request.
+     */
```
jaxdesmarais, line 22: Nit: update to /// doc strings.

jaxdesmarais:
Can we add some details to the PR description around what this feature is? Not everyone on the Braintree team knows what shipping callbacks are and why/how a merchant would use shipping callbacks. This will allow for a better review process from Braintree team members that have not worked on/with native checkout previously.

Demo/Application/Features/PayPal - Native Checkout/BraintreeDemoPayPalNativeCheckoutViewController.swift
stephaniechiu: Thanks for the feedback! That makes sense that we don't want merchants to import `PayPalCheckout` directly. I'll make the changes accordingly

stephaniechiu:
Closing until conflicts resolve

PR #840, opened 2022-07-13
Add Concurrency Flag to Build Workflow

jaxdesmarais:
### Summary of changes

- Currently each time you push to the repo in a PR, the previous builds continue to run. This change should allow us to cancel previously running CI processes as we push new commits to the branch allowing our CI to be faster and not potentially waiting on previously running processes. This will be helpful at times when we are making and pushing several small changes in quick succession to only run CI on the latest commit.

### Checklist

- @mattwylder 
- @jaxdesmarais

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,7 @@
 name: Build
 on: [pull_request, workflow_dispatch]
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }} 
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }} 
   cancel-in-progress: true
 jobs:
   cocoapods:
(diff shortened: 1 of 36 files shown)


jaxdesmarais:
We did a bunch of testing on this yesterday and it _seemed_ to be working inconsistently. Matt talked to some Venmo teams that were using this in the office and they said what we have is correct. 🤷  So we re-opened this PR and did some additional testing this morning and things are working as expected. 🥳

It's worth noting that we have observed that there is a delay with the running processes and when they are canceled. If you are unsure if an action actually was canceled, clicking in to the action tab > clicking into the job will provide the most accurate status. It seems like there is a slight delay between when jobs are canceled and when the spinning activity indicators are updated.

andrespelaezp (approved):

PR #841, opened 2022-07-19
Convert BTJSON to Swift

jaxdesmarais:
**Note**: this diff looks bigger than it is. We use `BTJSON` heavily in the SDK, so the majority of this PR is the Swift import statements. The main files of note are:
- `BTJSON.m` + `BTJSON.h` - removed Obj-C files
- `BTJSON.swift` - new Swift file replace the .`m` and `.h`

### Summary of changes

- Convert BTJSON to Swift

For this conversion the public interface did not change at all since it's used heavily throughout the SDK. We can look into ways to optimize this in future versions once the entire SDK is in Swift. For now, we just went with a 1:1 approach on the conversion of this file.

Internally, the major thing that changed is the following. In Obj-C we were doing this:
```
- (id)value {
    id value = _value;
    for (id key in self.subscripts) {
        if ([value isKindOfClass:[NSArray class]]) {
            if (![key isKindOfClass:[NSNumber class]]) {
                value = [self chainedErrorOrErrorWithCode:BTJSONErrorAccessInvalid userInfo:nil];
                break;
            }

            NSUInteger idx = [(NSNumber *)key unsignedIntegerValue];
            if (idx >= [(NSArray *)value count]) {
                value = nil;
                break;
            }

            value = [value objectAtIndexedSubscript:idx];
        } else if ([value isKindOfClass:[NSDictionary class]]) {
            if (![key isKindOfClass:[NSString class]]) {
                value = [self chainedErrorOrErrorWithCode:BTJSONErrorAccessInvalid userInfo:nil];
                break;
            }

            value = [value objectForKeyedSubscript:key];
        } else {
            value = [self chainedErrorOrErrorWithCode:BTJSONErrorValueInvalid userInfo:@{ NSLocalizedFailureReasonErrorKey: [NSString stringWithFormat:@"Attempted to index into a value that is neither an object nor an array using key (%@).", key] }];
            break;
        }
    }
    return value;
}
```

We instead now directly copy things vs mutating and changing the value directly. This also removes the need for the `subscripts` internal property that was used in Objective-C. 

### Checklist

- [x] Added a changelog entry

- @mattwylder @jaxdesmarais


Sources/BraintreeCoreSwift/BTJSON.swift
```diff
@@ -0,0 +1,241 @@
...
+    ///  Indexes into the JSON as if the current value is an object
+    ///
+    /// Notably, this method will always return successfully; however, if the value is not an object, the JSON will wrap an error.
+    public subscript(index: Int) -> BTJSON {
+        if value is NSError {
+            return self
+        }
+
+        guard let value = value as? [Any],
+                index < value.count else {
```
sshropshire, line 133:
```suggestion
        guard let value = value as? [Any], index < value.count else {
```
🧶 : we can flatten this out to improve readability.

UnitTests/BraintreeCoreTests/BTJSON_Tests.swift
```diff
@@ -254,21 +255,21 @@ class BTJSON_Tests: XCTestCase {
...
 
-        XCTAssertTrue(obj.isObject);
+        XCTAssertTrue(obj.isObject)
     }
 
     func testIsObjectForNonObject() {
         let JSON = "[]".data(using: String.Encoding.utf8)!
         let obj = BTJSON(data: JSON)
 
-        XCTAssertFalse(obj.isObject);
```
sshropshire, line 271: Semicolons! 😝 😭

.github/workflows/build.yml
```diff
@@ -37,21 +37,17 @@ jobs:
...
-      - name: Check out repository
-        uses: actions/checkout@v2
-      - name: Use Xcode 13
-        run: sudo xcode-select -switch /Applications/Xcode_13.0.app
-      - name: Use current branch
-        run: sed -i '' 's/branch = .*/branch = \"'"$GITHUB_HEAD_REF"'\";/' SampleApps/SPMTest/SPMTest.xcodeproj/project.pbxproj
-      - name: Run swift package resolve
-        run: cd SampleApps/SPMTest && swift package resolve
-      - name: Build & archive SPMTest
-        run: xcodebuild -project 'SampleApps/SPMTest/SPMTest.xcodeproj' -scheme 'SPMTest' clean build archive CODE_SIGNING_ALLOWED=NO
```
mattwylder, line 57: Commenting out SPM stuff because SPM can't build with mixed Objc and Swift modules. So SPM won't be able to build until the entire module is converted.

demerino (approved):
Looks good! Would be interesting to see how this changes when everything is swift.

PR #847, opened 2022-08-10
Convert `BTConfiguration`, `BTHTTPErrors` to Swift

mattwylder:
### Summary of changes

- Converted `BTConfiguration` and `BTHTTPErrors` to Swift
- Removed `BTErrors`

### Checklist

- @jaxdesmarais @mattwylder


Sources/BraintreeCoreSwift/BTConfiguration+GraphQL.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+public extension BTConfiguration {
+    var isGraphQLEnabled: Bool {
+        json?["graphQL"]["url"].asString()?.count ?? 0 > 0
```
sshropshire, line 5:
```suggestion
        (json?["graphQL"]["url"].asString()?.count ?? 0) > 0
```
nit 🧶

Sources/BraintreeCoreSwift/BTConfiguration.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+/// Contains information specific to a merchant's Braintree integration
+@objcMembers public class BTConfiguration: NSObject {
```
sshropshire, line 4: random q: Do the classes that aren't meant for the merchant to use directly need to have ObjC annotations?

sshropshire (approved):
🌊

Sources/BraintreeCoreSwift/BTConfiguration.swift
mattwylder: For the moment, yes. That way we can use the code in our own un-converted Objc code. But once all our stuff is converted we could remove it in certain classes.

kaylagalway (approved):
:+1::+1:

Sources/BraintreePayPal/BTPayPalClient.m
```diff
@@ -143,7 +143,7 @@ - (void)tokenizePayPalAccountWithPayPalRequest:(BTPayPalRequest *)request comple
                 if (error.code == NETWORK_CONNECTION_LOST_CODE) {
                     [self.apiClient sendAnalyticsEvent:@"ios.paypal.tokenize.network-connection.failure"];
                 }
-                NSString *errorDetailsIssue = ((BTJSON *)error.userInfo[BTHTTPJSONResponseBodyKey][@"paymentResource"][@"errorDetails"][0][@"issue"]).asString;
+                NSString *errorDetailsIssue = ((BTJSON *)error.userInfo[BTHTTPError.jsonResponseBodyKey][@"paymentResource"][@"errorDetails"][0][@"issue"]).asString;
```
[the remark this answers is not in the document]
jaxdesmarais: Yeah, eventually all of the modules will be converted to Swift and we should be able to clean these instances up a bit more.

PR #848, opened 2022-08-11
IMPROVEMENT - Added c header search path in Package.swift

kabirkhaan:

### Summary of changes
During working with [tuist](https://tuist.io) we found a 🐞 regarding the header search path with SPM Brain tree dependency 

Using `BrainTree SPM` package and run `tuist test` will throw the following error

Import headrs are release 
Screenshot 2022-08-11 at 10 45 16

Reason behind this is `tuist generate` and `tuist test` create different graphs for project as per following table. 
Due to different structure of `tuist test`  all file that impot`../BraintreeCore{ANY_CLASS}.h`  will not found.

| `tuist generate` | `tuist test`
| :---: | :---: | 
| Screenshot 2022-08-11 at 23 29 39 | Screenshot 2022-08-11 at 23 28 25 |

from the [tuist test](https://docs.tuist.io/commands/test/)
> Tuist also automatically caches successful test runs - this means that the subsequent tuist test calls will test only what has changed! You will see which test targets are skipped in the log. This is extremely useful especially on CI - if you make a simple change in a module that other targets do not depend on, CI can test only this module and nothing else.

# CHANGELOG

- Remove `#import "../BraintreeCore/{ANY_CLASS}.h`
- Use Same header for Carthage and SPM 
- To resolve the headers use added following in Package.swift
```
cSettings: [
                .headerSearchPath("../")
            ]
``` 

### Checklist

- [x] Added a changelog entry

- @kabirkhaan

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## unreleased
 * BraintreeSEPADirectDebit
   * Update nonce to pull in ibanLastFour and customerID as expected
+* Package.swift improvement - Added C Header setting in Package.swift
 
 ## 5.11.0 (2022-07-20)
 * BraintreeSEPADirectDebit
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -45,10 +45,7 @@ In general, we avoid importing **internal headers from other modules**, but occa
 #if __has_include(<Braintree/BraintreeAmericanExpress.h>) // CocoaPods
 #import <Braintree/BTAPIClient_Internal.h>
 
-#elif SWIFT_PACKAGE // SPM
-#import "../BraintreeCore/BTAPIClient_Internal.h"
-
-#else // Carthage
+#else // Carthage & SPM
 #import <BraintreeCore/BTAPIClient_Internal.h>
 #endif
 ```
--- a/Package.swift
+++ b/Package.swift
@@ -141,7 +141,10 @@ let package = Package(
             name: "BraintreeThreeDSecure",
             dependencies: ["BraintreePaymentFlow", "BraintreeCard", "CardinalMobile", "PPRiskMagnes"],
             publicHeadersPath: "Public",
-            cSettings: [.headerSearchPath("V2UICustomization")]
+            cSettings: [
+                .headerSearchPath("V2UICustomization"),
+                .headerSearchPath("../")
+            ]
         ),
         .binaryTarget(
             name: "CardinalMobile",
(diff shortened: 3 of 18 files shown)


mattwylder:
Hi there! Could you provide more context for the bug you were experiencing with tuist?

Also, we have fixed an issue with GitHub actions that was caused the builds to fail. Can you pull the latest from `master` into your branch?

kabirkhaan:
> Hi there! Could you provide more context for the bug you were experiencing with tuist?
> 
> Also, we have fixed an issue with GitHub actions that was causing the builds to fail. Can you pull the latest from `master` into your branch?


Thanks for the comments and fixing the GitHub action for fork branches 👍 
I update the PR description and put more details.

kabirkhaan:
> Hi there! Could you provide more context for the bug you were experiencing with tuist?
> 
> Also, we have fixed an issue with GitHub actions that was causing the builds to fail. Can you pull the latest from `master` into your branch?

Hello @mattwylder seems like we still have issue with CI with the [line](https://github.com/braintree/braintree_ios/blob/c41fb40d8d5bb7287647c26a5a7ff180b069956c/.github/workflows/build.yml#L122) and[line](https://github.com/braintree/braintree_ios/blob/c41fb40d8d5bb7287647c26a5a7ff180b069956c/.github/workflows/build.yml#L60) 😢

mattwylder:
Hey there! Thanks for your patience. The issue was with the slash in the branch name. We've pushed some changes to `master` that should fix the issue. Can you merge the lastest from `master` into your branch so we run CI?

jaxdesmarais:
Hello @kabirkhaan -

I'd like to take a step back here as the root of the problem seems to be a difference in configuration with how `tuist generate` and `tuist test` create graphs for projects based on this comment in the PR description:

> Reason behind this is tuist generate and tuist test create different graphs for project as per following table.
Due to different structure of tuist test all file that impot../BraintreeCore{ANY_CLASS}.h will not found.

Is there a specific reason why this change resolves the packages as expected with `tuist generate` and `tuist test`? The line running into the error you are seeing is this import:
```
#import "../BraintreeCore/BTAPIClient_Internal.h"
```

The change in this PR prepends `../` to the path:
```
#import <BraintreeCore/BTAPIClient_Internal.h>
```
Which should result in the path now being the same as the previously used `../BraintreeCore/BTAPIClient_Internal.h`.

Additionally, are you able to provide a sample project with this branch pulled in that runs both `tuist generate` and `tuist test` successfully? We aren't familiar with this external dependency so would be helpful to see the difference with running off of your branch vs otherwise.

kabirkhaan:
Hello @jaxdesmarais, Can you please help me if I am missing any context that will help review the PR?

jaxdesmarais:
Hello @kabirkhaan -

Thanks for reaching back out and providing all of those details. At this time we support integrating through Swift Package Manager, CocoaPods, and Carthage and do not officially support `tuist`. At this time, we want to gauge interest in supporting `tuist` before picking up adding this to the SDK. Would you mind opening a [feature request via our GitHub issues](https://github.com/braintree/braintree_ios/issues/new/choose)? This will allow us to track interest in supporting `tuist` officially throughout the community.

PR #762, opened 2021-12-08
Convert Objective-C AmericanExpress module to Swift

jaxdesmarais:
### Summary of changes

- Removed BraintreeAmericanExpress/Public directory
- Converted .h and .m files with single .swift files
- Updated podspec path to point to the updated file path
- Remove `BTAmericanExpressErrorDomain` global constant
- Remove `BTAmericanExpressErrorType`
- Make `BTAmericanExpressRewardsBalance` initializer private
- Created `BTAmericanExpressError.swift` that houses the domain, error code, and localized description of all the possible errors

### Testing

- All method names remained the same for both Objective-C and Swift
- Tested build with Cocoapods, Swift Package Manager, and Carthage
- Added additional unit test for error case due to optional unwrapping

### Checklist

- [x] Added a changelog entry

- @sarahkoop @scannillo @jaxdesmarais

PR #852, opened 2022-08-15
Convert `BTBinData` `BTClientToken`

mattwylder:
### Summary of changes

- Converted `BTBinData` to Swift
- Converted `BTClientToken` to Swift and moved errors to `BTClientTokenError`

### Checklist

- @mattwylder @jaxdesmarais


sshropshire (approved):

Sources/BraintreeCore/BTAnalyticsService.m
```diff
@@ -4,13 +4,6 @@
 #import "BTHTTP.h"
 #import "BraintreeCoreSwiftImports.h"
 
-// Objective-C Module Imports
-#if __has_include(<Braintree/BraintreeCore.h>)
-#import <Braintree/BTClientToken.h>
-#else
-#import <BraintreeCore/BTClientToken.h>
-#endif
-
```
scannillo, line 13: 🥲 TEARS OF JOY

scannillo:
I noticed [we commented out](https://github.com/braintree/braintree_ios/blob/next-major-version/.github/workflows/build.yml) the SPM and CocoaPods build jobs for our `next-major-version` branch. Also I notice this PR is breaking the `BraintreeCore` conversion up by file, which makes sense since `BraintreeCore` is so massive!

Here's a thought - we could create an additional feature branch off of `next-major-version` for the `BraintreeCore` swift conversion work (ex: `btcore-swift-conversion`). This way our commented out `build.yml` file is associated with the work that required it, versus turning those builds off for all `next-major-version` PRs. I think it's much less likely to slip through the cracks that way.

Sources/BraintreeCoreSwift/BTClientToken.swift
```diff
@@ -0,0 +1,118 @@
+import Foundation
+
+/// An authorization string used to initialize the Braintree SDK
+@objcMembers public class BTClientToken: NSObject, NSCoding, NSCopying {
+
+    /// The client token as a BTJSON object
+    public let json: BTJSON?
+
+    /// The extracted authorization fingerprint
+    public let authorizationFingerprint: String?
```
scannillo, line 10: Enforce non-optional via init

Sources/BraintreeCoreSwift/BTClientTokenError.swift
```diff
@@ -0,0 +1,53 @@
...
+    case invalidFormat
+
+    /// Unsupported client token version
+    case unsupportedVersion
+    
+    /// Invalid UTF8 Encoding
+    case expectedUTF8Encoding
+    
+    /// Invalid Base64 Encoding
+    case expectedBase64Encoding
```
scannillo, line 25: Add new error names & corresponding code int value to the CHANGELOG

Sources/BraintreeCoreSwift/BTClientToken.swift
```diff
@@ -0,0 +1,118 @@
...
+    private static func toBTJSON(data: Data, isBase64: Bool) throws -> BTJSON? {
+        guard let clientTokenJSON = try? JSONSerialization.jsonObject(with: data) as? [String: Any] else {
+            throw BTClientTokenError.invalidJSON
+        }
+
+        guard let version = clientTokenJSON["version"] as? Int else {
+            throw BTClientTokenError.invalidFormat
+        }
+
+        // Version 1 must be utf8, versions 2 & 3 must be base64
```
scannillo, line 66: Check if we need to support v1 client tokens anymore?

Sources/BraintreeCoreSwift/BTClientToken.swift
```diff
@@ -0,0 +1,118 @@
...
+    @objc(copyWithZone:)
+    public func copy(with zone: NSZone? = nil) -> Any {
+        do {
+            return try BTClientToken(clientToken: self.originalValue)
+        } catch {
+            return error
+        }
+    }
+
+    // MARK: - isEqual override
```
scannillo, line 108:
```suggestion
    // MARK: - NSObject override
```

Sources/BraintreeCoreSwift/BTClientToken.swift
jaxdesmarais: Sounds like we still need to support v1 for now 😢

jaxdesmarais:
> I noticed [we commented out](https://github.com/braintree/braintree_ios/blob/next-major-version/.github/workflows/build.yml) the SPM and CocoaPods build jobs for our `next-major-version` branch. Also I notice this PR is breaking the `BraintreeCore` conversion up by file, which makes sense since `BraintreeCore` is so massive!
> 
> Here's a thought - we could create an additional feature branch off of `next-major-version` for the `BraintreeCore` swift conversion work (ex: `btcore-swift-conversion`). This way our commented out `build.yml` file is associated with the work that required it, versus turning those builds off for all `next-major-version` PRs. I think it's much less likely to slip through the cracks that way.

Discussed IRL, but since we are fairly far into the conversion we are going to leave this as is for now. We do have a ticket to come back to this and also should keep this in mind for future Swift conversions where we may have similar issues with package managers being angry temporarily.

Sources/BraintreeCoreSwift/BTClientTokenError.swift
jaxdesmarais: Updated the CHANGELOG but since several error enums have been updated kept it slightly more vague vs calling out every error and its new value. Thoughts on the update or should we be even more explicit?

Sources/BraintreeCoreSwift/BTClientToken.swift
jaxdesmarais: Ope, sounds like maybe not? 🎢

scannillo: 🙏

Sources/BraintreeCoreSwift/BTClientTokenError.swift
scannillo: I'm always in favor of an explicit CHANGELOG. I think it's also really helpful for us to remember when we did things. _Especially_ with our converting to Swift, it will be really hard to track actual implementation changes/history since this will all be entirely new commit history from our ObjC SDK.

So - maybe something like:
```
  * BraintreeCore
    * Remove `BTAPIClientErrorDomain` global constant
    * Add `BTClientTokenError.expectedBase64Encoding` error type 
    * Add `BTClientTokenError.expectedUTF8Encoding ` error type 
    ...
```

Sources/BraintreeCoreSwift/BTClientTokenError.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
+
+///  Error codes associated with a client token.
+@objc public enum BTClientTokenError: Int, Error, CustomNSError, LocalizedError {
```
scannillo, line 4: I actually don't think we want these error classes to be public. I think we want our merchant to just access the returned `NSError` or `Error` type, instead of our custom types. (We do that in [this PR](https://github.com/braintree/braintree_ios/pull/762/files)).

Sources/BraintreeCoreSwift/BTClientTokenError.swift
```diff
@@ -0,0 +1,53 @@
...
+    case invalidJSON
+
+    /// Invalid client token format
+    case invalidFormat
+
+    /// Unsupported client token version
+    case unsupportedVersion
+    
+    /// Invalid UTF8 Encoding
+    case expectedUTF8Encoding
```
scannillo, line 22: Super nitpick - but these would make more sense to me phrased like `failedUTF8Decoding`, since the decoding is what failed. We don't necessarily know the encoding is wrong.

Sources/BraintreeCoreSwift/BTClientTokenError.swift
jaxdesmarais: Currently the BTHTTP tests are in Obj-C. If we need to unit test internal Swift methods in Obj-C tests, we will have to write our tests in Swift. We can certainly do that in this PR since we know the Obj-C tests pass, but that is why this is currently public.

We currently have ~7 Obj-C test files in Core and most (if not all) are for the remaining conversions. We can certainly convert the tests along with the files, but will likely end up with some ~1k+ line conversions for the remaining PRs.

Sources/BraintreeCoreSwift/BTClientToken.swift
jaxdesmarais: Sounds like they Gateway still returns v1 client tokens, so there is likely a chance that we could get a v1 client token from the fetch or return remote config call or other calls to the GW. It sounds like for now it may be best to continue supporting v1 client tokens until the backend services fully drop support.

UnitTests/BraintreeCoreTests/BTClientToken_Tests.swift
```diff
@@ -0,0 +1,198 @@
...
+
+        let clientToken = try XCTUnwrap(BTClientToken(clientToken: clientTokenString1))
+        let clientToken2 = try XCTUnwrap(BTClientToken(clientToken: clientTokenString2))
+
+        XCTAssertNotNil(clientToken)
+        XCTAssertNotNil(clientToken2)
+        XCTAssertNotEqual(clientToken, clientToken2)
+    }
+
+    // MARK: - NSCopying tests
```
scannillo, line 178: These MARKs are very helpful for organizing this file 👍

Sources/BraintreeCoreSwift/BTClientToken.swift
```diff
@@ -0,0 +1,118 @@
+import Foundation
+
+/// An authorization string used to initialize the Braintree SDK
+@objcMembers public class BTClientToken: NSObject, NSCoding, NSCopying {
+
+    // MARK: - Public Properties
+
+    /// The client token as a BTJSON object
+    public let json: BTJSON
```
scannillo, line 9: Shall we add a TODO to revisit the access level of these properties?

Sources/BraintreeCoreSwift/BTClientToken.swift
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+        case 2, 3:
+            if !isBase64 {
+                throw BTClientTokenError.failedDecoding("Base64 encoding is required for Client Token versions 2 & 3.")
+            }
+            
+        default:
+            throw BTClientTokenError.unsupportedVersion
+        }
```
scannillo, line 80: This is a take-it-or-leave-it comment, we could avoid a level of nesting without using a switch. Whichever you find more readable.

```suggestion
        if version == 1 && isBase64 {
            throw BTClientTokenError.failedDecoding("UTF8 encoding is required for Client Token version 1.")
        } else if (version == 2 || version == 3) && !isBase64 {
            throw BTClientTokenError.failedDecoding("Base64 encoding is required for Client Token versions 2 & 3.")
        } else if (version > 3) {
            throw BTClientTokenError.unsupportedVersion
        }
```

scannillo (approved):
🎸

Sources/BraintreeCoreSwift/BTClientToken.swift
jaxdesmarais: Good call, added a TODO for this, looks like most can be internal once BTAPIClient + BTAnalyticsService are converted

Sources/BraintreeCoreSwift/BTClientToken.swift
jaxdesmarais: Yeah, I think this is a lot cleaner than a nested if in a switch. I think in the final else if we probably want to do `else if version < 1 || version > 3` to catch any unlikely weirdness in the `Int` we get back from the guard above this.

PR #855, opened 2022-08-22
Convert BTHTTP +  BTCacheDateValidator

jaxdesmarais:
### Summary of changes

- Convert BTHTTP +  BTCacheDateValidator to Swift
- There are some TODOs in this PR that will be addressed in the follow up PR to convert BTAPIHTTP + BTGraphQLHTTP to Swift
- We have not deleted BTHTTP and left the classes suffixed with Swift as the Obj-C BTAPIHTTP + BTGraphQLHTTP files cannot subclass the Swift class - in the follow up PR, BTHTTPSwift will be renamed to BTHTTP and the BTHTTP.m/.h files will be deleted

### Checklist

- ~[ ] Added a changelog entry~ there are no merchant facing API changes

- @mattwylder @jaxdesmarais


Sources/BraintreeCoreSwift/BTCoreConstants.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+public class BTCoreConstants {
+    public static var braintreeVersion: String = "5.11.0"
+}
```
scannillo, line 5: Add `nodoc` string for

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,557 @@
+import Foundation
+import Security
+
+/// Performs HTTP methods on the Braintree Client API
+// TODO: once BTAPIHTTP + BTGraphQLHTTP are converted this can be internal + more Swift-y
+// TODO: When BTAPIHTTP + BTGraphQL are converted we should update the dictionaries to [String: Any]
+@objcMembers public class BTHTTPSwift: NSObject, NSCopying, URLSessionDelegate {
+
```
scannillo, line 8:
```suggestion
// TODO: - Mark interval vs private properties accordingly
// MARK: - Properties
```

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,557 @@
...
+    let cacheDateValidator: BTCacheDateValidator
+    let baseURL: URL
+
+    var authorizationFingerprint: String? = ""
+    var tokenizationKey: String? = ""
+
+    private var _dispatchQueue: DispatchQueue?
+    
+    // MARK: Initializers
```
scannillo, line 35:
```suggestion
    // MARK: - Initializers
```

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,557 @@
...
+            return _dispatchQueue ?? DispatchQueue.main
+        }
+        set {
+            _dispatchQueue = newValue
+        }
+    }
+
+    let cacheDateValidator: BTCacheDateValidator
+    let baseURL: URL
+
```
scannillo, line 29:
```suggestion

// TODO: - once BTAPIHTTP + BTGraphQLHTTP are converted to Swift, handle error cases for empty or nil credentials before API request is made
```

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,557 @@
...
+        self.session = URLSession(configuration: configuration, delegate: self, delegateQueue: delegateQueue)
+        self.pinnedCertificates = BTAPIPinnedCertificates.trustedCertificates()
+    }
+
+    /// Initialize `BTHTTP` with the authorization fingerprint from a client token
+    /// - Parameter clientToken: The client token
+    @objc(initWithClientToken:)
+    public convenience init(clientToken: BTClientToken) {
+        guard let clientApiURL = clientToken.json?["clientApiUrl"].asURL() else {
+            fatalError("clientApiURL contained in client token is not a valid URL")
```
scannillo, line 90: Throw error instead of fatalError

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,557 @@
...
+            guard let request = request else {
+                self.handleRequestCompletion(data: nil, request: nil, shouldCache: false, response: nil, error: error, completion: completion)
+                return
+            }
+
+            let task: URLSessionTask? = self.session?.dataTask(with: request) { [weak self] data, response, error in
+                self?.handleRequestCompletion(data: data, request: request, shouldCache: false, response: response, error: error, completion: completion)
+            }
+
+            task?.resume()
```
scannillo, line 200: I think it would be safer to not allow an optional `URLSessionTask` and to call the completion with an error if it is.

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,557 @@
...
+        }
+    }
+
+    func createRequest(
+        method: String?,
+        path: String?,
```
scannillo, line 206: Do we want to add a TODO to re-evaluate optionality across all function signatures as well, once the other files are converted to Swift?

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,557 @@
...
+        var request: URLRequest
+
+        if method == "GET" || method == "DELETE" {
+            if isNotDataURL {
+                components.percentEncodedQuery = BTURLUtils.queryString(from: parameters ?? [:])
+            }
+            guard let urlFromComponents = components.url else {
+                let error = constructError(
+                    code: .urlStringInvalid,
+                    userInfo: [NSLocalizedDescriptionKey: "The URL absolute string is malformed or invalid."]
```
scannillo, line 291: New errors that we add should be included in the CHANGELOG

scannillo: If it's merchant facing*^. Which, we can double check if this one would be, but I am _assuming_ we bubble this up to the merchant

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+/// Performs HTTP methods on the Braintree Client API
+// TODO: once BTAPIHTTP + BTGraphQLHTTP are converted this can be internal + more Swift-y
+// TODO: When BTAPIHTTP + BTGraphQL are converted we should update the dictionaries to [String: Any]
+@objcMembers public class BTHTTPSwift: NSObject, NSCopying, URLSessionDelegate {
+// TODO: - Mark interval vs private properties accordingly
+    public typealias RequestCompletion = (BTJSON?, HTTPURLResponse?, Error?) -> Void
+
+    // MARK: - Public Properties
+    /// An array of pinned certificates, each an NSData instance consisting of DER encoded x509 certificates
+    public let pinnedCertificates: [NSData]
```
mattwylder, line 13: Made this non-optional and moved the initialization to `init(url: URL)`. Stylewise, would we prefer to initialize it on this line?

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+    public lazy var session: URLSession = {
+        let configuration: URLSessionConfiguration = URLSessionConfiguration.ephemeral
+        configuration.httpAdditionalHeaders = defaultHeaders()
+        
+        let delegateQueue: OperationQueue = OperationQueue()
+        delegateQueue.name = "com.braintreepayments.BTHTTP"
+        delegateQueue.maxConcurrentOperationCount = OperationQueue.defaultMaxConcurrentOperationCount
+        
+        return URLSession(configuration: configuration, delegate: self, delegateQueue: delegateQueue)
+    }()
```
mattwylder, line 26: Made session non-optional by moving its initialization from the `convenience` inits into a lazy var declaration.

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+
+    // TODO: Make internal with Swift test?
+    /// DispatchQueue on which asynchronous code will be executed. Defaults to `DispatchQueue.main`.
+    public var dispatchQueue: DispatchQueue = DispatchQueue.main
```
mattwylder, line 30: removed the computed var with the `get` and `set` blocks. This simplifies the code and doesn't change any behavior.

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+
+    /// Initialize `BTHTTP` with the URL from Braintree API and the authorization fingerprint from a tokenizationKey
+    /// - Parameters:
+    ///   - url: The base URL for the Braintree Client API
+    ///   - tokenizationKey: The authorization fingerprint HMAC from a client token
+    @objc(initWithBaseURL:tokenizationKey:)
+    public convenience init(url: URL, tokenizationKey: String) {
+        self.init(url: url)
+        self.tokenizationKey = tokenizationKey
+    }
```
mattwylder, line 69: For this and the previous `convenience init` can we do something like this?

```suggestion
    @objc(initWithBaseURL:tokenizationKey:error)
    public convenience init(url: URL, tokenizationKey: String) throws {
        if tokenizationKey.isEmpty {
            throw BTHTTPError.invalidTokenizationKey
        }
        self.init(url: url)
        self.tokenizationKey = tokenizationKey
    }
```

mattwylder: maybe we could use some kind of enum to determine whether we're using auth fingerprint or tokenization key?

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+
+    // TODO: Make internal with Swift test?
+    /// DispatchQueue on which asynchronous code will be executed. Defaults to `DispatchQueue.main`.
+    public var dispatchQueue: DispatchQueue = DispatchQueue.main
+
+    // MARK: - Internal Properties
+    let cacheDateValidator: BTCacheDateValidator
+    let baseURL: URL
+
+// TODO: - once BTAPIHTTP + BTGraphQLHTTP are converted to Swift, handle error cases for empty or nil credentials before API request is made
```
scannillo, line 36:
```suggestion
    // TODO: - once BTAPIHTTP + BTGraphQLHTTP are converted to Swift, handle error cases for empty or nil credentials before API request is made
```

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+
+            completion(returnedError)
+        } else {
+            completion(json.asError())
+        }
+    }
+
+    static func constructError(code: BTHTTPErrorCode, userInfo: [String: Any]) -> NSError {
+        NSError(domain: BTHTTPError.domain, code: code.rawValue, userInfo: userInfo)
+    }
```
scannillo, line 456: I think we want to remove this in favor of using our new error creation pattern in swift. For example [BTDataCollectorError.swift](https://github.com/braintree/braintree_ios/blob/next-major-version/Sources/BraintreeDataCollector/BTDataCollectorError.swift).

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+        request: URLRequest?,
+        shouldCache: Bool,
+        response: URLResponse?,
+        error: Error?,
+        completion: RequestCompletion?
+    ) {
+        guard let completion = completion else {
+            return
+        }
+        /// Handle errors for which the response is irrelevant e.g. SSL, unavailable network, etc.
```
scannillo, line 330:
```suggestion
        // Handle errors for which the response is irrelevant e.g. SSL, unavailable network, etc.
```

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+        ) { request, error in
+            completion(request, error)
+        }
+    }
+
+    func buildHTTPRequest(
+        method: String,
+        url: URL,
+        parameters: NSMutableDictionary? = [:],
+        isNotDataURL: Bool,
```
scannillo, line 254: Nitpick - The negative in the bool naming is a little confusing. `isDataURL` would be easier to read IMO.

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+        if request != nil && shouldCache && successStatusCode, let request = request {
+            let cachedURLResponse: CachedURLResponse = CachedURLResponse(response: response, data: data)
+
+            URLCache.shared.storeCachedResponse(cachedURLResponse, for: request)
+        }
+
+        callCompletionBlock(with: completion, body: json, response: httpResponse, error: nil)
+    }
+    
+    func callCompletionBlock(with completion: @escaping RequestCompletion, body: BTJSON?, response: HTTPURLResponse?, error: Error?) {
```
scannillo, line 380: Nitpick

```suggestion
    func callCompletionAsync(with completion: @escaping RequestCompletion, body: BTJSON?, response: HTTPURLResponse?, error: Error?) {
```

Sources/BraintreeCoreSwift/BTHTTP.swift
scannillo: Actually, I personally don't find this comment really useful. We can 🗑 it?

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+            callCompletionBlock(with: completion, body: nil, response: nil, error: error)
+            return
+        }
+
+        guard let response = response,
+              let httpResponse = createHTTPResponse(response: response) else {
+            let error = Self.constructError(
+                code: .httpResponseInvalid,
+                userInfo: [NSLocalizedDescriptionKey : "Unable to create HTTPURLResponse from response data."]
+            )
```
scannillo, line 341: Same comment re: moving error construction to BTHTTPError.swift file

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+            handleHTTPResponseError(response: httpResponse, data: data) { json, error in
+                self.callCompletionBlock(with: completion, body: json, response: httpResponse, error: error)
+            }
+            return
+        }
+
+        // Empty response is valid
+        let json: BTJSON = data.isEmpty ? BTJSON() : BTJSON(data: data)
+        if json.isError {
+            handleJSONResponseError(json: json, response: response) { error in
```
scannillo, line 362:
```suggestion
            handleJSONResponseError(json: json, response: response) { [weak self] error in
```

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+        }
+
+        guard let data = data else {
+            let error = Self.constructError(code: .dataNotFound, userInfo: [NSLocalizedDescriptionKey: "Data unexpectedly nil."])
+            callCompletionBlock(with: completion, body: nil, response: nil, error: error)
+            return
+        }
+
+        if httpResponse.statusCode >= 400 {
+            handleHTTPResponseError(response: httpResponse, data: data) { json, error in
```
scannillo, line 353:
```suggestion
            handleHTTPResponseError(response: httpResponse, data: data) { [weak self] json, error in
```

scannillo:
Also, might be worth looking at the networking layer for Northstar. It was recently pretty recently and is definitely more "Swift-y". It might make our lives easier too if the 2 SDKs underlying networking is similar. If we like this idea - we can make it a separate PR/task?

Sources/BraintreeCoreSwift/BTHTTP.swift
mattwylder: I made `method` non-optional. Let's look at path next!

mattwylder: did path!

Sources/BraintreeCoreSwift/BTHTTP.swift
jaxdesmarais: This one is a bit trickier as we use `domain`, `urlResponseKey`,  and `jsonResponseBodyKey` and the `BTHTTPErrorCode` enum in a bunch of the other modules. It seems like converting this and extracting those other uses out could be a larger piece of work since we use both BTHTTPError and BTHTTPErrorCode in `Card`, `PaymentFlow`, `PayPal`, `UnionPay`, and `ThreeDSecure` modules.

I think this will require a larger discussion on how we want to handle shared constants/errors across modules as a number of errors in BTHTTP will require an associated type to pass through the use info. Are we okay creating a follow up ticket for this since this class + enum are used across multiple modules?

mattwylder: I think making a follow-up ticket makes sense. Could be something to do after all of the other modules have been converted.

scannillo: Agreed - a follow up ticket makes sense to me. Do we have one made already?

Sources/BraintreeCoreSwift/BTHTTP.swift
scannillo: See https://github.com/braintree/braintree_ios/pull/855/files#r953979091

Sources/BraintreeCoreSwift/BTHTTP.swift
jaxdesmarais: Yep! Already created one for this (ends in 1898).

Sources/BraintreeCoreSwift/BTHTTP.swift
scannillo: Addressed in this comment - https://github.com/braintree/braintree_ios/pull/855#discussion_r955000739

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -0,0 +1,542 @@
...
+
+        let mutableParameters: NSMutableDictionary = NSMutableDictionary(dictionary: parameters ?? [:])
+
+        if case .authorizationFingerprint(let fingerprint) = clientAuthorization {
+            mutableParameters["authorization_fingerprint"] = fingerprint
+        }
+
+        guard let fullPathURL = fullPathURL else {
+            // baseURL can be non-nil (e.g. an empty string) and still return nil for -URLByAppendingPathComponent:
+            // causing a crash when NSURLComponents.componentsWithString is called with nil.
```
scannillo, line 235:
```suggestion
            // baseURL can be non-nil (e.g. an empty string) and still return nil for appendingPathComponent(_:)
            // causing a crash when URLComponents(string:_) is called with nil.
```

Update to swift method names

scannillo (approved):

sshropshire (approved):
🌊

andrespelaezp (approved):

PR #856, opened 2022-08-23
Create ClientAuthorization enum

mattwylder:
### Summary of changes

- Replaced authorizationFingerprint and tokenizationKey with a private enum and associated values.

### Checklist

-

### DIFF
--- a/Sources/BraintreeCoreSwift/BTHTTP.swift
+++ b/Sources/BraintreeCoreSwift/BTHTTP.swift
@@ -8,13 +8,14 @@ import Security
 // TODO: - Mark interval vs private properties accordingly
     public typealias RequestCompletion = (BTJSON?, HTTPURLResponse?, Error?) -> Void
 
-    private enum AuthorizationType: Equatable {
+    private enum ClientAuthorization: Equatable {
         case authorizationFingerprint(String), tokenizationKey(String)
     }
     
     // MARK: - Public Properties
+    
     /// An array of pinned certificates, each an NSData instance consisting of DER encoded x509 certificates
-    public let pinnedCertificates = BTAPIPinnedCertificates.trustedCertificates()
+    public let pinnedCertificates: [NSData] = BTAPIPinnedCertificates.trustedCertificates()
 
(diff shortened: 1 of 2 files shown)


Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -490,7 +483,26 @@ import Security
...
+            }
+            
+        case .tokenizationKey(let key):
+            switch http.authType {
+            case .authorizationFingerprint:
+                return false
+            case .tokenizationKey(let httpKey):
+                return key == httpKey
+            }
+        }
```
mattwylder, line 505: this gets a lot longer and gross but idk how else to compare associated values

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -30,42 +34,31 @@ import Security
...
     // MARK: - Initializers
-
-    /// Initialize `BTHTTP` with the URL for the Braintree API
-    /// - Parameter url: The base URL for the Braintree Client API
-    @objc(initWithBaseURL:)
-    public init(url: URL) {
-        self.baseURL = url
-        self.cacheDateValidator = BTCacheDateValidator()
-        self.pinnedCertificates = BTAPIPinnedCertificates.trustedCertificates()
-    }
```
mattwylder, line 49: this init was never used, so I removed it and removed `convenience` from the two other inits.

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -306,8 +299,8 @@ import Security
             headers["Content-Type"] = "application/json; charset=utf-8"
         }
 
-        if tokenizationKey != "" {
-            headers["Client-Key"] = tokenizationKey
+        if case .tokenizationKey(let key) = authType {
+            headers["Client-Key"] = key
         }
```
mattwylder, line 304: `if case let` is unusual syntax but I like that the enum forces you to check whether we're using an authorizationFingerprint or a tokenizationKey before you can even access it.

PR #860, opened 2022-08-29
Add support for Xcode 14 & iOS 16

scannillo:
### Summary of changes

- Add support for Xcode 14 & iOS 16
- Tested all flows on device running iOS 16
- Removed deprecation warnings in Xcode 14

Note: These updates do not drop bitcode support - once we bump the minimum Xcode version to Xcode 14 for merchants, we will drop bitcode support.

### Checklist

- [x] Added a changelog entry

- @scannillo @mattwylder @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## unreleased
 * BraintreeSEPADirectDebit
   * Update nonce to pull in ibanLastFour and customerID as expected
+* Remove unused `presentationContextProvider` (fixes #854)
+  * Adds support for Xcode 14 and iOS 16 
 
 ## 5.11.0 (2022-07-20)
 * BraintreeSEPADirectDebit
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* Adds support for Xcode 14 and iOS 16 
 * BraintreeSEPADirectDebit
   * Update nonce to pull in ibanLastFour and customerID as expected
-* Remove unused `presentationContextProvider` (fixes #854)
-  * Adds support for Xcode 14 and iOS 16 
+  * Remove unused `presentationContextProvider` (fixes #854)
 
 ## 5.11.0 (2022-07-20)
 * BraintreeSEPADirectDebit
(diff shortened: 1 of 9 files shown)


Sources/BraintreeDataCollector/BTDataCollector.m
```diff
@@ -56,16 +61,19 @@ - (void)setUpKountWithDebugOn:(BOOL)debugLogging {
     self.kount = [KDataCollector sharedCollector];
     self.kount.debug = debugLogging;
 
-    CLAuthorizationStatus locationStatus = kCLAuthorizationStatusNotDetermined;
+    if ((locationStatus != kCLAuthorizationStatusAuthorizedWhenInUse && locationStatus != kCLAuthorizationStatusAuthorizedAlways) || ![CLLocationManager locationServicesEnabled]) {
```
scannillo, line 64: 🤔 How do we know that `locationManagerDidChangeAuthorization` will be called to set the `locationStatus` before this method (`setUpKountWithDebugOn`) is called?

jaxdesmarais: From the docs for [`locationManagerDidChangeAuthorization`](https://developer.apple.com/documentation/corelocation/cllocationmanagerdelegate/3563956-locationmanagerdidchangeauthoriz) it states:
```
Core Location always calls locationManagerDidChangeAuthorization(_:) when the user’s action results in an authorization status change, and when your app creates an instance of [CLLocationManager](https://developer.apple.com/documentation/corelocation/cllocationmanager), whether your app runs in the foreground or in the background.
```
That said, doing some testing with adding breakpoints this seems somewhat inconsistent on initial load, so we may need to revisit. 🤔

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## unreleased
 * BraintreeSEPADirectDebit
   * Update nonce to pull in ibanLastFour and customerID as expected
+* Remove unused `presentationContextProvider` (fixes #854)
```
sshropshire, line 6:
```suggestion
  * Remove unused `presentationContextProvider` (fixes #854)
```

sshropshire (approved):
🌊

andrespelaezp (approved):

scannillo:
LGTM!

PR #862, opened 2022-09-06
Add Analytics for PayPal Native Checkout

jaxdesmarais:
### Summary of changes

- add analytics events for PayPal Native Checkout
- add tests for analytics

### Checklist

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
+++ b/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
@@ -33,7 +33,7 @@ import PayPalCheckout
     ) {
         apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.started")
         guard let request = nativeRequest as? (BTPayPalRequest & BTPayPalNativeRequest) else {
-            apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.invalid-request.failure")
+            apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.invalid-request.failed")
             completion(nil, BTPayPalNativeError.invalidRequest)
             return
         }
@@ -64,7 +64,7 @@ import PayPalCheckout
                         completion(nil, BTPayPalNativeError.canceled)
                     },
                     onError: { error in
-                        self?.apiClient.sendAnalyticsEvent("ios.paypal-native.on-error.failure")
+                        self?.apiClient.sendAnalyticsEvent("ios.paypal-native.on-error.failed")
                         completion(nil, BTPayPalNativeError.checkoutSDKFailed)
                     },
                     environment: order.environment
@@ -73,7 +73,7 @@ import PayPalCheckout
                 PayPalCheckout.Checkout.set(config: payPalNativeConfig)
                 PayPalCheckout.Checkout.start()
             case .failure(let error):
-                self?.apiClient.sendAnalyticsEvent("ios.paypal-native.create-order.failure")
+                self?.apiClient.sendAnalyticsEvent("ios.paypal-native.create-order.failed")
                 completion(nil, error)
             }
         }
@@ -84,10 +84,10 @@ import PayPalCheckout
         tokenizationClient.tokenize(request: request, returnURL: approval.data.returnURL!.absoluteString) { result in
             switch result {
             case .success(let nonce):
-                self.apiClient.sendAnalyticsEvent("ios.paypal-native.on-approve.success")
+                self.apiClient.sendAnalyticsEvent("ios.paypal-native.on-approve.succeeded")
                 completion(nonce, nil)
             case .failure(let error):
-                self.apiClient.sendAnalyticsEvent("ios.paypal-native.on-approve.failure")
+                self.apiClient.sendAnalyticsEvent("ios.paypal-native.on-approve.failed")
                 completion(nil, error)
             }
         }
--- a/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
+++ b/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
@@ -31,9 +31,9 @@ import PayPalCheckout
         with nativeRequest: BTPayPalRequest,
         completion: @escaping (BTPayPalNativeCheckoutAccountNonce?, Error?) -> Void
     ) {
-        apiClient.sendAnalyticsEvent("ios.paypal-native-checkout.tokenize.started")
+        apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.started")
         guard let request = nativeRequest as? (BTPayPalRequest & BTPayPalNativeRequest) else {
-            apiClient.sendAnalyticsEvent("ios.paypal-native-checkout.tokenize.invalid-request.failure")
+            apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.invalid-request.failure")
             completion(nil, BTPayPalNativeError.invalidRequest)
             return
         }
@@ -42,7 +42,7 @@ import PayPalCheckout
         orderCreationClient.createOrder(with: request) { [weak self] result in
             switch result {
             case .success(let order):
-                self?.apiClient.sendAnalyticsEvent("ios.paypal-native-checkout.create-order.started")
+                self?.apiClient.sendAnalyticsEvent("ios.paypal-native.create-order.started")
                 let payPalNativeConfig = PayPalCheckout.CheckoutConfig(
                     clientID: order.payPalClientID,
                     createOrder: { action in
@@ -56,15 +56,15 @@ import PayPalCheckout
                         }
                     },
                     onApprove: { [weak self] approval in
-                        self?.apiClient.sendAnalyticsEvent("ios.paypal-native-checkout.on-approve.started")
+                        self?.apiClient.sendAnalyticsEvent("ios.paypal-native.on-approve.started")
                         self?.tokenize(approval: approval, request: request, completion: completion)
                     },
                     onCancel: {
-                        self?.apiClient.sendAnalyticsEvent("ios.paypal-native-checkout.canceled")
+                        self?.apiClient.sendAnalyticsEvent("ios.paypal-native.canceled")
                         completion(nil, BTPayPalNativeError.canceled)
                     },
                     onError: { error in
-                        self?.apiClient.sendAnalyticsEvent("ios.paypal-native-checkout.on-error.failure")
+                        self?.apiClient.sendAnalyticsEvent("ios.paypal-native.on-error.failure")
                         completion(nil, BTPayPalNativeError.checkoutSDKFailed)
                     },
                     environment: order.environment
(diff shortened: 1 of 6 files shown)


sshropshire (approved):

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
```diff
@@ -31,7 +31,9 @@ import PayPalCheckout
         with nativeRequest: BTPayPalRequest,
         completion: @escaping (BTPayPalNativeCheckoutAccountNonce?, Error?) -> Void
     ) {
+        apiClient.sendAnalyticsEvent("ios.paypal-native-checkout.tokenize.started")
```
scannillo, line 34: Can we align with the existing BTNXO analytic event naming on Android? [Example here](https://github.com/braintree/braintree_android/blob/6e003485313db9eafe32fb55966ddeed444da46f/PayPalNativeCheckout/src/main/java/com/braintreepayments/api/PayPalNativeCheckoutClient.java#L113). 

So `paypal-native` instead of `paypal-native-checkout`? This will help us when searching for analytics across platforms.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
```diff
@@ -40,6 +42,7 @@ import PayPalCheckout
         orderCreationClient.createOrder(with: request) { [weak self] result in
             switch result {
             case .success(let order):
+                self?.apiClient.sendAnalyticsEvent("ios.paypal-native-checkout.create-order.started")
```
scannillo, line 45: Also, it's ideal that we have the same analytic logging b/w platforms per payment method feature/flow. Not just in terms of naming, but also in terms of timing of analytic sendoff in the flow.

This PR adds more events than currently exist for Android NXO flow. This is a good opportunity to make sure we are tracking the same events across both platforms.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeOrderCreationClient.swift
```diff
@@ -28,16 +28,19 @@ class BTPayPalNativeOrderCreationClient {
     ) {
         apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
             guard let config = configuration, error == nil else {
+                self.apiClient.sendAnalyticsEvent("ios.paypal-native-checkout.create-order.fetch-configuration.failure")
```
scannillo, line 31: I don't think we need any config fetch analytics. None of our other payment method features have them, since the actual config fetch call already has analytics baked into it.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
jaxdesmarais: Good call - will update this

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
jaxdesmarais: Yeah I think we need a follow up ticket on the Android side to add additional events. Right now, we don't track success or failure at all for NXO on Android, just which flow was started. 🙈 I'll create an Android follow up ticket for an audit on that platform for NXO analytics events.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeTokenizationClient.swift
```diff
@@ -33,15 +33,18 @@ class BTPayPalNativeTokenizationClient {
         ) { body, _, error in
             guard let json = body, error == nil else {
                 let underlyingError = error ?? BTPayPalNativeError.invalidJSONResponse
+                self.apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.failure")
                 completion(.failure(.tokenizationFailed(underlyingError)))
                 return
             }
 
             guard let accountNonce = BTPayPalNativeCheckoutAccountNonce(json: json) else {
+                self.apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.parsing-result.failure")
```
scannillo, line 42: I don't think we need to get this nuanced with the failure analytics. I think `"ios.paypal-native.tokenize.failed"` is sufficient for all cases. Especially since we're already returning specific errors to the merchant for these error use cases - we will be able to learn from their integration which error they hit.

I think at the analytics level, we just want to know if it was success or not. Historically, we have over-complicated our analytics logging resulting in too many events with different names that we lose track of.

scannillo: We also created this [naming guide](https://github.braintreeps.com/team-sdk/sdk-knowledge/blob/main/analytics/events/event-names.md) to try to remedy some of that, and create some uniformity amongst platforms.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeOrderCreationClient.swift
```diff
@@ -33,11 +33,13 @@ class BTPayPalNativeOrderCreationClient {
             }
 
             guard config.json["paypalEnabled"].isTrue else {
+                self.apiClient.sendAnalyticsEvent("ios.paypal-native.create-order.paypal-not-enabled.failure")
```
scannillo, line 36: See [naming guide](https://github.braintreeps.com/team-sdk/sdk-knowledge/blob/main/analytics/events/event-names.md), prefer past tense such as `failed`.

scannillo: Also similar comment here regarding just needing `ios.paypal-native.create-order. succeeded` and `...failed` cases.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeTokenizationClient.swift
jaxdesmarais: > I don't think we need to get this nuanced with the failure analytics.

This came up during the merchant call for caching errors and the analytics team we worked with during post mortem recommended us getting more nuanced with the events to better track where we are failing when working with those teams. I don't really have a strong preference one way or the other though. 🤷

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeOrderCreationClient.swift
jaxdesmarais: Gotcha - can update this, had not seen this before, was just going off of venmo which uses success/failure

scannillo: Sounds good!

Yeh, there has been a lot of inconsistency with naming. We did an audit awhile ago and looked for the most common naming convention used and decided we should try to use that going forward. Re-naming existing analytics isn't usually worth it though because then we have 2 event names in the wild that mean the same thing, and that gets 🤯.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeTokenizationClient.swift
scannillo: Ah gotcha! Which post mortem was it for just curious? Which situation caused us to get that feedback?

scannillo: But if that's the feedback we got - makes sense to take it 👍

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
scannillo: Good spot!

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeOrderCreationClient.swift
jaxdesmarais: Ah, that makes sense - I am glad we have a guide for this!

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeTokenizationClient.swift
scannillo: @hollabaq86 - could we get your opinion here? I know you've been doing a handful of analytics work on open dev.

TLDR - do we favor more nuanced analytics events such as having several `ios.paypal-native.tokenize.<SPECIFIC_FAILURE_REASON>.failed` events, or instead favor just one `ios.paypal-native.tokenize.failed` event name that covers all tokenization failures?

hollabaq86: given that this is one of our larger payment methods, let's go granular and cite the specific failure reason if possible. Thanks for pinging me for my thoughts!

scannillo: 👍 Let's do it!

jaxdesmarais:
The UI test `testTokenizeVenmo_whenSignInSuccessfulWithPaymentContext_returnsToApp` is failing in main as well as this branch. Digging into what may have changed to cause this test to succeed yesterday but not today. Since it's also happening on main it is not related to the changes in this PR.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeOrderCreationClient.swift
```diff
@@ -62,6 +65,7 @@ class BTPayPalNativeOrderCreationClient {
             ) { json, response, error in
                 guard let hermesResponse = BTPayPalNativeHermesResponse(json: json), error == nil else {
                     let underlyingError = error ?? BTPayPalNativeError.invalidJSONResponse
+                    self.apiClient.sendAnalyticsEvent("ios.paypal-native.create-order.order-creation.failed")
```
scannillo, line 68: `.create-order.order-creation.failed` is a little confusing. Maybe `.create-order.hermes-request.failed`?

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeTokenizationClient.swift
```diff
@@ -33,15 +33,18 @@ class BTPayPalNativeTokenizationClient {
         ) { body, _, error in
             guard let json = body, error == nil else {
                 let underlyingError = error ?? BTPayPalNativeError.invalidJSONResponse
+                self.apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.failed")
```
scannillo, line 36: Similar comment to above. If the other events are granular, this one also might as well be?

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeTokenizationClient.swift
```diff
@@ -33,15 +33,18 @@ class BTPayPalNativeTokenizationClient {
         ) { body, _, error in
             guard let json = body, error == nil else {
                 let underlyingError = error ?? BTPayPalNativeError.invalidJSONResponse
+                self.apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.request-to-tokenize.failed")
```
scannillo, line 36: Nitpick

```suggestion
                self.apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.tokenize-url-request.failed")
```

scannillo (approved):

PR #865, opened 2022-09-08
Update Venmo UI Test

jaxdesmarais:
### Summary of changes

- Due to security updates to Venmo we can no longer use the static payment context ID used previously as the ID is tied to a different merchant ID than the one used in the test.

### Checklist

- @jaxdesmarais

### DIFF
--- a/Demo/MockVenmo/AppSwitcher.swift
+++ b/Demo/MockVenmo/AppSwitcher.swift
@@ -14,7 +14,7 @@ class AppSwitcher {
 
         successComponents?.queryItems = [
             URLQueryItem(name: "x-source", value: "Venmo"),
-            URLQueryItem(name: "resource_id", value: "cGF5bWVudGNvbnRleHRfaGg0Y3BjMzl6cTRyZ2pjZyMxYmU5OTBiNC02YTE1LTQ2ZTQtYmNmOS1iYmIwMzY4OWMwMmI=")
+            URLQueryItem(name: "resource_id", value: "cGF5bWVudGNvbnRleHRfZGNwc3B5MmJyd2RqcjNxbiM4NjE4ZThkYi0xZDJkLTQwYjktYWJjOC0zNTVlNTk5YzliNTg=")
         ]
 
         return successComponents?.url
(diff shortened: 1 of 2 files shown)


Demo/UI Tests/Venmo UI Tests/Venmo_UITests.swift
```diff
@@ -22,11 +22,11 @@ class Venmo_UITests: XCTestCase {
         demoApp.buttons["Venmo (custom button)"].tap()
     }
     
-    func testTokenizeVenmo_whenSignInSuccessfulWithPaymentContext_returnsToApp() {
+    func testTokenizeVenmo_whenSignInSuccessfulWithPaymentContext_returnsNonce() {
```
jaxdesmarais, line 25: We updated this test in [this PR with the following comment](https://github.com/braintree/braintree_ios/pull/807/commits/c24079ad1279ce6a7d825f8c8801d6826571001d#r848898247). Seems like we should be fine to revert this since we are updating the payment context ID to now match the merchant ID from the client token.

Demo/MockVenmo/AppSwitcher.swift
```diff
@@ -14,7 +14,7 @@ class AppSwitcher {
 
         successComponents?.queryItems = [
             URLQueryItem(name: "x-source", value: "Venmo"),
-            URLQueryItem(name: "resource_id", value: "cGF5bWVudGNvbnRleHRfaGg0Y3BjMzl6cTRyZ2pjZyMxYmU5OTBiNC02YTE1LTQ2ZTQtYmNmOS1iYmIwMzY4OWMwMmI=")
+            URLQueryItem(name: "resource_id", value: "cGF5bWVudGNvbnRleHRfZGNwc3B5MmJyd2RqcjNxbiM4NjE4ZThkYi0xZDJkLTQwYjktYWJjOC0zNTVlNTk5YzliNTg=")
```
jaxdesmarais, line 17: The client token used in the demo app is the same as the one used in the tests. I ran through the Venmo flow and pulled a payment context ID from the demo app to now be used here.

scannillo (approved):
LGTM - I wonder if that payment contextID we pulled will expire? But for now this seems like the best option.

jaxdesmarais:
> LGTM - I wonder if that payment contextID we pulled will expire? But for now this seems like the best option.

I asked about this in the slack thread and the old one we were using previously expired in 2021. They said the only issue would be if we tried to approve or update the payment context, so we should be good since we aren't doing that in the test.

PR #866, opened 2022-09-09
Update CocoaPods to pin NXO version

scannillo:
### Background

- For SPM integrations, we pin each version release of `braintree_ios` to the exact version of NXO ([see Package.swift](https://github.com/braintree/braintree_ios/blob/7da79965033cfdf95c9261aeefd973d90db38a05/Package.swift#L64)).
- From talking with the NXO team, this is because they aren't on SemVer yet and new releases aren't guaranteed to be non-breaking.

### Changes
- This PR updates CocoaPods to use the same NXO dependency resolution behavior as SPM - exact version pinning.

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/Braintree.podspec
+++ b/Braintree.podspec
@@ -80,7 +80,7 @@ Pod::Spec.new do |s|
     s.source_files = "Sources/BraintreePayPalNativeCheckout/*.swift"
     s.dependency "Braintree/Core"
     s.dependency "Braintree/PayPal"
-    s.dependency "PayPalCheckout", '~> 0.100.0'
+    s.dependency "PayPalCheckout", '= 0.100.0'
   end
 
   s.subspec "ThreeDSecure" do |s|
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPalNativeCheckout (BETA)
+  * Fix CocoaPods integrations to pin exact `PayPalCheckout` version `0.100.0` to match Swift Package Manager
+
 ## 5.12.0 (2022-09-07)
 * Adds support for Xcode 14 and iOS 16 
 * BraintreeSEPADirectDebit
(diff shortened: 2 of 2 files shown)


Braintree.podspec
```diff
@@ -80,7 +80,7 @@ Pod::Spec.new do |s|
     s.source_files = "Sources/BraintreePayPalNativeCheckout/*.swift"
     s.dependency "Braintree/Core"
     s.dependency "Braintree/PayPal"
-    s.dependency "PayPalCheckout", '~> 0.100.0'
+    s.dependency "PayPalCheckout", '= 0.100.0'
```
jaxdesmarais, line 83: nit: Is this is the same as doing:
```
    s.dependency "PayPalCheckout", '0.100.0'
```
Is the `=` needed?

sshropshire (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPalNativeCheckout (BETA)
+  * Fix CocoaPods integrations to pin exact `PayPalCheckout` version `0.100.0`
```
jaxdesmarais, line 5: Do we want this to be specific to just this version or do something like `Fix CocoaPods integrations to pin exact PayPalCheckout version` since this will also be fixed for all future versions as well?

jaxdesmarais (approved):
Couple of very small questions but overall this looks good!

Braintree.podspec
scannillo: Good question - the [Podspec](https://guides.cocoapods.org/syntax/podspec.html) docs don't call it out, but let me try locally with a `pod lib lint`

CHANGELOG.md
scannillo: Done!

Braintree.podspec
scannillo: Done!

PR #868, opened 2022-09-19
Convert `BTAPIHTTP` + `BTGraphQLHTTP` to Swift

jaxdesmarais:
### Summary of changes

- Convert BTAPIHTTP and BTGraphQLHTTP to Swift
- There are some TODOs in here that will be addressed in follow up PRs
- Delete BTHTTP, BTGraphQLHTTP, and BTAPIHTTP Objective-C files

### Checklist

- @sshropshire @scannillo @mattwylder @jaxdesmarais


IntegrationTests/Helpers/Helpers.swift
```diff
@@ -1,16 +1,16 @@
 import Foundation
 
-@objc class Helpers : NSObject {
+@objcMembers public class Helpers: NSObject {
```
scannillo, line 3: Curious why test files had to change?

Sources/BraintreeCoreSwift/BTAPIHTTP.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+// TODO: Make internal once rest of core is in Swift
+@objcMembers public class BTAPIHTTP: BTHTTP {
+
+    let accessToken: String?
+
+    @objc(initWithBaseURL:accessToken:)
+    public init(url: URL, accessToken: String? = "") {
```
scannillo, line 9: Why default to an empty string? Can we make this a required param instead?

Sources/BraintreeCoreSwift/BTCoreConstants.swift
```diff
@@ -5,4 +5,10 @@ public class BTCoreConstants {
 
     /// :nodoc:
     public static var braintreeSDKVersion: String = "5.11.0"
```
scannillo, line 7: Add TODO for updating our release script to find this new constant to update.

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -69,7 +74,13 @@ import Security
     /// - Parameter clientToken: The client token
     @objc(initWithClientToken:error:)
     public convenience init(clientToken: BTClientToken) throws {
-        guard let clientApiURL = clientToken.json["clientApiUrl"].asURL() else {
+        let url: URL
+
+        if clientToken.json["clientApiUrl"].asURL() != nil, let clientApiURL = clientToken.json["clientApiUrl"].asURL() {
+            url = clientApiURL
+        } else if clientToken.json["configUrl"].asURL() != nil, let configURL = clientToken.json["configUrl"].asURL() {
+            url = configURL
```
scannillo, line 82: Why are we adding this logic here? Not seeing where that comes from in the original BTHTTP.m?

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -476,19 +491,22 @@ import Security
         "Braintree/iOS/\(BTCoreConstants.braintreeSDKVersion)"
     }
 
+    func acceptString() -> String {
+        "application/json"
+    }
```
scannillo, line 496: Use property instead of function for String constant

Sources/BraintreeCoreSwift/BTHTTP.swift
```diff
@@ -476,19 +491,22 @@ import Security
...
+
     func acceptLanguageString() -> String {
         "\(Locale.current.languageCode ?? "en")-\(Locale.current.regionCode ?? "US")"
```
scannillo, line 499: Similar comment here - can we use a computed property?

Sources/BraintreeCoreSwift/BTAPIHTTP.swift
scannillo, line 9: Is [this logic](https://github.com/braintree/braintree_ios/blob/91cfc9567622099af7764b512f1104af9c15a9a5/Sources/BraintreeCore/BTAPIHTTP.m#L17-L25) from the ObjC implementation preserved?

Sources/BraintreeCoreSwift/BTGraphQLHTTP.swift
```diff
@@ -0,0 +1,244 @@
...
+    public override func delete(_ path: String, parameters: NSDictionary? = nil, completion: RequestCompletion?) {
+        NSException(name: exceptionName, reason: "DELETE is unsupported").raise()
+    }
+
+    // MARK: - Internal methods
+    
+    func httpRequest(
+        method: String,
+        parameters: NSDictionary? = [:],
+        completion: RequestCompletion?
```
scannillo, line 51: `RequestionCompletion` can be non-optional?

Sources/BraintreeCoreSwift/BTGraphQLHTTP.swift
```diff
@@ -0,0 +1,244 @@
...
+    ) {
+        var errorUserInfo: [String: Any] = [:]
+
+        if self.baseURL.absoluteString.isEmpty || self.baseURL.absoluteString == "" {
+            errorUserInfo["method"] = method
+            errorUserInfo["parameters"] = parameters
+            let error = Self.constructError(code: .missingBaseURL, userInfo: errorUserInfo)
```
scannillo, line 58: Do we have plans to convert this to something like ....

```
let error = BTHTTPError.missingBaseURL(userInfo)
```

like we're using for some other `Error.swift` files?

Sources/BraintreeCoreSwift/BTAPIHTTP.swift
```diff
@@ -0,0 +1,23 @@
...
+        self.accessToken = accessToken
+        super.init(url: url)
```
scannillo, line 11: Making the `accessToken` non-optional would also let us remove the `BTHTTP.init(url: URL)` internal helper method

Sources/BraintreeCoreSwift/BTGraphQLHTTP.swift
```diff
@@ -0,0 +1,244 @@
...
+        var request: URLRequest
+    
+        do {
+            let bodyData = try JSONSerialization.data(withJSONObject: parameters ?? [:])
+            request = URLRequest(url: urlFromComponents)
+            request.httpBody = bodyData
+            request.allHTTPHeaderFields = headers
+            request.httpMethod = method
+            
+            let task: URLSessionTask = session.dataTask(with: request) { [weak self] data, response, error in
```
scannillo, line 107:
```suggestion
            // Perform the actual request
            let task: URLSessionTask = session.dataTask(with: request) { [weak self] data, response, error in
```

Nitpick comment - can we preserve this comment? It always helped drill down to the actual request, since it's pretty obfuscating and hard to find.

Sources/BraintreeCoreSwift/BTGraphQLHTTP.swift
```diff
@@ -0,0 +1,244 @@
...
+                guard let self = self else { return }
+                self.handleRequestCompletion(data: data, response: response, error: error, completion: completion)
+            }
+
+            task.resume()
```
scannillo, line 112: Nitpick
```suggestion
            }.resume()
```

Sources/BraintreeCoreSwift/BTGraphQLHTTP.swift
```diff
@@ -0,0 +1,244 @@
...
+    @objc(handleRequestCompletion:response:error:completionBlock:)
+    public func handleRequestCompletion(
+        data: Data?,
+        response: URLResponse?,
+        error: Error?,
+        completion: RequestCompletion?
+    ) {
+        guard let completion = completion else {
+            return
+        }
```
scannillo, line 127:
```suggestion
        guard let completion = completion else { return }
```

If the guard's is just returning, we can make it a 1-liner. Thoughts?

Sources/BraintreeCoreSwift/BTAPIHTTP.swift
sshropshire: There was some wild stuff going on where the tests where crashing, and it all hinged on the [config fetch code](https://github.com/braintree/braintree_ios/blob/91cfc9567622099af7764b512f1104af9c15a9a5/Sources/BraintreeCore/BTAPIClient.m#L303). We got crashes in both unit and integration tests when the different API and HTTP clients are instantiated without access tokens. Swift was less forgiving than Obj-C in this case.

This is a workaround – we tried some other stuff but it was turning into a rabbit hole. We felt that `BTAPIClient` constructing dependencies in `fetchOrReturnRemoteConfiguration` was a potential root cause to consider refactoring since the method does more than its name implies.

IntegrationTests/Helpers/Helpers.swift
jaxdesmarais: It doesn't need to be - reverting this

Sources/BraintreeCoreSwift/BTAPIHTTP.swift
jaxdesmarais: We can make this non-optional now that we are handling this with the updated check in `fetchOrReturnRemoteConfiguration`. Good call.

Sources/BraintreeCoreSwift/BTHTTP.swift
jaxdesmarais: Now that the URL is required as part of the init here: `self.init(url: url, authorizationFingerprint: clientToken.authorizationFingerprint)` we need to get this URL from the client token. In v1 and v2 client tokens the URL is in the JSON at `clientApiUrl`. In v3 client tokens the URL is in the JSON at `configUrl`. 

We did not catch this when converting `BTHTTP` because we just converted the file but hadn't updated anything else to use the Swift version of `BTHTTP`. Once we updated the SDK to use all of the Swift classes, the integration tests using `SANDBOX_CLIENT_TOKEN_VERSION_3` began to fail. Decoding the client tokens in `IntegrationTestsConstants` showed the change in the URL between v1/v2 and v3 client tokens.

jaxdesmarais: The v1/v2 client tokens contain both a `clientApiUrl` and `configUrl` while v3 just contains a `configUrl`. So it's possible that we can just use the `configUrl` and remove the reference to `clientApiUrl`?

Sources/BraintreeCoreSwift/BTAPIHTTP.swift
jaxdesmarais: Yep, this is all done in `BTHTTP` now since it is the same across all of the classes that subclass `BTHTTP`

Sources/BraintreeCoreSwift/BTGraphQLHTTP.swift
jaxdesmarais: Right now `BTHTTPError` is also used in `BTCardClient`, `PayPalClient`, `ThreeDSecure`, and `UnionPay` modules. We likely can update it in those places, but that's a larger piece of work. There is already a ticket for this work so it's certainly something we should tackle in next major version (ticket ends in `1898`)

Sources/BraintreeCoreSwift/BTAPIHTTP.swift
jaxdesmarais: We would still need to call one of the super inits here which would mean we would need a client token, tokenization key, or an authorization fingerprint. I don't think we'd have any of these from the access token correct?

Sources/BraintreeCoreSwift/BTGraphQLHTTP.swift
jaxdesmarais: Is this the approach we want to take in the entire SDK? We currently do this across several of the HTTP clients, so should update them all if we want to do:
```
session.dataTask(with: request) { [weak self] data, response, error in
    // Do request stuff
}.resume()
```
over the current:
```
let task: URLSessionTask = session.dataTask(with: request) { [weak self] data, response, error in
    // Do request stuff
}

task.resume()
```
We should add whichever we prefer to the style guide since I could see it potentially being done both ways.

Sources/BraintreeCoreSwift/BTGraphQLHTTP.swift
jaxdesmarais: Agree, I'm a fan of the single line for these

UnitTests/BraintreeTestShared/FakeHTTP.swift
```diff
@@ -35,102 +35,104 @@ import BraintreeCore.Private
...
         GETRequestCount += 1
-        lastRequestEndpoint = endpoint
+        lastRequestEndpoint = path
         lastRequestParameters = parameters
         lastRequestMethod = "GET"
 
         if cannedError != nil {
             dispatchQueue.async {
-                completionBlock!(nil, nil, self.cannedError)
+                completion!(nil, nil, self.cannedError)
```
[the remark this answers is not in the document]
jaxdesmarais: Good catch, we do a lot of force unwrapping in our tests in general that need to be addressed at some point. 🙃 For these, since we are defaulting them to `nil` I think we can make them all optional completions vs force unwrapping them.

Sources/BraintreeCoreSwift/BTHTTP.swift
scannillo: Got it - thank you for the thorough explanation! Yeh, it would be interesting to know which should take priority. Did we check what Android and Web are doing?

Sources/BraintreeCoreSwift/BTGraphQLHTTP.swift
scannillo: Good call to note this in the style guide! I prefer the less-line option, as it's sometimes easy to forget that the `resume()` call is what actually kicks off the session. What do you prefer?

jaxdesmarais: Added to the style guide! Wasn't sure totally how to word it so feel free to make edits.

Sources/BraintreeCoreSwift/BTAPIHTTP.swift
scannillo: Yeh - I think my comment gets into the territory of rearchitecting our network layer and beyond converting to Swift. We can disregard for now!

jaxdesmarais: Ah, gotcha. Yeah we should certainly look at that more holistically at some point - would be great to modernize our networking layer but agree that is a large task that touches a lot of pieces. 🚀

Sources/BraintreeCoreSwift/BTGraphQLHTTP.swift
jaxdesmarais: Added a TODO for this here: https://github.com/braintree/braintree_ios/blob/876a040bf14922b9875b4f1c3d12936a57c47ee7/Sources/BraintreeCoreSwift/BTHTTP.swift#L490

jcnoriega (approved):

kgangineni (approved):

PR #869, opened 2022-09-26
Remove ENABLE_BITCODE from targets

scannillo:
### Background

- We got [this merchant inbound](https://braintree.slack.com/archives/C02CXD2LC/p1663855948470229) which shows an app getting rejected when compiling with Xcode 14, and PPRiskMagnes is built with bitcode.
    - [Apple's release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes) say _the App Store no longer accepts bitcode submissions from Xcode 14_.
- I think the reason that none of the Braintree modules were marked as "bitcode" issues in that merchant's AppStore Connect log is because SPM & CocoaPods do their own handling of bitcode support when compiling the frameworks.
   - For ex: in CocoaPods a merchant can specify their bitcode setting like [this](https://stackoverflow.com/questions/32640149/disable-bitcode-for-project-and-cocoapods-dependencies-with-xcode-7).
   - However, since PPRiskMagnes ships as a binary, CocoaPods & SPM can't control if it's built with bitcode or not.

### This PR
- This PR removes the ENABLE_BITCODE setting (aka set ENABLE_BITCODE to NO) for all of our framework targets.
    - This will only affect Carthage merchants.

### Checklist

- [X] Added a changelog entry


jaxdesmarais (approved):
Excited to get rid of all of those pesky bitcode warnings. Do we know if Cardinal has bitcode enabled by default? If so we should probably create a ticket with them since I could see this coming up for 3DS merchants as well.

sshropshire (approved):
🌊

scannillo:
> Do we know if Cardinal has bitcode enabled by default? If so we should probably create a ticket with them since I could see this coming up for 3DS merchants as well.

Great question. So I tried to replicate this merchant's app rejection by uploading to TestFlight. TestFlight unfortunately did not trigger the same warning, so we don't have a great way to end-to-end test this.

In terms of which other binaries have bitcode enabled, doing a grep for `bitcode` in the repo only highlights Magnes & Kount. After googling a bit, I can't come up with a better way to determine if a pre-packaged binary was built with bitcode. (I honestly wouldn't be surprised if Cardinal never enabled bitcode in the first place).

I will open an issue with Kount, and we can decide if we want to open one for safety with Cardinal, as well.

scannillo:
@jaxdesmarais Cardinal ticket created [here](https://support.cardinalcommerce.com/s/case/5008b000027buX0AAI/xcode-14-bitcode-apps-rejected-from-app-store)!

PR #872, opened 2022-09-29
Update PayPalCheckout binary, cartfile, podspec to 0.108.0. Also adds a PostNotification to update analytics config internal to the SDK to track Braintree as source.

jwarmkessel:
### Summary of changes

Update PayPalCheckout binary, cartfile, pod spec to 0.108.0. Also adds a PostNotification to update analytics config internal to the SDK to track Braintree as source.

### Checklist

- [X] Added a changelog entry

### Authors
@jwarmkessel


Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
```diff
@@ -71,6 +71,9 @@ import PayPalCheckout
                 )
 
                 PayPalCheckout.Checkout.set(config: payPalNativeConfig)
+                
+                NotificationCenter.default.post(name: Notification.Name("brain_tree_source_event"), object: nil)
```
jaxdesmarais, line 75: Should this be `braintree_source_event`? The company name is all one word or do we also separate out PayPal like `pay_pal` for these events?

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## unreleased
 * Remove `ENABLE_BITCODE` from framework target build settings
   * _The App Store no longer accepts bitcode submissions from Xcode 14_
+* BraintreePayPalNativeCheckout (BETA)
+   * Post Notification to PayPalCheckout SDK to update Analytics Config
```
jaxdesmarais, line 7: I don't think we need a changelog since this change is not merchant facing

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
jwarmkessel: This string has already been merged in the SDK and a release is being cut. Will have to change it next cycle but this string is not used in reporting/recording. Only used as a key.

CHANGELOG.md
jwarmkessel: Changelog is removed

scannillo:
Currently, we're including local copies of NXO in this repo [here](https://github.com/braintree/braintree_ios/tree/master/Frameworks). Until we modify our release process, we will need to update this PR to include those binaries.

jaxdesmarais:
Looking at the `.plist` of `PayPalCheckout.framework` it's still pointing to 0.106.0:
```
<string>0.106.0</string>
```

Please update the `.framework` as well to point to the correct version in `braintree_ios` > `Frameworks` > `FatFrameworks` > `PayPalCheckout.framework`

jwarmkessel:
> Looking at the `.plist` of `PayPalCheckout.framework` it's still pointing to 0.106.0:
> 
> ```
> <string>0.106.0</string>
> ```
> 
> Please update the `.framework` as well to point to the correct version in `braintree_ios` > `Frameworks` > `FatFrameworks` > `PayPalCheckout.framework`

Just pushed a change @jaxdesmarais

jaxdesmarais (approved):
👻  thanks for making these updates!

Podfile.lock
```diff
@@ -37,12 +37,12 @@ SPEC REPOS:
...
+  InAppSettingsKit: 37df0b44132380d4c7db6fc7cded92997e29873a
+  OCMock: 9491e4bec59e0b267d52a9184ff5605995e74be8
+  OHHTTPStubs: 90eac6d8f2c18317baeca36698523dc67c513831
+  Specta: b79d84043684b35ffdc2680df578dc318ec2efc2
+  xcbeautify: 663b4c0c6e1ba22b5dfaf9382202473d3bd52046
 
 PODFILE CHECKSUM: 75163f16229528991a9364c7c1a44cd57a30cac6
 
-COCOAPODS: 1.11.2
+COCOAPODS: 1.10.2
```
scannillo, line 48: 👋 We probably don't want to downgrade our CocoaPods version.

_For some background, we use this Podfile for internal dependencies on our Demo app and unit test suite. We also use [Bundler](https://github.com/braintree/braintree_ios/blob/master/Gemfile.lock) to manage our CocoaPods versions, so that each dev contributing is using the same version of CocoaPods (`bundle exec pod install`)._

For the purpose of this PR - I think we can just not commit the changes to this file and can handle bumping some of our internal pod dependencies in another PR.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
```diff
@@ -48,6 +48,7 @@ import PayPalCheckout
                     createOrder: { action in
                         switch request.paymentType {
                         case .checkout:
+                          print("JUSTIN ------ \(order.orderID)")
```
scannillo, line 51:
```suggestion
```

jwarmkessel: Thank you, really good catch!

mattwylder (approved):

PR #874, opened 2022-10-05
convert BTAPIClient_Tests to Swift

jaxdesmarais:
### Summary of changes

- Convert `BTAPIClient_Tests.m` to Swift - this will allow us to make properties internal off the bat vs revisiting and potentially missing some properties that shouldn't be public
- Rename `BTAPIClient_SwiftTests` to `BTAPIClient_Tests` - renaming this file caused it to show as deleted even though we just added to and renamed the previous file
- All tests have been converted 1:1 and moved to the Swift test file
- General test file cleanup to force unwrapping, removing nil completion handles, fixing typos, etc

### Checklist

- @jaxdesmarais


sshropshire (approved):
🌊

PR #876, opened 2022-10-12
Convert BTAPIClient to Swift

jaxdesmarais:
### Summary of changes

- Convert the majority of BTAPIClient to Swift
- NOTE: In order to keep this PR at a manageable size, the Analytics methods have not been converted. This is because they require the analytics classes to be converted as well. These are marked with TODOs and will be converted in a follow up PR where the Objective-C files for BTAPIClient will also be deleted.
- This is branched off of #874 to allow us to move properties to internal in the same way they are currently in BTAPIClient_Internal.h. The test PR will need to be merged in before this PR can be merged.

### Checklist

- ~[ ] Added a changelog entry~ No public interface changes were made in this PR

- @jaxdesmarais


Sources/BraintreeCoreSwift/BTAPIClientHTTPType.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+/// :nodoc:
+@objc public enum BTAPIClientHTTPTypeSwift: Int {
```
jaxdesmarais, line 4: This needs to be public since it is used in the get and post public functions. Those functions are used by other modules in the SDK to make requests.

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
```diff
@@ -1,5 +1,440 @@
...
+    public func fetchPaymentMethodNonces(completion: @escaping ([BTPaymentMethodNonce]?, Error?) -> Void) {
+        fetchPaymentMethodNonces(false, completion: completion)
+    }
+
+    /// Fetches a customer's vaulted payment method nonces.
+    /// Must be using client token with a customer ID specified.
+    ///  - Parameters:
+    ///   - defaultFirst: Specifies whether to sort the fetched payment method nonces with the default payment method or the most recently used payment method first
+    ///   - completion: Callback that returns either an array of payment method nonces or an error
+    public func fetchPaymentMethodNonces(_ defaultFirst: Bool, completion: @escaping ([BTPaymentMethodNonce]?, Error?) -> Void) {
```
jaxdesmarais, line 196: In the Objective-C file we have a TODO to move this function to the Drop-in. That said, with the Drop-in getting less love these days I could see this function being useful for core merchants. Curious to hear other thoughts!

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
```diff
@@ -1,5 +1,440 @@
...
+    func metaParametersWith(parameters: [String: Any], forHTTPType: BTAPIClientHTTPTypeSwift) -> [String: Any] {
+        // TODO: Implement when BTAnalyticsService is converted to Swift
+        return [:]
+    }
+
+    // MARK: - Internal Static Methods
+
+    static func baseURLFromTokenizationKey(_ tokenizationKey: String) -> URL? {
+        let pattern: String = "([a-zA-Z0-9]+)_[a-zA-Z0-9]+_([a-zA-Z0-9_]+)"
+        guard let regularExpression = try? NSRegularExpression(pattern: pattern) else { return nil }
```
jaxdesmarais, line 325: Looked into using Swift Regex, but it seems to only be available with iOS 15+ so it seems that we may need to continue to use NSRegularExpression for the time being.

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
sshropshire: Yeah we moved this to DropIn because of circular dependencies on Android but we did get an inbound for a merchant asking to bring it back. If a merch wanted to build their own DropIn like UI it would be useful.

sshropshire (approved):
🌊

jaxdesmarais:
Some 3DS UI Tests are blowing up, seems like Cardinal broke a bunch of the test cards (again). Gonna let these tests fail for now as it typically resolves itself and will kick it again in a bit.

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
jaxdesmarais: Yeah, even thinking for building a UI in general would be a nice helper function to have in the SDK.

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
```diff
@@ -1,5 +1,441 @@
...
+    /// Perfom an HTTP GET on a URL composed of the configured from environment and the given path.
+    /// - Parameters:
+    ///   - path: The endpoint URI path.
+    ///   - parameters: Optional set of query parameters to be encoded with the request.
+    ///   - completion:  A block object to be executed when the request finishes.
+    ///   On success, `body` and `response` will contain the JSON body response and the
+    ///   HTTP response and `error` will be `nil`; on failure, `body` and `response` will be
+    ///   `nil` and `error` will contain the error that occurred.
+    @objc(GET:parameters:completion:)
+    public func get(_ path: String, parameters: NSDictionary? = nil, completion: @escaping RequestCompletion) {
```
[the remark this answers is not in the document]
jaxdesmarais: We can now! We were waiting until BTAPIClient was converted to update since it uses the HTTP methods under the hood. Pushing up this change and removing the TODO we had for it. 🚀

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
```diff
@@ -1,5 +1,441 @@
...
+    func httpForType(_ httpType: BTAPIClientHTTPTypeSwift) -> BTHTTP? {
+        switch httpType {
+        case .gateway:
+            return http
+        case .braintreeAPI:
+            return apiHTTP
+        case .graphQLAPI:
+            return graphQLHTTP
+        }
+    }
```
mattwylder, line 440: Is this reasonable? The optional return type leads to optional chaining in the `post` and `get` requests which would be nice to avoid. we don't have to repeat the line `return http` as I suggest below, I just wanted to make sure the `gateway` case was explicitly decleared, and figured that `http` makes the most sense as a default.

I also moved the word `for` into the params just as a nitpick. 

```suggestion
    func http(for _ httpType: BTAPIClientHTTPTypeSwift) -> BTHTTP {
           switch httpType {
         case .gateway:
             return http
         case .braintreeAPI:
             return apiHTTP
         case .graphQLAPI:
             return graphQLHTTP
         default:
             return .http
         }
        }
    }
```

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
```diff
@@ -1,5 +1,441 @@
...
+            "session_id": metadata?.sessionID ?? ""
+        ]
+
+        get("v1/payment_methods", parameters: parameters) { body, response, error in
+            if let error {
+                completion(nil, error)
+                return
+            }
+
+            var paymentMethodNonces: [BTPaymentMethodNonce]? = []
```
mattwylder, line 214: Could this be non-optional? Or is there a chance it will be set to nil?

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
mattwylder: Actually... could this be a computed variable on the `BTAPIClientHTTPTypeSwift` enum itself?

mattwylder (changes requested):

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
```diff
@@ -1,5 +1,441 @@
...
+
+    /// Client metadata that is used for tracking the client session
+    var metadata: BTClientMetadata?
+
+    /// Used to fetch and store configurations in the URL Cache of the session
+    var configurationHTTP: BTHTTP?
+
+    var http: BTHTTP?
+    var apiHTTP: BTAPIHTTP?
+    var graphQLHTTP: BTGraphQLHTTP?
```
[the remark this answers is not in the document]
jaxdesmarais: Right now they need to be optional since we don't know the values coming into this class for any of these and this class essentially acts as a pass-thru to BTHTTP depending on what is set. 😕 Making them optional requires us to rewrite this networking layer from scratch, which we _should_ do at some point but probably not in this PR/major version.

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
```diff
@@ -1,5 +1,441 @@
...
+
+    var session: URLSession {
+        let configurationQueue: OperationQueue = OperationQueue()
+        configurationQueue.name = "com.braintreepayments.BTAPIClient"
```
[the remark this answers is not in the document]
jaxdesmarais: That's a great question, the URLSession is sandboxed per app so the name of the queue doesn't matter and each merchant app has their own instance of the SDK and cannot share types or URLSession/OperationQueue from other apps under the hood. We'd just have a problem with multiple classes sharing the same OperationQueue name across the SDK.

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
jaxdesmarais: 🤔  so this is a bit tricky because here we are actually returning the property defined here:
```
    var http: BTHTTP?
    var apiHTTP: BTAPIHTTP?
```

I don't think we want the enum to have access to the properties defined here. Making this required would require us to nil coalesce instead of provide a default and I don't think we want to provide a property when the properties defined at the top are nil, right?

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
jaxdesmarais: It can be non-optional. Updating.

Sources/BraintreeCoreSwift/BTAPIClientSwift.swift
mattwylder: ooooh I read this as returning enum instances

mattwylder (approved):

PR #882, opened 2022-10-19
Restructure Pulling in Native Checkout for SPM

jaxdesmarais:
### Summary of changes

When we pull in the Braintree SDK via SPM, it actually does a `git clone -mirror` [behind the scenes](https://github.com/apple/swift-package-manager/blob/629064c733fd680994a1e487984e63f8c4c8ebd8/Sources/SourceControl/GitRepository.swift#L85)  so when a merchant pulls in our SDK via SPM, they clone the entire history of our SDK as well. This means if we continue down the current path of adding the large NXO binaries directly to the repo download times will take increasingly longer with each update due to NXO being close to 100mb per version.

This PR allows us to do the following:
1. Not keep the large NXO binary in repo
2. Avoid issue where the NXO dependency was being included with ALL Braintree modules (such as BraintreeCard, and BraintreeDropIn which don't required NXO) for SPM users
3. Avoids having to move to GitHubs large file storage due to the size of NXO

This PR updates our Package.swift to reference the [.xcframework.zip files in the PayPal Checkout release tags](https://github.com/paypal/paypalcheckout-ios/releases/tag/0.108.0) vs keeping a local reference to the binary.

**Before:**
```
        .binaryTarget(
            name: "PayPalCheckout",
            path: "Frameworks/XCFrameworks/PayPalCheckout.xcframework"
        ),
```

**Proposed in PR:**
```
        .binaryTarget(
            name: "PayPalCheckout",
            url: "https://github.com/paypal/paypalcheckout-ios/releases/download/0.108.0/PayPalCheckout.xcframework.zip",
            checksum: "f186036ec1b180f9cb84a48ea00ca5835a3fce3af8d112b1e3067fa9d56dcf78"
        ),
```

There are no changes to how CocoaPods or Carthage pulls in NXO. Additionally this change will happen under the hood and will not be visible to merchants using SPM, so not merchant changes are needed.

Once this PR is approved, I will open a PR to update our internal release steps for NXO.

This PR also updates to use SPM to pull in NXO for the Braintree Native Checkout targets as well as the Demo app. While this adds additional steps to updating the Native Checkout version, it allows to avoid keeping a local copy of the framework in the SDK. This reduces the overall size and time taken to pull in the SDK via SPM due to SPM doing clone mirror under the hood.

### Testing
- A sample app external to the repo pointing to this branch has been tested to ensure we are not pulling in NXO for merchants only using BraintreeCard
- The demo app has been uploaded to App Store Connect to ensure there are no issues with App Store uploads caused by this change

#### Note
- This does not affect merchant integrations in any way. This only affects internal build & development processes.
- We will potentially need to look into how to solve this for all dependencies in a future PR if we want them all to handle pulling in dependencies in the same manner. This PR only solves the issue for having the large local binary for NXO in the repo.

### Checklist

- [x] Added a changelog entry

### Authors
- @jaxdesmarais


scannillo:
👏 👏 👏 to how incredibly thorough this PR and description is, & for documenting the testing steps you took!

I took a stab at using CocoaPods to pull in PayPalCheckout, in attempt to consolidate us to using just 1 package manager for all our xcworkspace - but it ultimately didn't end up working for Carthage (see #883 if curious).

I think going forward with your approach makes sense. In a follow up PR, we can consider moving all dependencies to SPM (InAppSettingsKit, Expect, etc). _We could even mess around with moving them all to Carthage too, if the DerivedData storing of SPM packages bothers anyone else 😆._

Changes look good to me! Only thing is I think we can update the [README](https://github.com/braintree/braintree_ios#demo) instructions for running the Demo to include that you have resolve the SPM packages, in addition to `pod install`. I usually do this with `File > Packages > Resolve Package Versions`, because `swift package resolve` on the command line doesn't always work for me.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPalNativeCheckout (BETA)
+  * Update Package.swift to use binary dependency of `PayPalCheckout` and pulling in the `PayPalCheckout.xcframework.zip` from the `PayPalCheckout` release directly
```
scannillo, line 5: Take it or leaving wording nitpick - 
```suggestion
  * Update Package.swift to fetch `PayPalCheckout` binary dependency directly instead of hosting copy in `braintree_ios` repo
```

jaxdesmarais:
> 👏 👏 👏 to how incredibly thorough this PR and description is, & for documenting the testing steps you took!
> 
> I took a stab at using CocoaPods to pull in PayPalCheckout, in attempt to consolidate us to using just 1 package manager for all our xcworkspace - but it ultimately didn't end up working for Carthage (see #883 if curious).
> 
> I think going forward with your approach makes sense. In a follow up PR, we can consider moving all dependencies to SPM (InAppSettingsKit, Expect, etc). _We could even mess around with moving them all to Carthage too, if the DerivedData storing of SPM packages bothers anyone else 😆._
> 
> Changes look good to me! Only thing is I think we can update the [README](https://github.com/braintree/braintree_ios#demo) instructions for running the Demo to include that you have resolve the SPM packages, in addition to `pod install`. I usually do this with `File > Packages > Resolve Package Versions`, because `swift package resolve` on the command line doesn't always work for me.

Good call! I will update the README as well with this callout. Yeah would be interesting to explore if moving it all to Carthage eventually would work since SPM still has a mind of it's own sometimes (though I wonder if that will continue to improve).

scannillo (approved):
🎸

jcnoriega (approved):

PR #883, opened 2022-10-20
[DO NOT REVIEW] Pull in NXO via CocoaPods for xcworkspace

scannillo:
### Summary of Changes

This is a proposal to https://github.com/braintree/braintree_ios/pull/882

- Instead of using both SPM and CocoaPods to pull dependencies into our xcworkspace (SPM for NXO & CocoaPods for InAppSettingsKit, OCMock, xcbeautify, etc), this PR proposes to use _just_ CocoaPods for _all_ dependencies.
- A few reasons I think this could be useful:
   - Let's us stick to [1 instruction step](https://github.com/braintree/braintree_ios#demo) for running the Demo.
   - SPM is annoying in that it stores local package references in Derived Data. We need to nuke Derived Data often due to Xcode weirdness, and it's always a pain to need to wait for re-fetch NXO every time.
   - I know we had some SPM timeout issues with fetching NXO into our project on CI. This fetch via CocoaPods instead could alleviate some of that.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


scannillo:
I tried this in a few different ways, and kept running into an issue on `carthage update --use-xcframeworks` from the merchant perspective.
 
Building `BraintreePayPalNativeCheckout` would fail with: 
```
Build description path: /Users/scannillo/Library/Caches/org.carthage.CarthageKit/DerivedData/14.0_14A309/braintree_ios/697d0ed003c0884aa16f8442d349a4b055b9bd66/Build/Intermediates.noindex/ArchiveIntermediates/BraintreePayPalNativeCheckout/IntermediateBuildFilesPath/XCBuildData/1785f5201774349ea5dbdaa135c9e4bd-desc.xcbuild

note: Building targets in dependency order
error: There is no XCFramework found at '/Users/scannillo/bt/braintree_ios/SampleApps/CarthageTest/Carthage/Checkouts/braintree_ios/Pods/PayPalCheckout/PayPalCheckout.xcframework'. (in target 'BraintreePayPalNativeCheckout' from project 'Braintree')
** ARCHIVE FAILED **
```

Our `BraintreePayPalNativeCheckout` target was referencing the dependency of `PayPalCheckout` which was pulled in via CocoaPods. But when building for Carthage, Carthage can't find the `PayPalCheckout` binary in the `Pods` directory, which makes sense. 

What's weird is - I don't really know why this works when pulling in `PayPalCheckout` with SPM instead of CocoaPods in #882. I assume the path that SPM points to when it stores the `PayPalCheckout` dep is the same as if we were to have included the framework manually. We probably _could_ get this working with CocoaPods, but it would be wonky/hacky to manipulate the `Pods/` path just for Carthage builds 😷.

PR #887, opened 2022-10-31
Convert BTAnalyticsService, BTAnalyticsEvent, and BTAnalyticsSession to Swift

jaxdesmarais:
### Summary of changes

- Convert BTAnalyticsService, BTAnalyticsEvent, and BTAnalyticsSession to Swift
- Remove APIClient and Analytics Obj-C files
- Update imports (since a lot of things imported BTAPIClient.h and BTAPIClient_Internal.h we needed to updated some forward declarations in some of the Obj-C files
- In addition to tests passing, I also tested several flows in the demo app to ensure all analytics events are still being sent as expected - these tests can be found in Splunk and use the sdkVersion "666" 👿  so we can differentiate from our other events
- Since this PR also includes the deletion of BTAPIClient in Obj-C all payment flows in the Demo app have also been tested

### Checklist

- ~[ ] Added a changelog entry~ there are no public facing changes in this PR as Analytics were all internal and remain internal

- @jaxdesmarais


Sources/BraintreeCoreSwift/BTAnalyticsService.swift
```diff
@@ -0,0 +1,174 @@
...
+                    self.http = BTHTTP(url: analyticsURL, tokenizationKey: tokenizationKey)
+                }else {
+                        completion(BTAnalyticsServiceError.invalidAPIClient)
+                    return
+                }
+            }
+
+            // A special value passed in by unit tests to prevent BTHTTP from actually posting
+            if self.http?.baseURL.absoluteString == "test://do-not-send.url" {
+                    completion(nil)
```
scannillo, line 78:
```suggestion
                completion(nil)
```

Sources/BraintreeCoreSwift/BTAnalyticsService.swift
```diff
@@ -0,0 +1,174 @@
...
+                return
+            }
+
+            self.sessionsQueue.async {
+                if self.analyticsSessions.count == 0 {
+                        completion(nil)
```
scannillo, line 84:
```suggestion
                    completion(nil)
```

Sources/BraintreeCoreSwift/BTAnalyticsService.swift
```diff
@@ -0,0 +1,174 @@
...
+                            completion(error)
+                        }
+                    }
+                }
+
+                if !willPostAnalyticsEvent {
+                    completion(nil)
+                }
+
+                    completion(nil)
```
scannillo, line 130:
```suggestion
                completion(nil)
```

Sources/BraintreeCard/Public/BraintreeCard/BTCardNonce.h
```diff
@@ -6,6 +6,7 @@
 
 @class BTThreeDSecureInfo;
 @class BTAuthenticationInsight;
+@class BTBinData;
```
scannillo, line 9: I see a handful of files have new added `@class` directives (other examples in the PR: BTConfiguration and BTPostalAddress). In this example of the `BTCardNonce.h`, it doesn't look like any changes to `BTCardNonce.m` were made. What were these new directives added for?

Sources/BraintreeCoreSwift/BTAPIClient.swift
```diff
@@ -22,6 +26,14 @@ import Foundation
     var apiHTTP: BTAPIHTTP?
     var graphQLHTTP: BTGraphQLHTTP?
 
+    /// Exposed for testing analytics
+    /// By default, the `BTAnalyticsService` instance is static/shared so that only one queue of events exists.
+    /// The "singleton" is managed here because the analytics service depends on `BTAPIClient`.
+    var analyticsService: BTAnalyticsService? {
```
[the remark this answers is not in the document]
jaxdesmarais: Good call - yeah it was present in Obj-C as well. My initial thought would be to make this one weak but I feel like from an ownership perspective it could potentially go either way. Thoughts?

Sources/BraintreeCoreSwift/BTAnalyticsSession.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+/// Encapsulates analytics events for a given session
+struct BTAnalyticsSession {
+
+    let sessionID: String
+    let source: String
+    let integration: String
+
+    var events: [BTAnalyticsEvent] = []
```
scannillo, line 10: Nitpick comment - but I was looking at Northstar and we don't really have consistency around when new structs (data model types) get new files, vs get added to existing files.

I feel like a good rule we can follow is - if the type is only used in one other model, it can live in the same file (so in this case, moving `BTAnalyticsEvent` definition to live in this file). If the type is used within various models, it should live in it's own file.

What are your thoughts on this? We can come up with a standard and include it in our Swift Style Guide.

Sources/BraintreeCard/Public/BraintreeCard/BTCardNonce.h
jaxdesmarais: I think some of these can be cleaned up. I'll go through and revert any ones that aren't needed. 👍

jaxdesmarais: Updated - the ones remaining are needed because of deleting the references to the apiclient.h files which handled some of the forward declarations and imports for other files previously

Sources/BraintreeCoreSwift/BTAnalyticsSession.swift
jaxdesmarais: Yeah, I don't feel strongly one way or another. Maybe it makes sense to have a line length requirement where we put them in the same file vs not? I feel like if a model is only used in once place but is like 100 lines long it could make more sense to separate out into its own file, but also that get's into a grey area.

I feel like it is nice when looking at something like BTAnalyticsSession to not need to scroll past ~30 lines of model structs and instead can just hop right into the implementation, but also can see that having a dedicated file for 2 pretty small structs makes the overall project feel busier. Maybe it makes sense to have both `BTAnalyticsEvent` and `BTAnalyticsSession` in one file called `BTAnalyticsSession+Models` (not sure if that's a great idea just thinking out loud).

I can go both ways on it, so whatever we feel is best we should add to the style guide and implement here! It would be great to have this consistent across both SDKs. 😄

Sources/BraintreeCoreSwift/BTAnalyticsService.swift
```diff
@@ -0,0 +1,174 @@
...
+
+                    postParameters["_meta"] = metadataParameters
+
+                    if let authorizationFingerprint = self.apiClient.clientToken?.authorizationFingerprint {
+                        postParameters["authorization_fingerprint"] = authorizationFingerprint
+                    } else if let tokenizationKey = self.apiClient.tokenizationKey {
+                        postParameters["tokenization_key"] = tokenizationKey
+                    }
+
+                    session?.events.removeAll()
```
mattwylder, line 117: Could it make sense to factor this out into a method called something like `createAnalyticsEvent`?

Sources/BraintreeCoreSwift/BTAnalyticsService.swift
```diff
@@ -0,0 +1,174 @@
...
+        sessionsQueue.sync {
+            analyticsSessions.values.forEach { analyticsSession in
+                eventCount += analyticsSession.events.count
+            }
+        }
+
+        if eventCount >= flushThreshold {
+            flush()
+        }
+    }
```
mattwylder, line 167: This function name doesn't fully express that it calls `flush()`. What about `flushIfAtThreshold`?

mattwylder (commented):
All my comments are nitpicks. Mostly just readability-related suggestions.

Sources/BraintreeCoreSwift/BTAPIClientHTTPType.swift
```diff
@@ -1,8 +1,7 @@
 import Foundation
 
 /// :nodoc:
-// TODO: rename once Objective-C files are deleted
-@objc public enum BTAPIClientHTTPTypeSwift: Int {
+@objc public enum BTAPIClientHTTPType: Int {
```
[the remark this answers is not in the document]
jaxdesmarais: I think this makes a lot of sense - updating

Sources/BraintreeCoreSwift/BTAnalyticsEvent.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+/// Encapsulates a single analytics event
+struct BTAnalyticsEvent {
+    var kind: String
+    var timestamp: Double
```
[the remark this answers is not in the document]
jaxdesmarais: We could likely update this in next major version/this PR if we wanted them to be consistent. I am not sure if there are any implications to making this update now that I may not be thinking of. Curious if @scannillo knows of any reasons we may want to wait to change this that I am not thinking of?

Sources/BraintreeCoreSwift/BTAnalyticsEvent.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+/// Encapsulates a single analytics event
+struct BTAnalyticsEvent {
+    var kind: String
```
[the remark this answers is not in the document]
jaxdesmarais: It's all used internally. I am not sure if there is a historic reason time was chosen over event name but I think we can update this for clarity and consistency across the SDKs.

Sources/BraintreeCoreSwift/BTAnalyticsService.swift
```diff
@@ -0,0 +1,174 @@
...
+                    completion(BTAnalyticsServiceError.missingAnalyticsURL)
+                return
+            }
+
+            if self.http == nil {
+                if let clientToken = self.apiClient.clientToken {
+                    self.http = BTHTTP(url: analyticsURL, authorizationFingerprint: clientToken.authorizationFingerprint)
+                } else if let tokenizationKey = self.apiClient.tokenizationKey {
+                    self.http = BTHTTP(url: analyticsURL, tokenizationKey: tokenizationKey)
+                }else {
```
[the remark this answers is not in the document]
jaxdesmarais: Was already updated in this commit: https://github.com/braintree/braintree_ios/pull/887/commits/948cac8fb9ef484990f15b7c72a5b43d40f8e7eb

Sources/BraintreeCoreSwift/BTAnalyticsEvent.swift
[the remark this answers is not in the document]
mattwylder: I don't think there's a rush to change - I just wanted to make note.

mattwylder:
Do we handle flushing analytics stored up events in cases like `applicationWillTerminate()`?

Sources/BraintreeCoreSwift/BTAnalyticsEvent.swift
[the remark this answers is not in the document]
jaxdesmarais: It's up to us! I don't know of a compelling reason _not_ to change it now if changing it will allow us to be better aligned across the SDKs unless I am missing some context for why it needed to be a Double initially.

jaxdesmarais:
> Do we handle flushing analytics stored up events in cases like `applicationWillTerminate()`?

That's a good question! I don't believe we do (and were not previously either) 🤔

Sources/BraintreeCoreSwift/BTAnalyticsEvent.swift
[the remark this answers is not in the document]
scannillo: It looks like [we were using `Uint64`](https://github.com/braintree/braintree_ios/blob/8964e496f38487439a70dc1d7fd0c1e9a91508f8/Sources/BraintreeCore/BTAnalyticsService.m#L246) in Objective-C and we're using a `long` in Java. We should hold off on changing this to a string, until we're ready for FPTI migration. We want to keep consistent with other platform analytics using Arachne.

Also, I think this might have been used for easy conversion to Redshift's DATE or TIMESTAMP type.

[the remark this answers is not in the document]
scannillo: Epoch timestamps will only be in whole numbers, so we don't need to use a floating-point type (ex: Double or Float). `UInt64` I think will be our best bet. [This site](https://www.tutorialspoint.com/swift/swift_data_types.htm) does a good job of explaining Swift's numeric types.

scannillo:
> Do we handle flushing analytics stored up events in cases like applicationWillTerminate()?

Good question! We don't do this since it would require a hook into a merchant's App/SceneDelegate.

Sources/BraintreeCoreSwift/BTAnalyticsSession.swift
scannillo: Those are good points!

At this point, I think I would prefer slightly longer model files over more noise/files to sift through in the Xcode project. We could always change our preference if we see a pattern of model files getting huge, but I imagine it would be unlikely.

@mattwylder - any preference?

Sources/BraintreeCoreSwift/BTAPIClient.swift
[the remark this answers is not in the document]
scannillo: I agree - a weak reference to `BTAnalyticsService` makes sense to me.

Sources/BraintreeCoreSwift/BTAPIClient.swift
```diff
@@ -22,6 +26,14 @@ import Foundation
...
+        get { BTAPIClient._analyticsService }
+        set { BTAPIClient._analyticsService = newValue }
```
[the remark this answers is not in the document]
scannillo: Got it - thanks for explaining! Off the top of my head don't have any other ideas, so down for leaving as-is.

Sources/BraintreeCoreSwift/BTAnalyticsEvent.swift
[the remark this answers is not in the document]
scannillo: It looks like doing `Date().timeIntervalSince1970 * 1000` without any cast, defaults to `Double` and spits out something like `1667488950096.996` 🤔. I thought that Epoch times in seconds are 10-digit INTs, milliseconds are 13-digit INTs and microseconds are 16-digit INTs.

I double checked, and on master, we have something like `1667489463800` (10 digit INT), without decimals.

I think doing something like `UInt64(Date().timeIntervalSince1970 * 1000)`, would match with our Obj-C implementation.

Sources/BraintreeCoreSwift/BTAPIClient.swift
```diff
@@ -22,6 +26,14 @@ import Foundation
...
+    }
+
```
scannillo, line 36: _Take it or leave it comment_ - moving the private analtyicsService property to be right below the internal one for testing.

```suggestion
    }

    private static var _analyticsService: BTAnalyticsService?
```

Sources/BraintreeCoreSwift/BTAnalyticsSession.swift
jaxdesmarais: Updated this in this commit: https://github.com/braintree/braintree_ios/pull/887/commits/a7e45b96a4083fe3a114b266fb812d4044ac7b99

Sources/BraintreeCoreSwift/BTAPIClient.swift
```diff
@@ -264,7 +285,7 @@ import Foundation
 
     /// :nodoc:
     @objc(POST:parameters:httpType:completion:)
-    public func post(_ path: String, parameters: [String: Any]? = nil, httpType: BTAPIClientHTTPTypeSwift, completion: @escaping RequestCompletion) {
+    public func post(_ path: String, parameters: [String: Any]? = nil, httpType: BTAPIClientHTTPService, completion: @escaping RequestCompletion) {
```
scannillo, line 288: Curious why the move to `String` here? What if our `POST` params include some bools or ints?

Sources/BraintreeCoreSwift/BTAnalyticsService.swift
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+
+    func flush(_ completion: @escaping (Error?) -> Void = { _ in }) {
+        apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
+            guard let configuration, error == nil else {
+                if let error {
+                    completion(error)
+                }
+                return
```
scannillo, line 89: If both `configuration` and `error` are nil, the completion wouldn't be called. I don't think we have that case in `APIClient.fetchOrReturnRemoteConfiguration()` though. Is this something we want to worry about/consider?

Sources/BraintreeCoreSwift/BTAnalyticsService.swift
```diff
@@ -0,0 +1,175 @@
...
+                } else if let tokenizationKey = self.apiClient.tokenizationKey {
+                    self.http = BTHTTP(url: analyticsURL, tokenizationKey: tokenizationKey)
+                } else {
+                    completion(BTAnalyticsServiceError.invalidAPIClient)
+                    return
+                }
+            }
+
+            // A special value passed in by unit tests to prevent BTHTTP from actually posting
+            if self.http?.baseURL.absoluteString == "test://do-not-send.url" {
```
scannillo, line 109: _Nitpick_ - instead of `if self.http == nil` on line 65 we could do a guard let, so that we can remove the optional chaining here.

Sources/BraintreeCoreSwift/BTAnalyticsService.swift
```diff
@@ -0,0 +1,175 @@
...
+                            completion(error)
+                        }
+                    }
+                }
+
+                if !willPostAnalyticsEvent {
+                    completion(nil)
+                }
+
+                completion(nil)
```
scannillo, line 135: This check isn't necessary if we'll just call the `completion(nil)` even if the check fails.

scannillo: I see in ObjC though we don't call the completion at all unless `!willPostAnalyticsEvent` passes - which is a little odd.

UnitTests/BraintreeCoreTests/BTAnalyticsService_Tests.swift
```diff
@@ -218,14 +218,14 @@ final class BTAnalyticsService_Tests: XCTestCase {
             XCTAssertEqual(mockAnalyticsHTTP.POSTRequestCount, 1)
 
             let parameters = mockAnalyticsHTTP.lastRequestParameters?["analytics"] as? [[String: Any]]
-            let timestampOne = parameters?[0]["timestamp"] as! Int
-            let timestampTwo = parameters?[1]["timestamp"] as! Int
+            let timestampOne = Int(parameters?[0]["timestamp"] as! Double)
+            let timestampTwo = Int(parameters?[1]["timestamp"] as! Double)
```
scannillo: We can revert the need to double cast here, once we update the source to use Int (UInt64) instead of Double.

Sources/BraintreeCoreSwift/BTAnalyticsEvent.swift
```diff
@@ -0,0 +1,20 @@
...
+    var timestamp: Double
+
+    var description: String {
+        "\(eventName) at \(timestamp)"
+    }
+
+    /// Event serialized to JSON
+    var json: [String: Any] {
+        [
+            "eventName": eventName,
```
scannillo: I notice this changed from `kind` to `eventName`. I think we want to keep parity with existing Arachne implementations. (Until we're ready to go full FPTI).

UnitTests/BraintreeCoreTests/BTAnalyticsService_Tests.swift
jaxdesmarais: Was already updated: https://github.com/braintree/braintree_ios/pull/887/commits/bc9cead62a7f3a9bca5bb74387f0501091c9971c

Sources/BraintreeCoreSwift/BTAnalyticsEvent.swift
scannillo: Some background here - we'll be forwarding Arachne events to FPTI at some point, and the less ways we have of doing something on Arachne, the less logic we'll need to implement later to make sure we're handling all the varied structure of analytics to FPTI.

jaxdesmarais: Ah, you're right. I did not resize this actually maps to the event: `"analytics":[{"kind":"ios.started.client-key","timestamp":1667492509855}]}}`. What do we think of keeping the eventName for the internal properties and reverting the JSON back to kind? So:
```
var json: String: Any {
    [
        "kind": eventName,
       "timestamp": timestamp
    ]
}
```

scannillo:
One thing I was looking for in the `next-major-version` branch was the Analytics testing granularity which used to be housed in[ `BTAnalyticsMetadataSpec.m`](https://github.com/braintree/braintree_ios/blob/master/UnitTests/BraintreeCoreTests/BTAnalyticsMetadataSpec.m). Here we assert (mostly using Regex and some OCMock) that analytics metadata values match some expected value.

It looks like [`BTAnalyticsMetadata_Tests.swift`](https://github.com/braintree/braintree_ios/blob/next-major-version/UnitTests/BraintreeCoreTests/BTAnalyticsMetadata_Tests.swift) (PR #879) is the replacement, but is mostly non-null checks.

I was implementing [the equivalent testing on Northstar](https://github.com/paypal/iOS-SDK/commit/e20f10ec6dfd5bb8bbe8496687f602493eb1415b#diff-38e2a44ec06e9f7541fb3223e5942469055f229451d71cc1203c6fd6b0c96f9a), when I noticed. _(The regex ones are easy to port over, but the OCMock one's will be a pain)_. Is this something on our radar to go back to? If not - let's add a ticket?

Sources/BraintreeCoreSwift/BTAPIClient.swift
jaxdesmarais: All of our post dictionaries currently are [String: Any] types otherwise we would be getting errors during build since all Objective-C post requests are now using this Swift class. It seems like we could maybe cross that if we get to a point where a dictionary deviates from the currently established pattern but would likely be a larger discussion with gateway teams as this seems like the pattern based on all of their parameter dictionaries currently, unless I am missing a use case. Thoughts?

Sources/BraintreeCoreSwift/BTAnalyticsService.swift
jaxdesmarais: We can remove the `error == nil` portion and create an error for if error is nil if we want to guard against this case. I don't _think_ it's likely but better safe than sorry. Thoughts?

jaxdesmarais: Actually, running the tests and looking through the rest of the code it doesn't look like we would ever run up against this case as in obj-c we always do when calling `fetchOrReturnRemoteConfiguration`:
```
        if (error) {
            completionBlock(nil, error);
            return;
        }
```
So we should always have either an error or a config (and never a case where both are nil) so I think we are good here.

jaxdesmarais:
> One thing I was looking for in the `next-major-version` branch was the Analytics testing granularity which used to be housed in[ `BTAnalyticsMetadataSpec.m`](https://github.com/braintree/braintree_ios/blob/master/UnitTests/BraintreeCoreTests/BTAnalyticsMetadataSpec.m). Here we assert (mostly using Regex and some OCMock) that analytics metadata values match some expected value.
> 
> It looks like [`BTAnalyticsMetadata_Tests.swift`](https://github.com/braintree/braintree_ios/blob/next-major-version/UnitTests/BraintreeCoreTests/BTAnalyticsMetadata_Tests.swift) (PR #879) is the replacement, but is mostly non-null checks.
> 
> I was implementing [the equivalent testing on Northstar](https://github.com/paypal/iOS-SDK/commit/e20f10ec6dfd5bb8bbe8496687f602493eb1415b#diff-38e2a44ec06e9f7541fb3223e5942469055f229451d71cc1203c6fd6b0c96f9a), when I noticed. _(The regex ones are easy to port over, but the OCMock one's will be a pain)_. Is this something on our radar to go back to? If not - let's add a ticket?

Ticket is: DTBTSDK-2062 @scannillo

Sources/BraintreeCoreSwift/BTAPIClient.swift
scannillo: Ah, I meant to comment this at the `[String: Any]` --> `[String: String]` change above in line `273` for our `GET`.

Sources/BraintreeCoreSwift/BTAnalyticsService.swift
scannillo: Hm - weird, maybe the tests are timing out and false-positive passing? I would need to look at which specific test you mean. 

But what I meant was if we always want to call the completion(nil), we can just remove the 
```
                if !willPostAnalyticsEvent {
                    completion(nil)
                }
```

entirely.

jaxdesmarais: Yep, I knew what you meant, was just kind of a odd discovery. The tests aren't false positive passing, I checked that as well, but it seems like there is some other _weird magic_ (probably not actually magic just weirdness)

Sources/BraintreeCoreSwift/BTAPIClient.swift
jaxdesmarais: Ah! I was working under the assumption that it was supposed to be unless our mock override is wrong in the implementation on main: https://github.com/braintree/braintree_ios/blob/master/UnitTests/BraintreeTestShared/MockAPIClient.swift#L26. I updated it to base it off of the mock implementation. If we think the mock has been wrong all along I can revert to `[String: Any]`

Sources/BraintreeCoreSwift/BTAnalyticsEvent.swift
jaxdesmarais: This was updated with the above in: https://github.com/braintree/braintree_ios/pull/887/commits/d78d7b0e9ef97f846c0f6f36aa52447615c24615

mattwylder (approved):

scannillo (approved):
Thanks for the thorough responses to my questions, as always! 🎸

PR #888, opened 2022-11-01
Remove unsupported ID Token reference

scannillo:
### Summary of changes

- This docstring was missed when removing support for PayPal UATs/ID Tokens

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


jaxdesmarais (approved):
👻

PR #890, opened 2022-11-04
Convert BTPreferredPaymentMethods, BTPreferredPaymentMethodsResult, BTCardNetwork and BTViewControllerPresentingDelegate

jaxdesmarais:
This is it, we have arrived at the final conversion files. From here, it's deleting the Objective-C module, cleanup, and general tidying but _this is it for implementation_. 🥳 

### Summary of changes

- Convert `BTPreferredPaymentMethods`, `BTPreferredPaymentMethodsResult`, `BTCardNetwork` and `BTViewControllerPresentingDelegate` to Swift
- Delete the Obj-C files
- Delete `UnitTests/BraintreeTestShared/BraintreeTestShared.h` that's no longer needed
- Update imports based on the above changes in a few test files

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais


Sources/BraintreeCard/BTCardNonce.m
```diff
@@ -74,7 +74,7 @@ + (NSString *)typeStringFromCardNetwork:(BTCardNetwork)cardNetwork {
             return @"Solo";
         case BTCardNetworkSwitch:
             return @"Switch";
-        case BTCardNetworkUKMaestro:
+        case BTCardNetworkUkMaestro:
```
jaxdesmarais, line 77: When converting the `BTCardNetwork` enum to Swift there is not a way to give the Obj-C case a different name than the Swift case. Due to how Obj-C maps enum names to Swift this worked and gave us `.ukMaestro` in Swift. there is not an equivalent to this mapping in Swift so this case name changed and was added to the CHANGELOG.

jcnoriega (approved):

Sources/BraintreeCoreSwift/BTPreferredPaymentMethodResult.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+///  Contains information about which payment methods are preferred on the device.
+///  This class is currently in beta and may change in future releases.
+@objcMembers public class BTPreferredPaymentMethodsResult: NSObject {
```
scannillo, line 5:
```suggestion
///  :nodoc: Contains information about which payment methods are preferred on the device.
///  This class is currently in beta and may change in future releases.
@objcMembers public class BTPreferredPaymentMethodsResult: NSObject {
```

Sources/BraintreeCoreSwift/BTPreferredPaymentMethods.swift
```diff
@@ -0,0 +1,76 @@
...
+    var application: AnyObject
+
+    private let apiClient: BTAPIClient
+
+    // MARK: - Initializer
+
+    /// Creates an instance of BTPreferredPaymentMethods.
+    /// - Parameter apiClient: An API client
+    @objc(initWithAPIClient:)
+    public init(apiClient: BTAPIClient) {
```
scannillo, line 20: We should probably add `nodoc` to all of the funcs on this class to be safe

scannillo (approved):
This LGTM! A few nitpicks re: adding `nodoc` to all functions in PreferredPaymentMethods files

Sources/BraintreeCoreSwift/BTPreferredPaymentMethods.swift
```diff
@@ -0,0 +1,76 @@
...
+                result.isPayPalPreferred = false
+                result.isVenmoPreferred = isVenmoInstalled
+
+                let errorEvent: String = "ios.preferred-payment-methods.\(error != nil ? "api-error" : "api-disabled")"
+                self.apiClient.sendAnalyticsEvent(errorEvent)
+                completion(result)
+                return
+            }
+
+            let parameters: [String: Any] = ["query": "query PreferredPaymentMethods { preferredPaymentMethods { paypalPreferred } }"]
```
scannillo, line 58: Side note - but there has been re-investment in getting this to work, on a new team called "Payment Intelligence" which is exciting! Hopefully we can expose it someday soon 🥳

Sources/BraintreeCoreSwift/BTPreferredPaymentMethods.swift
jaxdesmarais: Updated here: https://github.com/braintree/braintree_ios/pull/890/commits/57cdffb3685c054df240490b628e53b925549b72

PR #891, opened 2022-11-09
Bump Cardinal 2.2.5-5

jaxdesmarais:
### Summary of changes

- Update Cardinal `.framework` and `.xcframework` with version to 2.2.5-5
- Add demo app for 3D Secure tests to allow keychain entitlements 
  - Cardinal now requires keychain access so this forces the 3D Secure tests to run in a demo with keychain entitlements enabled
- Update `Cartfile` with new version

### Checklist

- [x] Added a changelog entry

- @scannillo @jaxdesmarais


mattwylder (approved):

sshropshire (approved):

sshropshire:
> Cardinal now requires keychain access so this forces the 3D Secure tests to run in a demo with keychain entitlements enabled

Sheesh do merchants need to add Keychain access to their app entitlements now?

scannillo (changes requested):
> Cardinal now requires keychain access so this forces the 3D Secure tests to run in a demo with keychain entitlements enabled

I'm confused by this (of course it also isn't mentioned in their [CHANGELOG](https://cardinaldocs.atlassian.net/wiki/spaces/CMSDK/pages/2002027228/Cardinal+Mobile+SDK+-+iOS)). This also sounds like a breaking change and all merchant apps would also need to enable the `keychain-access-groups` entitlement.

This feels odd. [Keychains](https://developer.apple.com/documentation/security/keychain_services/keychain_items/sharing_access_to_keychain_items_among_a_collection_of_apps) are used for sharing sensitive data (ex: card cards) b/w multiple apps in the same family. Cardinal shouldn't be requiring this 🤔.

Jax, do you have any other context? Sounds like we need to go back to Cardinal about this.

jaxdesmarais:
Closing this PR - per our meeting today cardinal is going to revert the keychain entitlement change in version 2.2.5-6. When the new version is released, we will pull in their updated SDK with this requirement removed.

PR #892, opened 2022-11-14
Rename BraintreeCoreSwift and Delete Obj-C Module

jaxdesmarais:
### Summary of changes

- Remove `BraintreeCore` Objective-C module
- Rename `BraintreeCoreSwift` module to `BraintreeCore`
- Update podspec
- Update Package.swift
- Update imports to reference renamed module
- Update Demo

**Note:** This does not fix Cocoapods and SPM imports yet - that will be tackled in a follow up PR

### Checklist

- ~[ ] Added a changelog entry~ these changes are all internal

- @jaxdesmarais


Sources/BraintreeCard/Public/BraintreeCard/BTCardNonce.h
```diff
@@ -1,9 +1,6 @@
-#if __has_include(<Braintree/BraintreeCard.h>)
-#import <Braintree/BraintreeCore.h>
-#else
-#import <BraintreeCore/BraintreeCore.h>
-#endif
+typedef NS_ENUM(NSInteger, BTCardNetwork);
```
sshropshire, line 1: random q: Was this in a different file initially?

jaxdesmarais: Yeah, this is part of core and was converted in PR: https://github.com/braintree/braintree_ios/pull/890. We need to forward declare the enum here to use it for the property further down. Previously, we were importing `Braintree.h` here which imported it as part of the `BTEnums.h` file here: https://github.com/braintree/braintree_ios/blob/master/Sources/BraintreeCore/Public/BraintreeCore/BraintreeCore.h#L16

IntegrationTests/Braintree-API-Integration-Specs/BTAPIClient_IntegrationTests.m
```diff
@@ -1,6 +1,5 @@
-#import <BraintreeCore/BraintreeCore.h>
 #import <XCTest/XCTest.h>
-@import BraintreeCoreSwift;
+@import BraintreeCore;
```
scannillo, line 2: I remember this pain in the butt formatting we had to do for importing Swift into Obj-C when we first converted PPDataCollector over. We can talk about this pairing later, but IIRC using the `@import` won't work for all package managers.

https://github.com/braintree/braintree_ios/blob/master/DEVELOPMENT.md#importing-a-swift-module-into-obj-c

Sources/BraintreeApplePay/BTApplePayCardNonce.m
```diff
@@ -1,4 +1,4 @@
-#import "BraintreeCoreSwiftImports.h"
+#import "Foundation/Foundation.h"
```
scannillo, line 1: I see a few new Foundation imports. Curious if these were moved over from `.m`s or why these were added?

Sources/BraintreeCore/Public/BraintreeCore/Braintree.h
```diff
@@ -1,8 +0,0 @@
-#ifndef Braintree_h
-#define Braintree_h
-
-/// This file added to resolve an issue with cross language module support using CocoaPods:
-/// https://github.com/CocoaPods/CocoaPods/issues/9432#issuecomment-570830515
-/// When compiling a Swift module, the `-Swift.h` header references `Braintree.h`, but requires nothing from it.
-
-#endif /* Braintree_h */
```
scannillo, line 8: 🤔 I think we might need to add this file back for CocoaPods. If we add it back, does this PR (local project) still build? Might be one less thing to hunt down when we follow up with the package managers

Demo/Application/Features/Amex/BraintreeDemoAmexViewController.m
```diff
@@ -1,6 +1,7 @@
 #import "BraintreeDemoAmexViewController.h"
 @import BraintreeAmericanExpress;
 @import BraintreeCard;
+@import BraintreeCore;
```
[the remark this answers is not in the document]
jaxdesmarais: Seems like there are some imports that were handled when importing other modules in Obj-C for us. Removing this gives us the error:
```
Receiver 'BTAPIClient' for class message is a forward declaration
```
On the following line in this file (and similarly in other files):
```
_apiClient = [[BTAPIClient alloc] initWithAuthorization:authorization];
```

IntegrationTests/Braintree-API-Integration-Specs/BTAPIClient_IntegrationTests.m
jaxdesmarais: I believe the `@import` should be okay for integration tests as the tests are passing currently. But I can update to `#import <BraintreeCore/BraintreeCore-Swift.h>` if that's our preferred way to import for these tests.

Sources/BraintreeApplePay/BTApplePayCardNonce.m
jaxdesmarais: Similarly, seems like this was handled in other files that are now removed when the Obj-C files were deleted from core, so removing this gives us unknown type errors.

Sources/BraintreeCore/Public/BraintreeCore/Braintree.h
jaxdesmarais: Yep, based on our findings yesterday we will. Adding back in this PR to keep the size of the follow up PR smaller. Also updating a number of the imports here vs in the follow up PR.

scannillo (approved):
Pulled this locally and it builds OK for me. Thanks for the thorough responses to my questions, too! Super helpful.

PR #897, opened 2022-11-29
Fix Subclassing CocoaPods Bug in Native Checkout Module

jaxdesmarais:
### Summary of changes

- Fix subclassing bug where merchants using CocoaPods were seeing the error "Cannot find interface declaration" with subclassed Obj-C class modules in Swift modules ([CocoaPods issue #11672](https://github.com/CocoaPods/CocoaPods/issues/11672))
- Rename `riskCorrelationId` to `riskCorrelationID`
- This fix was tested in a local cocoapods project pointing to the latest commit of this branch - the project now builds as expected without emitting any "Cannot find interface declaration" errors

### Checklist

- [x] Added a changelog entry

- @mattwylder @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 * BraintreePayPalNativeCheckout (BETA)
-  * Fix CocoaPods bug emitting "Cannot find interface declaration" error
+  * Fix CocoaPods bug emitting "Cannot find interface declaration" error ([CocoaPods issue #11672](https://github.com/CocoaPods/CocoaPods/issues/11672))
   * Rename `riskCorrelationId` to `riskCorrelationID`
   * Rename `nativeRequest` to `request` internally in `tokenizePayPalAccount`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 * BraintreePayPalNativeCheckout (BETA)
-  * Fix CocoaPods bug emiting "Cannot find interface declaration" error
+  * Fix CocoaPods bug emitting "Cannot find interface declaration" error
   * Rename `riskCorrelationId` to `riskCorrelationID`
 
 ## 5.16.0 (2022-10-27)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPalNativeCheckout (BETA)
+  * Fix CocoaPods bug emiting "Cannot find interface declaration" error
+  * Rename `riskCorrelationId` to `riskCorrelationID`
+
 ## 5.16.0 (2022-10-27)
 * BraintreePayPalDataCollector
   * Update PPRiskMagnes with a version of 5.4.0 with `ENABLE_BITCODE` removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * BraintreePayPalNativeCheckout (BETA)
   * Fix CocoaPods bug emitting "Cannot find interface declaration" error
   * Rename `riskCorrelationId` to `riskCorrelationID`
+  * Rename `nativeRequest` to `request` internally in `tokenizePayPalAccount`
 
 ## 5.16.0 (2022-10-27)
 * BraintreePayPalDataCollector
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
   * Fix CocoaPods bug emitting "Cannot find interface declaration" error ([CocoaPods issue #11672](https://github.com/CocoaPods/CocoaPods/issues/11672))
   * Rename `riskCorrelationId` to `riskCorrelationID`
   * Rename `nativeRequest` to `request` internally in `tokenizePayPalAccount`
+  * `tokenizePayPalAccount` now takes in a `request` of type `BTPayPalNativeRequest` instead of a `nativeRequest` of type `BTPayPalRequest`
 
 ## 5.16.0 (2022-10-27)
 * BraintreePayPalDataCollector
--- a/Demo/Application/Features/PayPal - Native Checkout/BraintreeDemoPayPalNativeCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPal - Native Checkout/BraintreeDemoPayPalNativeCheckoutViewController.swift
@@ -52,7 +52,6 @@ class BraintreeDemoPayPalNativeCheckoutViewController: BraintreeDemoPaymentButto
         sender.isEnabled = false
 
         let request = BTPayPalNativeVaultRequest()
-        request.activeWindow = self.view.window
 
         payPalNativeCheckoutClient.tokenizePayPalAccount(with: request) { nonce, error in
             sender.isEnabled = true
--- a/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
+++ b/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
@@ -24,7 +24,7 @@ import PayPalCheckout
     /// `BTPayPalNativeCheckoutAccountNonce`. On failure or user cancelation you will receive an error. If the user cancels
     /// out of the flow, the error code will equal `BTPayPalNativeError.canceled.rawValue`.
     /// - Parameters:
-    ///   - nativeRequest: Either a BTPayPalNativeCheckoutRequest or a BTPayPalNativeVaultRequest
+    ///   - request: Either a BTPayPalNativeCheckoutRequest or a BTPayPalNativeVaultRequest
     ///   - completion: The completion will be invoked exactly once: when tokenization is complete or an error occurs.
     @objc(tokenizePayPalAccountWithPayPalRequest:completion:)
     public func tokenizePayPalAccount(
--- a/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
+++ b/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
@@ -28,15 +28,10 @@ import PayPalCheckout
     ///   - completion: The completion will be invoked exactly once: when tokenization is complete or an error occurs.
     @objc(tokenizePayPalAccountWithPayPalRequest:completion:)
     public func tokenizePayPalAccount(
-        with nativeRequest: BTPayPalRequest,
+        with request: BTPayPalNativeRequest,
         completion: @escaping (BTPayPalNativeCheckoutAccountNonce?, Error?) -> Void
     ) {
         apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.started")
-        guard let request = nativeRequest as? (BTPayPalRequest & BTPayPalNativeRequest) else {
-            apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.invalid-request.failed")
-            completion(nil, BTPayPalNativeError.invalidRequest)
-            return
-        }
 
         let orderCreationClient = BTPayPalNativeOrderCreationClient(with: apiClient)
         orderCreationClient.createOrder(with: request) { [weak self] result in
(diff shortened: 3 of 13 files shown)


Demo/Application/Features/PayPal - Native Checkout/BraintreeDemoPayPalNativeCheckoutViewController.swift
```diff
@@ -52,7 +52,6 @@ class BraintreeDemoPayPalNativeCheckoutViewController: BraintreeDemoPaymentButto
         sender.isEnabled = false
 
         let request = BTPayPalNativeVaultRequest()
-        request.activeWindow = self.view.window
```
jaxdesmarais, line 55: This is not needed for the NXO flow so removed from our example

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
```diff
@@ -24,19 +24,14 @@ import PayPalCheckout
     /// `BTPayPalNativeCheckoutAccountNonce`. On failure or user cancelation you will receive an error. If the user cancels
     /// out of the flow, the error code will equal `BTPayPalNativeError.canceled.rawValue`.
     /// - Parameters:
-    ///   - nativeRequest: Either a BTPayPalNativeCheckoutRequest or a BTPayPalNativeVaultRequest
+    ///   - request: Either a BTPayPalNativeCheckoutRequest or a BTPayPalNativeVaultRequest
```
jaxdesmarais, line 27: Renamed this internal parameter and updated the doc string

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
```diff
@@ -24,19 +24,14 @@ import PayPalCheckout
...
-        with nativeRequest: BTPayPalRequest,
+        with request: BTPayPalNativeRequest,
         completion: @escaping (BTPayPalNativeCheckoutAccountNonce?, Error?) -> Void
     ) {
         apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.started")
-        guard let request = nativeRequest as? (BTPayPalRequest & BTPayPalNativeRequest) else {
-            apiClient.sendAnalyticsEvent("ios.paypal-native.tokenize.invalid-request.failed")
-            completion(nil, BTPayPalNativeError.invalidRequest)
-            return
-        }
```
jaxdesmarais, line 39: Since the `request` is always of type `BTPayPalNativeRequest` now we no longer need this guard as it would emit a warning stating that request is always of type `BTPayPalNativeRequest`

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutRequest.swift
```diff
@@ -2,12 +2,51 @@
...
+    /// Optional: Payment intent. Defaults to `.authorize`. Only applies to PayPal Checkout.
+    public var intent: BTPayPalNativeRequestIntent
+    
+    /// Used for a one-time payment.
+    ///
+    /// Amount must be greater than or equal to zero, may optionally contain exactly 2 decimal places separated by '.' and is limited to 7 digits before the decimal point.
+    public let amount: String
+    
+    /// Optional: Offers PayPal Pay Later if the customer qualifies. Defaults to false. Only available with PayPal Checkout.
+    public var offerPayLater: Bool
```
jaxdesmarais, line 36: The bool is defaulted in the init. Is this our preferred way to default values? We cannot use an optional `Bool` here since those cannot be represented in Objective-C.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeRequest.swift
```diff
@@ -6,17 +6,164 @@ import BraintreeCore
...
+
+    /// Defaults to false. Set to true to enable user editing of the shipping address.
+    /// - Note: Only applies when `shippingAddressOverride` is set.
+    public var isShippingAddressEditable: Bool
+
+    /// Optional: A valid shipping address to be displayed in the transaction flow. An error will occur if this address is not valid.
+    public var shippingAddressOverride: BTPostalAddress?
+
+    /// Optional: A risk correlation ID created with Set Transaction Context on your server.
+    public var riskCorrelationID: String?
```
jaxdesmarais, line 63: Updated this to `riskCorrelationID` vs riskCorrelationId` to match our preferred naming. This _is_ a breaking change but we are in beta currently. We can revert this and leave to be updated in next major version if preferred, but not sure what our pattern is on breaking changes in a beta. Thoughts?

sshropshire: On Android we broke stuff in beta when needed and we didn't get complaints from merchants. It's a good time to break stuff.

sshropshire (approved):

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeRequest.swift
sshropshire: Should the target branch for this PR be `next-major-version`?

jaxdesmarais: We had a merchant report this issue on master and were able to replicate when using cocoapods on master with other Obj-C modules. My question about beta was because PayPalNativeCheckout is still technically in beta/limited release. Since the bug can be fixed on master, we should fix it there then pull the changes into next major version. 

Not sure if that changes your comment about breaking changes in beta. This would be a breaking change on master to the `BraintreePayPalNativeCheckout` module which is in beta... if that makes sense.

sshropshire: Oh yeah I see now. It's probably best to defer the renaming until the beta release in that case to avoid introducing a breaking change.

scannillo: To clarify `BraintreePayPalNativeCheckout` module has already been released. We marked it as a "BETA" module in our docs and CHANGELOG.

Given that we've clearly documented it as a BETA module, I think we can make breaking changes to it (as long as they're called out in the CHANGELOG, of course).

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPalNativeCheckout (BETA)
+  * Fix CocoaPods bug emitting "Cannot find interface declaration" error
```
scannillo, line 5: I think it would be helpful for future us to describe the scenario that caused this error (the nested subclassing from ObjC: ObjC: Swift). We could add a note in the CHANGELOG or even in this PR description too just so we have a trace of it if a similar issue happens to us in the future.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
scannillo: Needs CHANGELOG entry?

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutRequest.swift
```diff
@@ -2,12 +2,51 @@
...
-    let hermesPath: String = "v1/paypal_hermes/create_payment_resource"
+    /// Order
+    case order
+}
+
+/// Options for the PayPal Checkout flow.
+@objcMembers public class BTPayPalNativeCheckoutRequest: BTPayPalNativeRequest {
+    
+    // MARK: - Public Properties
+    // next_major_version: subclass BTPayPalCheckoutRequest once BraintreePayPal is in Swift.
```
scannillo, line 25: Nitpick
```suggestion
    // NEXT_MAJOR_VERSION: subclass BTPayPalCheckoutRequest once BraintreePayPal is in Swift.
```

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeRequest.swift
```diff
@@ -6,17 +6,164 @@ import BraintreeCore
...
 
-protocol BTPayPalNativeRequest {
-    var hermesPath: String { get }
-    var paymentType: BTPayPalPaymentType { get }
+@objcMembers public class BTPayPalNativeRequest: NSObject {
 
-    func parameters(with configuration: BTConfiguration) -> [AnyHashable: Any]
-}
+    // MARK: - Public Properties
+    // next_major_version: subclass BTPayPalRequest once BraintreePayPal is in Swift.
```
scannillo, line 12: Can we make this note more clear that this class has duplicate logic of `BTPayPalRequest` and we should eventually remove the duplication?

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeRequest.swift
```diff
@@ -6,17 +6,164 @@ import BraintreeCore
...
+
+    // MARK: - Internal Properties
+
+    var hermesPath: String
+
+    var paymentType: BTPayPalPaymentType
+
+    // MARK: - Initializer
+
+    init(
```
scannillo, line 76: Needs docstring

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeVaultRequest.swift
```diff
@@ -3,42 +3,26 @@ import BraintreePayPal
 #endif
 
 /// Options for the PayPal Vault flow.
-@objc public class BTPayPalNativeVaultRequest: BTPayPalVaultRequest, BTPayPalNativeRequest {
+@objcMembers public class BTPayPalNativeVaultRequest: BTPayPalNativeRequest {
 
-    let hermesPath: String = "v1/paypal_hermes/setup_billing_agreement"
-    let paymentType: BTPayPalPaymentType = .vault
+    // MARK: - Public Properties
+    // next_major_version: subclass BTPayPalVaultRequest once BraintreePayPal is in Swift.
```
scannillo, line 9: Same comment here re: clarifying duplicate logic

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
```diff
@@ -24,19 +24,14 @@ import PayPalCheckout
     /// `BTPayPalNativeCheckoutAccountNonce`. On failure or user cancelation you will receive an error. If the user cancels
     /// out of the flow, the error code will equal `BTPayPalNativeError.canceled.rawValue`.
     /// - Parameters:
-    ///   - nativeRequest: Either a BTPayPalNativeCheckoutRequest or a BTPayPalNativeVaultRequest
+    ///   - request: Either a BTPayPalNativeCheckoutRequest or a BTPayPalNativeVaultRequest
     ///   - completion: The completion will be invoked exactly once: when tokenization is complete or an error occurs.
     @objc(tokenizePayPalAccountWithPayPalRequest:completion:)
     public func tokenizePayPalAccount(
-        with nativeRequest: BTPayPalRequest,
+        with request: BTPayPalNativeRequest,
```
scannillo, line 31: Also should call out that the underlying type changed too for this param in the CHANGELOG

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutRequest.swift
scannillo: I think this works well!

Once we're all in Swift, we should do a default `nil` audit since we'll no longer be constrained by our existing ObjC behavior.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeRequest.swift
jaxdesmarais: This init is internal. I know we've gone back and fourth on internal docstrings. Thoughts?

scannillo: Ah, yeh if it's not public I think its OK to omit!

Yeh I think in our style guide we _wanted_ to document everything including internal, but it ends up being a lot of work 😶

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPalNativeCheckout (BETA)
+  * Fix CocoaPods bug emitting "Cannot find interface declaration" error ([CocoaPods issue #11672](https://github.com/CocoaPods/CocoaPods/issues/11672))
+  * Rename `riskCorrelationId` to `riskCorrelationID`
+  * Rename `nativeRequest` to `request` internally in `tokenizePayPalAccount`
+  * `tokenizePayPalAccount` now takes in a `request` of type `BTPayPalNativeRequest` instead of a `nativeRequest` of type `BTPayPalRequest`
```
scannillo, line 8: This is awesome - thanks for making this update!

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeRequest.swift
jaxdesmarais: 🙈 agree

PR #898, opened 2022-12-01
Convert BTHTTP_SSLPinning_IntegrationTests, BTClientMetadataSpec, and BTGraphQLHTTP_Tests

jaxdesmarais:
### Summary of changes

- Convert `BTHTTP_SSLPinning_IntegrationTests` to Swift - this allowed us to make `BTHTTP` and `BTAPIHTTP` internal
  - All tests were converted 1:1
- Convert `BTClientMetadataSpec` to Swift
  - All tests were converted 1:1
  - Added helper method to String matches regex in `BreaintreeTestShared` this will allow us to use the same logic in other tests where we want to test regex matching
- Convert `BTGraphQLHTTP_Tests` to Swift - this allowed us to make `BTGraphQLHTTP` internal
  - All tests were converted 1:1
  - Helper method added to test `NSException` in Swift (doc string in `BTExceptionCatcher.m` explains further
- Update public access modifiers in `BTHTTP`, `BTAPIHTTP`, and `BTGraphQLHTTP` now that all tests are in Swift

### Checklist

- ~[ ] Added a changelog entry~ these classes were all internal in Objective-C, so there is no change in behavior from removing public access

- @jaxdesmarais


sshropshire (approved):

UnitTests/BraintreeCoreTests/BTGraphQLHTTP_Tests.swift
```diff
@@ -0,0 +1,540 @@
...
+
+        waitForExpectations(timeout: 2)
+    }
+
+    // MARK: - Unsupported requests
+
+    func testGETRequests_areUnsupported() {
+        do {
+            try BTExceptionCatcher.catchException {
+                self.http?.get("") { _, _, _ in
```
sshropshire, line 59: random q: can we remove these methods since all they do is throw?

jaxdesmarais: Do you mean the test methods or the actual method overrides in BTGraphQLHTTP?

mattwylder (approved):

PR #900, opened 2022-12-02
Remove `Bundle.main.bundleIdentifier` inspection from SEPA module

scannillo:
### Reason for changes

- We don't need to access the app's `Bundle.main.bundleIdentifier` for `ASWebAuthenticationSession` flows 🎉 

### Summary of changes

- Replace `Bundle.main.bundleIdentifier` inspection with custom SDK scheme (re-used the same scheme that we already use in BraintreePayPal and BraintreePayPalNativeCheckout)
- Added missing test for the URL and POST body of the request to `"v1/sepa_debit"`

### Questions
Jax, I know you mentioned there is difficulty testing this. I was able to go through the sandbox flow successfully. Let me know if there is anything else I should be testing.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


jaxdesmarais (approved):
This looks good. I was able to test as well and all of the flows (success and cancel) are completing as expected. 

We should replace the IBAN in in the SEPA ViewController with a valid IBAN in this PR as the one being used now is getting the error "The requested action could not be performed, semantically incorrect, or failed business validation. --- /payment_source/bank/sepa_debit/iban:INVALID_IBAN". This is expected and a limitation of our current testing story as they are invalidated after a certain number of uses.

We can use: `FR7630006000012283273522185` unless you have another one that was used for your testing.

sshropshire (approved):

scannillo:
@jaxdesmarais - at the time of testing, the IBAN already committed was working OK. We can update with the new one you provided in a follow-up!

PR #901, opened 2022-12-05
Re-Enable CocoaPods and SPM

jaxdesmarais:
### Summary of changes

- Re-Enable CocoaPods and SPM GitHub workflows - update imports so that all package managers work as expected
- Move all extensions of `BTConfiguration` into `BraintreeCore`
  - You cannot extend a Swift class in Objective-C
  - All imports were public so moving them to the Swift module that all Obj-C modules depend on allows us to keep the access of these values the same from the merchant perspective
  - Added notes to separate out each section as we convert other modules to Swift - at that time we can decide if these should remain public or move to internal
- Update Package.swift and Braintree.podspec
- Update imports in the Demo app and Tests as we received warnings for missing imports

### Checklist

- [x] Added a changelog entry

- @mattwylder @jaxdesmarais


jaxdesmarais:
Will address test failures - they are related to imports but the package managers are all happy 🥳

sshropshire (approved):
🌊

Sources/BraintreeCard/BTConfiguration+Card.m
```diff
@@ -1,29 +0,0 @@
-#import "BTConfiguration+Card.h"
-
-// Swift Module Imports
```
scannillo, line 3: What does `Swift Module Imports` mean?

scannillo: "Import all Braintree modules that are written in Swift?"

.github/workflows/build.yml
```diff
@@ -4,21 +4,21 @@ concurrency:
...
-#        with:
-#          ref: ${{ github.event.pull_request.head.ref }}
-#          repository: ${{ github.event.pull_request.head.repo.full_name }}
-#      - name: Use Xcode 14.0
-#        run: sudo xcode-select -switch /Applications/Xcode_14.0.app
-#      - name: Install CocoaPod dependencies
-#        run: pod install
-#      - name: Run pod lib lint
-#        run: pod lib lint
+  cocoapods:
```
scannillo, line 7: 👏  😍

Sources/BraintreeCard/BTConfiguration+Card.m
jaxdesmarais: `Swift Module Imports` is the header used for any of the imports of the Swift modules (Core or DataCollector) in any of the Obj-C files so we can hopefully easily sepearate where we are importing Obj-C vs Swift things in any of the Obj-C implementation files. If there is an easier way to separate these to make it clearer, happy to update them!

scannillo: Ah ok - I think what confused me is this comment looked only like it it applied to the CocoaPods line below, instead of the entire if else block.

Also I think the word "module" has several meanings. I thought first of the `module style imports` (@ syntax).

Maybe we can specify `// Import Braintree's Swift files` or something?

jaxdesmarais: For sure! We can do whatever will make it most clear where we are separating out imports for Obj-C modules/files vs Swift modules/files. We also currently have this for Obj-C imports ([example here](https://github.com/braintree/braintree_ios/blob/next-major-version/Sources/BraintreePaymentFlow/BTPaymentFlowClient.m#L4)) `// Objective-C Module Imports`. Should we update that as well?

jaxdesmarais: Matt and I were talking about this earlier. What do we think about replacing these:
```
// Objective-C Module Imports
// Swift Module Imports
```

With:
```
Option 1:
// Objective-C File Imports
// Swift File Imports

Option 2:
// MARK: - Objective-C File Imports
// MARK: - Swift File Imports

Option 3:
// Objective-C Imports for Package Managers
// Swift Imports for Package Managers
```
Thoughts on any of these or something else entirely?

scannillo: I like option #2! But I think it's also easy enough to decipher the ObjC vs Swift file imports since the swift ones all end in `-Swift.h`. But whatever you guys think is more readable works for me 👍

scannillo (approved):
👏

PR #903, opened 2022-12-06
Update SEPA Direct Debit IBAN in Demo App

jaxdesmarais:
### Summary of changes

- Update the IBAN used in our Demo app for SEPA Direct Debit - the one in there has been used too many times and is now invalid so we need to swap it out (again). We should have a better testing story for this but alas. 😢 

### Checklist

- @jaxdesmarais


scannillo (approved):
Any update on how those convos went for improving our merchant testing story for SEPA?

jaxdesmarais:
> Any update on how those convos went for improving our merchant testing story for SEPA?

No updates yet. We discussed with the PayPal team that's responsible for building it out and they are going to look into scoping out the work on their end and get back to us on if it's feasible or not. In the meantime, they wrote a script to randomly generate IBANs that they use in their tests that they offered to share with us but are still waiting to hear back on that as well. I'm planning to follow up on the script next week if we still haven't heard anything.

PR #909, opened 2022-12-14
Convert `BTPayPalLineItem` `BTPayPalCreditFinancing` `BTPayPalCreditFinancingAmount` to Swift

mattwylder:
### Summary of changes

- Converted `BTPayPalLineItem` `BTPayPalCreditFinancing` `BTPayPalCreditFinancingAmount` to Swift
- Extracted `BTConfiguration+PayPal` from `BTConfiguration+Extensions`

### Checklist

- @mattwylder @jaxdesmarais


jaxdesmarais:
We expect SPM to fail on this feature branch since `convert-paypal` is a mix of Objective-C and Swift. When we complete the conversion and open a PR to merge `convert-paypal` into `master` we will verify that SPM is passing again as expected. We took this approach to avoid commenting out a portion of the build process again during the conversion but can do that again if it's the more favorable approach vs having 1 red build step.

Sources/BraintreePayPal/BTPayPalCreditFinancing.swift
```diff
@@ -0,0 +1,39 @@
...
+    public let payerAcceptance: Bool
+    
+    /// Length of financing terms in months.
+    public let term: Int
+    
+    /// Estimated total payment amount including interest and fees the user will pay during the lifetime of the loan.
+    public let totalCost: BTPayPalCreditFinancingAmount?
+    
+    /// Estimated interest or fees amount the payer will have to pay during the lifetime of the loan.
+    public let totalInterest: BTPayPalCreditFinancingAmount?
```
mattwylder, line 21: We're not sure how to approach optionality for `monthlyPayment`, `totalCost` and `totalInterest`. [The tests](https://github.com/braintree/braintree_ios/blob/69f7caa0460d43595fc3b5c67cd7fc9fb5d73130/UnitTests/BraintreePayPalTests/BTPaymentMethodNonceParser_PayPal_Tests.swift#L69) expect the type to be optional, but asserts the value not to be `nil`. The [Android tests](https://github.com/braintree/braintree_android/blob/fca6c8ceae54d0f318bc23ad11d519c26a46946f/PayPal/src/test/java/com/braintreepayments/api/PayPalCreditFinancingUnitTest.java#L27) assert that the value should be `null` if the containing JSON object is `null`. A cursory search of the gateway repo for "monthlyPayment", "montly_payment", etc, shows test expectations that the value not be `null`. 

How do we want to handle the case where the gateway unexpectedly returns `null` for one of these properties? Crash the app or set it to `nil`?

How do others think we should handle this?

Sources/BraintreePayPal/BTPayPalCreditFinancing.swift
```diff
@@ -0,0 +1,39 @@
...
+    
+    // TODO: Make internal when all of PayPal module is converted to Swift
+    public init(
```
mattwylder, line 24: No docstring for this because it will be internal.

Sources/BraintreePayPal/BTPayPalCreditFinancingAmount.swift
```diff
@@ -0,0 +1,17 @@
...
+@objcMembers public class BTPayPalCreditFinancingAmount: NSObject {
+    
+    /// 3 letter currency code as defined by <a href="http://www.iso.org/iso/home/standards/currency_codes.htm">ISO 4217</a>.
+    public let currency: String
+    
+    /// An amount defined by <a href="http://www.iso.org/iso/home/standards/currency_codes.htm">ISO 4217</a> for the given currency.
+    public let value: String
+    
+    // TODO: Make internal when all of PayPal module is converted to Swift
+    public init(currency: String, value: String) {
```
mattwylder, line 13: No docstring for this because it will be internal.

Sources/BraintreePayPal/BTPayPalCreditFinancing.swift
scannillo: > How do we want to handle the case where the gateway unexpectedly returns null for one of these properties? Crash the app or set it to nil?

I'm a little confused by the question. Are you asking - do we need to handle the `null` case from the gateway at all, since the searching you did confirms that we'll never get a `null` value back?

> Crash the app or set it to nil?
We don't want to crash the merchant app if we unexpectedly get `null`. Our SDK shouldn't crash the app ever.

mattwylder: Yeah - I guess my question is: how essential are these to the merchant developer? My instinct is to handle a broken contract from the gateway when we parse - return the whole object as nil if essential properties are missing. The case is rare so it makes sense to me not to take the burden of optionalality off the merchant dev in these cases.

But I also agree with Jax, we should keep the current behavior but discuss a longer term strategy for situations like this once everything is in Swift.

sshropshire (approved):
🌊

Sources/BraintreePayPal/BTPayPalLineItem.swift
```diff
@@ -0,0 +1,82 @@
...
+    public let unitAmount: String
+
+    /// Item name. Maximum 127 characters.
+    public let name: String
+
+    /// Indicates whether the line item is a debit (sale) or credit (refund) to the customer.
+    public let kind: BTPayPalLineItemKind
+
+    /// Optional: Per-unit tax price of the item. Can include up to 2 decimal places. This value can't be negative or zero.
+    public let unitTaxAmount: String? = ""
```
scannillo, line 28: I think it would be a bit more readable to default these to `nil` instead of empty string. If defaulting to an empty string, and then checking against `""` later on (like line 64) there really is no point of these values being optional types.

Also if a merchant went to go fetch let's say `itemDescription`, when they haven't set it - they would get back an empty string, which I think is more confusing than just a `nil` value since the field is marked as optional.

Let me know if that makes sense / what you guys think?

mattwylder: Agreed. I'm gonna change the URL to default to nil as well.

jaxdesmarais: I think that makes a ton of sense!

scannillo (approved):
🎄

PR #910, opened 2022-12-14
Random IBAN Generation for the Demo App

jaxdesmarais:
### Summary of changes

- So we don't have to change the IBANs all the time added some helper functions to the demo app to randomly generate french IBANs
  - Resources: [Gateway helper functions for IBAN generation](https://github.braintreeps.com/braintree/gateway/pull/24498/files) and a [article about calculating values for IBAN generation](https://community.appway.com/screen/kb/article/generating-and-validating-an-iban-1683400256881)
  - We will also do the same on Android

### Checklist

- ~[ ] Added a changelog entry~ this is for the Demo app

- @jaxdesmarais


Demo/Application/Features/SEPA Direct Debit/BraintreeDemoSEPADirectDebitViewController.swift
```diff
@@ -81,5 +81,65 @@ class BraintreeDemoSEPADirectDebitViewController: BraintreeDemoBaseViewControlle
     private func generateRandomCustomerID() -> String {
         String(UUID().uuidString.replacingOccurrences(of: "-", with: "").prefix(20))
     }
+
+    // MARK: - Helper functions for IBAN generation
+
+    private func generateRandomIBAN() -> String {
+        let countryCode: String = "FR"
+        let bankCode: String = "30006"
+        let branchCode: String = "00001"
```
jaxdesmarais, line 90: There is some weird stuff that needed to happen where we need the leading 0's in some cases and anything not of a String type (such as Int) will strip these.

Demo/Application/Features/SEPA Direct Debit/BraintreeDemoSEPADirectDebitViewController.swift
```diff
@@ -81,5 +81,65 @@ class BraintreeDemoSEPADirectDebitViewController: BraintreeDemoBaseViewControlle
...
+        let checksum = 97 - calculateMod97(from: sum)
+
+        return "\(accountNumber)\(checksum)"
+    }
+
+    private func checksum(bankCode: String, branchCode: String, accountNumber: Int) -> String {
+        // 152700 is taken from the conversion table here: https://community.appway.com/screen/kb/article/generating-and-validating-an-iban-1683400256881#conversion-table
+        // and is representative of the characters "FR" with 00 being added to the end for all bban's to calculate the checksum
+        let bbanString: String = bankCode + branchCode + "\(accountNumber)" + "152700"
+        let modResult = (Decimal(string: bbanString) ?? 0) % 97
```
jaxdesmarais, line 110: This is where the decimal extension comes into play as we need to divide a 29 character digit which is not possible with Int types. It's also not possible with Decimal by default so there is an extension below that allows us to do calculations on large integers.

scannillo:
I wonder if this would make most sense to live on our sample merchant server to share the IBAN creation logic b/w iOS and Android 🤔 - or if that would be more headache?

Also I know there's been convos about increased testability for SEPA merchants in sand. Can you explain a little more what the current limitations for a merchant is?

jaxdesmarais:
> I wonder if this would make most sense to live on our sample merchant server to share the IBAN creation logic b/w iOS and Android 🤔 - or if that would be more headache?

Since merchants would need this or similar logic to test SEPA on their end, I think it makes more sense to have it live in the demo app vs the sample merchant server (since the merchant server is on GitHub Enterprise). Agree that it's a bit of duplicating the logic but it seems like this is the "testing strategy" for the foreseeable future so would be nice to at least be able to show merchants how they can test this feature in sandbox without getting the "invalid IBAN" errors.

Another potential benefit of putting this in the demo app is that we could feasibly now build out UI tests for this feature. We don't currently have these due to the IBAN expiration limitation and it being unpredictable when a given IBAN may expire as these expiration details are not something returned by the PayPal or Braintree back end.

> Also I know there's been convos about increased testability for SEPA merchants in sand. Can you explain a little more what the current limitations for a merchant is?

The current limitation is that if a merchant just grabs the static IBAN from the demo app and it's been used too many times they would get the error "invalid IBAN". The PayPal back end that ultimately backs the SEPA feature does not have a way to generate static non expiring testing values similar to the Braintree Gateway so instead rely on the same logic they use in production in sandbox. They are very early in the discussions of building out a similar system and then extending it to SEPA, but it's currently in the discussion phase and not on any teams roadmap.

From the merchant point of view, they could likely find articles ([like the one linked in the PR description](https://community.appway.com/screen/kb/article/generating-and-validating-an-iban-1683400256881)) and build out the logic themselves, but this is obviously a large lift from the merchant point of view vs having an example in our demo that merchants could feasibly use. I think also having a sample in our demo that is guaranteed to work vs an IBAN that may randomly expire in our demo app is a better merchant experience if they pull down the demo to test the feature before integrating.

scannillo:
> Agree that it's a bit of duplicating the logic but it seems like this is the "testing strategy" for the foreseeable future so would be nice to at least be able to show merchants how they can test this feature in sandbox without getting the "invalid IBAN" errors.

Got it - this makes sense to me. Maybe it would be helpful if our SDK wraps up this "sandbox IBAN generation" logic and provides merchants with a helper function to use for testing. This way they won't need to care about the logic and can just call `BTSEPAClient.generateSandboxIBAN()` or something. Thoughts?

Last question I have - so all randomly generated IBANs eventually return "invalid IBAN" from the Gateway if used too many times. This behavior seems odd to me - do you know more about what threshold / rate limit this is based off of? Because if an IBAN represents a bank account ID, you should be able to transact with it unlimited amount of times.

> since the merchant server is on GitHub Enterprise

Tangent - I'd love to move this to public at some point.

jaxdesmarais:
> Got it - this makes sense to me. Maybe it would be helpful if our SDK wraps up this "sandbox IBAN generation" logic and provides merchants with a helper function to use for testing. This way they won't need to care about the logic and can just call `BTSEPAClient.generateSandboxIBAN()` / `BTSEPATestHelper.generateIBAN()` or something. Thoughts?

I like that idea! So putting it directly in the module so that we can extend it to any tests (Integration/UI/Unit in the future) as well as make it more visible for merchants. Do you have a preference for putting it in the client directly vs a helper in the module? Not sure which would be easier for merchants to find/use.

> Last question I have - so all randomly generated IBANs eventually return "invalid IBAN" from the Gateway if used too many times. This behavior seems odd to me - do you know more about what threshold / rate limit this is based off of? Because if an IBAN represents a bank account ID, you should be able to transact with it unlimited amount of times.

With IBANs, my understanding is theres some logic on PayPal back end that invalidates IBANs that are used over and over, so it's ultimately a PayPal system that returns the error and the Gateway passes it through to us. Yes, in sandbox they would all eventually fail but randomly generating them in this manner where we create a new account number and checksum each time _should_ mean we have significantly longer to use them vs a static value. 

It's certainly possible this is a  only a sandbox feature (and wouldn't happen in prod) on the PayPal side to expire things regularly to prevent issues (maybe database related?), but I don't know enough about how that system works on the PayPal side to speak confidently about it. As far as what the threshold/rate limit is based off of, I am not totally sure. 

> Tangent - I'd love to move this to public at some point.

Agree, would be nice to write it whatever the most common BT merchant server language is potentially and allow merchants to easily discover it!

sshropshire (approved):

sshropshire:
> since the merchant server is on GitHub Enterprise

We should totally move the sample merchant server to GitHub Enterprise. We can re-evaluate if we want it to be nodejs as well. I assumed Javascript would be the most approachable but we should bring up language / architecture for debate so we can easily make updates to the sample merchant server when we need them.

scannillo:
> Do you have a preference for putting it in the client directly vs a helper in the module? Not sure which would be easier for merchants to find/use.

My first thought we be to have a separate helper class/file. But feel free to experiment!

Demo/Application/Features/SEPA Direct Debit/BraintreeDemoSEPADirectDebitViewController.swift
```diff
@@ -81,5 +81,65 @@ class BraintreeDemoSEPADirectDebitViewController: BraintreeDemoBaseViewControlle
     private func generateRandomCustomerID() -> String {
         String(UUID().uuidString.replacingOccurrences(of: "-", with: "").prefix(20))
     }
+
+    // MARK: - Helper functions for IBAN generation
+
+    private func generateRandomIBAN() -> String {
```
scannillo, line 87: Following up on our idea to have this live as public SDK helper func for merchants / something we can add in our docs to improve the testing story?

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitTestHelper.swift
```diff
@@ -0,0 +1,66 @@
+import Foundation
+
+///Provides helper functions to test the SEPA Direct Debit flow in sandbox.
```
scannillo, line 3:
```suggestion
/// Provides helper functions to test the SEPA Direct Debit flow in sandbox.
```

scannillo: Should we specify that this file + it's methos aren't under version control? Or should it adhere to SemVer?

scannillo: Similar / follow-up question - do we think this is worth calling out in the CHANGELOG?

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitTestHelper.swift
```diff
@@ -0,0 +1,66 @@
...
+@objcMembers public class BTSEPADirectDebitTestHelper: NSObject {
+    
+    /// Generates a valid 27-digit IBAN (International Bank Account Number) for testing SEPA Direct Debit flows.
+    /// - Returns: a valid IBAN
+    public static func generateValidIBAN() -> String {
+        let countryCode: String = "FR"
```
scannillo, line 9: Nitpick - would it be worth exposing the countryCode (maybe providing a default value) for merchants to set using this function? Not sure if there are implications where they support SEPA in some countries but not others.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitTestHelper.swift
jaxdesmarais: 🤔 Yeah it may be worth calling that out just incase we need to make changes that we aren't thinking of currently. Not sure what that would be but better safe than getting ourselves stuck. We can also add a changelog entry for this!

This is also currently targeting the `master` branch. We probably want this to go into `5.x` and merge that into `master` eventually correct?

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitTestHelper.swift
jaxdesmarais: I thought about that but all of the countries have different IBAN lengths/requirements/checks so we would need to build out logic for all of them: https://www.iban.com/structure.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitTestHelper.swift
scannillo: Braintree SEPA is not in beta, right? If not - then I think you're right we'll want to make the change on `5.x` and then be sure to regularly be merging `5.x` into master.

This way - we can cut a `5.x` release with this SEPA update (for example) on a stable v5 release.

jaxdesmarais: Cool - I thought it'd be easy-ish to change the base branch but looks like it's gonna be weird, so gonna close this PR and open a new one with these changes pointing to 5.x

jaxdesmarais:
Closing to make these changes off of the `5.x` branch - will open a new PR shortly

PR #912, opened 2022-12-19
Convert BTPayPalRequest, BTPayPalVaultRequest, and BTPayPalCheckoutRequest

jaxdesmarais:
### Summary of changes

- Convert BTPayPalRequest, BTPayPalVaultRequest, and BTPayPalCheckoutRequest to Swift

### Checklist

- ~[ ] Added a changelog entry~ no changes to the public interface

- @jaxdesmarais


Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalClient.h
```diff
@@ -1,11 +1,6 @@
-#if __has_include(<Braintree/BraintreePayPal.h>)
-#import <Braintree/BTPayPalCheckoutRequest.h>
-#import <Braintree/BTPayPalVaultRequest.h>
-#else
-#import <BraintreePayPal/BTPayPalCheckoutRequest.h>
-#import <BraintreePayPal/BTPayPalVaultRequest.h>
-#endif
+#import <Foundation/Foundation.h>
 
+typedef NS_ENUM(NSInteger, BTPayPalPaymentType);
```
sshropshire, line 3: 🧶 : Maybe a note mentioning this will be removed?

UnitTests/BraintreePayPalTests/BTPayPalRequest_Tests.swift
```diff
@@ -50,7 +50,7 @@ class BTPayPalRequest_Tests: XCTestCase {
 
         request.lineItems = [BTPayPalLineItem(quantity: "1", unitAmount: "1", name: "item", kind: .credit)]
 
-        let parameters = request.parameters(with: configuration)
+        let parameters = request.baseParameters(with: configuration)
```
sshropshire, line 53: 🧶 : Any reason for the method name change here?

sshropshire (approved):
🌊

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalClient.h
jaxdesmarais: This enum still exists! It's in `BTPayPalRequest.swift`. This is just us forward declaring it to use in this Obj-C file. Once this is converted to Swift, `BTPayPalPaymentType` will still exist in the same way as it does currently we just won't need to forward declare it here. Really just exists here because of how exposing Swift in Obj-C needs to be done.

UnitTests/BraintreePayPalTests/BTPayPalRequest_Tests.swift
jaxdesmarais: Mainly for clarity. We used `parameters(with: )` in `BTPayPalRequest` as well as `BTPayPalCheckoutRequest` and `BTPayPalVaultRequest`. But in `BTPayPalCheckoutRequest` and `BTPayPalVaultRequest` we were then adding to the parameters from `BTPayPalRequest`. Made more sense to clarify what each function was doing since they are just used internally and `BTPayPalRequest` is only responsible for constructing the `baseParameters` then `BTPayPalCheckoutRequest` and `BTPayPalVaultRequest` add to those parameters to create the parameters for the request.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -0,0 +1,157 @@
...
+}
+
+///  The call-to-action in the PayPal Checkout flow.
+///
+///  - Note: By default the final button will show the localized word for "Continue" and implies that the final amount billed is not yet known.
+///  Setting the BTPayPalRequest's userAction to `BTPayPalRequestUserActionCommit` changes the button text to "Pay Now", conveying to
+///  the user that billing will take place immediately.
+@objc public enum BTPayPalRequestUserAction: Int {
+    /// Default
+    case `default`
```
scannillo, line 32:
```suggestion
    case default
```

did we need this backticks?

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -0,0 +1,157 @@
...
+
+    /// Initializes a PayPal Native Checkout request
+    /// - Parameters:
+    ///   - amount: Used for a one-time payment. Amount must be greater than or equal to zero, may optionally contain exactly 2 decimal places separated by '.'
+    ///   - intent: Optional: Payment intent. Defaults to `.authorize`. Only applies to PayPal Checkout.
+    ///   and is limited to 7 digits before the decimal point.
+    ///   - offerPayLater: Optional: Offers PayPal Pay Later if the customer qualifies. Defaults to `false`. Only available with PayPal Checkout.
+    ///   - currencyCode: Optional: A three-character ISO-4217 ISO currency code to use for the transaction. Defaults to merchant currency code if not set.
+    ///   See https://developer.paypal.com/docs/api/reference/currency-codes/ for a list of supported currency codes.
+    ///   - requestBillingAgreement: Optional: If set to `true`, this enables the Checkout with Vault flow, where the customer will be prompted to consent to a billing agreement during checkout.
```
scannillo, line 103:
```suggestion
    ///   - requestBillingAgreement: Optional: If set to `true`, this enables the Checkout with Vault flow, where the customer will be prompted to consent to a billing agreement during checkout. Defaults to false.
```

Sources/BraintreePayPal/BTPayPalRequest.swift
```diff
@@ -0,0 +1,197 @@
...
+
+    /// Vault
+    case vault
+}
+
+/// Use this option to specify the PayPal page to display when a user lands on the PayPal site to complete the payment.
+@objc public enum BTPayPalRequestLandingPageType: Int {
+
+    /// Default
+    case `default` // Obj-C enums cannot be nil; this default option is used to make `landingPageType` optional for merchants
```
scannillo, line 28: Same question re: backticks

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalClient.h
scannillo: Should we add a TODO to remove this fwd declare then?

scannillo (approved):

Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalClient.h
jaxdesmarais: It will be removed when we convert `BTPayPalClient` to Swift in a follow up PR. The PayPal conversion is all going into a feature branch so will get cleaned up as we convert the remaining files.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
jaxdesmarais: Yeah we ned the backticks here if we want to use `default`. It's a [protected identifier and needs to be escaped](https://docs.swift.org/swift-book/ReferenceManual/LexicalStructure.html#ID413).

Sources/BraintreePayPal/BTPayPalRequest.swift
jaxdesmarais: See https://github.com/braintree/braintree_ios/pull/912#discussion_r1053554297

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
scannillo: Since this change is happening in the v6beta, we can make breaking changes if we want, right?

Maybe us needing to use the `` is a sign we could come up with a better name for this enum case option.

Maybe `.continue` and `.commit`, or `.continue` and `.payNow` even?

jaxdesmarais: Yeah! We can certainly make this a breaking change. Agree it's quite odd - on NXO we used `case none` for optional obj-c exposed enum defaults ([example here](https://github.paypal.com/nativexo/ios-native-checkout/blob/db68545f03ba400e23d21e4ece64b63a1ba09d92/Sources/PayPalCheckoutIO/Models/OrderApplicationContext.swift#L97)). I think that worked well and was clear. So we could do `.payNow` and `.none` as the case options here and switch the other use of default to none as well.

jaxdesmarais: Or like you said if we want continue to be more explicit `. continue` and `.payNow` works even better!

jaxdesmarais: Updated to what I think makes the most sense here with avoiding backticks but let me know your thoughts: https://github.com/braintree/braintree_ios/pull/912/commits/bd7439fa850ee1c74c1d76e84871a0d83f93d2ea

PR #914, opened 2023-01-06
Convert `BTPayPayClient` and `BTPayPalAccountNonce`to Swift

mattwylder:
### Summary of changes

- Converted `BTPayPalClient` to Swift
- Converted `BTPayPalAccountNonce` to Swift
- `BTPayPalClientToken.tokenizePayPalAccount()` takes in a request of type BTPayPalRequest & BTPayPalRequestable to enforce the usage of `BTPayPalCheckoutRequest` or `BTPayPalVaultRequest`, and not the abstract parent class `BTPayPalRequest`
- Added `.asAddress()` to `BTJSON` to centralize parsing of `BTPostalAddress` and make it available for Native Checkout
- `BTPayPalError`
    - Added more specific error cases with new descriptions
    - Implement `CustomNSError` and `LocalizedError` for cleaner usage
- Created `BTPayPalLocaleCode` enum

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais @mattwylder


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -0,0 +1,415 @@
...
+    ///
+    /// On success, you will receive an instance of `BTPayPalAccountNonce`; on failure or user cancelation you will receive an error.
+    /// If the user cancels out of the flow, the error code will be `.canceled`.
+    ///
+    /// - Parameters:
+    ///   - request: Either a `BTPayPalCheckoutRequest` or a `BTPayPalVaultRequest`
+    ///   - completion: This completion will be invoked exactly once when tokenization is complete or an error occurs.
+    @objc(tokenizePayPalAccountWithPayPalRequest:completion:)
+    public func tokenizePayPalAccount(
+        with request: BTPayPalRequest & BTPayPalRequestable,
```
mattwylder, line 69: Here we are requiring a request of type `BTPayPalRequest & BTPayPalRequestable` to enforce the usage of `BTPayPalCheckoutRequest` or `BTPayPalVaultRequest` and not the abstract `BTPayPalRequest`. Objective-C was more easily able to fudge the typing and [completed with an error](https://github.com/braintree/braintree_ios/blob/c449e0b81c78fcae47bb02cfb8f987aed2a19996/Sources/BraintreePayPal/BTPayPalClient.m#L107-L113) if the types did not match.

Sources/BraintreePayPal/BTPayPalError.swift
```diff
@@ -0,0 +1,82 @@
+import Foundation
+
+/// Error codes associated with PayPal.
+enum BTPayPalError: Error, CustomNSError, LocalizedError {
```
mattwylder, line 4: We created a number of new, specific error cases, but many of them aren't tested yet. We've created a ticket (DTBTSDK-2190) to add test coverage.

Sources/BraintreeCore/BTJSON.swift
```diff
@@ -237,4 +237,20 @@ import Foundation
 
         return result
     }
+    
+    /// The `BTJSON` as a `BTPostalAddress`
+    /// - Returns: A `BTPostalAddress` parsed from the key/value pairs inside the `BTJSON`
+    public func asAddress() -> BTPostalAddress? {
```
jaxdesmarais, line 243: We will also be able to use this in the NativeCheckout module when that is refactored to use the Swift classes, which is why we opted to move it into `BTJSON` directly. Internal ticket for this is: DTBTSDK-2131

Sources/BraintreePayPal/BTJSON+PayPal.swift
```diff
@@ -0,0 +1,36 @@
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+extension BTJSON {
+
+    func asCreditFinancingAmount() -> BTPayPalCreditFinancingAmount? {
```
sshropshire, line 7:
```suggestion
    func asPayPalCreditFinancingAmount() -> BTPayPalCreditFinancingAmount? {
```
🧶 : May be overkill, but it could be a nice hint to calling files that it's a `PayPal` module extension method.

Sources/BraintreePayPal/BTPayPalAccountNonce.swift
```diff
@@ -0,0 +1,56 @@
...
+    /// Will be provided for Vault and Checkout.
+    public let creditFinancing: BTPayPalCreditFinancing?
+    
+    init?(json: BTJSON) {
+        guard let nonce = json["nonce"].asString() else { return nil }
+        
+        let details = json["details"]
+        let payerInfo = details["payerInfo"]
+
+        self.email = payerInfo["email"].asString() ?? details["email"].asString()
```
sshropshire, line 45:
```suggestion
        email = payerInfo["email"].asString() ?? details["email"].asString()        
```
🧶 : take or leave.

sshropshire (approved):
🌊

Sources/BraintreePayPal/BTPayPalAccountNonce.swift
mattwylder: Seems like we use the explicit `self.` [in Northstar as well](https://github.com/paypal/iOS-SDK/blob/855802216f9fbddbc6709d101e8871b307999628/Sources/PaymentsCore/AnalyticsEventData.swift#L88). I personally prefer `self.` but my tastes can be a little wordy. What do others think?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -0,0 +1,415 @@
...
+
+#if canImport(BraintreeDataCollector)
+import BraintreeDataCollector
+#endif
+
+@objcMembers public class BTPayPalClient: NSObject, ASWebAuthenticationPresentationContextProviding {
+    
+    // MARK: - Internal Properties
+
+    ///Exposed for testing to get the instance of BTAPIClient
```
scannillo, line 16:
```suggestion
    /// Exposed for testing to get the instance of BTAPIClient
```

Sources/BraintreePayPal/BTPayPalAccountNonce.swift
jaxdesmarais: I personally am a fan of removing the explicit self where possible to clean things up, but we have done a mix of both throughout the Swift conversion. For example the [SEPA nonce](https://github.com/braintree/braintree_ios/blob/c449e0b81c78fcae47bb02cfb8f987aed2a19996/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitNonce.swift#L23-L25) does use explicit self while the [PayPalNativeCheckout nonce](https://github.com/braintree/braintree_ios/blob/c449e0b81c78fcae47bb02cfb8f987aed2a19996/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutAccountNonce.swift#L47-L52) does not. This is certainly something that we can solidify and implement in our style guide either was we decide.

Sources/BraintreePayPal/BTPayPalClient.swift
scannillo: It might be cleaner/easier for merchants to understand if we instead had 2 separate public method signatures. Something like: 
* `tokenizePayPalAccount(with vaultRequest: BTPayPalVaultRequest)`
* `tokenizePayPalAccount(with checkoutRequest: BTPayPalCheckoutRequest)`

(or whatever want to name them)

This way we can remove `BTPayPalRequestable`, which it seems just exists to enforce the type, rather than be usable to the merchant.

What do you think?

Sources/BraintreePayPal/BTJSON+PayPal.swift
mattwylder: Will do salute emoji

Sources/BraintreePayPal/BTPayPalAccountNonce.swift
mattwylder: It seems like generally we avoid using `self` when possible in most methods, but we tend to use it in initializers. I personally like the explicitness when initializer parameter names are the same as property names (e.g. `self.firstName = firstName`) but am open to whatever.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -0,0 +1,415 @@
...
+    @objc(initWithAPIClient:)
+    public init(apiClient: BTAPIClient) {
+        self.apiClient = apiClient
+        super.init()
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(applicationDidBecomeActive),
+            name: UIApplication.didBecomeActiveNotification,
+            object: nil
+        )
```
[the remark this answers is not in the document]
mattwylder: If we remove this notification observer, then we don't know if the user has canceled the flow through the alert or the browser cancel button. [Right now use only use that information for analytics](https://github.com/braintree/braintree_ios/blob/f5873c632428a292ce85a8ec1766b630387859ea/Sources/BraintreePayPal/BTPayPalClient.swift#L288-L295). Do we care about the difference between `authsession.browser.cancel` vs `authsession.alert.cancel`?

Sources/BraintreePayPal/BTPayPalClient.swift
mattwylder: I like this idea! Though if we want to share the code between the two we'll still need the protocol, but we can hide it from the merchant by making it private. I would also like to find a way to make `BTPayPalRequest` itself a protocol so we don't need both `BTPayPalRequest` and `BTPayPalRequestable`.

jaxdesmarais: Matt and I messed around with getting an all protocol approach for `BTPayPalRequest` to work under the current limitations of how the SDK is structured. As we move to codeable, we may have opportunities to move to more protocol oriented code as opposed to using subclassing.

Problems with protocol BTPayPalRequest (example in branch btpaypalrequest-protocol):
- Repetitive boilerplate for BTPayPalCheckoutRequest and Vault request as we cannot expose default implementations from a protocol extension in Objective-C as we can with all Swift - this limitation is a good callout if we ever need to allow Obj-C exposure on Northstar. Error with objc default protocol implementations: `Non-‘@objc’ property ‘riskCorrelationId’ does not satisfy requirement of ‘@objc’ protocol ‘BTPayPalRequest’`
- Big unwieldy initializer based on the above, we need the entirety of the init in each implementation - this makes it confusing to discern what is part of the protocol vs part of the subclass
- Cannot invoke static properties on a protocol - error `Static member 'callbackURLHostAndPath' cannot be used on protocol metatype '(any BTPayPalRequest).Type'`

Based on the above, we _did_ decide to remove the protocol `BTPayPalRequestable` all together to reduce concerns. This is reflected in commit: https://github.com/braintree/braintree_ios/pull/914/commits/a447ec02ee5f52940f2ab9c53fdd3dbf0f645c31. We decided to go with an inheritance approach because it is cleaner (less duplication of the `BTPayPalRequest` properties in the subclasses). The main downside being that we had to do some downcasting when constructing the dictionaries which we have added comments to as the alternative was to have an function that returns and empty dictionary in `BTPayPalRequest` that we would then override in the subclasses.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -0,0 +1,415 @@
...
+        }
+    }
+    
+    // MARK: - ASWebAuthenticationPresentationContextProviding protocol
+
+    @available(iOSApplicationExtension, unavailable, message: "Uses APIs (i.e UIApplication.sharedApplication) not available for use in App Extensions.")
+    public func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
+        if let activeWindow = payPalRequest?.activeWindow {
+            return activeWindow
+        }
```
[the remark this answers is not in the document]
jaxdesmarais: We have a seperate ticket for cleaning up the use of `ASWebAuthentication` and removing `activeWindow`. Ticket is: DTBTSDK-2200 for internal reference.

Sources/BraintreePayPal/BTPayPalClient.swift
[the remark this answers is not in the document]
mattwylder: I believe `UIApplication.didBecomeActiveNotification` is [posted by UIKit](https://developer.apple.com/documentation/uikit/uiapplication/1622953-didbecomeactivenotification), allowing us some hooks into the app lifecycle.

Sources/BraintreePayPal/BTPayPalLocaleCode.swift
```diff
@@ -0,0 +1,94 @@
+import Foundation
+
+///  A locale code to use for a transaction.
+@objc public enum BTPayPalLocaleCode: Int {
+    case none
```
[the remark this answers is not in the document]
jaxdesmarais: In Objective-C, you cannot represent an optional of primitive types such as enums. To expose this enum to Objective-C it must be non-optional. The `none` case is essentially the only way we can provide something similar to a `nil` default.

Sources/BraintreePayPal/BTPayPalAccountNonce.swift
jaxdesmarais: I am down with this as well as our style! Do we have a plan to "officially" move our style guide into the repo at some point?

Sources/BraintreePayPal/BTPayPalRequest.swift
```diff
@@ -157,6 +136,75 @@ import BraintreeCore
 
     // MARK: Internal Methods
 
+    func parameters(with configuration: BTConfiguration) -> [String: Any] {
+        let baseParameters: [String: Any] = baseParameters(with: configuration)
+
+        switch paymentType {
+        case .checkout:
+            // We downcast self to avoid an empty implementation in the parent class
+            // which would require us to override this function
```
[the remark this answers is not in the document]
mattwylder: We _are_ being a little naughty here. We were trying to avoid writing an "empty" implementation in the super class. We wanted to avoid writing this function:

```swift
func parameters(with configuration: BTConfiguration) -> [String: Any] {
    return [:]
}
```
and overriding it in both classes, since there is no way to _not_ implement it. Though, that solution is probably more standard.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -0,0 +1,429 @@
...
+            return approvalURL
+        }
+
+        let userActionValue = request.userAction.stringValue
+        guard userActionValue.count > 0 else {
+            return approvalURL
+        }
+        
+        let userActionQueryItem = URLQueryItem(name: "useraction", value: userActionValue)
+        var queryItems = approvalURLComponents.queryItems ?? []
```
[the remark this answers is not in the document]
jaxdesmarais: We want to set `var queryItems = approvalURLComponents.queryItems ?? []` to an empty array if it doesn't exist. If we append directly to `approvalURLComponents.queryItems` the `queryItems` are optional and it would not be appended to if it does not exist yet.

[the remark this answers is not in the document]
mattwylder: Yeah it's a little tough to explain -- our tests fail if we write `approvalURLComponents.queryItems?.append(userActionQueryItem)` because in some cases `approvalURLComponents.queryItems` is nil, so when it comes time to make the URL from the components, we have not appended anything to `queryItems`. In the current implementation we set it to an empty array to avoid that problem.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -33,6 +44,15 @@ import BraintreeCore
 
     /// Pay Now
     case payNow
```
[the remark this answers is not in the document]
jaxdesmarais: We _can_ do continue but that is a protected word so we would be back to having this with the backticks:
```
case `continue`
```
Previously we had discussed avoiding backticks and protected words, but curious what your thoughts are?

Sources/BraintreePayPal/BTPayPalRequest.swift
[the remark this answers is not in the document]
jaxdesmarais: I think to also add to this, from my perspective the above is less of a concern than having the logic all in 1 place to build the parameters dictionary. With the above (returning an empty dictionary and overriding it) we have the logic of how we construct the `parameters` living in `BTPayPalRequest` (the base dictionary), as well as `BTPayPalCheckoutRequest` and `BTPayPalVaultRequest` for appending the request specific pieces to the base parameters. This makes it harder to reason about what is being done where to construct the `parameters`.

Having the entirety of how we construct `parameters` live in one place makes it much easier to reason about how we are constructing this dictionary, imo. But of folks feel strongly about returning an empty dictionary here and overriding it in the other files we can do that instead!

Sources/BraintreePayPal/BTPayPalAccountNonce.swift
scannillo: Agreed yeh, we should move our style guide out of HackMD soon! Maybe the knowledge repo, or maybe the real repo.

Sources/BraintreePayPal/BTPayPalRequest.swift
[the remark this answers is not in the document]
scannillo: Hm, I think this violates "separation/responsibility" of concern, since BTPayPalRequest is doing more than it should be.

We could do something like [this](https://github.com/braintree/braintree_ios/compare/convert-paypal-client-accountnonce...refactor-paramaters-override). The superclass implementation is not empty, and each subclass just appends what it needs to on-top of the parent class implementation.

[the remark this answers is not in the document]
jaxdesmarais: Good point - yeah I think I was thinking of it in a weird way initially, but the diff you sent makes sense. I'll make this update and push up the changes. Thanks for talking through that 🙏 !

scannillo (approved):
👏 lots of great discussion raised here. Thanks for the thorough responses and ideas!

Sources/BraintreePayPal/BTPayPalRequest.swift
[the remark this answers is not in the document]
mattwylder: This is the best solution for sure!

PR #923, opened 2023-01-13
Remove Xcode 12 note in README

scannillo:
### Summary of changes

- Xcode 12 was deprecated in April and devs can no longer upload Apps with it.
- Since Apple doesn't support it, I don't think we need to keep this comment around for Xcode 12.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo
-


jaxdesmarais (approved):
👻

sshropshire (approved):

PR #924, opened 2023-01-17
Remove Union Pay module

scannillo:
### Summary of changes

Follow up PR to #922
- Remove UnionPay code under `Sources/`, `UnitTests/` and `IntegrationTests/`
- Remove UnionPay from Podspec and Package.swift
- Update BraintreeCard module to use non-deprecated `tokenize()` function
- Update V6_MIGRATION guide

### Testing
- Tested to confirm that a UnionPay sandbox card works fine tokenizing via BT Card. I tried getting a production card to test, but the cards team don't have one handy. They assured me that we've been processing UnionPay cards in production via tokenize() for years.

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo


Sources/BraintreeCard/Public/BraintreeCard/BTCardRequest.h
```diff
@@ -20,29 +20,6 @@ NS_ASSUME_NONNULL_BEGIN
...
-/**
- The enrollment verification code sent via SMS to the mobile phone number. The code is needed to tokenize a UnionPay card that requires enrollment.
-*/
-@property (nonatomic, copy, nullable) NSString *smsCode;
-
-/**
- The UnionPay enrollment ID
-*/
-@property (nonatomic, copy, nullable) NSString *enrollmentID;
-
```
scannillo, line 45: These need to be marked as deprecated in V5.

Sources/BraintreeCore/BTConfiguration+Extensions.swift
```diff
@@ -37,13 +37,6 @@ import PassKit
         json?["payWithVenmo"]["environment"].asString()
     }
 
-    // MARK: - BTConfiguration+UnionPay
-
-    /// Indicates whether UnionPay is enabled for the merchant account.
-    var isUnionPayEnabled: Bool {
```
jaxdesmarais, line 43: Since the properties in this extension are public should this go in the CHANGELOG?

jaxdesmarais (approved):
Goodbye forever UnionPay module! 🚀

Sources/BraintreeCore/BTConfiguration+Extensions.swift
scannillo: Good question. I was thinking since because in V5, this lived in the UnionPay module ([here](https://github.com/braintree/braintree_ios/blob/fdb5561f47d912e6629dc0ec500d4155958f84b1/Sources/BraintreeUnionPay/Public/BraintreeUnionPay/BTConfiguration%2BUnionPay.h#L13-L16)), that the CHANGELOG entry for nuking the whole module would cover it.

But you're right that our CHANGELOG is keeping track across the betas, so to follow that pattern we should probably call it out. Will update!

scannillo: Updated

mattwylder (approved):

Sources/BraintreeCard/Public/BraintreeCard/BTCardRequest.h
scannillo: Ticket for this work - DTBTSDK-2219

PR #925, opened 2023-01-17
Add Remaining PayPal Test Coverage

jaxdesmarais:
### Summary of changes

- After the conversion we had some missing test coverage in the PayPal module. This adds those tests in for the missing errors as well as cleans up some of the implementation. 

### Checklist

- @mattwylder @jaxdesmarais


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -290,7 +284,7 @@ import BraintreeDataCollector
                 let dataCollector = BTDataCollector(apiClient: self.apiClient)
                 self.clientMetadataID = self.payPalRequest?.riskCorrelationId ?? dataCollector.clientMetadataID(pairingID)
                 self.sendAnalyticsEvent(for: request.paymentType, success: error == nil)
-                self.handlePayPalRequest(with: approvalURL, error: nil, paymentType: request.paymentType, completion: completion)
+                self.handlePayPalRequest(with: approvalURL, paymentType: request.paymentType, completion: completion)
```
jaxdesmarais, line 287: We were always passing the error as nil here. Above, we cast `error` as `NSError` which [always succeeds](https://stackoverflow.com/questions/51602907/why-can-any-error-be-unconditionally-converted-to-nserror). We removed error from this function signature and return the error above (instead of `httpResponseMissingUserInfoJSON`). This matches the [Obj-C implementation more similarly](https://github.com/braintree/braintree_ios/blob/b694fc747142a465831b661c4e41732a73ae23e8/Sources/BraintreePayPal/BTPayPalClient.m#L140-L167) and the error was previously only passed here because the completion could be `nil` and it cannot in Swift. This allowed us to also remove `httpResponseMissingUserInfoJSON` and pass the error directly.

Sources/BraintreePayPal/BTPayPalError.swift
```diff
@@ -42,18 +39,16 @@ enum BTPayPalError: Error, CustomNSError, LocalizedError {
             return 1
         case .fetchConfigurationFailed:
             return 2
-        case .httpResponseMissingUserInfoJSON:
-            return 3
         case .httpPostRequestError:
-            return 4
+            return 3
```
jaxdesmarais, line 43: This only is changing in the `convert-paypal` branch. So there are no breaking changes to call out in the CHANGELOG with this change in raw value of the errors. The new errors are all listed in the CHANGELOG for `convert-paypal`.

sshropshire (approved):
🌊

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -94,13 +95,34 @@ class BTPayPalClient_Tests: XCTestCase {
         XCTAssertEqual(lastPostParameters["cancel_url"] as? String, "sdk.ios.braintree://onetouch/v1/cancel")
     }
 
-    func testTokenizePayPalAccount_whenPaymentResourceCreationFails_callsBackWithError() {
-        mockAPIClient.cannedResponseError = BTPayPalError.httpResponseMissingUserInfoJSON as NSError
+    func testTokenizePayPalAccount_whenPoseRequstContainsError_callsBackWithError() {
```
scannillo, line 98: Looks like a small typo in the test name. Is this supposed to be `whenPostRequestContainsError`?

scannillo (approved):
LGTM! Small comment on typo in test name

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
jaxdesmarais: 🤦

Issue #920, opened 2023-01-13
PPRiskMagnes has been changed back to a dynamic Framework


mrtnlst:
**Integration Details:**
 - Braintree SDK Version: 5.19.0

**Describe the bug**
In [Release 5.8.0](https://github.com/braintree/braintree_ios/releases/tag/5.8.0) you changed the PPRiskMagnes to dynamic framework: 
> This version of PPRiskMagnes replaces the dynamic framework/xcframework with a static framework/xcframework

On the most current one (5.19.0 and 6.0.0-beta) it has changed back to be a dynamically linked framework. Has this happened by accident? Is it possible that you change it back in a future release? 

**To Reproduce**
I checked the current version 5.19.0's PPRiskMagnes binary with `file` in Terminal: 
```
$ file /…/braintree_ios-5.19.0/Frameworks/XCFrameworks/PPRiskMagnes.xcframework/ios-arm64/PPRiskMagnes.framework/PPRiskMagnes
> /…/braintree_ios-5.19.0/Frameworks/XCFrameworks/PPRiskMagnes.xcframework/ios-arm64/PPRiskMagnes.framework/PPRiskMagnes: Mach-O 64-bit dynamically linked shared library arm64

```

scannillo:
👋 Hi @mrtnlst. Thank you for opening this issue. 

So a few things: in v5.8.0 we did switch to a static framework of Magnes. In [v5.16.0](https://github.com/braintree/braintree_ios/releases/tag/5.16.0), we updated Magnes back to a dynamic framework, unknowingly. This was our mistake for not documenting and something we will be sure to clearly document in the future and test prior to shipping.

In retrospect, we believe these changes should have more properly been versioned as breaking changes, since Carthage merchants who include the Magnes binary directly would need to change its embed settings.

In the meantime - we are working to get static versions of Magnes and Cardinal binaries from our partners. We will keep you posted on when and how those are available for release.

It would be helpful though if you could give more context on how your app pulls in Braintree and your use case for static v dynamic frameworks. Thanks!

mrtnlst:
@scannillo Thank you for the quick response. It does break in a weird way in a project where we include Magnes (through PayPalDataCollector) with Swift Package Manager. It does work while debugging, but crashes in the Simulator mentioning, that the dylib can't be found. 
[...]

PR #928, opened 2023-01-18
Amend CHANGELOG entry for 5.16.0

scannillo:
### Summary of changes

- This PR amends the CHANGELOG for v5.16.0 to call out that it reverts a "bug" that was introduced in v5.8.0, when we switched from dynamic to static Magnes, which is technically a breaking change.
- I called out in [GH Issue #920 ](https://github.com/braintree/braintree_ios/issues/920)that major version V6 is where we will migrate both binaries (Cardinal and Magnes) to static frameworks.

### Notes
- This go-forward plan was based on our discussions in Slack.
- Let me know what y'all think!
- If this PR gets approved, I can manually update the release notes associated with the 5.16.0 tag.

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo


jaxdesmarais (approved):
👻

sshropshire (approved):

mattwylder (approved):

PR #596, opened 2021-02-01
Fix memory leak in BTPayPalDriver

scannillo:
### Summary of changes

- This [GH issue](https://github.com/braintree/braintree_ios/issues/563) surfaced a memory leak in v4 of Braintree iOS for the PayPal flow
- This PR fixes the issue in v5 by making sure to set the `self.authenticationSession` property to nil for _both_ the success and cancel PayPal flows

### Checklist

- [X] Added a changelog entry

### Authors
- @scannillo 
- @sestevens

PR #929, opened 2023-01-18
Update ASWebAuthenticationSession Usage

jaxdesmarais:
### Summary of changes

- Removed `BTPayPalRequest.activeWindow`
- Removed merchant conformance to `ASWebAuthenticationPresentationContextProviding`
- Removed the `context` parameter from the `BTSEPADirectDebitClient.tokenize` method
- Moved `BTPayPalClient` and `BTSEPADirectDebitClient` conformance to `ASWebAuthenticationPresentationContextProviding` into an extension
- Updated demo app with the above changes

The above changes have been tested in the demo app to confirm everything is still working as it was previously 🚀 

### Checklist

- [x] Added a changelog entry

- @mattwylder @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,6 @@
   * BraintreeSEPADirectDebit
     * The `tokenize` method no longer takes in a `context` parameter
     * Merchants no longer need to conform to the `ASWebAuthenticationPresentationContextProviding` protocol
-      * This is now handled by the SDK in the `BTSEPADirectDebitClient`
 
 ## 5.19.0 (2022-12-19)
 * BraintreePayPalNativeCheckout (BETA)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 * Breaking Changes
   * BraintreePayPal
     * Rename `BTPayPalRequest.riskCorrelationId` to `BTPayPalRequest.riskCorrelationID`
+    * Removed `BTPayPalRequest.activeWindow` property
+      * The window will be set to the first window or a new `ASPresentationAnchor` if the first window is `nil`
     * Update `BTPayPalRequestLandingPageType` enum default case to `.none`
         * Update enum values
             * `.none` = 0
@@ -46,6 +48,10 @@
       * UnionPay cards can now be processed as regular cards (through the BraintreeCard module) due to their partnership with Discover
   * BraintreeCore
     * Remove `BTConfiguration+Extensions.isUnionPayEnabled` property
+  * BraintreeSEPADirectDebit
+    * The `tokenize` method no longer takes in a `context` parameter
+    * Merchants no longer need to conform to the `ASWebAuthenticationPresentationContextProviding` protocol
+      * This is now handled by the SDK in the `BTSEPADirectDebitClient`
 
 ## 5.19.0 (2022-12-19)
 * BraintreePayPalNativeCheckout (BETA)
--- a/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
+++ b/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
@@ -36,7 +36,6 @@ import BraintreeCore
     ///   - context: the ASWebAuthenticationPresentationContextProviding protocol conforming ViewController
     public func tokenize(
         request: BTSEPADirectDebitRequest,
-        context: ASWebAuthenticationPresentationContextProviding,
         completion:  @escaping (BTSEPADirectDebitNonce?, Error?) -> Void
     ) {
         apiClient.sendAnalyticsEvent("ios.sepa-direct-debit.selected.started")
@@ -59,7 +58,7 @@ import BraintreeCore
                 return
             } else if let url = URL(string: createMandateResult.approvalURL) {
                 self.apiClient.sendAnalyticsEvent("ios.sepa-direct-debit.create-mandate.success")
-                self.startAuthenticationSession(url: url, context: context) { success, error in
+                self.startAuthenticationSession(url: url, context: self) { success, error in
                     switch success {
                     case true:
                         self.tokenize(createMandateResult: createMandateResult, completion: completion)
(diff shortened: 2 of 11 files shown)


sshropshire (approved):

sshropshire:
It'd be interesting to test some of the window logic on an iPad to see what it does, that's the only multi window iOS device I can think of that we could wild out on.

jaxdesmarais:
> It'd be interesting to test some of the window logic on an iPad to see what it does, that's the only multi window iOS device I can think of that we could wild out on.

Yep, you are correct! We chatted about it here (I had a similar question): https://github.com/paypal/iOS-SDK/pull/104#discussion_r1062914756. @scannillo did some SPLUNK magic and found the following stats:

> SPLUNKing BT logs shows that 98% of all events are on iPhones vs iPads (multitasking is only supporting on iPads). So seeing as we don't get much traffic at all there combined with the fact that I think supporting multi-tasking for a checkout flow would be pretty rare for a merchant.

From there, we discussed removing it in the beta to see if we get any feedback on it being necessary. I had previously thought the property was added based on merchant feedback but Sammy confirmed that wasn't the case!

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -59,7 +58,7 @@ import BraintreeCore
                 return
             } else if let url = URL(string: createMandateResult.approvalURL) {
                 self.apiClient.sendAnalyticsEvent("ios.sepa-direct-debit.create-mandate.success")
-                self.startAuthenticationSession(url: url, context: context) { success, error in
+                self.startAuthenticationSession(url: url, context: self) { success, error in
```
scannillo, line 61: Might this cause a retain cycle? Between the SEPAClient.swift & the ASWebAuthSession?

We have[ this note](https://github.com/braintree/braintree_ios/blob/2a0a3ce2143bf8c3ce9194a9f06fb794f631f29f/Sources/BraintreePayPal/BTPayPalClient.m#L231-L232) for PayPalClient, which originated in this PR https://github.com/braintree/braintree_ios/pull/596.

jaxdesmarais: Good catch - yeah we can add this. It looks like the `authenticationSession` property is coupled to the `isAuthenticationSessionStarted` property in `BTPayPalClient`. Is that something we want to add as well to SEPA (looks like it is used to track if the auth session was started successfully since it would need to be optional to set to `nil` and prevent the retain cycle). In SEPA this is currently not optional so would need to make some updates, so may be good to also track this if we wanted.

scannillo: So going through the SEPA and PayPal flows a few times and using the Xcode memory graph debugger, I'm not able to replicate the same leak reported in that original GH Issue, using the code as-is on this branch.

I am not sure exactly why this is - but a handful has changed since that original bug. Both the demo & module are in Swift now + we have the wrapper WebAuthenticationSession around Apple's ASWeb, which could have something to do with it.

scannillo: Regarding `isAuthenticationSessionStarted`, I was just curious and ran a quick SPLUNK. `1.7%` of these ASWeb events are failures over the last day. I guess this isn't far off of what I expected, but I expected it to be slightly lower.

Screen Shot 2023-01-19 at 11 55 06 AM

In terms of adding these kinds of analytics for SEPA, on one hand, it is nice to maintain parity with our other ASWeb flows (ex: PayPal), but I know we have a history of tracking too much. I'd vote to add it in for now to maintain parity. Maybe at some point we'll audit the whole SDK analytics to streamline ... but I don't feel too strongly. Thoughts?

jaxdesmarais: Interesting that you aren't seeing it in the memory graphs. I tried poking around online to see if there were some updates somewhere but am coming up empty. But yeah could have also been tied to something else. 🤷 

I think that makes sense, we can add for now, make it `nil` just incase there is something that could go sideways somewhere. And yeah, I think for now we add the events so we have parity and we can always get rid of it down the road... or if it makes our life a pain with FPTI. 😅

mattwylder: What does the leak look like in the memory graph debugger? I don't think I see it on `5.x` either.

scannillo: > What does the leak look like in the memory graph debugger? I don't think I see it on 5.0.0-beta2 either.

If I comment out [this line](https://github.com/braintree/braintree_ios/blob/2a0a3ce2143bf8c3ce9194a9f06fb794f631f29f/Sources/BraintreePayPal/BTPayPalClient.m#L231-L232) (basically revert [this PR](https://github.com/braintree/braintree_ios/pull/596)), you can trigger the leak.

It will look like this after 1 PayPal checkout:
Screen Shot 2023-01-19 at 9 43 02 PM

And then after going through the PayPal flow 5 times it will look like this:
Screen Shot 2023-01-19 at 9 42 40 PM
With the fix, the count of `BTPayPalClient` and `BTPayPalCheckoutRequest` will _not_ continue increasing, and will just remain at 1.

scannillo: Since I can't trigger any issues with the current implementation, I don't think we need to make changes to the PR. I just don't really know exactly _why_ it is no longer an issue in this Swift version.

Also happy to go through it together, might be easier to make sure we're not missing something.

CHANGELOG.md
```diff
@@ -46,6 +48,10 @@
       * UnionPay cards can now be processed as regular cards (through the BraintreeCard module) due to their partnership with Discover
   * BraintreeCore
     * Remove `BTConfiguration+Extensions.isUnionPayEnabled` property
+  * BraintreeSEPADirectDebit
+    * The `tokenize` method no longer takes in a `context` parameter
+    * Merchants no longer need to conform to the `ASWebAuthenticationPresentationContextProviding` protocol
+      * This is now handled by the SDK in the `BTSEPADirectDebitClient`
```
scannillo, line 54: Take it or leave it comment - I don't think this is needed for the CHANGELOG
```suggestion
```

V6_MIGRATION.md
```diff
@@ -151,3 +154,13 @@ cardClient.tokenizeCard(card) { (tokenizedCard, error) in
     // Communicate the tokenizedCard.nonce to your server, or handle error
 }
 ```
+
+## SEPA Direct Debit
+`BTSEPADirectDebitClient.tokenize(request: BTSEPADirectDebitRequest, context: ASWebAuthenticationPresentationContextProviding)` has been renamed to `BTSEPADirectDebitClient.tokenize(request: BTSEPADirectDebitRequest)`. Additionally, conformance to the `ASWebAuthenticationPresentationContextProviding` protocol is no longer needed.
```
scannillo, line 159: Another take it or leave it comment - instead of "method was renamed to" this makes more sense to me as "removed parameter from".

scannillo (approved):
Changes look good to me! 🎸

Two minor take-it-or-leave-it comments.

Also happy to pair on the memory graph stuff. Can't hurt for us to all get more familiar with the tool.

CHANGELOG.md
jaxdesmarais: Yeah we added this because we weren't sure if it was clear that the responsibility of this shifted and there isn't a change in behavior. If we don't feel it's helpful we can remove it!

V6_MIGRATION.md
mattwylder: done!

PR #930, opened 2023-01-18
BraintreeCore Swift API Audit

mattwylder:
### Summary of changes

- Cleaned up TODO comments
- Moved  `callbackURLScheme` into `BTCoreConstants`
- Removed a force unwrap from `BTGraphQLErrorTree`

### Checklist

- @jaxdesmarais @mattwylder

### DIFF
--- a/Sources/BraintreeCore/BTClientToken.swift
+++ b/Sources/BraintreeCore/BTClientToken.swift
@@ -3,7 +3,8 @@ import Foundation
 /// An authorization string used to initialize the Braintree SDK
 @objcMembers public class BTClientToken: NSObject, NSCoding, NSCopying {
 
-    // TODO: resolve public access modifiers once rest of core is in Swift
+    // NEXT_MAJOR_VERSION (v7): properties exposed for Objective-C interoperability.
+    // Once the entire SDK is in Swift, determine if we want public properties to be internal
     // MARK: - Public Properties
 
     /// The client token as a BTJSON object
--- a/Sources/BraintreeCore/BTGraphQLErrorTree.swift
+++ b/Sources/BraintreeCore/BTGraphQLErrorTree.swift
@@ -20,7 +20,7 @@ class BTGraphQLErrorTree {
             if !parentNode.hasChild(forField: field) {
                 parentNode.insertChild(BTGraphQLMultiErrorNode(field: field))
             }
-            parentNode = parentNode.getChild(forField: field) as! BTGraphQLMultiErrorNode
+            parentNode = parentNode.getChild(forField: field) as? BTGraphQLMultiErrorNode ?? BTGraphQLMultiErrorNode(field: "Failed to find BTGraphQLMultiErrorNode for field: \(field)")
         }
         
         // add child node at end of key path
--- a/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
+++ b/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
@@ -33,7 +33,6 @@ import BraintreeCore
     ///   - mandateType: Optional. The `BTSEPADebitMandateType`. If not set, defaults to `.oneOff`
     ///   - billingAddress: Required. The user's billing address.
     ///   - merchantAccountID: Optional. A non-default merchant account to use for tokenization.
-    // NEXT_MAJOR_VERSION consider refactoring public request initializers to include required parameters instead of defaulting everything to optional
     public init(
         accountHolderName: String? = nil,
         iban: String? = nil,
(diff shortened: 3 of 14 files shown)


Sources/BraintreeCore/BTClientToken.swift
```diff
@@ -3,7 +3,8 @@ import Foundation
 /// An authorization string used to initialize the Braintree SDK
 @objcMembers public class BTClientToken: NSObject, NSCoding, NSCopying {
 
-    // TODO: resolve public access modifiers once rest of core is in Swift
+    // NEXT_MAJOR_VERSION (v7): properties exposed for Objective-C interoperability.
+    // Once the entire SDK is in Swift, determine if we want public properties to be internal
```
scannillo, line 7: Should the docstrings below get a `nodoc` if not intended for merchants?

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
```diff
@@ -33,7 +33,6 @@ import BraintreeCore
     ///   - mandateType: Optional. The `BTSEPADebitMandateType`. If not set, defaults to `.oneOff`
     ///   - billingAddress: Required. The user's billing address.
```
scannillo, line 34: For the params documented as "Required", should those params be non-optional?

Sources/BraintreeCore/BTClientToken.swift
jaxdesmarais: In `5.x` everything here was also in the [BTClientToken.h public header](https://github.com/braintree/braintree_ios/blob/5.x/Sources/BraintreeCore/Public/BraintreeCore/BTClientToken.h). Though it looks like most of the properties may have only been public to expose them to other Obj-C modules. This is more-so just to denote that we should explore if these need to remain public or if it was mostly done for interoperability reasons.

Seems like the [ClientToken class in Android](https://github.com/braintree/braintree_android/blob/master/BraintreeCore/src/main/java/com/braintreepayments/api/ClientToken.java) has these properties as internal, so may be able to clean up public access once we don't need to worry about Obj-C interoperability.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
jaxdesmarais: So this brings into question how we want to initialize objects in the SDK (and it is a little goofy). These are all optional and defaulted to `nil` so that we can use the dot syntax like:
```
var sepaDirectDebitRequest = BTSEPADirectDebitRequest()
sepaDirectDebitRequest.accountHolderName = "John Doe"
sepaDirectDebitRequest.iban = "FR891751244434203564412313"
...
```

If we make these required we would _always_ require merchants to initialize the required parameters like:
```
var sepaDirectDebitRequest = BTSEPADirectDebitRequest(
    accountHolderName: "John Doe", 
    iban: "FR891751244434203564412313", 
    customerID: "1234", 
    billingAddress: <a_long_billing_address_that_i_dont_want_to_type>
)
```

Then you could still use the dot syntax after initializing the required params (or add them to the init). We _could_ do that but it breaks convention of what the rest of the SDK does, so seems like something we would want to do at a later point when the entire SDK is in Swift so that we are consistent across the SDK. Thoughts?

jaxdesmarais: While we were doing this work initially I had the same question! I opened a GitHub discussion to see if merchants had any options but no one responded 😭 : https://github.com/braintree/braintree_ios/discussions/800

scannillo: > We could do that but it breaks convention of what the rest of the SDK does, so seems like something we would want to do at a later point when the entire SDK is in Swift so that we are consistent across the SDK.

Gotcha, thank you for the explanation. I agree, we should 🏈 on it when we're ready to be consistent across all modules.

Sources/BraintreeCore/BTClientToken.swift
scannillo: Ah gotcha. 

I would guess that this entire file was just exposed for Drop-In. We never instruct merchants to interact directly with `BTClientToken`, they just pass us a raw string into the `BTAPIClient`. 

I can't see any case where we'd want a merchant to use these values, or even instantiate a `BTClientToken` at all - so some `nodoc`s probably couldn't hurt. What do you think?

jaxdesmarais: Ah, that makes sense that it's exposed for the Drop-in. Yep we can add nodoc for this and specifically call out that it's exposed for the Drop-in as well. Will get this updated. 🚀

jaxdesmarais: It looks like some of these are used in the Drop-in and others not - still worth adding a `nodoc` but added an additional note about evaluating this entire class at some point down the road.

scannillo (approved):

Sources/BraintreeCore/BTGraphQLErrorTree.swift
```diff
@@ -20,7 +20,7 @@ class BTGraphQLErrorTree {
             if !parentNode.hasChild(forField: field) {
                 parentNode.insertChild(BTGraphQLMultiErrorNode(field: field))
             }
-            parentNode = parentNode.getChild(forField: field) as! BTGraphQLMultiErrorNode
+            parentNode = parentNode.getChild(forField: field) as? BTGraphQLMultiErrorNode ?? BTGraphQLMultiErrorNode(field: "Failed to find BTGraphQLMultiErrorNode for field: \(field)")
```
sshropshire, line 23: I wonder if Lines 20-23 could be reduced to something like
```
parentNode = parentNode.createChildIfNotExists(forField: field)
```
The name is total garbage, but it'll avoid having to force unwrap. Technically the right side of the null coalesce won't ever run as it's written.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
sshropshire: We could throw and validate / throw on error when the request is made. We could use empty strings, etc. as default values to allow the types to be non-optional.

sshropshire (approved):
🌊

PR #931, opened 2023-01-19
Update SEPA and NXO Test Schemes

jaxdesmarais:
### Summary of changes

This is very much a quality of like fix that has been annoying me. The test schemes for NXO and SEPA were always greyed out in the Demo scheme unlike the other tests which can be run from the Demo scheme. In order to run the NXO and SEPA tests you'd have to switch to the UnitTest scheme - not a huge deal but annoying enough. Now we can run these tests from the same scheme as the others.

| Before | After |
|-----|-----|
|Screen Shot 2023-01-19 at 8 59 15 AM|Screen Shot 2023-01-19 at 8 31 21 AM|

### Checklist

- @jaxdesmarais


mattwylder (approved):

mattwylder (approved):
🚢

scannillo (approved):

PR #937, opened 2023-01-24
Update Release Workflow (5.x)

jaxdesmarais:
### Summary of changes

During the 5.x release today the [release blew up on the last step](https://github.com/braintree/braintree_ios/actions/runs/3998282595/jobs/6860714412) (womp womp). We got the following error:
```
sourcekitten: A full installation of Xcode.app 14.0 is required to compile
this software. Installing just the Command Line Tools is not sufficient.

Xcode 14.0 cannot be installed on macOS 11.
You must upgrade your version of macOS.
```
tl;dr is that [Sourcekitten released a new version](https://github.com/jpsim/SourceKitten/releases/tag/0.34.0) with the following requirement: "SourceKitten now requires Swift 5.7 or higher to build.". 

Our CI for version 5.x uses Xcode 13. We went on an adventure and found out that it's actually quite difficult to specify an old version with Homebrew (more womp womp). We found [this article](https://cmichel.io/how-to-install-an-old-package-version-with-brew/) and tested locally to confirm it is working as expected. [Homebrew discussions](https://github.com/Homebrew/discussions/discussions/3251) confirms pulling in the latest version of a package is expected behavior and `brew extract` is how you specify an older version. 

**Note:** we do not need to pick up this change in `master` as it uses Xcode 14 + MacOS 12. We can use the normal `brew install sourcekitten` on that branch.

### Checklist

- @mattwylder @jaxdesmarais

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -115,7 +115,9 @@ jobs:
       - name: Publish reference docs
         run: |
           gem install jazzy
-          brew install sourcekitten
+          brew tap-new $USER/local-sourcekitten
+          brew extract --version=0.33.1 sourcekitten $USER/local-sourcekitten
+          brew install sourcekitten@0.33.1
 
           # run sourcekitten on each Swift module individually
           sourcekitten doc -- -workspace Braintree.xcworkspace -scheme PayPalDataCollector -destination 'name=iPhone 11,platform=iOS Simulator' > pay-pal-data-collector.json
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -115,9 +115,8 @@ jobs:
       - name: Publish reference docs
         run: |
           gem install jazzy
-          brew tap-new $USER/local-sourcekitten
-          brew extract --version=0.33.1 sourcekitten $USER/local-sourcekitten
-          brew install sourcekitten@0.33.1
+          sudo xcode-select -switch /Applications/Xcode_14.0.app
+          brew install sourcekitten
 
           # run sourcekitten on each Swift module individually
           sourcekitten doc -- -workspace Braintree.xcworkspace -scheme PayPalDataCollector -destination 'name=iPhone 11,platform=iOS Simulator' > pay-pal-data-collector.json


.github/workflows/release.yml
```diff
@@ -115,7 +115,9 @@ jobs:
       - name: Publish reference docs
         run: |
           gem install jazzy
-          brew install sourcekitten
+          brew tap-new $USER/local-sourcekitten
```
jaxdesmarais, line 118: We _assume_ `$USER/local-sourcekitten` will work as expected but were not able to find anything online confirming one way or the other. This is the last step so if for some reason this blows up we should look at maybe moving the docs into its own workflow that can be run independent of the release script if needed. Thoughts?

scannillo:
Why can't we instead bump the Xcode version for the `5.x` branch release script?

scannillo:
> We can, we just didn't want to take the liberty. Some merchants using 5.x may still use Xcode 13, so we wanted to make sure the entire release workflow ran on Xcode 13 to ensure backwards compatibility.

Ah, you're right!

So I think SPM and CocoaPods would be fine if we bumped to Xcode 14. For those, we specify the swift tools version to build with in the [Package.swift](https://github.com/braintree/braintree_ios/blob/04fd2351c2ab7e65ec2196bf07339fba48e1144f/Package.swift#L1) and the [Podspec](https://github.com/braintree/braintree_ios/blob/04fd2351c2ab7e65ec2196bf07339fba48e1144f/Braintree.podspec#L20), which we aren't changing. So long as Xcode 14 didn't drop support for those versions, our release script would be able to build those fine while maintaining backward compatibility.

But then there's Carthage 🙃. If we bump to Xcode 14, the `swift-compiler-version` of the frameworks we upload to our GH release will bump to `5.7.x` from `5.5.x`.

jaxdesmarais:
> Carthage

Carthage is _always_ the problem child 🤣

scannillo (approved):

sshropshire (approved):

PR #939, opened 2023-01-25
Remove Nested Dictionary Construction

jaxdesmarais:
## Summary of changes

- Using Xcode's timeline tool, we saw that a few files were taking 1+ second to compile. This is due to how [Swift/Xcode compiles nested dictionaries](https://github.com/apple/swift/issues/47007) - tl;dr much slower.
- This PR extracts nested dictionaries into their own constants to speed up this compile time.
- As an example, the update here to the dictionaries in `SEPADirectDebitAPI` saves us a whopping second of compile time!
- **Note for future us:** we should look into if SwiftLint has the ability to check for this anti-pattern _or_ add a custom regex rule for this - can also likely add this to our style guide

### Checklist

- @mattwylder @jaxdesmarais


mattwylder:
A regular expression that can be added to SwiftLint to help find nested dictionaries. Though it only works with proper newlines at the end of the square bracket.
`\".*\"\s*:\s*\[(?>\s*.*,\n*)+(?>\s+.*\n*)\s*\]`

KunJeongPark (approved):
Looks good.

jcnoriega (approved):

PR #940, opened 2023-01-25
Update Cocoapods version and Remove Gemfile

mattwylder:
### Summary of changes

- Update `Podfile.lock`
- Remove `Gemfile` and `Gemfile.lock`

### Checklist

- @jaxdesmarais @mattwylder


scannillo:
Looks like there's some places we'll need to update our instructions for running the Demo ([here](https://github.com/braintree/braintree_ios/blob/master/CONTRIBUTING.md)).

Also, maybe we should specify in [these instructions](https://github.com/braintree/braintree_ios#demo) that people need to download CocoaPods on their own. Do we think it would be useful to specify the version of CocoaPods our project requires too in these instructions, too?

scannillo:
Also without the Gemfile, the Rakefile will not be able to find its gem dependencies (it already wasn't working for other reasons). Maybe we make a separate JIRA to clean that up and replace outdated `rake` commands in our repo (ex [here](https://github.com/braintree/braintree_ios/blob/master/DEVELOPMENT.md#tests)). 🏈

mattwylder:
I was under the mistaken assumption that we were no longer using rake at all. Reverting

scannillo:
> I was under the mistaken assumption that we were no longer using rake at all.

Woops, sorry if I was unclear. We aren't using the Rakefile anymore. It hasn't been maintained since we made the migration to GH Actions for CI. (We used to run our Rakefile on Jenkins CI, but it was a pain in the butt to iterate on). What I meant was, we can make a separate task to remove the Rakefile entirely and all of it's references. We can also consider replacing the functionality it once provided (a way to run tests via command line) if we think it is useful. Clearly we don't do it very often if none of us have ever bothered trying in over a year since we moved to GH actions lol.

Let me know if that makes sense. I think it is OK to remove the Gemfile, but the Rakefile being outdated is also related, so it can be a fast follow.

jcnoriega (approved):

mattwylder:
@scannillo do you think we should go ahead and remove the Rakefile in this PR, since it's related, or should we leave it for now and make a ticket to replace it with something else like Fastlane?

scannillo:
> do you think we should go ahead and remove the Rakefile in this PR, since it's related, or should we leave it for now and make a ticket to replace it with something else like Fastlane?

I think we should 🏈 punt on it. Determine first what we want to use it for.

scannillo (approved):

PR #941, opened 2023-01-26
Update `BTAnalyticsMetadata_Tests` to use Regex

mattwylder:
### Summary of changes

- Added match regex assertions to `BTAnalyticsMetadata_Tests`

### Checklist

- @mattwylder


UnitTests/BraintreeCoreTests/BTAnalyticsMetadata_Tests.swift
```diff
@@ -4,22 +4,41 @@ import XCTest
...
+        assertRegexMatch(metadata["sdkVersion"] as? String, "^\\d+\\.\\d+\\.\\d+(-[0-9a-zA-Z-]+)?$")
+        XCTAssertNotNil(metadata["merchantAppId"] as? String) //OCMocked
+        XCTAssertNotNil(metadata["merchantAppName"] as? String) // OCMocked
+        XCTAssertNotNil(metadata["merchantAppVersion"] as? String) // OCMocked, could be regex
         XCTAssertEqual(metadata["deviceManufacturer"] as? String, "Apple")
-        XCTAssertNotNil(metadata["deviceModel"] as? String)
-        XCTAssertNotNil(metadata["deviceAppGeneratedPersistentUuid"] as? String)
-        XCTAssertNotNil(metadata["iosIdentifierForVendor"] as? String)
+        assertRegexMatch(metadata["deviceModel"] as? String, "iPhone\\d,\\d|i386|x86_64|arm64")
+        XCTAssertNotNil(metadata["deviceAppGeneratedPersistentUuid"] as? String) // Empty String because there's no BTKeychain "^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$"
```
mattwylder, line 18: `BTAnalyticsMetadata`[returns the empty String](https://github.com/braintree/braintree_ios/blob/59a69da80d5735967df8d1533df7d1022de66d94/Sources/BraintreeCore/BTAnalyticsMetadata.swift#L62-L67) because BTKeychain can't use methods from `Security`in the test environment. Did `OCMock` work around that issue?

jaxdesmarais: Afaik, the only way to do this in our tests would be what I had to do [here for the keychain requirements in Cardinal](https://github.com/braintree/braintree_ios/pull/891) - add demo app for Core tests to allow keychain entitlements. We can certainly do this but do we want to 🤷 . 

It also looks like on [PPCP we don't care about keychain](https://github.com/paypal/iOS-SDK/blob/855802216f9fbddbc6709d101e8871b307999628/Sources/PaymentsCore/Networking/AnalyticsService.swift#L10) only on BT. Should we care about it on BT or should we just always use the UUID like on PPCP?

mattwylder: In my opinion, it's probably not worth wrapping it all up like that. We may as well add a comment that certain values aren't tested because of their dependencies on keychain. We could also consider reworking the var in `BTAnalyticsMetadata` to fail differently, and see where that takes us.

jaxdesmarais: I'd lean towards just generating a UUID on BT like we do on PPCP. Curious though if @scannillo has any context about why we need to/care about putting the UUID into the keychain on BT - doesn't look like it's ever accessed anywhere, but I may be missing some context.

mattwylder: If we're not using the UUID in arachne I'm down to remove it.

jaxdesmarais: I wasn't suggesting removing it all-together, I assume we use the UUID it's more a question of do we use the current implementation of a persistent UUID via Keychain that I am not sure if. If we can remove the keychain requirement and just do `UUID().uuidString` vs accessing/putting it into the keychain is the question I think.

If we _do_ care about the persistent value, I think I'd lean towards not testing it vs standing up a demo with keychain entitlements enabled.

scannillo: @jaxdesmarais - It looks like BTKeychain is being used as a cache sort of? We're generating a unique ID for this device, storing it, and retrieving it when the user re-opens the app is my guess. Sort of as a way for us to _attempt_ to uniquely identify devices using our SDK.

I think we should remove it entirely, I've never seen this value prove useful, and doubt you have too either, right Jax?

scannillo: Another thing - V6 might be a good time for us to slim down our metadata to match how we're planning to send analytics metadata in PPCP (FPTI). We made [this doc](https://paypal.sharepoint.com/:x:/r/sites/BTSDK/_layouts/15/Doc.aspx?sourcedoc=%7B2BE96AB1-F4B4-4040-B4B0-3BBA636310C1%7D&file=FPTI%20Tags%20-%20Northstar%20Mobile%20Events.xlsx&action=default&mobileredirect=true&cid=59f3ce54-d020-4702-8ce3-9a3f6f4e4811)

which shows the tags/params we opted not useful (crossed out).

jaxdesmarais: @scannillo agreed, I've never used this value in this way, so think it makes sense to just return the UUID string. That also seems more useful as it'd still remain a constant through the session but agree that I can't really see the benefit of caching this in the keychain.

Also agree that this is a good opportunity to slim some of these down. Currently we have:
```
        [
            "platform": platform,
            "platformVersion": platformVersion,
            "sdkVersion": sdkVersion,
            "merchantAppId": merchantAppID,
            "merchantAppName": merchantAppName,
            "merchantAppVersion": merchantAppVersion,
            "deviceManufacturer": deviceManufacturer,
            "deviceModel": deviceModel,
            "iosDeviceName": iOSDeviceName,
            "iosSystemName": iOSSystemName,
            "iosBaseSDK": iOSBaseSDK,
            "iosDeploymentTarget": iOSDeploymentTarget,
            "iosIdentifierForVendor": iOSIdentifierForVendor,
            "iosPackageManager": iOSPackageManager,
            "deviceAppGeneratedPersistentUuid": deviceAppGeneratedPersistentUUID,
            "isSimulator": isSimulator,
            "deviceScreenOrientation": deviceScreenOrientation,
            "venmoInstalled": isVenmoInstalled,
            "dropinVersion": dropInVersion
        ]
```

My initial thought would be to update the following:
```
"platformVersion" + "iosSystemName" > "clientOS" like PPCP
"deviceAppGeneratedPersistentUuid" > "sessionUUID"

// Remove
"iosBaseSDK"
"iosDeploymentTarget"
"iosIdentifierForVendor" < also a UUID why idk!
"deviceScreenOrientation"
```
But curious to hear your thoughts on what has been helpful for you to use previously as there may just be some that I've simply not had the use-case for yet.

jaxdesmarais: I think it may make sense to do this in this PR since it's related (and relatively small currently). Does making these changes require us to notify anyone? More thinking about any of the ones that we decide to rename or remove.

scannillo: Not sure if you saw [this link above](https://paypal.sharepoint.com/:x:/r/sites/BTSDK/_layouts/15/Doc.aspx?sourcedoc=%7B2BE96AB1-F4B4-4040-B4B0-3BBA636310C1%7D&file=FPTI%20Tags%20-%20Northstar%20Mobile%20Events.xlsx&action=default&mobileredirect=true&cid=59f3ce54-d020-4702-8ce3-9a3f6f4e4811), where Matt and I basically wrote out all the BT tags sent for analytics, and removed the ones we deemed not useful.

The list mostly lines up with your list, with just 3 others we opted not to add in PPCP, in addition to the ones you called out:
```
venmoInstalled
paypalInstalled
deviceAppGeneratedPersistentUuid
```

scannillo: I think we are OK to remove params in V6, but we make things a bit messy if we start adding/re-naming things. This is because of the plan to migrate incoming Arachne data --> FPTI. 

If we rename in V6, then Arachne will have to map both the naming conventions for V4/5 & V6 to the appropriate FPTI tags. It will be easier of Arachne only has 1 set of mappings to follow for all iOS events. I think when the SDKs themselves are ready to move to FPTI though, we should definitely align with PPCP naming.

Does that make sense? What do you think?

jaxdesmarais: Yeah, I poked through that doc. I feel like `venmoInstalled` could be helpful, but I'm not seeing `paypaInstalled` even on [v5 analytics metadata](https://github.com/braintree/braintree_ios/blob/5.x/Sources/BraintreeCore/BTAnalyticsMetadata.m). I also removed `deviceAppGeneratedPersistentUUID` since if we aren't keeping it persistent there doesn't seem to be a compelling reason to keep it around. We already have the session ID which is the UUID, but curious if the duplication is helpful. (commit https://github.com/braintree/braintree_ios/pull/941/commits/2cb5eeae2ed82a6001c09f21ee6ff8f269ef1dc7 and https://github.com/braintree/braintree_ios/pull/941/commits/7f0479f1a441e90ce374bbb8341decf53c42f541)

What you said for removing but not renaming makes total sense. This also will set us up in a good place to rename them if/when we fully cut over to FPTI to align them with PPCP.

scannillo: Ah, I think we removed `paypalInstalled` from iOS but it still lives on Android. Leaving `venmoInstalled` sounds good to me.

scannillo (approved):
LGTM! But I think the removal of these 5 params is worthy of a CHANGELOG entry, mostly for our future selves.

CHANGELOG.md
```diff
@@ -49,6 +50,7 @@
   * BraintreeCore
     * Remove `BTConfiguration+Extensions.isUnionPayEnabled` property
     * Remove `BTPreferredPaymentMethods` and `BTPreferredPaymentMethodResult`
+    * Remove `BTKeychain` class and methods
```
scannillo, line 53: Nit - This wasn't public, was it? Don't think it needs an entry.

jaxdesmarais: It was public before ([here in v6](https://github.com/braintree/braintree_ios/blob/master/Sources/BraintreeCore/BTKeychain.swift) but [wasn't in v5](https://github.com/braintree/braintree_ios/blob/5.x/Sources/BraintreeCore/BTKeychain.h)). Genuinely no idea why we made it public unless it was for some weird interop reason during the conversion? But even then 🤷 . Worth calling it out since it technically a breaking change between betas or just remove this line?

sshropshire (approved):

PR #943, opened 2023-01-27
Fix NXO Dictionary Construction

jaxdesmarais:
### Summary of changes

- A merchant reported an error where setting `BTPayPalNativeVaultRequest.shippingAddressOverride` as `nil` was throwing an error. Looking at the dictionary construction we were incorrectly constructing the dictionaries in this case resulting in an empty dictionary being passed as the "shipping_address" which is invalid.
- There were some other dictionaries where we weren't checking for `nil` first so made some updates in both the `BTPayPalNativeVaultRequest` and BTPayPalNativeCheckoutRequest` to ensure the dictionaries were constructed as appropriate for the values in the requests.
- Added a unit tests for the specific `request.shippingAddressOverride = nil` case initially reported

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeRequest.swift
+++ b/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeRequest.swift
@@ -145,15 +145,15 @@ import BraintreePayPal
             }
 
             // Should only include shipping params if they exist
-            if let shippingAddressOverride {
+            if shippingAddressOverride != nil {
                 let shippingAddressParameters: [String: String?] = [
-                    "line1": shippingAddressOverride.streetAddress,
-                    "line2": shippingAddressOverride.extendedAddress,
-                    "city": shippingAddressOverride.locality,
-                    "state": shippingAddressOverride.region,
-                    "postal_code": shippingAddressOverride.postalCode,
-                    "country_code": shippingAddressOverride.countryCodeAlpha2,
-                    "recipient_name": shippingAddressOverride.recipientName,
+                    "line1": shippingAddressOverride?.streetAddress,
+                    "line2": shippingAddressOverride?.extendedAddress,
+                    "city": shippingAddressOverride?.locality,
+                    "state": shippingAddressOverride?.region,
+                    "postal_code": shippingAddressOverride?.postalCode,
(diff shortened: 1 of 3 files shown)


Sources/BraintreePayPalNativeCheckout/BTPayPalNativeRequest.swift
```diff
@@ -105,58 +105,59 @@ import BraintreePayPal
...
         case .checkout:
             guard let request = request as? BTPayPalNativeCheckoutRequest else { return [:] }
+            var checkoutParameters: [String: Any] = [
+                "intent": request.intentAsString,
+                "amount": request.amount,
+                "offer_pay_later": request.offerPayLater
+            ]
 
-            var billingAgreementDictionary: [AnyHashable: Any]? = [:]
+            let currencyCode = request.currencyCode != nil ? request.currencyCode : configuration.json["paypal"]["currencyIsoCode"].asString()
```
scannillo, line 114:
```suggestion
            let currencyCode = request.currencyCode ?? configuration.json["paypal"]["currencyIsoCode"].asString()
```

I think we can use the nil coalescing operator here

mattwylder (approved):

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeRequest.swift
```diff
@@ -105,58 +105,59 @@ import BraintreePayPal
...
-            var billingAgreementDictionary: [AnyHashable: Any]? = [:]
+            let currencyCode = request.currencyCode ?? configuration.json["paypal"]["currencyIsoCode"].asString()
 
-            if request.billingAgreementDescription != nil {
-                billingAgreementDictionary?["description"] = request.billingAgreementDescription
-            } else {
-                billingAgreementDictionary = nil
+            if currencyCode != nil {
+                checkoutParameters["currency_iso_code"] = currencyCode
             }
```
scannillo, line 118: Take it or a leave it comment to use an if-let combining with the line above:

```suggestion
            if let currencyCode = request.currencyCode ?? configuration.json["paypal"]["currencyIsoCode"].asString() {
                checkoutParameters["currency_iso_code"] = currencyCode
            }
```

scannillo (approved):

PR #947, opened 2023-01-31
[WIP] Move Publish Docs to Own Action/Workflow

jaxdesmarais:
DRAFT/DISCUSSION PR

### Summary of changes

The 5.x [publish reference docs step](https://github.com/braintree/braintree_ios/actions/runs/4055780046/jobs/6979378871) failed again. 😭 We were doing the following: `sudo xcode-select -switch /Applications/Xcode_14.0.app` but have the runner set to `macOS-11` which [does not support Xcode 14](https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#xcode). We will need to be in the [MacOS-12 runner](https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md#xcode) for Xcode 14 support.

Matt and I were talking and since we have had docs failures as part of the 6.x-beta's and now on 5.x due to changes to sourcekitten we thought it could be a good idea to separate the docs release into it's own action. This will allow us to:
1. Publish the reference docs as usual during the release process
2. Publish the docs independent of the release process if needed - this would be done in the same workflow section but selecting "publish docs" instead of release (this will _not_ show in GH Actions tab until we merge this change in and merge the 5.x branch into `master`)

This PR is currently a draft because we wanted to discuss with the team their thoughts on moving this to it's own workflow.

### Checklist

- @mattwylder @jaxdesmarais

### DIFF
--- a/.github/actions/release_docs/action.yml
+++ b/.github/actions/release_docs/action.yml
@@ -1,0 +1,51 @@
+name: Publish reference docs
+description: 'Publish reference docs to GitHub Pages'
+inputs:
+  version:
+    description: "Version of the reference docs to release"
+    required: true
+runs:
+  using: actions/runner-images@macOS12
+  steps:
+    - name: Use Xcode 14
+      run: sudo xcode-select -switch /Applications/Xcode_14.0.app
+
+    - name: Publish reference docs
+      run: |
+          gem install jazzy
+          brew install sourcekitten
+
+          # run sourcekitten on each Swift module individually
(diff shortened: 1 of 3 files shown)


jaxdesmarais:
Closing in favor of a simplified approach in https://github.com/braintree/braintree_ios/pull/949 - we can make a decision on this approach as we potentially migrate to Fastlane and/or DocC which may eliminate the sourcekitten + jazzy issues we've ran into.

PR #948, opened 2023-02-02
[DRAFT] Remove 3DSecure Subclassing of PaymentFlow

jaxdesmarais:
### Summary of changes

While looking at the next conversion to Swift (PaymentFlow) we realized that we were doing some subclassing that we would not be able to do if PaymentFlow was in Swift and 3DSecure was in Obj-C (technical limitation of the languages that can go the other way but not the direction we would need). We discussed as the iOS group that we wanted to spike out the changes needed to make this work and see if we would like to move forward.

As a note the classes being subclassed today: `BTPaymentFlowRequest` and `BTPaymentMethodResult` are empty in their implementation and are solely used to inherit their superclass of `NSObject`

#### Changes covered in this PR
- Update completion blocks to use `NSObject`
- Update `startPaymentFlow` and `setupPaymentFlow` to take in and return an `NSObject` (this is the underlying type of the classes `BTPaymentFlowRequest` and `BTPaymentMethodResult` used previously
- Update `onPaymentComplete` to return an `NSObject`
- Update `setupPaymentFlow` to use an `NSObject` conforming to `BTPaymentFlowRequestDelegate` and return an `NSObject` in the completion
- `prepareLookup` now takes a `BTThreeDSecureRequest`
- `initializeChallengeWithLookupResponse` now take a `BTThreeDSecureRequest` and returns an `NSObject` in the completion
- `BTThreeDSecureRequest` and` BTThreeDSecureResult` now subclass `NSObject` directly

#### Changes not covered in this PR
Another limitation would be that the 3DS module has a category/extension on `BTPaymentFlowClient` - during the conversion we would likely need to move that extension into a public extension on `BTPaymentFlowClient` instead of in the 3DS module. This would be similar to what we needed to do for the `BTConfiguration` extensions. We could then move this back into the 3DS module when we convert it to Swift in v7. From a merchant perspective they would have the same exposure to the objects and methods in this extension that they do now.

### Testing
All tests are passing - additionally going through both flows in the Demo app is working just as it was previously.

### Checklist

Will add a changelog if we decide to move forward with this approach

- [ ] Added a changelog entry

- @KunJeongPark @jaxdesmarais

### DIFF
--- a/Sources/BraintreePaymentFlow/Public/BraintreePaymentFlow/BTPaymentFlowClient.h
+++ b/Sources/BraintreePaymentFlow/Public/BraintreePaymentFlow/BTPaymentFlowClient.h
@@ -56,7 +56,7 @@ typedef NS_ENUM(NSInteger, BTPaymentFlowErrorType) {
  @param result The BTPaymentFlowResult of the payment flow.
  @param error NSError containing details of the error.
  */
-- (void)onPaymentComplete:(BTPaymentFlowResult * _Nullable)result error:(NSError * _Nullable)error;
+- (void)onPaymentComplete:(NSObject * _Nullable)result error:(NSError * _Nullable)error;
 
 /**
  Returns the base return URL scheme used by the client.
@@ -85,7 +85,7 @@ typedef NS_ENUM(NSInteger, BTPaymentFlowErrorType) {
  @param request A BTPaymentFlowRequest request.
  @param delegate The BTPaymentFlowClientDelegate to handle response.
  */
-- (void)handleRequest:(BTPaymentFlowRequest *)request client:(BTAPIClient *)apiClient paymentClientDelegate:(id<BTPaymentFlowClientDelegate>)delegate;
+- (void)handleRequest:(NSObject *)request client:(BTAPIClient *)apiClient paymentClientDelegate:(NSObject<BTPaymentFlowClientDelegate>* )delegate NS_SWIFT_NAME(handle(_:client:paymentClientDelegate:));
 
 /**
  Check if this BTPaymentFlowRequestDelegate can handle the return URL
@@ -136,7 +136,7 @@ typedef NS_ENUM(NSInteger, BTPaymentFlowErrorType) {
  @param request A BTPaymentFlowRequest request.
  @param completionBlock This completion will be invoked exactly once when the payment flow is complete or an error occurs.
  */
-- (void)startPaymentFlow:(BTPaymentFlowRequest<BTPaymentFlowRequestDelegate> *)request completion:(void (^)( BTPaymentFlowResult * _Nullable result,  NSError * _Nullable error))completionBlock;
+- (void)startPaymentFlow:(NSObject<BTPaymentFlowRequestDelegate> *)request completion:(void (^)( NSObject * _Nullable result,  NSError * _Nullable error))completionBlock;
 
 /**
  A required delegate to control the presentation and dismissal of view controllers.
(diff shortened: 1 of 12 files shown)


Sources/BraintreePaymentFlow/Public/BraintreePaymentFlow/BTPaymentFlowClient.h
```diff
@@ -136,7 +136,7 @@ typedef NS_ENUM(NSInteger, BTPaymentFlowErrorType) {
  @param request A BTPaymentFlowRequest request.
  @param completionBlock This completion will be invoked exactly once when the payment flow is complete or an error occurs.
  */
-- (void)startPaymentFlow:(BTPaymentFlowRequest<BTPaymentFlowRequestDelegate> *)request completion:(void (^)( BTPaymentFlowResult * _Nullable result,  NSError * _Nullable error))completionBlock;
+- (void)startPaymentFlow:(NSObject<BTPaymentFlowRequestDelegate> *)request completion:(void (^)( NSObject * _Nullable result,  NSError * _Nullable error))completionBlock;
```
scannillo, line 139: I don't think using `NSObject` in the merchant-facing API is expressive enough.

scannillo: For example:

The Swift version of this func would be:  
```
func startPaymentFlow(_ request: NSObjectProtocol & BTPaymentFlowRequestDelegate, completion completionBlock: @escaping (NSObject?, Error?) -> Void)
```
---
An example of the Xcode auto-fill when in Swift:
Screen Shot 2023-02-03 at 10 09 29 AM

jaxdesmarais: That's a good point - the other option would be to override this method (or call the super method with the more generic interface) in the individual client implementations. The main advantage being that it would be even more strongly typed than it is now - so for 3DS the merchant facing API interface could _only_ use and return BTThreeDSecureRequest/Result vs even what we have today using BTPaymentFlowRequest/Result. 

It will ultimately require moving some things around throughout the conversion because we'd likely want a BTLocalPaymentMethodClient and to create a BTThreeDSecureClient (vs a category that we have now). That seems to be our preference though in other module to keep the merchant facing API more strongly typed than it is today, if possible. This is a similar approach to what we took in the PayPal/PayPalNative modules with the tokenize methods. Thoughts?

jaxdesmarais: I started spiking what I was thinking above out and it's gonna be a nightmare and a half. Technically possible but will require more changes than I think it's worth. Especially since it'd involve breaking changes (while small but still breaking) to 3DS, LPMs, and PaymentFlow. My thoughts are we should scrap converting PaymentFlow for now and make it dependent on converting 3DS first as that will minimizing the breaking changes in all payment methods that depend on PaymentFlow. Then just move forward with:
1. Convert another small module that has no dependancies to get us to the "halfway point"
2. Still switch payment flow to use ASWeb as this will get us in a good place down the line for the future conversion of modules in v7

Curious to hear your thoughts @scannillo!

scannillo: > That seems to be our preference though in other module to keep the merchant facing API more strongly typed than it is today, if possible.

I agree. Currently, merchants need to cast the result type, which is a little odd.

```
 self.paymentFlowDriver.startPaymentFlow(request) { (result, error) in
    guard let result = result as? BTThreeDSecureResult else {
        // Handle error
    }
}
```

scannillo: I think that sounds like an excellent plan forward! Thank you for the thorough investigation on this 👏

jaxdesmarais:
Closing based on the above conversation - will convert this module in v7 after we convert 3DS

PR #949, opened 2023-02-03
GitHub Actions Release Update to use MacOS-12 for Reference Docs (5.x)

jaxdesmarais:
### Summary of changes

The 5.x [publish reference docs step](https://github.com/braintree/braintree_ios/actions/runs/4055780046/jobs/6979378871) failed again. 😭 We were doing the following: sudo xcode-select -switch /Applications/Xcode_14.0.app but have the runner set to macOS-11 which [does not support Xcode 14](https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#xcode). We will need to be in the [MacOS-12 runner](https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md#xcode) for Xcode 14 support.

This PR separates the docs publishing into it's own job ([we do something similar in Android release](https://github.com/braintree/braintree_android/blob/master/.github/workflows/release.yml) and the [iOS build](https://github.com/braintree/braintree_ios/blob/master/.github/workflows/build.yml)). This way we can control the macOS version used for this step. As we look into potentially moving things to Fastlane or something similar as well as potentially moving to DocC we can evaluate if it makes sense to move the docs into their own runnable job (though the local development of DocC I've been playing with may make this specific issue we've been running into irrelevant). 

### Checklist

- @jaxdesmarais

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -112,10 +112,19 @@ jobs:
           COCOAPODS_TRUNK_TOKEN: ${{ secrets.COCOAPODS_TRUNK_TOKEN }}
         run: pod trunk push Braintree.podspec
 
+  docs:
+    name: Publish Reference Docs
+    runs-on: macOS-12
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+
+      - name: Use Xcode 14
+        run: sudo xcode-select -switch /Applications/Xcode_14.0.app
+
       - name: Publish reference docs
         run: |
           gem install jazzy
-          sudo xcode-select -switch /Applications/Xcode_14.0.app
           brew install sourcekitten
 
           # run sourcekitten on each Swift module individually


sshropshire (approved):

scannillo (approved):
🎸

scannillo:
IMO - [this](https://forums.swift.org/t/nodoc-attribute-for-hiding-symbols-from-the-symbol-graph/59227) is the main thing blocking us from moving our reference docs DocC - no equivalent for `nodoc`. It's coming in swift 5.8 though, the next Xcode release likely!

Also, I remember reading that it will be hard (or impossible) to preserve our reference docs URL due to the way DocC builds for GH pages. So the URL for the reference docs may have to change. 
---
Found the [article](https://www.jessesquires.com/blog/2022/04/22/docc-on-github-pages/) if you're interested under the `DocC URL schemes are… strange` section. (example: `braintree.github.io/braintree_ios` --> `braintree.github.io/braintree_ios/documentation/Braintree`). Maybe things have improved since last April though!

jaxdesmarais:
> IMO - [this](https://forums.swift.org/t/nodoc-attribute-for-hiding-symbols-from-the-symbol-graph/59227) is the main thing blocking us from moving our reference docs DocC - no equivalent for nodoc. It's coming in swift 5.8 though, the next Xcode release likely!

Yeah, this is true. Right now they just generate with the `:nodoc:` which is maybe helpful but also not ideal... ideally they wouldn't be there at all. Glad to hear it's coming in Swift 5.8 though!

> Also, I remember reading that it will be hard (or impossible) to preserve our reference docs URL due to the way DocC builds for GH pages. So the URL for the reference docs may have to change.

Yeah, this is true ([and touched on in the WWDC video for DocC as well](https://developer.apple.com/videos/play/wwdc2022/110368/?time=586)). It's not fully impossible but it's certainly harder than it should be. It looks like a lot of folks use some sort of redirect from the original page to the new one [like in the example here using the meta tag](https://www.jessesquires.com/blog/2022/04/22/docc-on-github-pages/). Though this could also be improved in the next version.

It does look like the [DocC SPM plugin](https://apple.github.io/swift-docc-plugin/documentation/swiftdoccplugin/) will be super helpful for documenting all of our targets vs 1 at a time via the command line, so that's neat.

PR #952, opened 2023-02-07
Update PaymentFlow to Use ASWebAuthenticationSession

jaxdesmarais:
### Summary of changes

- Remove `BTViewControllerPresentingDelegate`
- Cleanup unused methods in Integration tests
- Remove `returnURLScheme` and instead use `BTCoreConstants. callbackURLScheme` in request
- Update request to use `asAddress` helper on `BTJSON`
- `BTPaymentFlowClient` Updates
    - `BTPaymentFlowClient` only conforms to `ASWebAuthenticationPresentationContextProviding` now
    - Removed conformance to `SFSafariViewControllerDelegate` and `BTAppContextSwitchClient`
    - Remove `load` method
    - Remove unneeded properties and methods now that we use `ASWebAuthenticationSession`
    - Update to use `ASWebAuthenticationSession`
    - Update internal header properties
    - Removed `onPaymentCancel` and `returnURLScheme` methods from `BTPaymentFlowClientDelegate`
    - Removed `canHandleAppSwitchReturnURL` method from `BTPaymentFlowRequestDelegate`
    - Removed `viewControllerPresentingDelegate` property from `BTPaymentFlowClient`
- A bunch of updates to the unit tests now that we removed `SFSafari` specific methods
    - Also updated some syntax while we were at it

### Checklist

- [x] Added a changelog entry

- @KunJeongPark @jaxdesmarais


sshropshire (approved):
🌊

sshropshire:
The only 🧶: should there be a few CHANGELOG callouts for this?

Sources/BraintreePaymentFlow/Public/BraintreePaymentFlow/BTPaymentFlowClient.h
```diff
@@ -46,24 +44,13 @@ typedef NS_ENUM(NSInteger, BTPaymentFlowErrorType) {
  */
 - (void)onPaymentWithURL:(NSURL * _Nullable) url error:(NSError * _Nullable)error;
 
-/**
- Use when the payment flow was canceled.
- */
-- (void)onPaymentCancel;
```
scannillo, line 52: I think we still want to give merchants a hook into if the flow was cancelled

V6_MIGRATION.md
```diff
@@ -164,3 +166,17 @@ sepaDirectDebitClient.tokenize(request: sepaDirectDebitRequest) { sepaDirectDebi
...
+
+## Local Payment Methods
+We have replaced the deprecated `SFAuthenticationSession` with `ASWebAuthenticationSession` in the Local Payment Method flow.
+
+Your view no longer needs to conform to the `BTViewControllerPresentingDelegate` protocol. The methods `BTPaymentFlowClient.paymentClient(BTPaymentFlowClient, requestsPresentationOfViewController: UIViewController)` and `BTPaymentFlowClient.paymentClient(BTPaymentFlowClient, requestsDismissalOfViewController: UIViewController)` have been removed. 
+
+Additionally, you do not need to assign the `BTPaymentFlowClient.viewControllerPresentingDelegate` property in your view.
+
+## Three D Secure
+We have replaced the deprecated `SFAuthenticationSession` with `ASWebAuthenticationSession` in the Local Three D Secure flow.
```
scannillo, line 178:
```suggestion
We have replaced the deprecated `SFAuthenticationSession` with `ASWebAuthenticationSession` in the Three D Secure flow.
```

V6_MIGRATION.md
```diff
@@ -164,3 +166,17 @@ sepaDirectDebitClient.tokenize(request: sepaDirectDebitRequest) { sepaDirectDebi
     // handle response
 }
 ```
+
+## Local Payment Methods
```
scannillo, line 170: It looks like these instructions are the same for both headers. Maybe we can instead consolidate them

jaxdesmarais:
> The only 🧶: should there be a few CHANGELOG callouts for this?

I am seeing it in the diff/files changed. Is there a specific callout you were referencing that is missing?

sshropshire:
> I am seeing it in the diff/files changed. Is there a specific callout you were referencing that is missing?

Oh wait nah nm. I missed those. 🖖 carry on

V6_MIGRATION.md
jaxdesmarais: Yeah we went back and fourth on this. At first we used a "Payment Flow" header but weren't sure if we needed to be more explicit about calling out LPMs and 3DS specifically. I guess we could also use the "Payment Flow" header and under it callout that it applies to 3DS and LPMs before going into the changes. Thoughts?

Sources/BraintreePaymentFlow/Public/BraintreePaymentFlow/BTPaymentFlowClient.h
jaxdesmarais: I'm probably misunderstanding the intent of this method but it looks like this was called in `safariViewControllerDidFinish` which was an internal implementation delegate that was a part of `SFSafariViewController`. Was this exposed to merchants and just not used in our demo app? I also didn't see it in the docs so may just be unsure of how this was used previously.

scannillo: Ah - I see!

I was confused because this is in a public header, but you're right - it looks like `BTPaymentFlowDriverDelegate` only exists for internal conformance, not merchant implementation.

It looks like we return cancellations as Error types in the `startPaymentFlow` completion. 

It looks like it was made public, because it is used in the public header for `BTPaymentFlowClient.startPaymentFlow()`.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## unreleased
 * Remove `iosBaseSDK`, `iosDeploymentTarget`, `iosIdentifierForVendor`, `deviceAppGeneratedPersistentUuid`, and `deviceScreenOrientation` from `BTAnalyticsMetadata`
+* Breaking Changes
+  * BraintreePaymentFlow
+    * Replaced deprecated `SFAuthenticationSession` with `ASWebAuthenticationSession` in payment flow
```
scannillo, line 7:
```suggestion
    * Replaced `SFSafariViewController` with `ASWebAuthenticationSession`
```

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
+    * Removed `BTViewControllerPresentingDelegate` protocol and correlating methods
+    * Removed `load` method from `BTPaymentFlowClient`
```
scannillo, line 9: Don't think this needs a CHANGELOG, since not a public method

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
...
+    * Removed `onPaymentCancel` and `returnURLScheme` methods from `BTPaymentFlowClientDelegate`
+    * Removed `canHandleAppSwitchReturnURL` method from `BTPaymentFlowRequestDelegate`
+    * Removed `viewControllerPresentingDelegate` property from `BTPaymentFlowClient`
```
scannillo, line 12: I think this is covered in the bullet above for "Removed `BTViewControllerPresentingDelegate` protocol"

scannillo, line 12: More CHANGELOG nit picks - do we need to add these, if they aren't intended for merchant use? Good opportunity to add no-docs to these?

Also another nitpick -  if we did end up keeping these, we can reduce some wordiness by doing syntax like "Remove `BTPaymentFlowRequestDelegate.canHandleAppSwitchReturn(_:)`"

V6_MIGRATION.md
scannillo: Gotcha - I personally prefer the later option. Or maybe even the header could be `## Local Payments & 3DS`?

Another thing - shall we add details that they can remove the [custom URL scheme](https://developer.paypal.com/braintree/docs/guides/client-sdk/setup/ios/v5#register-a-url-type) they previously added for BT?

Sources/BraintreePaymentFlow/Public/BraintreePaymentFlow/BTPaymentFlowClient.h
jaxdesmarais: The delegate on delegate on delegate pattern is confusing, so wanted to make sure I wasn't fully getting confused as well. 🙈

CHANGELOG.md
jaxdesmarais: Cool - wasn't sure since I know we've had merchants comment on compile times with the load method previously (I think we had a GH issue for the one in PayPalDriver at one point) if it was worth calling out.

CHANGELOG.md
jaxdesmarais: That's a good call - I think removing them and adding `nodoc` to the delegates and methods is a good call. Probably not worth calling out in the changelog as it may confuse merchants as to what they were/were not supposed to be using in the flow.

V6_MIGRATION.md
jaxdesmarais: Good call on the URL scheme - forgot about it because it's still used for Venmo so it still exists in the demo. Updated with the above changes here: https://github.com/braintree/braintree_ios/pull/952/commits/c78e3af922d931f8a56e5cce9a057f77b11eb9ad

CHANGELOG.md
jaxdesmarais: Updated here: https://github.com/braintree/braintree_ios/pull/952/commits/c78e3af922d931f8a56e5cce9a057f77b11eb9ad

scannillo (approved):
😍

PR #955, opened 2023-02-13
Update BTHTTPError Construction

jaxdesmarais:
### Summary of changes

When we converted 1:1 from Obj-C the error construction was a little odd. This majority of this PR is to update from the Obj-C implementation structure to using an `enum` like we do for the other errors in Swift modules.

- Move `httpErrorDomain`, `urlResponseKey`, and `jsonResponseBodyKey ` into `BTCoreConstants` - we can likely move this into the error enum directly and make the enum public if we want when the entire SDK is in Swift - these are used across other modules so they need to be somewhere public for now.
- Update `BTHTTPError` to match other error enum construction
- Remove `BTHTTPErrorCode`
- Update tests and implementations to use new `BTHTTPError` enum

### Checklist

- @jaxdesmarais


Sources/BraintreeCore/BTHTTPErrorCode.swift
```diff
@@ -1,36 +0,0 @@
-import Foundation
-
-@objc public enum BTHTTPErrorCode: Int {
```
jaxdesmarais, line 3: This was exposed publicly for use in Obj-C modules. We can add a changelog for removing this enum though if we feel it's ncessary, but probably should have been a `nodoc`.

Sources/BraintreeThreeDSecure/BTPaymentFlowClient+ThreeDSecure.m
```diff
@@ -127,12 +127,12 @@ - (void)performThreeDSecureLookup:(BTThreeDSecureRequest *)request
                     [self.apiClient sendAnalyticsEvent:@"ios.three-d-secure.lookup.network-connection.failure"];
                 }
                 // Provide more context for card validation error when status code 422
-                if ([error.domain isEqualToString:BTHTTPError.domain] &&
-                    error.code == BTHTTPErrorCodeClientError &&
-                    ((NSHTTPURLResponse *)error.userInfo[BTHTTPError.urlResponseKey]).statusCode == 422) {
+                if ([error.domain isEqualToString:BTCoreConstants.httpErrorDomain] &&
+                    error.code == 2 &&
```
jaxdesmarais, line 131: This doesn't feel _great_ but is what we will need to do until this module is in Swift. We cannot have access to the Swift enum in this module because the enum contains associated values that cannot be represented (or exposed) to Obj-C. Alternatively we could extract the client error into an`Int` property on `BTCoreConstants` and use that both here and in the enums `errorCode` for this value (similar to what we do for the domain). Thoughts?

Sources/BraintreeCore/BTCoreConstants.swift
```diff
@@ -17,4 +17,17 @@ import Foundation
...
+    /// The error domain for BTHTTP errors
+    public static let httpErrorDomain: String = "com.braintreepayments.BTHTTPErrorDomain"
+
+    /// :nodoc:
+    /// Key for userInfo dictionary that contains the NSHTTPURLResponse from server when it returns an HTTP error
+    public static let urlResponseKey: String = "com.braintreepayments.BTHTTPURLResponseKey"
+
+    /// :nodoc:
+    ///Key for userInfo dictionary that contains the BTJSON body of the HTTP error response
+    public static let jsonResponseBodyKey: String = "com.braintreepayments.BTHTTPJSONResponseBodyKey"
```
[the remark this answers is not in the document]
scannillo: This is a great callout. Can we leave a TODO with this info?

Sources/BraintreeCore/BTCoreConstants.swift
```diff
@@ -17,4 +17,17 @@ import Foundation
...
+    /// :nodoc:
+    /// The error domain for BTHTTP errors
+    public static let httpErrorDomain: String = "com.braintreepayments.BTHTTPErrorDomain"
+
+    /// :nodoc:
+    /// Key for userInfo dictionary that contains the NSHTTPURLResponse from server when it returns an HTTP error
+    public static let urlResponseKey: String = "com.braintreepayments.BTHTTPURLResponseKey"
+
+    /// :nodoc:
+    ///Key for userInfo dictionary that contains the BTJSON body of the HTTP error response
```
scannillo, line 30:
```suggestion
    /// Key for userInfo dictionary that contains the BTJSON body of the HTTP error response
```

Sources/BraintreeThreeDSecure/BTPaymentFlowClient+ThreeDSecure.m
scannillo: Maybe we can add a comment for readability that the 2 here matches with `BTHTTPError.errorCode.clientError`?

scannillo (approved):

Sources/BraintreeCore/BTHTTPError.swift
```diff
@@ -1,12 +1,96 @@
...
+        case .serverError(let errorDictionary):
+            return errorDictionary
+        case .missingBaseURL(let errorDictionary):
+            return ["Base URL constructed with invalid properties": errorDictionary]
+        case .rateLimitError(let errorDictionary):
+            return errorDictionary
+        case .dataNotFound:
+            return [NSLocalizedDescriptionKey: "Data unexpectedly nil."]
+        case .httpResponseInvalid:
+            return [NSLocalizedDescriptionKey : "Unable to create HTTPURLResponse from response data."]
```
sshropshire, line 87:
```suggestion
            return [NSLocalizedDescriptionKey: "Unable to create HTTPURLResponse from response data."]
```

sshropshire (approved):

PR #956, opened 2023-02-13
Convert Apple Pay to Swift

jaxdesmarais:
### Summary of changes

- Update `Braintree.podspec` and `Package.swift` now that module is in Swift
- Convert `BTApplePayError` to Swift
    -  Remove `BTApplePayErrorDomain` global constant
    - Remove `BTApplePayErrorType`
- Convert `BTApplePayCardNonce` to Swift
    - Make `BTApplePayCardNonce` initializer internal
    - Subclass `BTPaymentMethodNonce` vs duplicating the nonce types
- Convert `BTApplePayClient` to Swift
    - Rename `tokenizeApplePay` to `tokenize`
    - Add missing tests for new error cases 

### Checklist

- [x] Added a changelog entry

- @KunJeongPark @jaxdesmarais


sshropshire (approved):

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -0,0 +1,129 @@
...
+            paymentRequest.currencyCode = configuration.applePayCurrencyCode ?? ""
+            paymentRequest.merchantIdentifier = configuration.applePayMerchantIdentifier ?? ""
+            paymentRequest.supportedNetworks = configuration.applePaySupportedNetworks ?? []
+
+            self.completionHandler(onMainThreadWithPaymentRequest: paymentRequest, error: nil, completion: completion)
+        }
+    }
+
+    @objc(tokenizeApplePayPayment:completion:)
+    public func tokenize(_ payment: PKPayment, completion: @escaping (BTApplePayCardNonce?, Error?) -> Void) {
```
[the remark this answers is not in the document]
sshropshire: CHANGELOG entry is good enough for me. It may be worth calling out in a migration guide if `tokenize` becomes a common naming convention across modules within the iOS SDK. I'll defer to y'all though.

Braintree.xcodeproj/project.pbxproj
```diff
@@ -4704,6 +4677,8 @@
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SKIP_INSTALL = YES;
 				SUPPORTS_MACCATALYST = NO;
+				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
```
scannillo, line 4680: This change probably wasn't intentional?

Braintree.xcodeproj/project.pbxproj
```diff
@@ -5386,6 +5362,7 @@
 		A948D6E524FAC6C000F4F178 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
```
scannillo, line 5365: We shouldn't need this, [since Swift 5](https://www.swift.org/blog/abi-stability-and-apple/) (Xcode 10.2/ iOS 12.2) ships the Swift standard libraries with the OS, instead of needing it to be included with each app.

Package.swift
```diff
@@ -62,7 +62,7 @@ let package = Package(
         .target(
             name: "BraintreeApplePay",
             dependencies: ["BraintreeCore"],
-            publicHeadersPath: "Public"
+            path: "Sources/BraintreeApplePay"
```
scannillo, line 65: Also don't think we need to specify a path here since we're following the SPM default package structure `Sources/<TARGET_NAME>`

[Apple docs on path](https://developer.apple.com/documentation/packagedescription/target/path)

Braintree.xcodeproj/project.pbxproj
jaxdesmarais: This was generated automagically by Xcode when we added the Swift files. It looks like it exists for `BraintreeSEPADirectDebit`, `BraintreeAmericanExpress`, `BraintreePayPal`, `BraintreeCore`, `BraintreePayPalNativeCheckout`, and `BraintreeDataCollector` (so all of the Swift modules). Looking at the [Apple Swift Docs](https://github.com/apple/swift/blob/main/docs/OptimizationTips.rst#enabling-optimizations) for this it states:
> This is meant for normal development. It performs minimal optimizations and preserves all debug info.

If we want a different optimization level for `Debug` we should likely update the entire SDK.

It's worth noting our `Release` build is set to `-O`: 
> This is meant for most production code. The compiler performs aggressive optimizations that can drastically change the type and amount of emitted code. Debug information will be emitted but will be lossy.

Braintree.xcodeproj/project.pbxproj
jaxdesmarais: Same thing here, this was generated automagically by Xcode. This line is for `BraintreeApplePayTests` in `Debug` and the other is for `Release`. It looks like this also matches all of our Swift tests. We should likely change this across the board in a follow up PR if we don't want this setting as it is.

Package.swift
jaxdesmarais: Yeah it looks like some of our Swift packages contain and `path` (SEPA and NXO) and some do not. We can remove it here and should have a follow up to remove from SEPA and NXO as well!

Braintree.xcodeproj/project.pbxproj
scannillo: Interesting, thanks for the docs link! I wonder what kind of "debug info" specifically you miss out on in -0 mode.

Braintree.xcodeproj/project.pbxproj
scannillo: Got it - thanks for clarifying this is for the test targets

Sources/BraintreeApplePay/BTApplePayCardNonce.swift
```diff
@@ -0,0 +1,22 @@
...
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+/// Contains information about a tokenized Apple Pay card.
+@objcMembers public class BTApplePayCardNonce: BTPaymentMethodNonce {
+
+    /// The BIN data for the card number associated with this nonce.
+    let binData: BTBinData?
```
scannillo, line 11: Should this be public?

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -0,0 +1,129 @@
...
+import PassKit
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+/// Used to process Apple Pay payments
+@objcMembers public class BTApplePayClient: NSObject {
+
+    //MARK: - Internal Properties
```
scannillo, line 11:
```suggestion
    // MARK: - Internal Properties
```

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -0,0 +1,129 @@
...
+        apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
+            if let error {
+                self.apiClient.sendAnalyticsEvent("ios.apple-pay.error.configuration")
+                completion(nil, error)
+                return
+            }
+
+            if let configuration,
+               configuration.json?["applePay"]["status"].isString == false ||
+               configuration.json?["applePay"]["status"].asString() == "off" {
```
scannillo, line 65: Can we re-use [this logic](https://github.com/braintree/braintree_ios/blob/dd0cbeae0b48606885ed3684b7463ad799d8ee13/Sources/BraintreeCore/BTConfiguration%2BExtensions.swift#L57-L60) for this check?

Sources/BraintreeApplePay/BTApplePayCardNonce.swift
jaxdesmarais: Good catch, yes it should

Sources/BraintreeApplePay/BTApplePayClient.swift
jaxdesmarais: Yep, will update this. I am not sure why it was duplicated in Obj-C in the [configuration here](https://github.com/braintree/braintree_ios/blob/230d8f0fd7dbdfc3fd9d9230e9472f3b705e3669/Sources/BraintreeApplePay/BTConfiguration%2BApplePay.m#L9-L12) and the [implementation here](https://github.com/braintree/braintree_ios/blob/230d8f0fd7dbdfc3fd9d9230e9472f3b705e3669/Sources/BraintreeApplePay/BTApplePayClient.m#L115-L116). Will update this as well as move the Apple Pay configuration extension into the Apple Pay module as we have with the other Swift modules.

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -0,0 +1,129 @@
...
+
+                completion(applePayNonce, nil)
+                self.apiClient.sendAnalyticsEvent("ios.apple-pay.success")
+            }
+        }
+    }
+
+    // MARK: - Internal Methods
+
+    func parametersForPaymentToken(token: PKPaymentToken) -> [String: Any?] {
```
scannillo, line 111: Nitpick - I think we know all the below 4 values below will be Strings, so we could do ` -> [String: String]`. Might get us a tiny improvement in type checking type if we can avoid `Any`.

Sources/BraintreeApplePay/BTApplePayCardNonce.swift
scannillo: Also, looks like BTBinData was [non-optional](https://github.com/braintree/braintree_ios/blob/dd0cbeae0b48606885ed3684b7463ad799d8ee13/Sources/BraintreeApplePay/Public/BraintreeApplePay/BTApplePayCardNonce.h#L1) before

scannillo (approved):
🚀 

A few small comments, but otherwise LGTM!

Sources/BraintreeApplePay/BTApplePayClient.swift
jaxdesmarais: The `token.paymentMethod.network` is a `PKPaymentNetwork` and not a `String` so we need to use `[String: Any?]` 😞

Sources/BraintreeApplePay/BTApplePayCardNonce.swift
jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/956/commits/c71b1a7f3bdffadddab2ec3a177213c102b808c5

KunJeongPark: why does it need to be public? for testing?

Sources/BraintreeApplePay/BTConfiguration+ApplePay.swift
```diff
@@ -0,0 +1,60 @@
...
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+public extension BTConfiguration {
+
+    /// Indicates whether Apple Pay is enabled for your merchant account.
+    var isApplePayEnabled: Bool {
+        guard let applePayConfiguration: BTJSON = json?["applePay"] else { return false }
+        return applePayConfiguration["status"].isString && applePayConfiguration["status"].asString() != "off"
```
KunJeongPark, line 13: This is much cleaner way!

Sources/BraintreeApplePay/BTApplePayCardNonce.swift
KunJeongPark: Jax, didn't we decide to make this optional because it objective-c, this property could be nil?

jaxdesmarais: That's a great question! It needs to be public so that merchants can use it in their integration. In the link Sammy posted [above](https://github.com/braintree/braintree_ios/pull/956#discussion_r1109889198) in Obj-C we declare this files as "assume nonnull". This means that in Obj-C we were assuming we'd never get a null object back here. This is a different pattern from the other nonces in Obj-C [here is PayPal as an example](https://github.com/braintree/braintree_ios/blob/5.x/Sources/BraintreePayPal/Public/BraintreePayPal/BTPayPalAccountNonce.h) where each of the properties are `nullable`.

KunJeongPark: Thanks Jax!

PR #959, opened 2023-02-17
Audit Apple Pay Analytics Events

jaxdesmarais:
### Summary of changes

These changes are going into the fpti-feature branch. Which will get merged in once we audit all modules and make the switch to the FPTI endpoints

- Update ApplePay analytics events to prepare for switch to FPTI
- Update tests

### Checklist

- @scannillo @KunJeongPark @jaxdesmarais

### DIFF
--- a/Sources/BraintreeApplePay/BTApplePayAnalytics.swift
+++ b/Sources/BraintreeApplePay/BTApplePayAnalytics.swift
@@ -1,5 +1,6 @@
 import Foundation
 
+// NEXT_MAJOR_VERSION: Move this class to an enum once the entire SDK is in Swift
 class BTApplePayAnalytics {
 
     // MARK: - Payment Request Events
--- a/Sources/BraintreeApplePay/BTApplePayAnalytics.swift
+++ b/Sources/BraintreeApplePay/BTApplePayAnalytics.swift
@@ -1,0 +1,17 @@
+import Foundation
+
+class BTApplePayAnalytics {
+
+    // MARK: - Payment Request Events
+
+    static let paymentRequestStarted = "apple-pay:payment-request:started"
+    static let paymentRequestFailed = "apple-pay:payment-request:failed"
+    static let paymentRequestSucceeded = "apple-pay:payment-request:succeeded"
+
+    // MARK: - Tokenize Request Events
+
+    static let tokenizeStarted = "apple-pay:tokenize:started"
+    static let tokenizeFailed = "apple-pay:tokenize:failed"
+    static let tokenizeNetworkConnectionLost = "apple-pay:tokenize:network-connection:failed"
+    static let tokenizeSucceeded = "apple-pay:tokenize:succeeded"
+}
--- a/Sources/BraintreeApplePay/BTApplePayClient.swift
+++ b/Sources/BraintreeApplePay/BTApplePayClient.swift
@@ -29,14 +29,12 @@ import BraintreeCore
         self.apiClient.sendAnalyticsEvent("apple-pay:payment-request:started")
         apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
             if let error {
-                self.apiClient.sendAnalyticsEvent("apple-pay:payment-request:failed")
-                self.completionHandler(onMainThreadWithPaymentRequest: nil, error: error, completion: completion)
+                self.paymentRequestCompletionWithAnalytics(for: .failure, error: error, completion: completion)
                 return
             }
 
             guard let configuration, configuration.isApplePayEnabled else {
-                self.apiClient.sendAnalyticsEvent("apple-pay:payment-request:failed")
-                self.completionHandler(onMainThreadWithPaymentRequest: nil, error: BTApplePayError.unsupported, completion: completion)
+                self.paymentRequestCompletionWithAnalytics(for: .failure, error: BTApplePayError.unsupported, completion: completion)
                 return
             }
 
@@ -46,8 +44,7 @@ import BraintreeCore
             paymentRequest.merchantIdentifier = configuration.applePayMerchantIdentifier ?? ""
             paymentRequest.supportedNetworks = configuration.applePaySupportedNetworks ?? []
 
-            self.apiClient.sendAnalyticsEvent("apple-pay:payment-request:succeeded")
-            self.completionHandler(onMainThreadWithPaymentRequest: paymentRequest, error: nil, completion: completion)
+            self.paymentRequestCompletionWithAnalytics(for: .success, paymentRequest: paymentRequest, completion: completion)
         }
     }
 
@@ -57,14 +54,12 @@ import BraintreeCore
 
         apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
             if let error {
-                self.apiClient.sendAnalyticsEvent("apple-pay:tokenize:failed")
-                completion(nil, error)
+                self.completionWithAnalytics(for: .failure, error: error, completion: completion)
                 return
             }
 
             guard let configuration, configuration.isApplePayEnabled else {
-                self.apiClient.sendAnalyticsEvent("apple-pay:tokenize:failed")
-                completion(nil, BTApplePayError.unsupported)
+                self.completionWithAnalytics(for: .failure, error: BTApplePayError.unsupported, completion: completion)
                 return
             }
 
@@ -83,28 +78,23 @@ import BraintreeCore
                 if let error = error as NSError? {
                     if error.code == BTCoreConstants.networkConnectionLostCode {
                         self.apiClient.sendAnalyticsEvent("apple-pay:tokenize:network-connection:failed")
-                    } else {
-                        self.apiClient.sendAnalyticsEvent("apple-pay:tokenize:failed")
                     }
 
-                    completion(nil, error)
+                    self.completionWithAnalytics(for: .failure, error: error, completion: completion)
                     return
                 }
 
                 guard let body else {
-                    self.apiClient.sendAnalyticsEvent("apple-pay:tokenize:failed")
-                    completion(nil, BTApplePayError.noApplePayCardsReturned)
+                    self.completionWithAnalytics(for: .failure, error: BTApplePayError.noApplePayCardsReturned, completion: completion)
                     return
                 }
 
                 guard let applePayNonce: BTApplePayCardNonce = BTApplePayCardNonce(json: body["applePayCards"][0]) else {
-                    self.apiClient.sendAnalyticsEvent("apple-pay:tokenize:failed")
-                    completion(nil, BTApplePayError.failedToCreateNonce)
+                    self.completionWithAnalytics(for: .failure, error: BTApplePayError.failedToCreateNonce, completion: completion)
                     return
                 }
 
-                completion(applePayNonce, nil)
-                self.apiClient.sendAnalyticsEvent("apple-pay:tokenize:succeeded")
+                self.completionWithAnalytics(for: .failure, nonce: applePayNonce, completion: completion)
             }
         }
     }
@@ -120,11 +110,30 @@ import BraintreeCore
         ]
     }
 
-    func completionHandler(
-        onMainThreadWithPaymentRequest paymentRequest: PKPaymentRequest?,
-        error: Error?,
+    // MARK: - Analytics Completion Helpers
+
+    enum AnalyticsResult {
+        case success
+        case failure
+    }
+
+    func completionWithAnalytics(
+        for analyticsResult: AnalyticsResult,
+        nonce: BTApplePayCardNonce? = nil,
+        error: Error? = nil,
+        completion: @escaping (BTApplePayCardNonce?, Error?) -> Void
+    ) {
+        apiClient.sendAnalyticsEvent("apple-pay:tokenize:\(analyticsResult == .success ? "succeeded" : "failed")")
+        completion(nonce, error)
+    }
+
+    func paymentRequestCompletionWithAnalytics(
+        for analyticsResult: AnalyticsResult,
+        paymentRequest: PKPaymentRequest? = nil,
+        error: Error? = nil,
         completion: @escaping (PKPaymentRequest?, Error?) -> Void
     ) {
+        apiClient.sendAnalyticsEvent("apple-pay:payment-request:\(analyticsResult == .success ? "succeeded" : "failed")")
         DispatchQueue.main.async {
             completion(paymentRequest, error)
         }
--- a/Sources/BraintreeApplePay/BTApplePayClient.swift
+++ b/Sources/BraintreeApplePay/BTApplePayClient.swift
@@ -112,7 +112,7 @@ import BraintreeCore
 
     // MARK: - Analytics Completion Helpers
 
-    func completionWithAnalytics(
+    private func completionWithAnalytics(
         for analyticsResult: BTAnalyticsResult,
         nonce: BTApplePayCardNonce? = nil,
         error: Error? = nil,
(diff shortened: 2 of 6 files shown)


Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -80,26 +82,29 @@ import BraintreeCore
             self.apiClient.post("v1/payment_methods/apple_payment_tokens", parameters: parameters) { body, _, error in
                 if let error = error as NSError? {
                     if error.code == BTCoreConstants.networkConnectionLostCode {
-                        self.apiClient.sendAnalyticsEvent("ios.apple-pay.network-connection.failure")
+                        self.apiClient.sendAnalyticsEvent("apple-pay:tokenize:network-connection:failed")
+                    } else {
+                        self.apiClient.sendAnalyticsEvent("apple-pay:tokenize:failed")
```
scannillo, line 87: Thinking through this again, if we want to calculate conversion failure rates, we need to do:
```
total_failed_events / (total_start_events + total_failed_events) X 100
```

We know that our start event is `apple-pay:payment-request:started` (simple, since only triggered once). I think if we keep the code as-is, then we'll need to add up both `# of apple-pay:tokenize:failed` + `# of apple-pay:tokenize:network-connection:failed` to get `total_failed_events`.

So maybe we _do_ want to send `apple-pay:tokenize:network-connection:failed` alongside a `apple-pay:tokenize:failed`? We can treat `failed` as an umbrella for all failures. If we do this route - what we need to be sure of, is just documenting the behavior of the all-encompassing failed event, and making sure it is not fired more than once per error route.

Does that make sense?

Whatever we do, it should be consistent & documented so we know exactly how to calculate conversion later on.

scannillo: I actually think this was more in-line with our plan for [PPCP events](https://paypal.sharepoint.com/:x:/r/sites/BTSDK/_layouts/15/Doc.aspx?sourcedoc=%7BAAA7C353-5744-4F21-AC7D-DF85D8605817%7D&file=Northstar%20Analytics%20Events.xlsx&action=default&mobileredirect=true&cid=1b0afccd-68ab-448f-9760-60ed14df9811). Have a single/high-level "failed" event that gets fired for all failures, for the purpose of calculating conversion.

jaxdesmarais: I think that makes a ton of sense and is a great idea! Then we can calculate conversion for all failures (including network connection failures). That will also allow us to calculate the network connection failure rate as a percentage of the total failures, which I think would likely be more useful as well so we can really drill into that data alongside the success/failure data.

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -115,11 +110,30 @@ import BraintreeCore
...
 
-    func completionHandler(
-        onMainThreadWithPaymentRequest paymentRequest: PKPaymentRequest?,
-        error: Error?,
+    // MARK: - Analytics Completion Helpers
+
+    enum AnalyticsResult {
+        case success
+        case failure
+    }
```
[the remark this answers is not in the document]
sshropshire: I know at one point support and API support used the entire `ios:event-name:event-type` string while searching for solutions to merchant inbounds. This may no longer be relevant, but something to consider. An enum will be less error prone and having a core type shared across each module will make the code more DRY.

sshropshire (approved):

Sources/BraintreeApplePay/BTApplePayClient.swift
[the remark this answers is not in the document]
jaxdesmarais: Yeah, when this feature branch is merged (or about to be merged) we will need to let Sales and Support know about this change, both in formatting and simplifying the events. Overall it should tell a more holistic story of a merchants conversion rate vs now is much harder to reason about.

Agree on the `enum` being less error prone, I'll make that change and if folks disagree we can move it back out.

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -115,11 +110,25 @@ import BraintreeCore
...
-        error: Error?,
+    // MARK: - Analytics Completion Helpers
+
+    private func completionWithAnalytics(
+        for analyticsResult: BTAnalyticsResult,
+        nonce: BTApplePayCardNonce? = nil,
+        error: Error? = nil,
+        completion: @escaping (BTApplePayCardNonce?, Error?) -> Void
+    ) {
+        apiClient.sendAnalyticsEvent("apple-pay:tokenize:\(analyticsResult == .success ? "succeeded" : "failed")")
```
scannillo, line 121: I am in favor of having our analytic events be searchable as raw strings in the SDKs, instead of the concat logic making them harder to search / visualize.

Sources/BraintreeCore/BTAnalyticsService.swift
```diff
@@ -1,5 +1,11 @@
 import Foundation
 
+/// :nodoc: Enum used for status to send to analytics event
+@objc public enum BTAnalyticsResult: Int {
```
scannillo, line 4: We don't know if this will be applicable to all payment methods. For example, some have cancel cases, some don't. Also there might be one off completion types like `paypalShippingCallback` called or something.

So maybe a global enum for all payment method modules won't fit our needs?

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -115,11 +110,25 @@ import BraintreeCore
...
+        completion(nonce, error)
+    }
+
+    private func paymentRequestCompletionWithAnalytics(
```
scannillo, line 125: I also think we should sync with PPCP on our naming convention & pattern for returning these compeltions & analytic events, whatever we do land on.

Sources/BraintreeApplePay/BTApplePayClient.swift
scannillo: Also, if we stick _did_ with the enum - we could give the cases associated String values which would simplify the ternary operator logic, as well as make sure we have consistency with the naming of "succeeded" "failed" "canceled" spelling and tense.

jaxdesmarais: I think that's fine if we feel it'd be more helpful to have it be more searchable as the raw string. I think we can still use 1 method to capture `success/failure` vs having 2. Something like this in this method would achieve both:
```
if analyticsResult == .success {
    apiClient.sendAnalyticsEvent("apple-pay:tokenize:succeeded")
} else {
    apiClient.sendAnalyticsEvent("apple-failed")
}

completion(nonce, error)
```

Agree that having the `stringValue` on the enum case would be cleaner if we want to go that route. I personally feel that it's more important to ensure we are using the same tense across all modules over the search-ability for the raw string, but happy to do that as well if folks feel strongly about that aspect. With the string values this could be simplified to:
```
apiClient.sendAnalyticsEvent("apple-pay:tokenize:\(analyticsResult.stringValue)")
```

Thoughts?

Sources/BraintreeApplePay/BTApplePayClient.swift
jaxdesmarais: Yeah this one is a bit of an edge case because there are 2 public methods in apple pay returning different results (one a nonce + error and the other a paymentRequest + error). I think we should align them for completions that don't notify a delegate but _do_ think the notify language makes more sense for methods that do bubble up to a delegate as it makes it more clear what is actually happening (notifying a delegate vs just calling a completion).

Sources/BraintreeApplePay/BTApplePayClient.swift
KunJeongPark: I like that idea of using associated String values. So are we adding a third option of "canceled"? 
That option is only applicable for some cases, not most, right?

Sources/BraintreeCore/BTAnalyticsService.swift
jaxdesmarais: I think it could make sense for the tense of the common cases (started/network connection lost/failed/succeeded) but then we could still handle edge cases on a case by case basis. Again, if we feel strongly against this, we can remove this but thought it worth discussing.

KunJeongPark: whatever we do, I'm not a fan of associated values

jaxdesmarais: Is there a specific reason you are not a fan of associated values, @KunJeongPark?

scannillo: +1 - associated values are really powerful, curious too why you say that?

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -115,11 +110,25 @@ import BraintreeCore
...
-        onMainThreadWithPaymentRequest paymentRequest: PKPaymentRequest?,
-        error: Error?,
+    // MARK: - Analytics Completion Helpers
+
+    private func completionWithAnalytics(
+        for analyticsResult: BTAnalyticsResult,
+        nonce: BTApplePayCardNonce? = nil,
+        error: Error? = nil,
+        completion: @escaping (BTApplePayCardNonce?, Error?) -> Void
+    ) {
```
scannillo, line 120: I think these helper methods are a bit over-loaded, especially since we expect either only an error or a nonce, but never both.

Maybe a good compromise is reverting back to using two separate methods for calling success vs fail cases, but then also performing the concat using the associated string value from an enum.

jaxdesmarais: That sounds good to me. Are we tied to the names `notifySuccess/notifyFailure` or can we do something like `failureCompletionWithAnalytics/successCompletionWithAnalytics`? Reason being that notify still feels tied to something like a delegate to me. I’m fine either way, just wanted to give a why. 🙂

For the enum are you thinking doing `.stringValue` or actually passing the whole string in like `.success(apple-pay:tokenize:succeeded)`. Just want to make sure we are on the same page since we’ve talked about both here!

I can make the updates tomorrow in any case.

Sources/BraintreeCore/BTAnalyticsService.swift
sshropshire: It depends for me but there are definitely some cases where associated values get [weird](https://stackoverflow.com/a/32154586). A lot of gymnastics is required to work with them when outside of a `switch` statement.

Sources/BraintreeApplePay/BTApplePayAnalytics.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+class BTApplePayAnalytics {
```
jaxdesmarais, line 3: We _can_ use an enum here instead like:
```
enum BTApplePayAnalytics: String {
    case paymentRequestStarted = "apple-pay:payment-request:started"
    ...
}
```

Then at the callsite we'd do `apiClient.sendAnalyticsEvent(BTApplePayAnalytics.paymentRequestFailed.rawValue)`

The advantage to a class with static properties is it's easy to bridge to our Obj-C modules vs enums (which need to be Int type in Obj-C). We would then need a separate stringValue - which may muddy up the integration pattern an hopping between Obj-C and Swift.

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -114,14 +119,4 @@ import BraintreeCore
...
     }
-
-    func completionHandler(
-        onMainThreadWithPaymentRequest paymentRequest: PKPaymentRequest?,
-        error: Error?,
-        completion: @escaping (PKPaymentRequest?, Error?) -> Void
-    ) {
-        DispatchQueue.main.async {
-            completion(paymentRequest, error)
-        }
```
jaxdesmarais, line 125: Doing some testing, we don't seem to need to call just this completion from the main queue. Ran a series of tests on device and simulator and `paymentRequest` is not throwing warnings. Should be good to remove this and call the completion the same way in both methods.

Sources/BraintreeApplePay/BTApplePayClient.swift
jaxdesmarais: Implemented what we had discussed here: https://github.com/braintree/braintree_ios/pull/960#discussion_r1122402533

If we like this pattern, we can make this update in the other 2 PRs, but we didn't want to make the change in all 3 places since the discussion was still ongoing. This PR now reflects the changes discussed above and added a test file for our event names.

Sources/BraintreeCore/BTAnalyticsService.swift
KunJeongPark: Yes, for someone seeing it for first time, it can be a bit confusing. You can have different data types as associated values, having to click in to see what type it is. You can't iterate over all cases (which we don't need in our case)  and I think changes are sometimes harder to reason about.

jaxdesmarais: > It depends for me but there are definitely some cases where associated values get [weird](https://stackoverflow.com/a/32154586). A lot of gymnastics is required to work with them when outside of a switch statement.

This is an odd example for sure - typically you'd do something like the function `doToLine` in an extension on the enum (in later versions of Swift).

> You can't iterate over all cases (which we don't need in our case)

🤔  I am not sure this is true. This works as expected in a playground - it just cannot automatically synthesize `allCases` so you need to provide it:
```
enum Cases: CaseIterable {
    static var allCases: [Cases] {
        [.one(""), .two(""), .three("")]
    }

    case one(String)
    case two(String)
    case three(String)
}

Cases.allCases.forEach { print($0) }
```

Agree they can get complicated once you start passing in tuples, a ton of types, etc that aren't easily able to be reasoned about when looking through the callsite code (vs simple types - String, Error, Int, etc). Also agree they can get weird, but I do think that they are a powerful language feature when used correctly and don't want to discount them entirely in our codebases when they are the right language tool to use for a given case.

Sources/BraintreeApplePay/BTApplePayAnalytics.swift
scannillo: Great callout! Should we leave a NEXT_MAJOR_VERSION note for when we're entirely in Swift to use the enum type?

UnitTests/BraintreeApplePayTests/BTApplePay_Tests.swift
```diff
@@ -271,7 +271,7 @@ class BTApplePay_Tests: XCTestCase {
         }
         waitForExpectations(timeout: 2)
         
-        XCTAssertTrue(mockClient.postedAnalyticsEvents.contains("ios.apple-pay.network-connection.failure"))
+        XCTAssertTrue(mockClient.postedAnalyticsEvents.contains("apple-pay:tokenize:network-connection:failed"))
```
scannillo, line 274: Lol this lonely unit test reveals how much we aren't unit testing analytics. Shall we make a JIRA for this as a 🏈 punt ?

Another thing - we sort of inconsistently test analytics. Sometimes we have separate tests, other times we bake analytic event checks into existing tests. Would be nice to standardize our testing strategy.

Sources/BraintreeApplePay/BTApplePayAnalytics.swift
jaxdesmarais: Good call - I'll add a note for this so we remember to re-evaluate once we are all Swift!

UnitTests/BraintreeApplePayTests/BTApplePay_Tests.swift
jaxdesmarais: 😢 yeah I noticed that too. We should come up with a strategy then audit all of the tests. Having all of the analytics in 1 file will have the advantage of being able to quickly see test coverage for out events, which will be nice! Spike ticket to determine our pattern is: DTBTSDK-2441

PR #964, opened 2023-03-02
Xcode 14.3 Import Error

jaxdesmarais:
### Summary of changes

- Doing some testing in Xcode 14.3 we see the error `Fixes error `@objcMembers attribute used without importing module 'Foundation'` in Xcode 14.3+` - we need to import Foundation in these places to avoid this error.

From [release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-14_3-release-notes):
> An explicit import of Foundation is now required if APIs from Foundation are used in a manifest. 

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais


sshropshire (approved):

scannillo (approved):
Great catch!

PR #969, opened 2023-03-07
Convert VenmoRequest, BTVenmoAccountNonce, BTVenmoAppSwitchReturnURL, BTVenmoAppSwitchRequestURL and  to Swift

jaxdesmarais:
These changes are going into the `convert-venmo` feature branch. We **expect** SPM to fail as the target contains a mixed Obj-C and Swift module. When the remaining conversion is complete SPM should pass as expected and we will open a PR to merge the feature branch into main.

### Summary of changes

- Convert `BTVenmoRequest` to Swift
- Convert `BTVenmoAppSwitchReturnURL` to Swift
- Convert `BTVenmoAppSwitchRequestURL` to Swift
- Convert `BTVenmoAccountNonce` to Swift
- Create `BTVenmoError` in Swift

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais


Sources/BraintreeVenmo/BTVenmoAppSwitchRequestURL.swift
```diff
@@ -0,0 +1,104 @@
...
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+// TODO: Entire class be internal and likely a struct once rest of Venmo is in Swift
+// TODO: methods can all be non-static likely when we convert the rest of Venmo to Swift
+@objcMembers public class BTVenmoAppSwitchRequestURL: NSObject {
+
+    static let kXCallbackTemplate: String = "scheme://x-callback-url/path"
+    static let kVenmoScheme: String = "com.venmo.touch.v2"
```
sshropshire, line 12:
```suggestion
    static let xCallbackTemplate: String = "scheme://x-callback-url/path"
    static let venmoScheme: String = "com.venmo.touch.v2"
```
🧶 : Swifty

Sources/BraintreeVenmo/BTVenmoAppSwitchRequestURL.swift
```diff
@@ -0,0 +1,104 @@
...
+                let errorReturnURL,
+                let cancelReturnURL,
+                let accessToken,
+                let metadata,
+                let bundleDisplayName,
+                let environment,
+                let merchantID
+        else {
+            return nil
+        }
```
sshropshire, line 56:
```suggestion
        guard [
          successReturnURL,
          errorReturnURL,
          cancelReturnURL,
          accessToken,
          metadata,
          bundleDisplayName,
          environment,
          merchantID
        ].allSatisfy({ $0 != nil }) else {
            return nil
        }
```
Take or leave. It is cool that `guard let` offers the same functionality. This alternate method would prevent having to create shadow variables. I've never used `allSatisfy` before either seems kinda cool.

Sources/BraintreeVenmo/BTVenmoAppSwitchRequestURL.swift
```diff
@@ -0,0 +1,104 @@
...
+
+        let venmoMetadata: [String: String?] = [
```
sshropshire, line 58: 🧶 : If this can be `[String: String]` that'd be clean. Understand though if there are potentially optional values.

sshropshire (approved):
🌊

Sources/BraintreeVenmo/BTVenmoAppSwitchRequestURL.swift
jaxdesmarais: We use all of these later on in `appSwitchParameters` which is why I guard let-ed them - but `allSatisfy` is dope and I am sure we can find a use for it at some point!

Sources/BraintreeVenmo/BTVenmoAppSwitchRequestURL.swift
jaxdesmarais: You're right - this one can be `[String: String]` we can unwrap all of these at this point. Good catch!

Sources/BraintreeVenmo/BTVenmoAccountNonce.swift
```diff
@@ -0,0 +1,65 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
```
scannillo, line 5: I'm really gonna miss this:

```
// MARK: - Swift File Imports for Package Managers
#if __has_include(<Braintree/Braintree-Swift.h>) // Cocoapods-generated Swift Header
#import <Braintree/Braintree-Swift.h>

#elif SWIFT_PACKAGE                              // SPM
/* Use @import for SPM support
 * See https://forums.swift.org/t/using-a-swift-package-in-a-mixed-swift-and-objective-c-project/27348
 */
@import BraintreeCore;

#elif __has_include("Braintree-Swift.h")         // CocoaPods for ReactNative
/* Use quoted style when importing Swift headers for ReactNative support
 * See https://github.com/braintree/braintree_ios/issues/671
 */
#import "Braintree-Swift.h"

#else // Carthage or Local Builds
#import <BraintreeCore/BraintreeCore-Swift.h>
#endif
```

Sources/BraintreeVenmo/BTVenmoAccountNonce.swift
```diff
@@ -0,0 +1,65 @@
...
+    /// The phone number associated with the Venmo account
+    public var phoneNumber: String?
+
+    /// The username associated with the Venmo account
+    public var username: String?
+
+    // MARK: - Initializers
+
+    init(with nonce: String, username: String, isDefault: Bool) {
+        self.username = username
```
scannillo, line 33: I think we also have to set `isDefault` and `nonce` properties

Sources/BraintreeVenmo/BTVenmoAppSwitchRequestURL.swift
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        var components = appSwitchBaseURLComponents()
+        components?.percentEncodedQuery = BTURLUtils.queryString(from: appSwitchParameters as NSDictionary)
+        return components?.url
+    }
+
+    // MARK: - Internal Helper Methods
+
+    static func returnURL(with scheme: String?, result: String) -> URL? {
```
scannillo, line 91: We can probably update this to require a `scheme` and return non-optional too. I think we could do it now and move the nil check for `scheme` to the `init`, or lump it in with the TODO you have above the `init`.

Sources/BraintreeVenmo/BTVenmoAppSwitchRequestURL.swift
```diff
@@ -0,0 +1,104 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+// TODO: Entire class be internal and likely a struct once rest of Venmo is in Swift
+// TODO: methods can all be non-static likely when we convert the rest of Venmo to Swift
+@objcMembers public class BTVenmoAppSwitchRequestURL: NSObject {
```
scannillo, line 9: Total 🏈 question - we use the phrases "redirect URL", "request URL", and "return URL" mixed throughout the SDK. Would be good to only use 2 and clearly define which is used for switching TO the app and which is for switching FROM the app.

Sources/BraintreeVenmo/BTVenmoRequest.swift
```diff
@@ -0,0 +1,54 @@
...
+    public var vault: Bool = false
+
+    /// If set to `.multiUse`, the Venmo payment will be authorized for future payments and can be vaulted.
+    ///  If set to `.singleUse`, the Venmo payment will be authorized for a one-time payment and cannot be vaulted.
+    public var paymentMethodUsage: BTVenmoPaymentMethodUsage
+
+    /// Optional. The business name that will be displayed in the Venmo app payment approval screen. Only used by merchants onboarded as PayFast channel partners.
+    public var displayName: String?
+
+    // TODO: move into enum once VenmoClient is in Swift
```
scannillo, line 36: 👏 Awesome comments throughout this PR!

scannillo (approved):

Sources/BraintreeVenmo/BTVenmoAppSwitchRequestURL.swift
jaxdesmarais: Yeah, updating the scheme to be non-optional makes sense. I'll make this change. For the return, `URLComponents` [url](https://developer.apple.com/documentation/foundation/urlcomponents/1780032-url) is optional so I think we want to keep this as an optional `URL` return.

Sources/BraintreeVenmo/BTVenmoAppSwitchRequestURL.swift
jaxdesmarais: Yeah, these will be internal once the client is in Swift, so updating the names to be more clear makes sense to me. What names do you prefer?

scannillo: Chatted IRL - going forward with "redirectURL" to represent the URL used to switch _to_ the checkout flow from a merchant app, and "returnURL" when coming back to the merchant app _from_ the web view.

jaxdesmarais: Note for self to update the error case `BTVenmoErrorTypeInvalidRequestURL` in follow up client PR

scannillo (approved):

Sources/BraintreeVenmo/BTVenmoRequest.swift
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    // MARK: - Initializer
+
+    /// Initialize a Venmo request with a payment method usage.
+    /// - Parameter paymentMethodUsage: a `BTVenmoPaymentMethodUsage`
+    @objc(initWithPaymentMethodUsage:)
+    public init(paymentMethodUsage: BTVenmoPaymentMethodUsage) {
+        self.paymentMethodUsage = paymentMethodUsage
```
KunJeongPark, line 52: I'm comparing with the objective-c initializer and I am wondering why we needed checking "if (self = [super init])" in objective-c?

jaxdesmarais: That line in Objective-C is just checking that `self` is not nil

PR #970, opened 2023-03-14
Convert BTVenmoClient to Swift

jaxdesmarais:
### Summary of changes

- Remove `BTVenmoErrorDomain` global constant
- Renamed `BTVenmoClient.tokenizeVenmoAccount(with:completion:)` to `BTVenmoClient.tokenize(_:completion:)`
- Renamed `BTVenmoClient.isiOSAppAvailableForAppSwitch()` to `BTVenmoClient.isVenmoAppInstalled()`
- Address TODOs from initial conversion PR

### Checklist

- [x] Added a changelog entry

- @scannillo @jaxdesmarais


Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -0,0 +1,357 @@
...
+
+            self.apiClient.sendAnalyticsEvent("ios.pay-with-venmo.vault.success")
+            self.appSwitchCompletion(venmoAccountNonce, venmoAccountJSON.asError())
+            return
+        }
+    }
+
+    // MARK: - App Switch Methods
+
+    func verifyAppSwitch(with configuration: BTConfiguration, error: inout Error?) -> Bool {
```
sshropshire, line 323: 🧶 : is the `inout` still necessary? A `try / catch` could be more expressive.

sshropshire (approved):

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -0,0 +1,357 @@
...
+    public func isVenmoAppInstalled() -> Bool {
+        if let _ = application as? UIApplication {
+            guard let appSwitchURL = BTVenmoAppSwitchRedirectURL().baseAppSwitchURL else {
+                return false
+            }
+
+            return UIApplication.shared.canOpenURL(appSwitchURL)
+        } else {
+            return application.canOpenURL(BTVenmoAppSwitchRedirectURL().baseAppSwitchURL ?? URL(string: "")!)
+        }
```
scannillo, line 158: This is hard to read, IMO - as well as the way we inject the `application` mock. 

I think mocking `UIApplication` is a good case for using `OCMock`, and re-writing[ `FakeApplication`](https://github.com/braintree/braintree_ios/blob/master/UnitTests/BraintreeTestShared/FakeApplication.swift) in ObjC. This would allow us to remove the `AnyObject` type on line 20 we have now to work around Swift's mocking limitations.

Eventually, this would allow us to remove testing details from the source code:
```
        guard let appSwitchURL = BTVenmoAppSwitchRedirectURL().baseAppSwitchURL else {
            return false
        }
        
        return application.canOpenURL(appSwitchURL)
```

Definitely a punt 🏈 but wanted to leave a note. One example of this is in [BTMockOpenURLContext](https://github.com/braintree/braintree_ios/blob/master/UnitTests/BraintreeCoreTests/Helpers/BTMockOpenURLContext.h).

CasandraDevivaW1i: It appears that the conversation is related to improving the readability and maintainability of the codebase, specifically in relation to mocking the UIApplication and the BTVenmoAppSwitchRedirectURL. The proposed solution involves using OCMock and rewriting the FakeApplication class in Objective-C to remove Swift's mocking limitations, which would allow for the removal of testing details from the source code. This suggestion has been noted for a follow-up PR, with BTMockOpenURLContext given as an example.

KayceKomarekFUJ:
Great job! The changes look good. Don't forget to update the documentation and release notes as well. Also, it might be helpful to add a brief description of each change in the changelog entry.

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: Done! https://github.com/braintree/braintree_ios/pull/970/commits/8e8bc6fcf0a99a23f7bb5a18e78ba84217ae7d9b

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: Totally agree - this is much harder to reason about in Swift. It wasn't great in Obj-C either, but was obfuscated by the... Objective-C-ness. 

It looks like there could be some ways around this in Swift. I found [this blog post](http://swiftyjimmy.com/mock-uiapplication-swift/) but may also be worth a punt to explore if there is a better way to do this so we can avoid these weird switches in the implementation.

scannillo: Ah yeh, that's like what we're doing on PPCP for `URLSession`. Still a pain to need to make those protocols, but such is life! Swift can't be better at _everything_ :D 

Ticket for improving this later - DTBTSDK-2517

V6_MIGRATION.md
```diff
@@ -52,12 +52,16 @@ The possible values for `BTVenmoPaymentMethodUsage` include:
 * `.multiUse` - the Venmo payment will be authorized for future payments and can be vaulted.
 * `.singleUse` - the Venmo payment will be authorized for a one-time payment and cannot be vaulted.
 
+`BTVenmoClient.tokenizeVenmoAccount(with: BTVenmoRequest, completion: @escaping (BTVenmoAccountNonce?, Error?) -> Void)` has been renamed to `BTVenmoClient.tokenize(_ request: BTVenmoRequest, completion: @escaping (BTVenmoAccountNonce?, Error?) -> Void)`
+
```
scannillo, line 56: Nitpick - do we want to use the shorthand syntax (like we looked at at the top of Apple's dev docs) here instead?

I think it would be:
```suggestion
`BTVenmoClient.tokenizeVenmoAccount(with:completion:)` has been renamed to `BTVenmoClient.tokenize(_:completion:)`

```

scannillo (approved):

V6_MIGRATION.md
jaxdesmarais: It looks like we've followed the pattern in this PR at least for the rest of the V6_MIGRATION guide. I'm going to 🏈  for now and add a ticket for us to decide and update everything in here to match the decided pattern. Ticket here: DTBTSDK-2518. Also included an audit of the CHANGELOG in the ticket since it's probably be good to make sure we are consistent there for v6 as well.

PR #975, opened 2023-03-15
Add Explicit Async/Await Support for Swift Methods

jaxdesmarais:
### Summary of changes

With Objective-C, if we follow a [certain naming convention for methods](https://developer.apple.com/documentation/swift/calling-objective-c-apis-asynchronously), we get “for free” an async/await equivalent when calling the methods in Swift. We need to add this support back in explicitly as the modules are converted to Swift. This gets all of the current Swift modules back to the baseline, but we will need to ensure that as we convert the remaining modules we are adding this explicit support.

Add async/await support to the following methods:
- `BTAmericanExpressClient.getRewardsBalance()`
- `BTApplePayClient.paymentRequest()` and `BTApplePayClient.tokenize()`
- `BTDataCollector.collectDeviceData()`
- Both the Vault and Checkout method for `BTPayPalClient.tokenize()`
- Both the Vault and Checkout method for `BTPayPalNativeCheckoutClient.tokenize()`
- `BTSEPADirectDebitClient.tokenize()`

Renamed the following methods (to match the style used in the rest of the v6 SDK):
- `BTAmericanExpressClient. getRewardsBalance(forNonce: currencyIsoCode:completion:)` to `BTAmericanExpressClient. getRewardsBalance(for: currencyISOCode:completion:)`
- `BTSEPADirectDebitClient.tokenize(request:completion)` to `BTSEPADirectDebitClient.tokenize(_:completion:)`

Other work done in this PR:
- Added missing or update doc strings in a few places
- Added missing `MARK` in SEPA file

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais


jaxdesmarais:
Note: Venmo async/await support was added in [this commit on the client conversion PR](https://github.com/braintree/braintree_ios/pull/970/commits/33e10b2959e79c69329b4941cda682dd2379cd3d). Moving forward once we have the baseline in this PR, we should be sure to add these methods during the conversion process.

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -24,6 +24,9 @@ import BraintreeCore
 
     // MARK: - Public Methods
 
+    /// Creates a `PKPaymentRequest` with values from your Braintree Apple Pay configuration.
```
jaxdesmarais, line 27: We were missing docstring for all the public methods in here so added them back in

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -72,6 +72,30 @@ import BraintreeDataCollector
...
+    /// server when this method completes without any additional user interaction.
+    ///
+    /// On success, you will receive an instance of `BTPayPalAccountNonce`; on failure or user cancelation you will receive an error.
+    /// If the user cancels out of the flow, the error code will be `.canceled`.
+    ///
+    /// - Parameter request: A `BTPayPalVaultRequest`
+    /// - Returns: A `BTPayPalAccountNonce` if successful
+    /// - Throws: An `Error` describing the failure
+    @objc(tokenizeWithVaultRequest:completionHandler:)
+    public func tokenize(_ request: BTPayPalVaultRequest) async throws -> BTPayPalAccountNonce {
```
jaxdesmarais, line 87: Testing these methods made me realize there are some holes in our testing strategy for PayPal and we currently do some weird things to test tokenize. I created ticket DTBTSDK-2512 to revisit our PayPal testing strategy as a whole and maybe use something like the MockASWebSession that we use in SEPA tests to test this more easily. Seems like something we can 🏈  for now though.

For the un-tested methods I tested them in the demo app to ensure they are working as expected.

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
```diff
@@ -40,11 +41,34 @@ import PayPalCheckout
...
+    ///
+    /// On success, you will receive an instance of `BTPayPalNativeCheckoutAccountNonce`.
+    /// On failure or user cancelation you will receive an error. If the user cancels
+    /// out of the flow, the error code will equal `BTPayPalNativeError.canceled.rawValue`.
+    ///
+    /// - Parameter request: A `BTPayPalNativeCheckoutRequest`
+    /// - Returns: A `BTPayPalNativeCheckoutAccountNonce` if successful
+    /// - Throws: An `Error` describing the failure
+    @objc(tokenizeWithNativeCheckoutRequest:completionHandler:)
+    public func tokenize(_ request: BTPayPalNativeCheckoutRequest) async throws -> BTPayPalNativeCheckoutAccountNonce {
```
jaxdesmarais, line 54: Same thing here with future testing being captured in DTBTSDK-2512 - we currently only test 1 thing (an error) in BTPayPalNativeCheckoutClient_Tests. 😱 Testing this may be harder/impossible due to not being able to easily mock NXO methods, so this may be the best we can do, but worth revisiting.

Also tested this in the demo app though to ensure it's working as expected since we can't easily unit test this.

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
```diff
@@ -48,4 +49,23 @@ import BraintreeCore
...
         }
     }
+
+    ///  Gets the rewards balance associated with a Braintree nonce. Only for American Express cards.
+    /// - Parameters:
+    ///   - nonce: A nonce representing a card that will be used to look up the rewards balance.
+    ///   - currencyIsoCode: The currencyIsoCode to use. Example: 'USD'
+    /// - Returns: A `BTAmericanExpressRewardsBalance` object with information about the rewards balance
+    /// - Throws: An `Error` describing the failure
+    @objc(getRewardsBalanceForNonce:currencyIsoCode:completionHandler:)
```
scannillo, line 59: I don't think we want to expose both of these to ObjC - just the non-async await one

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -49,6 +52,27 @@ import BraintreeCore
         }
     }
 
+    /// Creates a `PKPaymentRequest` with values from your Braintree Apple Pay configuration.
+    /// It populates the following values of `PKPaymentRequest`: `countryCode`, `currencyCode`, `merchantIdentifier`, `supportedNetworks`.
+    /// - Returns: A `PKPaymentRequest`
+    /// - Throws: An `Error` describing the failure
```
scannillo, line 58: 🤔 Stinks to have to duplicate and maintain 2 docstrings for each method now. More room for error.

Off the top of my head I don't have ideas for remedying this, but we can think on it!

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -104,6 +128,23 @@ import BraintreeCore
         }
     }
 
+    /// Tokenizes an Apple Pay payment.
+    /// - Parameter payment: A `PKPayment` instance, typically obtained by presenting a `PKPaymentAuthorizationViewController`
+    /// - Returns: A `BTApplePayCardNonce`
+    /// - Throws: An `Error` describing the failure
+    @objc(tokenizeApplePayPayment:completionHandler:)
```
scannillo, line 135: Same comment throughout - re not exposing async style to ObjC

Sources/BraintreeDataCollector/BTDataCollector.swift
```diff
@@ -36,7 +36,7 @@ import BraintreeCore
     ///  We recommend that you call this method as early as possible, e.g. at app launch. If that's too early,
     ///  calling it when the customer initiates checkout is also fine.
     ///  Use the return value on your server, e.g. with `Transaction.sale`.
-    ///  - Parameter completion:  A completion block that returns a device data string that should be passed into server-side calls, such as `Transaction.sale`.
+    ///  - Parameter completion:  A completion block that returns either a device data string that should be passed into server-side calls, such as `Transaction.sale` or an error with the failure reason.
```
scannillo, line 39:
```suggestion
    ///  - Parameter completion:  A completion block that returns either a device data string that should be passed into server-side calls, such as `Transaction.sale`, or an error with the failure reason.
```

Sources/BraintreeApplePay/BTApplePayClient.swift
scannillo: DocC might have some capabilities to allow for docstring re-use

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
scannillo: Yeh, our testing for PP Native modules on BT and PPCP is really limited - mostly due to NXO limitations (no @testable import support, and all classes are `final`). 

We can punt for now, but at some point should outline what exactly we want from them in terms of testing capability.

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
jaxdesmarais: The async/await ones are automatically exposed regardless because we use `@objcMembers` and `async/await` is automatically bridged to a `completionHandler`. We can certainly move away from `@objcMembers` and just use `@objc` across these classes though. Thoughts?

Sources/BraintreeApplePay/BTApplePayClient.swift
jaxdesmarais: Yeah agree, lots of potential room for error, we can certainly look into other options though in the future!

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
```diff
@@ -24,8 +24,9 @@ import BraintreeCore
     ///   `rewardsBalance` will contain information about the rewards balance and `error` will be `nil` (see exceptions in note);
     ///   if it fails, `rewardsBalance` will be `nil` and `error` will describe the failure.
     ///  - Note: If the nonce is associated with an ineligible card or a card with insufficient points, the rewardsBalance will contain this information as `errorMessage` and `errorCode`.
-    public func getRewardsBalance(forNonce nonce: String, currencyIsoCode: String, completion: @escaping (BTAmericanExpressRewardsBalance?, Error?) -> Void) {
-        let parameters = ["currencyIsoCode": currencyIsoCode, "paymentMethodNonce": nonce]
+    @objc(getRewardsBalanceForNonce:currencyIsoCode:completion:)
+    public func getRewardsBalance(for nonce: String, currencyISOCode: String, completion: @escaping (BTAmericanExpressRewardsBalance?, Error?) -> Void) {
```
sshropshire, line 28: The `forNonce` variant might be preferred here since `String` is an opaque type in this method. The `nonce` parameter just happens to contain a string representation of a nonce.

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -49,6 +52,27 @@ import BraintreeCore
         }
     }
 
+    /// Creates a `PKPaymentRequest` with values from your Braintree Apple Pay configuration.
+    /// It populates the following values of `PKPaymentRequest`: `countryCode`, `currencyCode`, `merchantIdentifier`, `supportedNetworks`.
+    /// - Returns: A `PKPaymentRequest`
+    /// - Throws: An `Error` describing the failure
+    public func paymentRequest() async throws -> PKPaymentRequest {
```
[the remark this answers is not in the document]
jaxdesmarais: The non-async/await method is called `paymentRequest` and I believe we want to keep the names the same for both methods. We change both if this is our preference - thoughts?

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
jaxdesmarais: Yeah, technically, in an ideal world, we'd want this module to have merchants just pass in a `BTPaymentMethodNonce` type here instead of a `String`. That way merchants could pass in a `BTCardNonce` here and on the SDK side we would extract the nonce value out for them. Not really possible in it's current state but we should revisit when `Card` is in Swift since it doesn't currently subclass `BTPaymentMethodNonce` but could.

I can revert this for now, but going to add a TODO when we convert Card to revisit this.

Sources/BraintreeApplePay/BTApplePayClient.swift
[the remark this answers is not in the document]
sshropshire: I'd say yeah to follow Swift guidelines but we do get to make the decision ultimately. I remember this being a thing in the guidelines but honestly never seen a ton of people adopt this naming convention for factory methods.

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
scannillo: > The async/await ones are automatically exposed regardless because we use @objcMembers and async/await is automatically bridged to a completionHandler. 

Interesting. So I just messed around taking a look at the auto-generated Swift header. With the current implementation, we only have one `getRewardsBalanceForNonce`, because our ObjC naming annotation matches for both functions. If the naming _didn't_ match for both, you're right that one will get auto-generated due to our `objCMembers`. This results in two `getRewardsBalanceForNonce` methods.

IMO the most explicit / lowest margin of error route here is to not use objCMemembers, so we always know exactly which function is being exposed and it's naming convention. Otherwise we'll always need to make sure that both our completionHandler and async-await functions have the same objC naming annotation - does that make sense? what do you think, Jax?

Sources/BraintreeApplePay/BTApplePayClient.swift
scannillo: It's surprising that ObjC does a lot of translation out of the box - but Swift requires 2 explicit function defs.

A few options/brainstorming:
* 1 docstring
    * Only offer async-await functions. What would the conversion process look like if they need to use completion syntax? At this point, maybe most Swift merchants are prepared to use async-await.
    * Only offer completionHandler style functions. Let merchants wrap to async-await.
* 2 docstrings
    * What we have now - offering both out of the box from our SDK. This _seems_ to me like the best merchant experience. But this is assuming that some Swift merchants don't want to use async-await.
    * But also, at what cost? If our reference docs messy lost in the shuffle, this isn't worth it.
    * What internal practices and enforcements can we put in place to ensure consistency in our docstrings. More of a 🏈 punt, but worth maybe doing some high-level investigation on.

Sources/BraintreeApplePay/BTApplePayClient.swift
scannillo: Resolving to use that other thread

Sources/BraintreeApplePay/BTApplePayClient.swift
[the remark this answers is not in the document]
scannillo: Good call! I'm always down for following the Swift style guide as much as possible

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
jaxdesmarais: Yeah that totally makes sense. I am also inclined to go the route of not using `@objCMembers` in these classes at least and just rely on `@objc`. It's mainly just on the clients, we don't need to nuke it everywhere. I'll make this update.

Sources/BraintreeApplePay/BTApplePayClient.swift
jaxdesmarais: Yeah - I think for now I would vote 2 docstring. The first go around it's a lot to maintain, but most of these were ripped right from Obj-C during the conversion process, so once they are there they don't change a ton it seems. Maybe we can start a discussion with some merchants or on GitHub discussions to see what sort of impact we'd have if we moved to just async/await in v7. May be an even higher adoption rate at that point as well that will make us ripping the completions out easier?

UnitTests/BraintreeSEPADirectDebitTests/BTSEPADirectDebitClient_Tests.swift
```diff
@@ -492,4 +492,90 @@ class BTSEPADirectDebitClient_Tests: XCTestCase {
             }
         }
     }
+
+    func testTokenize_withSEPADirectDebitRequest_returnsSEPADirectDebitNonce() async throws {
```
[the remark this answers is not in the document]
jaxdesmarais: Ah - switching all of the tests over is going to probably make the diff in this PR pretty huge, thoughts on either just updating a few tests for now instead of having new tests or removing these new tests and punting as a whole on updating the tests to use sync await?

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
jaxdesmarais: Updated: https://github.com/braintree/braintree_ios/pull/975/commits/eb9e6ef1d1b8a87a2437a6b64b45cf30c0b7cdde

Sources/BraintreeApplePay/BTApplePayClient.swift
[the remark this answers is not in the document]
jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/975/commits/0aaa0b8fbb0cc719b73b6461a8727c8e19772f70

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -9,7 +9,7 @@ import BraintreeCore
 import BraintreeDataCollector
 #endif
 
-@objcMembers public class BTPayPalClient: NSObject {
+@objc public class BTPayPalClient: NSObject {
```
KunJeongPark, line 12: What is the difference between these two at class level?

jaxdesmarais: From the Swift docs:

[`@objcMembers`](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes/#objcMembers):
> Apply this attribute to a class declaration, to implicitly apply the objc attribute to all Objective-C compatible members of the class, its extensions, its subclasses, and all of the extensions of its subclasses.

[`@objc`](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes/#objc):
> Apply this attribute to any declaration that can be represented in Objective-C — for example, nonnested classes, protocols, nongeneric enumerations (constrained to integer raw-value types), properties and methods (including getters and setters) of classes, protocols and optional members of a protocol, initializers, and subscripts. The objc attribute tells the compiler that a declaration is available to use in Objective-C code.

We do not want to expose both the async and completion methods to Objective-C (discussed [here](https://github.com/braintree/braintree_ios/pull/975#discussion_r1138484648)) so opted for the explicit `@objc` in the clients.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -90,6 +113,29 @@ import BraintreeDataCollector
...
+    /// - Note: You can use this as the final step in your order/checkout flow. If you want, you may create a transaction from your
+    /// server when this method completes without any additional user interaction.
+    ///
+    /// On success, you will receive an instance of `BTPayPalAccountNonce`; on failure or user cancelation you will receive an error.
+    /// If the user cancels out of the flow, the error code will be `.canceled`.
+    ///
+    /// - Parameter request: A `BTPayPalCheckoutRequest`
+    /// - Returns: A `BTPayPalAccountNonce` if successful
+    /// - Throws: An `Error` describing the failure
+    public func tokenize(_ request: BTPayPalCheckoutRequest) async throws -> BTPayPalAccountNonce {
```
KunJeongPark, line 128: Do we also need to make changes to demo app for this?

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -164,6 +163,7 @@ import BraintreeCore
     }
 
     /// Returns true if the proper Venmo app is installed and configured correctly, returns false otherwise.
+    @objc
     public func isVenmoAppInstalled() -> Bool {
```
scannillo, line 167: Total nitpick - but like we do with the class-level `objc` annotations (`@objc class Name`), we can leave the method ones inline like `@objc public func ...`.

Whatever we land on we should document in the style-guide.

I personally usually do `@objc` on the line above only if the naming is different for ObjC. Thoughts?

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: We can punt on this along with tests ([conversation here](https://github.com/braintree/braintree_ios/pull/975#discussion_r1142276481)) I did test them all manually though and you're more than welcome to pull the branch down locally to test as well!

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: Nah - for some reason I just reverted to what is done on the NXO SDK here, not sure what my brain was doing. Will update to single line it. Agree we should style guide it "officially"!

jaxdesmarais: Updated: https://github.com/braintree/braintree_ios/pull/975/commits/e7ace6889b9afb46a9e9d82db5d66a53cb8f5189

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -29,13 +33,16 @@ import BraintreeCore
...
     
     /// Initiates an `ASWebAuthenticationSession` to display a mandate to the user. Upon successful mandate creation, tokenizes the payment method and returns a result
     /// - Parameters:
-    ///   - request: a BTSEPADebitRequest
+    ///   - request: a `BTSEPADebitRequest`
     ///   - completion: This completion will be invoked exactly once when tokenization is complete or an error occurs
+    @objc(tokenizeWithSEPADirectDebitRequest:completion:)
     public func tokenize(
-        request: BTSEPADirectDebitRequest,
+        _ request: BTSEPADirectDebitRequest,
```
KunJeongPark, line 45: Does this need to go in Changelog?

jaxdesmarais: Yes- it is in there already. In the files changed for this PR.

KunJeongPark: Oop, I c it now

Sources/BraintreeApplePay/BTApplePayClient.swift
scannillo: > Maybe we can start a discussion with some merchants or on GitHub discussions to see what sort of impact we'd have if we moved to just async/await in v7.

Yeh for sure - I'd love to try to get some feedback on this. Though the only real way might be to remove it and see who/if anyone raises concerns.

scannillo (approved):

PR #979, opened 2023-03-16
[v6] Remove 3DS Version 1 Classes

jaxdesmarais:
These changes are going into the `convert-three-d-secure` feature branch.

### Summary of changes

- We have a separate ticket to throw errors with 3DSv1 in iOS v5 but have decided for v6+ we will remove 3DSv1 classes entirely
- Before we start the Swift conversion of these files, we want remove files related to 3DSv1 that we don't want to convert - bye forever 3DSv1

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais


Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
```diff
@@ -99,8 +98,6 @@ - (NSString *)shippingMethodAsString {
 
 - (NSString *)versionRequestedAsString {
     switch (self.versionRequested) {
```
scannillo, line 100: We can remove the `versionRequested` enum entirely now, and update the docstrings for the entry-point function that it only does 3DS 2

scannillo:
I think there is a bit more we can remove too. Line 177 of BTThreeDSecureRequest.m [here](https://github.com/braintree/braintree_ios/blob/ffc326d95e14b6ba9d2fdb53f7d9f49e466fadd8/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m#L177) is where we have the 3DS v1 fallback divergent path for if either A) the merchant didn't specify v2 or B) we didn't get back a JWT from Cardinal.

I think that `startRequest` func (and all helper funcs it may use) can go bye-bye.

jaxdesmarais:
> I think there is a bit more we can remove too. Line 177 of BTThreeDSecureRequest.m [here](https://github.com/braintree/braintree_ios/blob/ffc326d95e14b6ba9d2fdb53f7d9f49e466fadd8/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m#L177) is where we have the 3DS v1 fallback divergent path for if either A) the merchant didn't specify v2 or B) we didn't get back a JWT from Cardinal.
> 
> I think that `startRequest` func (and all helper funcs it may use) can go bye-bye.

🤔  actually it looks like we use `startRequest` [both for 3DS v2 and v1](https://github.com/braintree/braintree_ios/blob/ffc326d95e14b6ba9d2fdb53f7d9f49e466fadd8/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m#L169-L174) - looks like we can remove the else but maybe not `startRequest` and it's correlating methods.

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BTThreeDSecureRequest.h
```diff
@@ -157,6 +154,7 @@ typedef NS_ENUM(NSInteger, BTThreeDSecureRequestedExemptionType) {
 /**
  Optional. Set to BTThreeDSecureVersion2 if ThreeDSecure V2 flows are desired, when possible. Defaults to BTThreeDSecureVersion2
  */
+// TODO: as the conversion progresses of 3DS to Swift, consider if we want to keep this property - it may be helpful for an eventual v3?
```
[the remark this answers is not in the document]
jaxdesmarais: I'm not able to find anything related to 3DSv3 being a thing even talked about online or in our internal wiki. I'm going to remove it entirely for now and if we see fit at a later point in time, we can add this enum and property back in. Removed here: https://github.com/braintree/braintree_ios/pull/979/commits/c4ce49fb20cd69118aa63ff6857053d6babb8ade

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
```diff
@@ -165,16 +146,14 @@ - (void)handleRequest:(BTPaymentFlowRequest *)request
...
             [self prepareLookup:apiClient completion:^(NSError * _Nullable error) {
                 if (error != nil) {
                     [delegate onPaymentComplete:nil error:error];
                 } else {
                     [self startRequest:request configuration:configuration];
                 }
             }];
-        } else {
-            [self startRequest:request configuration:configuration];
         }
```
scannillo, line 157: Since there is no more `else` here, I think we need to handle the case of no `configuration.cardinalAuthenticationJWT` and return some error to the merchant, vs right now this no-ops.

sshropshire: +1

jaxdesmarais: Good catch - updated: https://github.com/braintree/braintree_ios/pull/979/commits/fce558979ad35575c473a4de74d2d0ae63271c04

scannillo (approved):
🔥

PR #980, opened 2023-03-17
Convert 3D Secure V2Customization Classes to Swift

jaxdesmarais:
This work is branched off of the `remove-3ds-version1` branch and will ultimately go into the `convert-three-d-secure` feature branch. We expect SPM to fail as the target contains a mixed Obj-C and Swift module. When the remaining conversion is complete SPM should pass as expected and we will open a PR to merge the feature branch into main.

### Summary of changes

- Convert `BTThreeDSecureV2BaseCustomization` to Swift
- Convert `BTThreeDSecureV2UICustomization` to Swift
- Convert `BTThreeDSecureV2ButtonCustomization` to Swift
- Convert `BTThreeDSecureV2LabelCustomization` to Swift
- Convert `BTThreeDSecureV2ToolbarCustomization` to Swift
- Convert `BTThreeDSecureV2TextBoxCustomization` to Swift
- Convert `BTThreeDSecureV2ButtonType` enum to Swift
- Update unit test imports
- Update enum cases in demo app and unit test

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais


Sources/BraintreeThreeDSecure/V2UICustomization/BTThreeDSecureV2ButtonCustomization.swift
```diff
@@ -0,0 +1,41 @@
...
+    // MARK: - Public Properties
+
+    /// Color code in Hex format. For example, the color code can be “#999999”.
+    public var backgroundColor: String? {
+        get { _backgroundColor }
+        @objc(setBackgroundColor:) set {
+            _backgroundColor = newValue
+            (cardinalValue as? ButtonCustomization)?.backgroundColor = newValue
+        }
+    }
```
jaxdesmarais, line 16: This pattern exists in every class that was converted. Currently these properties are accessed like this in each language:
```
// Swift
let labelCustomization = BTThreeDSecureV2LabelCustomization()
labelCustomization.headingTextColor = "#A020F0"

// Objective-C
BTThreeDSecureV2LabelCustomization *labelCustomization = [BTThreeDSecureV2LabelCustomization new];
[labelCustomization setHeadingTextColor:@"#A020F0"];
```
Since there are going to be other breaking changes in the conversion of 3D Secure, it did not seem compelling to also break this interface. From the merchant point of view if they are using any of these customization classes, they will have the exact same experience pre and post the conversion outlined here. 

That said if someone feels incredibly strongly about this, we can certainly discuss!

scannillo: Thanks for the explanation here - down to minimize breaking changes at this point, and agree to keep!

scannillo (approved):
🎉

Sources/BraintreeThreeDSecure/V2UICustomization/BTThreeDSecureV2BaseCustomization.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    // MARK: - Internal Properties
+
+    var cardinalValue: Customization?
+
+    // MARK: - Fileprivate Properties
+
+    /// Used as a holder for Obj-C interoperability
+    fileprivate var _textFontName: String?
```
sshropshire, line 43: random q: Can `private` be used here?

jaxdesmarais: Yeah we can use either - if we have a preference I can update these to private. I guess it's unlikely something in this file needs access to these properties.

scannillo (approved):

PR #981, opened 2023-03-20
Convert BTAuthenticationInsight, BTCardRequest and BTThreeDSecureInfo

jaxdesmarais:
This work will go into the `convert-card` feature branch. We expect SPM to fail as the target contains a mixed Obj-C and Swift module. When the remaining conversion is complete SPM should pass as expected and we will open a PR to merge the feature branch into main.

### Summary of changes

- Convert `BTAuthenticationInsight` to Swift
- Convert `BTCardRequest` to Swift
- Convert `BTThreeDSecureInfo` to Swift

### Checklist

- ~[ ] Added a changelog entry~ there are no changes to the public interface in this PR

- @jaxdesmarais


Sources/BraintreeCard/BTThreeDSecureInfo.swift
```diff
@@ -0,0 +1,109 @@
...
+    /// The 3DS version used in the authentication, example "1.0.2" or "2.1.0".
+    public var threeDSecureVersion: String?
+
+    /// Indicates if the 3D Secure lookup was performed.
+    public var wasVerified: Bool = false
+
+    /// Transaction identifier resulting from 3D Secure authentication. Uniquely identifies the transaction and sometimes required in the authorization message.
+    /// This field will no longer be used in 3DS 2 authentications.
+    // TODO: do we need this now that we only use 3ds v2?
+    public var xid: String?
```
jaxdesmarais, line 77: Added this TODO to check with other folks - based on the comment from Obj-C it seems like we can remove this but wanted to check if anyone had more insight into this field before we remove it.

KunJeongPark (approved):

Sources/BraintreeCard/BTThreeDSecureInfo.swift
jaxdesmarais: The [Cardinal response object docs](https://cardinaldocs.atlassian.net/wiki/spaces/CC/pages/98315/Response+Objects) don't seem to indicate that this is a 3dsv1 only thing just that it's only returned for Amex, I am going to remove this TODO.

Sources/BraintreeCard/BTThreeDSecureInfo.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    /// The 3DS version used in the authentication, example "1.0.2" or "2.1.0".
+    public var threeDSecureVersion: String?
+
+    /// Indicates if the 3D Secure lookup was performed.
+    public var wasVerified: Bool = false
+
+    /// Transaction identifier resulting from 3D Secure authentication. Uniquely identifies the transaction and sometimes required in the authorization message.
+    /// This field will no longer be used in 3DS 2 authentications.
+    // TODO: do we need this now that we only use 3ds v2?
```
jaxdesmarais, line 76:
```suggestion
```

PR #983, opened 2023-03-21
Convert BTCardClient to Swift

jaxdesmarais:
### Summary of changes

- Convert `BTCardClient` to Swift
- Add `BTCardError` in Swift
- Update `Braintree.podspec` and `Package.swift`

### Checklist

- [x] Added a changelog entry

- @scannillo @jaxdesmarais


Sources/BraintreeCard/BTCardClient.swift
```diff
@@ -0,0 +1,218 @@
...
+    // MARK: - Public Methods
+
+    /// Tokenizes a card
+    /// - Parameters:
+    ///    - card: The card to tokenize.
+    ///    - completion: A completion block that is invoked when card tokenization has completed. If tokenization succeeds,
+    ///    `tokenize` will contain a nonce and `error` will be `nil`; if it fails, `tokenize` will be `nil` and `error`will describe the failure.
+    @objc(tokenizeCard:completion:)
+    public func tokenize(_ card: BTCard, completion: @escaping (BTCardNonce?, Error?) -> Void) {
+        let request = BTCardRequest(card: card)
```
KunJeongPark, line 35: Didn't we have analytic event sent here?

jaxdesmarais: That's in the `fpti-feature` branch which has not been merged into `master` yet. We will have to resolve the merge conflicts and update `fpti-feature` with the analytics events once this feature branch is merged into `master`.

Sources/BraintreeCard/BTCardClient.swift
```diff
@@ -0,0 +1,218 @@
...
+
+        apiClient.fetchOrReturnRemoteConfiguration() { configuration, error in
+            guard let configuration, error == nil else {
+                completion(nil, error)
```
KunJeongPark, line 39: Are we going to send BTCardAnalytics.cardTokenizedFailed?

jaxdesmarais: Same answer as here - the updated analytics names don't exist on `master` yet: https://github.com/braintree/braintree_ios/pull/983#discussion_r1145098549

Sources/BraintreeCard/BTCardClient.swift
KunJeongPark: Since there are deletions of objective-c files where we had change in analytics, they wouldn't be conflicts but todo's?

Sources/BraintreeCard/BTCardClient.swift
```diff
@@ -0,0 +1,218 @@
...
+                    if let error = error as NSError? {
+                        if error.code == BTCoreConstants.networkConnectionLostCode {
+                            self.apiClient.sendAnalyticsEvent("ios.tokenize-card.graphQL.network-connection.failure")
+                        }
+
+                        let response: HTTPURLResponse? = error.userInfo[BTCoreConstants.urlResponseKey] as? HTTPURLResponse
+                        var callbackError: Error? = error
+
+                        if response?.statusCode == 422 {
+                            callbackError = self.constructCallbackError(with: error.userInfo, error: error)
```
KunJeongPark, line 61: can the error.userInfo be nil?

Sources/BraintreeCard/BTCardClient.swift
jaxdesmarais: Not sure I follow, conflicts vs TODOs. In any case, we should be able to resolve it, will just involve looking at the diff and moving the events from the `.m` file to the `.swift` file. It will be _slightly_ more complicated but not impossible. We should though prioritize adding analytics for the modules already in Swift though to avoid similar issues in the future while we wrap up the conversions.

Sources/BraintreeCard/BTCardClient.swift
jaxdesmarais: Good catch! Yes, there could potentially be a situation where when we cast `error` to `NSError` above and the error is `nil` it'd still be cast successfully (while unlikely, certainly possible). Since a 422 would indicate an error from the gateway, I think we have always just assumed the gateway would return an error containing `userInfo`. I think we can just make the dictionary optional in the method to match what we were doing in Obj-C. Updated here: https://github.com/braintree/braintree_ios/pull/983/commits/d623ffc077af4bb9171937021efe95cf08972690

Sources/BraintreeThreeDSecure/Public/BraintreeThreeDSecure/BraintreeThreeDSecure.h
```diff
@@ -10,7 +10,6 @@ FOUNDATION_EXPORT const unsigned char BraintreeThreeDSecureVersionString[];
 
 #if __has_include(<Braintree/BraintreeThreeDSecure.h>)
 #import <Braintree/BraintreePaymentFlow.h>
-#import <Braintree/BraintreeCard.h>
```
KunJeongPark, line 13: 👋🏻

Sources/BraintreeCard/BTCardClient.swift
```diff
@@ -0,0 +1,218 @@
...
+
+    private func sendAnalyticsEvent(with success: Bool) {
+        let integration = apiClient.metadata.integrationString
+        let status = success ? "succeeded" : "failed"
+        let event = "ios.\(integration).card.\(status)"
+
+        apiClient.sendAnalyticsEvent(event)
+    }
+
+    private func sendGraphQLAnalyticsEvent(with success: Bool) {
```
sshropshire, line 165:
```suggestion
    private func sendGraphQLTokenizeAnalyticsEvent(with success: Bool) {
```
🧶

Sources/BraintreeCard/BTCardClient.swift
```diff
@@ -0,0 +1,218 @@
...
+            parameters["authenticationInsight"] = true
+            parameters["merchantAccountId"] = request.card.merchantAccountID
+        }
+
+        return parameters
+    }
+
+    // MARK: - Analytics
+
+    private func sendAnalyticsEvent(with success: Bool) {
```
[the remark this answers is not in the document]
jaxdesmarais: These methods have both been removed in [`fpti-feature` branch](https://github.com/braintree/braintree_ios/compare/fpti-feature). Gonna just leave as is for now and resolve this entirely when we merge that branch in.

Sources/BraintreeCard/BTCardClient.swift
jaxdesmarais: These methods have both been removed in [`fpti-feature` branch](https://github.com/braintree/braintree_ios/compare/fpti-feature). Gonna just leave as is for now and resolve this entirely when we merge that branch in.

PR #984, opened 2023-03-22
Update Migration Guide

jaxdesmarais:
### Summary of changes

- Update method signature style of the `V6_MIGRATION` guide to match the Apple method signature syntax (example [here](https://developer.apple.com/documentation/passkit/pkpaymentauthorizationcontroller/1649455-canmakepayments))
- Other general cleanup

### Checklist

- @jaxdesmarais

### DIFF
--- a/V6_MIGRATION.md
+++ b/V6_MIGRATION.md
@@ -54,7 +54,7 @@ The possible values for `BTVenmoPaymentMethodUsage` include:
 * `.multiUse` - the Venmo payment will be authorized for future payments and can be vaulted.
 * `.singleUse` - the Venmo payment will be authorized for a one-time payment and cannot be vaulted.
 
-`BTVenmoClient.tokenizeVenmoAccount(with: BTVenmoRequest, completion: @escaping (BTVenmoAccountNonce?, Error?) -> Void)` has been renamed to `BTVenmoClient.tokenize(_ request: BTVenmoRequest, completion: @escaping (BTVenmoAccountNonce?, Error?) -> Void)`
+`BTVenmoClient.tokenizeVenmoAccount(with:completion:)` has been renamed to `BTVenmoClient.tokenize(_:completion:)`
 
 `BTVenmoClient.isiOSAppAvailableForAppSwitch()` has been renamed to `BTVenmoClient.isVenmoAppInstalled()`
 
@@ -76,9 +76,9 @@ venmoClient.tokenize(venmoRequest) { venmoAccountNonce, error in
 
 The property `BTPayPalRequest.activeWindow` has been removed
 
-Removed `BTPayPalDriver.requestOneTimePayment` and `BTPayPalDriver.requestBillingAgreement` in favor of `BTPayPalClient.tokenizePayPalAccount`:
+Removed `BTPayPalDriver.requestOneTimePayment` and `BTPayPalDriver.requestBillingAgreement` in favor of `BTPayPalClient.tokenize`:
 ```
-let payPalClient = BTPayPalClient(apiClient: <MY_BTAPICLIENT>)
+let payPalClient = BTPayPalClient(apiClient: <MY_BTAPIClient>)
 let request = BTPayPalCheckoutRequest(amount: "1")
 
 payPalClient.tokenize(request) { payPalAccountNonce, error in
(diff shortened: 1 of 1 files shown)


V6_MIGRATION.md
```diff
@@ -76,9 +76,9 @@ venmoClient.tokenize(venmoRequest) { venmoAccountNonce, error in
 
 The property `BTPayPalRequest.activeWindow` has been removed
 
-Removed `BTPayPalDriver.requestOneTimePayment` and `BTPayPalDriver.requestBillingAgreement` in favor of `BTPayPalClient.tokenizePayPalAccount`:
+Removed `BTPayPalDriver.requestOneTimePayment` and `BTPayPalDriver.requestBillingAgreement` in favor of `BTPayPalClient.tokenize`:
 ```
-let payPalClient = BTPayPalClient(apiClient: <MY_BTAPICLIENT>)
+let payPalClient = BTPayPalClient(apiClient: <MY_BTAPIClient>)
```
jaxdesmarais, line 81: Updated this but not totally sure what our preferred syntax is for this. All caps? Just `<BTAPIClient>`? Something else entirely? Hoping someone has a strong preference and we can update these all accordingly.

V6_MIGRATION.md
```diff
@@ -149,23 +148,22 @@ The `BraintreeUnionPay` module, and all containing classes, was removed in v6. U
...
-let cardClient = BTCardClient(apiClient: braintreeClient)
+let cardClient = BTCardClient(apiClient: <MY_BTAPIClient>)
 
 let card = BTCard()
 card.number = "4111111111111111"
 card.expirationMonth = "12"
 card.expirationYear = "2025"
 
-cardClient.tokenizeCard(card) { (tokenizedCard, error) in
+cardClient.tokenize(card) { tokenizedCard, error in
```
jaxdesmarais, line 158: Below this we have `// Communicate the tokenizedCard.nonce to your server, or handle error` but in the rest of this file we just do `// handle response`. Do folks have a preference on this? We should update them all to be the same, imo.

V6_MIGRATION.md
sshropshire: Could we instantiate a BTAPIClient with a "<TOKENIZATION_KEY_OR_CLIENT_TOKEN>" auth string? Not sure if that's too much but it'll have a similar effect as above and be valid for the copy/paste.

sshropshire (approved):

V6_MIGRATION.md
scannillo: I think in general the `<SOME_VALUE>` syntax is all-caps, and is usually for strings.

I would do something like:
```
let apiClient = BTAPIClient("<TOKENIZATION_KEY_OR_CLIENT_TOKEN>")
let payPalClient = BTPayPalClient(apiClient: apiClient)
```

scannillo: And then in the examples below, just use `apiClient` as the param.

V6_MIGRATION.md
```diff
@@ -54,7 +54,7 @@ The possible values for `BTVenmoPaymentMethodUsage` include:
 * `.multiUse` - the Venmo payment will be authorized for future payments and can be vaulted.
 * `.singleUse` - the Venmo payment will be authorized for a one-time payment and cannot be vaulted.
 
-`BTVenmoClient.tokenizeVenmoAccount(with: BTVenmoRequest, completion: @escaping (BTVenmoAccountNonce?, Error?) -> Void)` has been renamed to `BTVenmoClient.tokenize(_ request: BTVenmoRequest, completion: @escaping (BTVenmoAccountNonce?, Error?) -> Void)`
+`BTVenmoClient.tokenizeVenmoAccount(with:completion:)` has been renamed to `BTVenmoClient.tokenize(_:completion:)`
```
scannillo, line 57: This makes things a lot easier to read, too! We could also audit the CHANGELOG entries for the beta cleaning these up too.

jaxdesmarais: I looked through the CHANGELOG entries and wasn't able to find anywhere where we had the other format, it looked like so far we've always used this syntax there.

V6_MIGRATION.md
jaxdesmarais: Dope! I'll update these - I dig that too.

V6_MIGRATION.md
jaxdesmarais: Updated to `handle response`: https://github.com/braintree/braintree_ios/pull/984/commits/5842ae352f8324be34cc5c372319765e22a2aa72

V6_MIGRATION.md
jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/984/commits/5842ae352f8324be34cc5c372319765e22a2aa72

KunJeongPark (approved):

PR #991, opened 2023-03-28
Convert BTThreeDSecureV2Provider and BTThreeDSecureAuthenticateJWT to Swift

jaxdesmarais:
This work is going into the `convert-three-d-secure` feature branch. We expect SPM to fail as the target contains a mixed Obj-C and Swift module. When the remaining conversion is complete SPM should pass as expected and we will open a PR to merge the feature branch into main.

### Summary of changes

- Convert `BTThreeDSecureV2Provider` to Swift
- Convert `BTThreeDSecureAuthenticateJWT` to Swift
- Add `BTThreeDSecureError`

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais


Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
```diff
@@ -0,0 +1,65 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+class BTThreeDSecureAuthenticateJWT {
+
+    static func authenticate(
+        jwt cardinalJWT: String,
```
KunJeongPark, line 10: Does this automatically translate to objective-c function name "authenticateJWT"?

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
```diff
@@ -0,0 +1,160 @@
...
+@objcMembers public class BTThreeDSecureV2Provider: NSObject {
+
+    // MARK: - Internal Properties
+
+    let cardinalSession: CardinalSession
+    let apiClient: BTAPIClient
+
+    var lookupResult: BTThreeDSecureResult? = nil
+
+    static var completionHandler: (BTThreeDSecureResult?, Error?) -> Void = { _, _ in }
```
KunJeongPark, line 18: I've never seen this before. Makes this completion handler optional, right?

Sources/BraintreeThreeDSecure/BTThreeDSecureError.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    /// 3D Secure was not configured correctly
+    case configuration
+
+    /// A body was not returned from the API during the request.
+    case noBodyReturned
+
+    static var errorDomain: String {
+        "com.braintreepayments.BTThreeDSecureFlowErrorDomain"
+    }
```
sshropshire, line 22:
```suggestion
    static let errorDomain: String = "com.braintreepayments.BTThreeDSecureFlowErrorDomain"
```
🧶 : does this compile as a plain constant?

Sources/BraintreeThreeDSecure/BTThreeDSecureError.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    var errorCode: Int {
+        rawValue
+    }
```
sshropshire, line 26:
```suggestion
    let errorCode: Int = rawValue
```
🧶: +1 as a constant

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
```diff
@@ -0,0 +1,160 @@
...
+        cardinalSession.continueWith(
+            transactionId: lookupResult.lookup?.transactionID ?? "",
+            payload: lookupResult.lookup?.paReq ?? "",
+            validationDelegate: self
+        )
+    }
+
+    // MARK: - Private Methods
+
+    private func callFailureHandler(
```
sshropshire, line 84:
```suggestion
    private func notifyError(
```
Take or leave, we've been using `notifyX` on Android to callout when we pass a message back to the merchant.

sshropshire (approved):

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
```diff
@@ -0,0 +1,65 @@
...
+
+            let threeDSecureResult: BTThreeDSecureResult = BTThreeDSecureResult(json: body)
+
+            if threeDSecureResult.tokenizedCard != nil && threeDSecureResult.errorMessage == nil {
+                apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.upgrade-payment-method.succeeded")
+            } else {
+                apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.upgrade-payment-method.failure.returned-lookup-nonce")
+
+                // If authentication wasn't successful, add the BTCardNonce from the lookup result to the authentication result
+                // so that merchants can transact with the lookup nonce if desired.
```
KunJeongPark, line 57: I know that this comment was just carried over, but what can the merchants do with the BTCardNonce saved in the result?

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
```diff
@@ -168,7 +167,7 @@ - (void)prepareLookup:(BTAPIClient *)apiClient completion:(void (^)(NSError * _N
         }
 
         if (configuration.cardinalAuthenticationJWT) {
-            self.threeDSecureV2Provider = [BTThreeDSecureV2Provider initializeProviderWithConfiguration:configuration
```
KunJeongPark, line 171: I wasn't sure why a static function was initially used to return an instance in this deleted line.

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        cardinalConfiguration.deploymentEnvironment = cardinalEnvironment
+        cardinalSession.configure(cardinalConfiguration)
+        cardinalSession.setup(
+            jwtString: cardinalAuthenticationJWT,
+            completed: { consumerSessionID in
+                apiClient.sendAnalyticsEvent("ios.three-d-secure.cardinal-sdk.init.setup-completed")
+                completion(["dfReferenceId": consumerSessionID])
+            }, validated: { _ in
```
KunJeongPark, line 57: how come this works? I thought it would be didValidate: { ...}

Sources/BraintreeThreeDSecure/BTThreeDSecureError.swift
jaxdesmarais: It won't compile because enums can't contain stored properties

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
jaxdesmarais: Good question - yeah we use this pattern a few times in the SDK (BTVenmoClient.appSwitchCompletion is another example). This initializes the completion with a default implementation, though it doesn't do anything until we use it. It doesn't need to be static though I am now realizing so going to update that.

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
jaxdesmarais: A merchant could still use that nonce from the `lookupResult` in a transaction, it would just mean the card was ineligible for 3D Secure. They would just take on additional liability that would typically be covered by 3D Secure.

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.m
jaxdesmarais: Not sure either! Makes more sense to have it as an `init` vs a method that is essentially an init. I am guessing there was a reason at some point but now we have the opportunity to streamline it!

KunJeongPark (approved):

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -1058,7 +1058,7 @@
 			baseConfigurationReference = 73498B4265CA7D315E2FBF38 /* Pods-Demo.debug.xcconfig */;
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
-				CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER = YES_ERROR;
+				CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER = NO;
```
jaxdesmarais, line 1061: I was getting the same error that the UI tests are getting ([example here](https://github.com/braintree/braintree_ios/actions/runs/4556828816/jobs/8040317812)) last week as well. This should resolve the issue but we should open a ticket with Cardinal to update their SDK to not use quoted includes as this would potentially be change merchants would have to make as well.

Posting the error for posterity since the CI run will be deleted after some time:
```
double-quoted include "<CARDINAL_CLASS_NAME>.h" in framework header, expected angle-bracketed instead [-Werror,-Wquoted-include-in-framework-header]
```

PR #993, opened 2023-03-29
Add capture lists to `fetchOrReturnRemoteConfiguration()` calls

scannillo:
### Summary of changes

- Add missing [weak capture lists ](https://www.hackingwithswift.com/articles/179/capture-lists-in-swift-whats-the-difference-between-weak-strong-and-unowned-references)when holding references in `BTAPIClient.fetchOrReturnRemoteConfiguration()`'s non-escaping closure to avoid retain cycle.
- Use new Swift 5.8 `guard let self` shorthand.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


jaxdesmarais (approved):
👏

sshropshire (approved):

scannillo:
Turns out there was a strong reference cycle happening in the Demo app due to the way that the `PayPalClient` instance was deallocated too early (within the scope of a single helper method).

This is a case where we vend a specific merchant error. Ticket created - DTBTSDK-2544.

PR #989, opened 2023-03-27
PaypalNativeCheckout analytics audit

KunJeongPark:
### Summary of changes
These changes are going into the fpti-feature branch which will get merged once we audit all modules and make the switch to the FPTI endpoints
- Update PayPalNativeCheckout analytic events to prepare switch to FPT
- Update tests
- Bucketed all analytic events sent before callbacks into 3 categories of final events: tokenizeFailed, tokenizeSucceeded, tokenizeCanceled.

### Checklist

~- [ ] Added a changelog entry~

### Authors
@KunJeongPark

PR #994, opened 2023-03-30
Venmo analytics audit

KunJeongPark:
Summary of changes
These changes are going into the fpti-feature branch. Which will get merged in once we audit all modules and make the switch to the FPTI endpoints

-Update Venmo analytics events to prepare for switch to FPTI
-Update tests
-Bucketed all analytic events sent before callbacks into 3 categories counted in conversion rates:
tokenizeFailed, tokenizeSucceeded and appSwitchCanceled

Checklist
~[ ] Added a changelog entry~
Authors
@KunJeongPark @sshropshire

### DIFF
--- a/Sources/BraintreeVenmo/BTVenmoAnalytics.swift
+++ b/Sources/BraintreeVenmo/BTVenmoAnalytics.swift
@@ -6,12 +6,12 @@ class BTVenmoAnalytics {
     
     static let tokenizeStarted = "venmo:tokenize:started"
     static let tokenizeFailed = "venmo:tokenize:failed"
-    static let tokenizeSucceeded = "venmo:tokenize:succeeded"
+    static let tokenizeSucceeded = "venmo:tokenize:succeeded"    
+    static let appSwitchCanceled = "venmo:tokenize:app-switch:canceled"
     
     // MARK: - Additional Detail Events
     
     static let tokenizeNetworkConnectionLost = "venmo:tokenize:network-connection:failed"
     static let appSwitchSucceeded = "venmo:tokenize:app-switch:succeeded"
     static let appSwitchFailed = "venmo:tokenize:app-switch:failed"
-    static let appSwitchCanceled = "venmo:tokenize:app-switch:canceled"
 }
--- a/Sources/BraintreeVenmo/BTVenmoAnalytics.swift
+++ b/Sources/BraintreeVenmo/BTVenmoAnalytics.swift
@@ -2,14 +2,14 @@ import Foundation
 
 class BTVenmoAnalytics {
     
-    // MARK: - Tokenize Request events
+    // MARK: - Conversion Events
     
     static let tokenizeStarted = "venmo:tokenize:started"
     static let tokenizeFailed = "venmo:tokenize:failed"
     static let tokenizeNetworkConnectionLost = "venmo:tokenize:network-connection:failed"
     static let tokenizeSucceeded = "venmo:tokenize:succeeded"
     
-    // MARK: - App Switch Events
+    // MARK: - Additional Detail Events
     
     static let appSwitchSucceeded = "venmo:tokenize:app-switch:succeeded"
     static let appSwitchFailed = "venmo:tokenize:app-switch:failed"
--- a/Sources/BraintreeVenmo/BTVenmoAnalytics.swift
+++ b/Sources/BraintreeVenmo/BTVenmoAnalytics.swift
@@ -1,0 +1,17 @@
+import Foundation
+
+class BTVenmoAnalytics {
+    
+    // MARK: - Tokenize Request events
+    
+    static let tokenizeStarted = "venmo:tokenize:started"
+    static let tokenizeFailed = "venmo:tokenize:failed"
+    static let tokenizeNetworkConnectionLost = "venmo:tokenize:network-connection:failed"
+    static let tokenizeSucceeded = "venmo:tokenize:succeeded"
+    
+    // MARK: - App Switch Events
+    
+    static let appSwitchSucceeded = "venmo:tokenize:app-switch:succeeded"
+    static let appSwitchFailed = "venmo:tokenize:app-switch:failed"
+    static let appSwitchCanceled = "venmo:tokenize:app-switch:canceled"
+}
--- a/Sources/BraintreeVenmo/BTVenmoAnalytics.swift
+++ b/Sources/BraintreeVenmo/BTVenmoAnalytics.swift
@@ -6,11 +6,11 @@ class BTVenmoAnalytics {
     
     static let tokenizeStarted = "venmo:tokenize:started"
     static let tokenizeFailed = "venmo:tokenize:failed"
-    static let tokenizeNetworkConnectionLost = "venmo:tokenize:network-connection:failed"
     static let tokenizeSucceeded = "venmo:tokenize:succeeded"
     
     // MARK: - Additional Detail Events
     
+    static let tokenizeNetworkConnectionLost = "venmo:tokenize:network-connection:failed"
     static let appSwitchSucceeded = "venmo:tokenize:app-switch:succeeded"
     static let appSwitchFailed = "venmo:tokenize:app-switch:failed"
     static let appSwitchCanceled = "venmo:tokenize:app-switch:canceled"
--- a/Sources/BraintreeVenmo/BTVenmoClient.swift
+++ b/Sources/BraintreeVenmo/BTVenmoClient.swift
@@ -197,8 +197,7 @@ import BraintreeCore
     // MARK: - App Switch Methods
 
     func handleOpen(_ url: URL) {
-        guard let returnURL = BTVenmoAppSwitchReturnURL(url: url) else {
-          
+        guard let returnURL = BTVenmoAppSwitchReturnURL(url: url) else {          
             apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeFailed)
             appSwitchCompletion(nil, BTVenmoError.invalidReturnURL(""))
             return
--- a/Sources/BraintreeVenmo/BTVenmoClient.swift
+++ b/Sources/BraintreeVenmo/BTVenmoClient.swift
@@ -69,11 +69,13 @@ import BraintreeCore
 
         apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
             if let error {
+                self.apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeFailed)
                 completion(nil, error)
                 return
             }
 
             guard let configuration else {
+                self.apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeFailed)
                 completion(nil, BTVenmoError.fetchConfigurationFailed)
                 return
             }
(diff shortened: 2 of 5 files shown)


Sources/BraintreeVenmo/BTVenmoAnalytics.swift
```diff
@@ -0,0 +1,17 @@
...
+    static let tokenizeStarted = "venmo:tokenize:started"
+    static let tokenizeFailed = "venmo:tokenize:failed"
+    static let tokenizeNetworkConnectionLost = "venmo:tokenize:network-connection:failed"
+    static let tokenizeSucceeded = "venmo:tokenize:succeeded"
+    
+    // MARK: - App Switch Events
+    
+    static let appSwitchSucceeded = "venmo:tokenize:app-switch:succeeded"
+    static let appSwitchFailed = "venmo:tokenize:app-switch:failed"
+    static let appSwitchCanceled = "venmo:tokenize:app-switch:canceled"
```
scannillo, line 16: See [comment](https://github.com/braintree/braintree_ios/pull/989/files#r1154509130).

```
    // MARK: - Conversion Events
    
    static let tokenizeStarted = "venmo:tokenize:started"
    static let tokenizeFailed = "venmo:tokenize:failed"
    static let tokenizeSucceeded = "venmo:tokenize:succeeded"
    static let appSwitchCanceled = "venmo:tokenize:app-switch:canceled"

    
    // MARK: - Additional Detail Events
    
    static let tokenizeNetworkConnectionLost = "venmo:tokenize:network-connection:failed"
    static let appSwitchSucceeded = "venmo:tokenize:app-switch:succeeded"
    static let appSwitchFailed = "venmo:tokenize:app-switch:failed"
    static let appSwitchCanceled = "venmo:tokenize:app-switch:canceled"
```

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -295,11 +300,12 @@ import BraintreeCore
...
 
         if success {
-            apiClient.sendAnalyticsEvent("ios.pay-with-venmo.appswitch.initiate.success")
+            apiClient.sendAnalyticsEvent(BTVenmoAnalytics.appSwitchSucceeded)
             BTVenmoClient.venmoClient = self
             self.appSwitchCompletion = completion
         } else {
-            apiClient.sendAnalyticsEvent("ios.pay-with-venmo.appswitch.initiate.error.failure")
+            apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeFailed)
+            apiClient.sendAnalyticsEvent(BTVenmoAnalytics.appSwitchFailed)
```
jaxdesmarais, line 308: We want to send `appSwitchFailed` before `tokenizeFailed`

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -313,23 +319,24 @@ import BraintreeCore
...
+                self.apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeFailed)
                 self.appSwitchCompletion(nil, BTVenmoError.invalidBodyReturned)
                 return
             }
 
             let venmoAccountJSON: BTJSON = body["venmoAccounts"][0]
             let venmoAccountNonce: BTVenmoAccountNonce = BTVenmoAccountNonce.venmoAccount(with: venmoAccountJSON)
 
-            self.apiClient.sendAnalyticsEvent("ios.pay-with-venmo.vault.success")
+            self.apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeSucceeded)
```
jaxdesmarais, line 339: The completion below sends either a nonce or an error if present - we should check `venmoAccountJSON.asError()` first and send the `tokenizeFailed` event there with a nil nonce in the completion. Right now we'd send a `tokenizeSucceeded` event even if there is an error. I think we did a similar check in Card.

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -207,45 +212,44 @@ import BraintreeCore
...
-                apiClient.sendAnalyticsEvent("ios.pay-with-venmo.appswitch.handle.client-failure")
+                apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeFailed)
                 appSwitchCompletion(nil, BTVenmoError.invalidReturnURL("username"))
                 return
             }
 
-            apiClient.sendAnalyticsEvent("ios.pay-with-venmo.appswitch.handle.success")
-
             if shouldVault && apiClient.clientToken != nil {
                 vault(nonce)
```
jaxdesmarais, line 254: I think we probably need to add the tokenize events to the `vault` method since if we hit this block we'd never end up sending a success/failure.

Sources/BraintreeVenmo/BTVenmoClient.swift
KunJeongPark: Great catch! I missed that it would be sending an error from checking the JSON body.

Sources/BraintreeVenmo/BTVenmoClient.swift
KunJeongPark: I was wondering about the second else block in the case the clientToken is nil. That would be a fail, right?

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: Nope, we don't care about those - only the ones on the doc are the ones we want to retain

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: Ah, I think I must have missed those. 👍

jaxdesmarais: I am only seeing checks that `clientToken != nil` for vaulting. It looks like we check for the existence of a client token to vaultVenmo and if we don't vault we continue on the non-vault flow. If I misunderstood the conditional you were talking about feel free to link the code block you were referencing here!

KunJeongPark: yeah, that's what i was thinking. We need to add the case clientToken is nil on top of both of these? wdyt?

KunJeongPark: @scannillo is this logic checking for client token not being nil only for vault option ok as it is?

Sources/BraintreeVenmo/BTVenmoAnalytics.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+class BTVenmoAnalytics {
+    
+    // MARK: - Conversion Events
+    
+    static let tokenizeStarted = "venmo:tokenize:started"
+    static let tokenizeFailed = "venmo:tokenize:failed"
+    static let tokenizeNetworkConnectionLost = "venmo:tokenize:network-connection:failed"
```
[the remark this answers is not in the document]
KunJeongPark: sorry, I had them originally grouped by "tokenize" events and forgot to change the grouping.

Sources/BraintreeVenmo/BTVenmoAnalytics.swift
```diff
@@ -0,0 +1,17 @@
...
+    static let tokenizeStarted = "venmo:tokenize:started"
+    static let tokenizeFailed = "venmo:tokenize:failed"
+    static let tokenizeSucceeded = "venmo:tokenize:succeeded"
+    
+    // MARK: - Additional Detail Events
+    
+    static let tokenizeNetworkConnectionLost = "venmo:tokenize:network-connection:failed"
+    static let appSwitchSucceeded = "venmo:tokenize:app-switch:succeeded"
+    static let appSwitchFailed = "venmo:tokenize:app-switch:failed"
+    static let appSwitchCanceled = "venmo:tokenize:app-switch:canceled"
```
scannillo, line 16: The canceled case is a `conversion` event, so it should move up.

scannillo: Another thought I was having, maybe the generic cancels would be more clear as `venmo:tokenize:canceled`, and have that syntax apply to all modules. So `<feature>:<method_name>:canceled`.

We can tackle this in a follow up PR though if we like it.

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -207,45 +214,44 @@ import BraintreeCore
...
-                apiClient.sendAnalyticsEvent("ios.pay-with-venmo.appswitch.handle.client-failure")
+                apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeFailed)
                 appSwitchCompletion(nil, BTVenmoError.invalidReturnURL("nonce"))
                 return
             }
 
             guard let username = returnURL.username else {
-                apiClient.sendAnalyticsEvent("ios.pay-with-venmo.appswitch.handle.client-failure")
+                apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeFailed)
                 appSwitchCompletion(nil, BTVenmoError.invalidReturnURL("username"))
```
scannillo, line 251: I don't want to beat a dead horse - but I still prefer the original idea of using a single "notify" function for each of the 3 end cases (success, fail, cancel). It's annoying to need to duplicate the line `apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeFailed)` everywhere and is more prone to error.

[We do this in PPCP.](https://github.com/paypal/iOS-SDK/blob/2a79160e81d8757d9c1891ad4e3dc1bf98f64187/Sources/PayPalNativePayments/PayPalNativeCheckoutClient.swift#L103)

scannillo: @KunJeongPark & @sshropshire thoughts?

KunJeongPark: I am open to either. I thought we had a conversation to move away from that to make it clear where the messages were coming from but I may be wrong. I don't have a personal preference with this.

Sources/BraintreeVenmo/BTVenmoAnalytics.swift
KunJeongPark: Great catch with that. Will change it. If we don't need specifics of how/where cancel happened with these anlytic messages, I am fine to change it later.

Sources/BraintreeVenmo/BTVenmoClient.swift
sshropshire: I do like the PPCP way. I think it's good our culture allows these discussions to make sure we're well aligned 🙏

KunJeongPark: So other analytic tickets we have finished so far - ApplePay, Card, PayPal, PayPalNativeCheckout, we've taken them out of the helper function to do it this way. Is this suggestion for all future change or start change with this ticket?

scannillo: > So other analytic tickets we have finished so far - ApplePay, Card, PayPal, PayPalNativeCheckout, we've taken them out of the helper function to do it this way. Is this suggestion for all future change or start change with this ticket?

It's up to us, really. You can finish the rest of the modules with the same format you're already doing, and then address the syntax all at once across the SDK. Or each subsequent module you pick up could be using the helper-method style. I would go with the former just to get things out the door, and then make tickets for refactoring with the helper methods. Also feel free to 2nd pass on it at stand-up if you want to get the whole teams opinion.

PR #996, opened 2023-03-31
Convert BTThreeDSecureRequest to Swift

jaxdesmarais:
This work is going into the `convert-three-d-secure` feature branch. We expect SPM to fail as the target contains a mixed Obj-C and Swift module. When the remaining conversion is complete SPM should pass as expected and we will open a PR to merge the feature branch into main.

It's also worth noting `pod lib lint` is failing only for 3DS right now - we spent half a day time boxing a fix to get cocoapods back in the green, but it's an interop issue with Swift > Obj-C > Swift, so we are going to ignore it for now since it will be a non-issue once PaymentFlow is converted to Swift.

### Summary of changes

- Convert `BTThreeDSecureRequest` to Swift

### Checklist

- ~[ ] Added a changelog entry~ no public interface changes in this PR

- @scannillo @jaxdesmarais


Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
```diff
@@ -16,12 +16,12 @@ class BTThreeDSecureAuthenticateJWT {
...
         guard let nonce = lookupResult?.tokenizedCard?.nonce else {
             apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.upgrade-payment-method.errored")
-            completion(nil, BTThreeDSecureError.failedAuthentication)
+            completion(nil, BTThreeDSecureError.failedAuthentication("Tokenized card nonce is required."))
             return
         }
 
         guard let urlSafeNonce = nonce.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) else {
-            completion(nil, BTThreeDSecureError.failedAuthentication)
+            completion(nil, BTThreeDSecureError.failedAuthentication("Tokenized card nonce is required."))
```
scannillo, line 24: Should this error message be specific to `addingPercentEncoding` failing? At this point we know `nonce` is non-nil.

scannillo (approved):
🔥

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -0,0 +1,377 @@
...
+    func stringFor(_ boolean: Bool) -> String {
+        boolean ? "true" : "false"
+    }
+}
+
+// MARK: - BTPaymentFlowRequestDelegate Protocol Conformance
+
+extension BTThreeDSecureRequest: BTPaymentFlowRequestDelegate {
+
+    /// :nodoc:
```
KunJeongPark, line 260: What does this mean?

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -0,0 +1,377 @@
...
+
+            if let error {
+                self.paymentFlowClientDelegate?.onPaymentComplete(nil, error: error)
+                return
+            }
+
+            var integrationError: Error?
+
+            if configuration?.cardinalAuthenticationJWT == nil {
+                NSLog("%@ BTThreeDSecureRequest versionRequested is 2, but merchant account is not setup properly.", BTLogLevelDescription.string(for: .critical))
```
KunJeongPark, line 281: So we are not dealing with version 1 anymore, right?

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -0,0 +1,377 @@
...
+                integrationError = BTThreeDSecureError.configuration("BTThreeDSecureRequest versionRequested is 2, but merchant account is not setup properly.")
+            }
+
+            if self.amount?.decimalValue.isNaN == true {
```
KunJeongPark, line 285: nice

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
jaxdesmarais: We use Jazzy for our [reference docs](https://braintree.github.io/braintree_ios/current/). `:nodoc:` is Jazzy's way of excluding docs from the public reference documentation that is not for merchant consumption/use. This is public because the protocol is public to be used in both the Payment Flow and 3D Secure module.

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
jaxdesmarais: Not in v6, no - v5 will need to return an error for merchants still attempting to use 3DSv1. All v1 classes were removed here: https://github.com/braintree/braintree_ios/pull/979

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
KunJeongPark, line 285: I'm still thinking about case where amount is nil.

KunJeongPark: I c in Playground that this doesn't return true if amount is nil. It returns a nil.

jaxdesmarais: That's odd that it returns nil but I guess makes sense if amount is optional. Added an additional or nil check along with the `isNaN` check here: https://github.com/braintree/braintree_ios/pull/996/commits/80a883725a3c3e69c0ea8f8ec8018040b8f538fa

KunJeongPark: I think it's weird too.

KunJeongPark (approved):

PR #997, opened 2023-03-31
Convert BTThreeDSecureClient to Swift

jaxdesmarais:
It's also worth noting `pod lib lint` is failing only for 3DS right now - we spent half a day time boxing a fix to get cocoapods back in the green, but it's an interop issue with Swift > Obj-C > Swift, so we are going to ignore it for now since it will be a non-issue once `PaymentFlow` is converted to Swift.

### Summary of changes

- Convert `BTThreeDSecureClient` to Swift
- Address TODOs
- Update podspec and Package.swift
- Update tests

`BTThreeDSecureClient+ThreeDSecure` is being converted 1:1 for now, there is a followup ticket when `PaymentFlow` is converted to create a dedicated `ThreeDSecureClient` that will use the `BTThreeDSecure` types and results directly.

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais


Sources/BraintreeThreeDSecure/BTPaymentFlowClient+ThreeDSecure.swift
```diff
@@ -0,0 +1,241 @@
...
+    ///   - request: The `BTPaymentFlowRequest` object where prepareLookup was called.
+    ///   - completion: This completion will be invoked exactly once with the client payload string or an error.
+    @objc(prepareLookup:completion:)
+    public func prepareLookup(
+        _ request: BTPaymentFlowRequest & BTPaymentFlowRequestDelegate,
+        completion: @escaping (String?, Error?) -> Void
+    ) {
+        let threeDSecureRequest = request as? BTThreeDSecureRequest
+
+        guard apiClient().clientToken != nil else {
```
scannillo, line 27: Why the `apiClient()` method call?

KunJeongPark (approved):

PR #998, opened 2023-03-31
Convert BTLocalPaymentResult to Swift

scannillo:
### Summary of changes

JIRA ticket DTBTSDK-2276

- Convert `BTLocalPaymentResult` to Swift
- Convert `BTPaymentFlowResult` to Swift
- Note:
   - SPM & CocoaPod builds expected to fail until `PaymentFlow` is fully converted

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


Sources/BraintreePaymentFlow/BTLocalPaymentResult.swift
```diff
@@ -0,0 +1,57 @@
+@objcMembers public class BTLocalPaymentResult: BTPaymentFlowResult {
+    
+    /// The billing address.
+    public let billingAddress: BTPostalAddress
```
KunJeongPark, line 4: Are all these fields non-optional?

scannillo: These are non-optional, yes. `BTLocalPaymentResult` is instantiated internally (see TODO below in file).

Sources/BraintreePaymentFlow/BTLocalPaymentResult.swift
```diff
@@ -0,0 +1,58 @@
+@objcMembers public class BTLocalPaymentResult: BTPaymentFlowResult {
```
KunJeongPark, line 1: I am not 100% sure when we need to import @BraintreeCore or not.

sshropshire (approved):

Sources/BraintreePaymentFlow/BTLocalPaymentResult.swift
scannillo: You import a module when you need to access it's classes & methods. In this case, `BTPostalAddress` is imported in parent class `BTPaymentFlowResult`.

The `canImport` syntax is for building our Demo app, and also for SPM _I believe_

KunJeongPark (approved):

PR #999, opened 2023-04-05
Removing weak self on native order creation client

jonathajones:
### Summary of changes

- The native checkout order creation client was being weakly retained by the `fetchOrReturnRemoteConfiguration` request. This causes the order creation client to be deallocated, as it was created locally up the chain. The solution here is to strongly capture `self` in the completion of `fetchOrReturnRemoteConfiguration` - this does not risk a retain cycle as the closure is not being retained by the client

### Checklist

- [ ] Added a changelog entry

- @jonathajones


scannillo (approved):
👏 Good catch!

After we merge this, I'll revert - https://github.com/braintree/braintree_ios/pull/993.

PR #1002, opened 2023-04-10
Convert `BTPaymentFlowClient` --> Swift

scannillo:
### Summary of changes

- Remove top level `BraintreePaymentFlow.h`
- Convert `BTPaymentFlowClient` --> Swift
- Convert `BTPaymentFlowRequestDelegate` --> Swift
- Convert `BTPaymentFlowClientDelegate` --> Swift
- Add explicit `import Foundation` to each file where accessed

Note: SPM & CocoaPod builds should now pass that entire SDK is in Swift 🥲 

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo


Demo/Application/Features/3D Secure/BraintreeDemoThreeDSecurePaymentFlowViewController.m
```diff
@@ -197,7 +197,7 @@ - (void)tappedToVerifyNewCard {
             self.callbackCount++;
             [self updateCallbackCount];
             if (error) {
-                if (error.code == BTPaymentFlowErrorTypeCanceled) {
+                if (error.code == 5) {
```
sshropshire, line 200: Can we assign `5` to a named variable?

Demo/Application/Features/Ideal/BraintreeDemoIdealViewController.m
```diff
@@ -69,7 +69,7 @@ - (void)startPaymentWithBank {
 
     void (^paymentFlowCompletionBlock)(BTPaymentFlowResult *, NSError *) = ^(BTPaymentFlowResult * _Nullable result, NSError * _Nullable error) {
         if (error) {
-            if (error.code == BTPaymentFlowErrorTypeCanceled) {
+            if (error.code == 5) {
```
sshropshire, line 72: +1

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
```diff
@@ -0,0 +1,138 @@
...
+
+@objcMembers public class BTPaymentFlowClient: NSObject {
+    
+    // MARK: - Internal Properties
+    
+    var authenticationSession: ASWebAuthenticationSession?
+    
+    // MARK: - Private Properies
+    
+    private let _apiClient: BTAPIClient
```
sshropshire, line 16:
```suggestion
    private let apiClient: BTAPIClient
```

sshropshire (approved):

Demo/Application/Features/3D Secure/BraintreeDemoThreeDSecurePaymentFlowViewController.m
scannillo: Yeh, this change is annoying since merchants no longer have access to BT code constants.

Throughout the re-write we aren't making [these Error files](https://github.com/braintree/braintree_ios/blob/convert-payment-flow-client/Sources/BraintreePaymentFlow/BTPaymentFlowError.swift) per module public. If we did, merchants could easily access these error codes like `BTPaymentFlowError.canceled.code`. I think we should audit/revisit the error strategy throughout each module.

Another thing too, this format was kept to for compatibility with [`NSError`](https://developer.apple.com/documentation/foundation/nserror) which has code & domain. The only property [`Error`](https://developer.apple.com/documentation/swift/error) has is a localized description.

scannillo: How important do we think error codes are to merchants, or are detailed error strings sufficient?

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
scannillo: This has to have a different name than `apiClient` to satisfy this [protocol requirement](https://github.com/braintree/braintree_ios/blob/7cc5fde720d2b65e3f0f3bbcfa7385c7c9f617ae/Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift#L119-L121) or else swift complains with `Invalid redeclaration of 'apiClient'`.

In a follow-up PR I think we can remove that protocol requirement entirely

Demo/Application/Features/Ideal/BraintreeDemoIdealViewController.m
scannillo: Addressed in the demo!

If we get merchant feedback re: wanting to expose the error codes, we can easily make them public.

Demo/Application/Features/3D Secure/BraintreeDemoThreeDSecurePaymentFlowViewController.m
scannillo: Addressed in the demo!

If we get merchant feedback re: wanting to expose the error codes, we can easily make them public.

jcnoriega (approved):

PR #1004, opened 2023-04-11
Sepa analytics audit

KunJeongPark:

Summary of changes

Update SEPADirectPayments analytic events to prepare switch to FPT
Update tests
Bucketed all analytic events sent before callbacks into 3 categories of final events: tokenizeFailed, tokenizeSucceeded, tokenizeCanceled.
Checklist

Authors
@KunJeongPark


Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -25,6 +32,13 @@ import BraintreeCore
         self.apiClient = apiClient
         self.sepaDirectDebitAPI = SEPADirectDebitAPI(apiClient: apiClient)
         self.webAuthenticationSession =  WebAuthenticationSession()
+        super.init()
+        NotificationCenter.default.addObserver(
```
KunJeongPark, line 36: This notification is not getting triggered on launch of the ASWebSession browser as it is in the PayPal client for some reason.

KunJeongPark: I put debugger on casting the webAuthenticationSession.start() method to cast `presentationContextProvider` as `BTSEPADirectDebitClient` and it seems to be set properly.

KunJeongPark: I am wondering if it is related to the fact that alert view doesn't show on launching webView as it is for PayPal module.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -328,7 +328,7 @@ import BraintreeDataCollector
                 case ASWebAuthenticationSessionError.canceledLogin:
                     // User canceled by breaking out of the PayPal browser switch flow
                     // (e.g. System "Cancel" button on permission alert or browser during ASWebAuthenticationSession)
-                    if self.returnedToAppAfterPermissionAlert == true {
+                    if self.returnedToAppAfterPermissionAlert == false {
```
KunJeongPark, line 331: I had a misunderstanding of this variable. It means that it had returned from launch of the webView, so true would mean it was canceled from browser.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
KunJeongPark: I tried saving reference to `ASWebAuthenticationSession` in the `WebAuthenticationSession` class to see if it was problem of scope outside of the start function, but that didn't fix the issue.

Sources/BraintreeSEPADirectDebit/WebAuthenticationSession.swift
```diff
@@ -12,17 +12,19 @@ class WebAuthenticationSession: NSObject {
     func start(
         url: URL,
         context: ASWebAuthenticationPresentationContextProviding,
-        completion: @escaping (URL?, Error?) -> Void
+        sessionDidDisplay: @escaping (Bool) -> Void,
+        sessionDidComplete: @escaping (URL?, Error?) -> Void
     ) {
-        self.authenticationSession = ASWebAuthenticationSession(
+        let authenticationSession = ASWebAuthenticationSession(
```
jcnoriega, line 18: shouldnt `var authenticationSession` be deleted then?

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -15,6 +15,13 @@ import BraintreeCore
     var webAuthenticationSession: WebAuthenticationSession
         
     var sepaDirectDebitAPI: SEPADirectDebitAPI
+    
+    /// Exposed for testing, for determining if ASWebAuthenticationSession was started
+    var isAuthenticationSessionStarted: Bool = false
```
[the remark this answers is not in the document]
KunJeongPark: Yes I think I should change it to private since `returnedToAppAfterPermissionAlert` is private.

Sources/BraintreeSEPADirectDebit/WebAuthenticationSession.swift
KunJeongPark: yes. I was testing out persisting reference in the class to see if I can get the system notification on `applicationDidBecomeActive` on browser launch but it didn't work.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -34,6 +48,9 @@ import BraintreeCore
         self.sepaDirectDebitAPI = sepaDirectDebitAPI
     }
 
+    @objc func applicationDidBecomeActive(notification: Notification) {
+        returnedToAppAfterPermissionAlert = isAuthenticationSessionStarted
```
sshropshire, line 52: q: is this needed only for analytics?

KunJeongPark: yes, for special cancel from the alert.

sshropshire: Would tracking this be valuable? The only concern I have is we're adding application code only for analytics. We may have an opportunity to simplify tracking if we remove this event.

sshropshire: Ah ok 👍 .

Sources/BraintreeSEPADirectDebit/WebAuthenticationSession.swift
```diff
@@ -12,17 +12,19 @@ class WebAuthenticationSession: NSObject {
...
             callbackURLScheme: BTCoreConstants.callbackURLScheme,
-            completionHandler: completion
+            completionHandler: sessionDidComplete
         )
 
-        authenticationSession?.prefersEphemeralWebBrowserSession = true
-        authenticationSession?.presentationContextProvider = context
-
-        authenticationSession?.start()
+        authenticationSession.prefersEphemeralWebBrowserSession = true
```
KunJeongPark, line 24: I found something interesting, that this line suppresses alert button from showing up and from system notification of applicationDidBecomeActive. I am not sure why. 
But I am wondering if we need this for some European compliance and if so, we would not need to deal with alert view cancel.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
KunJeongPark: I found the culprit, it was this line in WebAuthenticaionSession class: `authenticationSession.prefersEphemeralWebBrowserSession = true`
It suppresses the alert view and also trigger of system notification applicationDidBecomeActive.

KunJeongPark: So this prefersEphemeralWebBrowserSession option requests that the browser doesn't share cookies or other browsing data between the authentication session and the user's normal browser session. Whether the request is honored depends on the user's default web browser and Safari always honors the request. 

So I guess I will leave this option as well as the notification for the browser launching. Another question I had was why this option is enabled on SEPA but not on Local Payment?

scannillo: I don't have context on why `prefersEphemeralWebBrowserSession` was set to true, as I wasn't here when SEPA was added. Either it was a requirement from the SEPA team, or we just added it because we thought it was more secure.

My guess is the later.

scannillo: I vote to remove it, and follow the same pattern as PayPal and Local Payments. Great find!

KunJeongPark: I tested it with Chrome on my device and it also behaves the same as what I observe on the simulator but I guess the behavior might be different for different browser. I want to make sure that I am clear on any business decision for both SEPA or Local Payments.

Sources/BraintreePayPal/BTPayPalClient.swift
scannillo: Why is this changing from true to false?

scannillo: I believe this is working as expected on `master`

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -15,6 +15,13 @@ import BraintreeCore
     var webAuthenticationSession: WebAuthenticationSession
         
     var sepaDirectDebitAPI: SEPADirectDebitAPI
+    
+   
+    // MARK: - Private Properties
+
+    private var isAuthenticationSessionStarted: Bool = false
+
```
scannillo, line 23:
```suggestion
```

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -150,26 +179,31 @@ import BraintreeCore
         if let error = error {
             switch error {
             case ASWebAuthenticationSessionError.canceledLogin:
-                self.apiClient.sendAnalyticsEvent("ios.sepa-direct-debit.web-flow.canceled")
+                if !returnedToAppAfterPermissionAlert {
+                    self.apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.challengeAlertCanceled)
```
scannillo, line 183: Can we add the same comment that we have for the PayPalClient? The whole `returnedToAppAfterPermissionAlert` code is confusing and it helps future readers

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -135,10 +153,21 @@ import BraintreeCore
...
+            url: url,
+            context: context,
+            sessionDidDisplay: { [weak self] didDisplay in
+                if didDisplay {
+                    self?.apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.challengePresentationSucceeded)
+                    self?.isAuthenticationSessionStarted = true
+                } else {
+                    self?.apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.challengePresentationFailed)
+                }
+            },
```
scannillo, line 166: We need to add a unit test for this added logic. That said, all of the analytic event audit PRs need follow up Unit Test PRs. There are several events that are sent, but never validated for when they should be sent in a unit test.

Victoria - this is something we should make tickets for.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -135,10 +153,21 @@ import BraintreeCore
...
-        }
+        self.webAuthenticationSession.start(
+            url: url,
+            context: context,
+            sessionDidDisplay: { [weak self] didDisplay in
+                if didDisplay {
+                    self?.apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.challengePresentationSucceeded)
+                    self?.isAuthenticationSessionStarted = true
+                } else {
+                    self?.apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.challengePresentationFailed)
```
KunJeongPark, line 164: I think I also need to add two errors in the BTSepaDirectDebitError and add it in Changelog.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -15,6 +15,12 @@ import BraintreeCore
     var webAuthenticationSession: WebAuthenticationSession
         
     var sepaDirectDebitAPI: SEPADirectDebitAPI
+
```
scannillo, line 18: Uneeded new line
```suggestion
```

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
scannillo: @KunJeongPark - can you link the ticket you made for unit testing this?

Sources/BraintreePayPal/BTPayPalClient.swift
KunJeongPark: I tested this several times.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
KunJeongPark: https://paypal.atlassian.net/browse/DTBTSDK-2653 => I made one for each module that we did analytics audit for. This one is for SEPA.

Sources/BraintreePayPal/BTPayPalClient.swift
scannillo: Ok I see - The true case ([used in PayPal](https://github.com/braintree/braintree_ios/blob/495738201776c5ce79d907c4ac9511a5120e0a05/Sources/BraintreePayPal/BTPayPalClient.swift#L327)) means the cancel button on the web view was clicked.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
scannillo: Awesome, thank you!

Sources/BraintreePayPal/BTPayPalClient.swift
scannillo: Ok I see, so this is fixing a mistake introduced in[ this PR](https://github.com/braintree/braintree_ios/pull/977)

scannillo (approved):

PR #1014, opened 2023-04-18
PaymentFlow Analytics Audit

KunJeongPark:
Summary of changes

Update PaymentFlow analytic events to prepare switch to FPT
Update tests
Bucketed all analytic events sent before callbacks into 3 categories of final events: paymentFailed, paymentSucceeded, paymentCanceled.
Checklist
- [x] Added a changelog entry

Authors
@KunJeongPark


Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
```diff
@@ -223,27 +230,31 @@ extension BTLocalPaymentRequest: BTPaymentFlowRequestDelegate {
                 "sessionId": metadata?.sessionID
             ]
             
-            paymentFlowClientDelegate?.apiClient().post("/v1/payment_methods/paypal_accounts", parameters: params, completion: { [weak self] body, response, error in
```
KunJeongPark, line 226: I didn't understand why there was need to do this when apiClient reference is passed into this function. To show separation of concern?

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
```diff
@@ -20,23 +20,36 @@ import BraintreeCore
...
+       }
+    
     /// Starts a payment flow using a BTPaymentFlowRequest (usually subclassed for specific payment methods).
     /// - Parameters:
     ///   - request: A BTPaymentFlowRequest request.
     ///   - completionBlock: This completion will be invoked exactly once when the payment flow is complete or an error occurs.
     public func startPaymentFlow(_ request: BTPaymentFlowRequest & BTPaymentFlowRequestDelegate, completion: @escaping (BTPaymentFlowResult?, Error?) -> Void) {
         setupPaymentFlow(request, completion: completion)
-        _apiClient.sendAnalyticsEvent("ios.\(paymentFlowName).start-payment.selected")
+        sendAnalyticsEvent(BTPaymentFlowAnalytics.paymentStarted)
```
KunJeongPark, line 52: @scannillo I realize that I am sending "three-d-secure:start-payment:started" here for 3Ds flow which deviates from "3ds:verify:started" that you specified in the list of events for 3DS.

KunJeongPark: I guess I can have two sets of analytic event names in BTFlowAnalytics and use a helper function to send the appropriate name depending not he flowName.

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
```diff
@@ -93,23 +124,41 @@ extension BTPaymentFlowClient: BTPaymentFlowClientDelegate {
...
-        authenticationSession?.start()
+        isAuthenticationSessionStarted = authenticationSession?.start() ?? false
+        if isAuthenticationSessionStarted {
+            sendAnalyticsEvent(BTPaymentFlowAnalytics.browserPresentationSucceeded)
+        } else {
+            sendAnalyticsEvent(BTPaymentFlowAnalytics.browserPresentationFailed)
+            sendAnalyticsEvent(BTPaymentFlowAnalytics.paymentFailed)
+            onPaymentComplete(nil, error: BTPaymentFlowError.webSessionFailedToLaunch)
+            return
+        }
```
scannillo, line 161: This result of this boolean value should only be used for analytics, and should not call the merchant's completion ([see PPCP](https://github.com/paypal/iOS-SDK/blob/3be6c44b1b5fad08d219886cf2d3bfeb7435cf40/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift#L60-L66)).

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
```diff
@@ -167,7 +167,7 @@ extension BTLocalPaymentRequest: BTPaymentFlowRequestDelegate {
             apiClient.post("v1/local_payments/create", parameters: params) { body, response, error in
                 if let error {
                     if (error as NSError).code == BTCoreConstants.networkConnectionLostCode {
-                        apiClient.sendAnalyticsEvent("ios.local-payment-methods.network-connection.failure")
+                        apiClient.sendAnalyticsEvent("local-payment:\(BTPaymentFlowAnalytics.paymentNetworkConnectionLost)")
```
scannillo, line 170: How come this isn't using the helper method defined on line 265?

KunJeongPark: I just wanted to clarify that this is using apiClient parameter in the `handle` function as opposed to the calls in `handleOpen` function where apiClient is not passed in as a parameter, it is referenced as `paymentFlowClientDelegate?.apiClient()`. They are the same but I wanted to keep the same logic.

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
KunJeongPark: We followed the same pattern for PayPalClient. https://github.com/braintree/braintree_ios/pull/977/files#diff-7593900bd413ee81d48fae61aeeab899a00cbfdc0d2a688e21d3acf3f7ef3360R358 Isn't failure to launch this authentication page an end event?

UnitTests/BraintreePaymentFlowTests/BTLocalPayment_Tests.swift
```diff
@@ -222,7 +223,77 @@ class BTLocalPayment_UnitTests: XCTestCase {
...
+        mockWebAuthenticationSession.cannedErrorResponse = ASWebAuthenticationSessionError(
+            _bridgedNSError: NSError(
+                domain: ASWebAuthenticationSessionError.errorDomain,
+                code: ASWebAuthenticationSessionError.canceledLogin.rawValue,
+                userInfo: ["Description": "Mock cancellation error description."]
+            )
+        )
+        
         let client = BTPaymentFlowClient(apiClient: mockAPIClient)
+        client.returnedToAppAfterPermissionAlert = true
```
KunJeongPark, line 295: I have to figure out a way that this is toggled during the ASWebSession due to system notification.

KunJeongPark: This test was just to toss an idea but this test will always pass since I am resetting returnedToAppAfterPermit to false before the webSession launches in the code.

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
scannillo: It's fine if the 3DS events are wonky until you get to it in your 3DS PR!

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
```diff
@@ -256,3 +267,9 @@ extension BTLocalPaymentRequest: BTPaymentFlowRequestDelegate {
         return "\(paymentType).local-payment"
     }
 }
+
+extension BTLocalPaymentRequest {
+    var paymentFlowAPIClient: BTAPIClient? {
+        return paymentFlowClientDelegate?.apiClient()
+    }
+}
```
scannillo, line 275: Nitpick - I personally don't think we need an extension here

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
```diff
@@ -73,43 +87,57 @@ extension BTPaymentFlowClient: BTPaymentFlowClientDelegate {
...
+                    self?._apiClient.sendAnalyticsEvent(BTPaymentFlowAnalytics.browserPresentationSucceeded)
+                } else {
+                    self?._apiClient.sendAnalyticsEvent(BTPaymentFlowAnalytics.browserPresentationFailed)
+                }
+            },
+            sessionDidComplete: { url, error in
+                if let error = error as? NSError {
+                    if error.domain == ASWebAuthenticationSessionError.errorDomain,
+                       error.code == ASWebAuthenticationSessionError.canceledLogin.rawValue {
+                        // User canceled by breaking out of the PayPal browser switch flow
```
scannillo, line 116:
```suggestion
                        // User canceled by breaking out of the LocalPayment browser switch flow
```

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecure_Tests.swift
```diff
@@ -392,7 +392,7 @@ class BTThreeDSecure_UnitTests: XCTestCase {
         client.startPaymentFlow(threeDSecureRequest) { _, _ in }
 
         waitForExpectations(timeout: 4)
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.start-payment.selected"))
+ //       XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("three-d-secure:start-payment:started"))
```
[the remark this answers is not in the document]
KunJeongPark: Good idea. I was thinking about order of doing this. Wait for 3DS analytics to get merged?

UnitTests/BraintreePaymentFlowTests/BTLocalPayment_Tests.swift
```diff
@@ -222,7 +223,43 @@ class BTLocalPayment_UnitTests: XCTestCase {
...
+                "paymentResource": [
+                    "redirectUrl": "https://www.somebankurl.com",
+                    "paymentToken": "123aaa-123-543-777",
+                ]
+            ]
+        )
+
+        client.startPaymentFlow(localPaymentRequest) { _, _ in }
+
+        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTPaymentFlowAnalytics.browserPresentationSucceeded))
```
scannillo, line 244: This test can also assert that the `start` event was called

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
KunJeongPark: I've refactored this to return sessionDidDisplay result on successful launch. So we wouldn't be able to send this specific error at that point since we are not returning to merchant app here. I will delete addition of webSessionFailedToLaunch and keep BTPaymentFlowError.webSessionError(error)

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
```diff
@@ -73,43 +87,57 @@ extension BTPaymentFlowClient: BTPaymentFlowClientDelegate {
...
-                    self._apiClient.sendAnalyticsEvent("ios.\(self.paymentFlowName)).authsession.browser.cancel")
+        returnedToAppAfterPermissionAlert = false
+        webAuthenticationSession.start(
+            url: url,
+            context: self,
+            sessionDidDisplay: { [weak self] didDisplay in
+                if didDisplay {
+                    self?._apiClient.sendAnalyticsEvent(BTPaymentFlowAnalytics.browserPresentationSucceeded)
+                } else {
+                    self?._apiClient.sendAnalyticsEvent(BTPaymentFlowAnalytics.browserPresentationFailed)
```
KunJeongPark, line 109: If it's really important to be able to send `webSessionFailedToLaunch` error back to merchant at end of the error block, we can set boolean flag here and reset before calling the webSession.

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
```diff
@@ -63,6 +63,9 @@ import BraintreeDataCollector
     var paymentID: String?
     weak var paymentFlowClientDelegate: BTPaymentFlowClientDelegate?
     var correlationID: String?
+    var paymentFlowAPIClient: BTAPIClient? {
+        return paymentFlowClientDelegate?.apiClient()
```
jaxdesmarais, line 67: We can rely on an implicit return here and remove the `return` keyword. This has been our preference throughout the SDK.

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
```diff
@@ -63,6 +63,9 @@ import BraintreeDataCollector
     var paymentID: String?
     weak var paymentFlowClientDelegate: BTPaymentFlowClientDelegate?
     var correlationID: String?
+    var paymentFlowAPIClient: BTAPIClient? {
```
jaxdesmarais, line 66: nit: can we group these properties together? It looks like in other files we would do vars > computed vars > weak vars.

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
```diff
@@ -20,23 +16,37 @@ import BraintreeCore
     private var paymentFlowName: String {
         return paymentFlowRequestDelegate?.paymentFlowName() ?? "local-payments"
     }
-    
+    public var returnedToAppAfterPermissionAlert: Bool = false
+
     // MARK: - Public Methods
     
+    var webAuthenticationSession: WebAuthenticationSession
```
jaxdesmarais, line 23: This looks to be internal, is there a reason it's under the `Public Methods` mark?

Sources/BraintreePaymentFlow/WebAuthenticationSession.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+import AuthenticationServices
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+class WebAuthenticationSession: NSObject {
```
jaxdesmarais, line 8: This is a punt for sure, but I still think it makes sense to move this class into Core to prevent needing to make updates across several modules if we want to make changes to our `WebAuthenticationSession` since they all use the exact same logic. We could potentially be in a scenario where we make an update here but forget to in SEPA and have different behavior across modules that's unintended. Nothing to worry about for this particular PR though. 😄

UnitTests/BraintreePaymentFlowTests/BTLocalPayment_Tests.swift
```diff
@@ -3,6 +3,7 @@ import XCTest
 @testable import BraintreePaymentFlow
 @testable import BraintreeCore
 @testable import BraintreeTestShared
+import AuthenticationServices
```
jaxdesmarais, line 6: nit: move this above the `@testable` imports with the other standard imports

UnitTests/BraintreePaymentFlowTests/MockWebAuthenticationSession.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+import AuthenticationServices
+@testable import BraintreePaymentFlow
+
+class MockWebAuthenticationSession: WebAuthenticationSession {
```
jaxdesmarais, line 5: Similar comment here on punting on duplication - we can likely move `MockWebAuthenticationSession` into `BraintreeTestShared` and share logic to prevent duplication errors.

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
KunJeongPark: Not sure what you mean by group the properties?

Sources/BraintreePaymentFlow/WebAuthenticationSession.swift
KunJeongPark: I did think about that doing SEPA but I think we don't have the two separate completion handler as we do in PPCP.

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
KunJeongPark: you are right, we only need it as public if we move WebAuthenticationSession to Core module, I think,

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
jaxdesmarais: Right now we have:
```swift
var paymentID: String?
weak var paymentFlowClientDelegate: BTPaymentFlowClientDelegate?
var correlationID: String?
var paymentFlowAPIClient: BTAPIClient? { ... }
```

Should be:
```swift
var paymentID: String?
var correlationID: String?

var paymentFlowAPIClient: BTAPIClient? { ... }

weak var paymentFlowClientDelegate: BTPaymentFlowClientDelegate?
```

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
scannillo: We should style-guide that about implicit `return`

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
KunJeongPark: Done

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
KunJeongPark: It's done here.

UnitTests/BraintreePaymentFlowTests/MockWebAuthenticationSession.swift
scannillo: Do we have a JIRA for that work?

KunJeongPark: Not yet. I will make it and link here.

Sources/BraintreePaymentFlow/WebAuthenticationSession.swift
KunJeongPark: Here is the ticket I made for it. https://paypal.atlassian.net/browse/DTBTSDK-2688

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
jaxdesmarais: > We should style-guide that about implicit `return`

Added a note about implicit return to the [style guide](https://hackmd.io/kIyOnIFIRrKY4b5wQ0dUpw). Also added a section on use of `self` since we've talked about it here and there but not actually wrote it down.

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
jaxdesmarais: I don't see the error `webSessionFailedToLaunch` anywhere in this codebase currently aside from in the CHANGELOG and error enum. Would you mind clarifying where this error is being sent?

KunJeongPark: I see it  in our feature branch I think, on BTPayPalError.

KunJeongPark: I did have it in previous commit but @scannillo thought that we shouldn't return to merchant app at this point with an error but it should be handled in completion handler. In PayPal module, we did return with completion with error on failure to launch and sent that webSessionFailedToLaunch. So I was wondering what to do with that error. How we make it consistent. Sorry, it took some digging to last week me.

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
```diff
@@ -20,23 +16,39 @@ import BraintreeCore
...
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(applicationDidBecomeActive),
+            name: UIApplication.didBecomeActiveNotification,
+            object: nil
+        )
     }
     
+    // MARK: - Internal Methods
+    var webAuthenticationSession: WebAuthenticationSession
```
sshropshire, line 39: This should be under `// MARK: - Internal Properties` right?

sshropshire (approved):

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
jaxdesmarais: In PayPal it looks like we send `webSessionFailedToLaunch` if `isAuthenticationSessionStarted = false`. It looks like we don't use `isAuthenticationSessionStarted` in `PaymentFlow` at all so I don't know that we need to add it.

I think for now we can go with what we have but should have a follow up PR to align how we are using `WebAuthenticationSession` as a whole. We shouldn't need 3 different implementations of `WebAuthenticationSession`. I think this brings us back to an earlier discussion we had around having a shared `WebAuthenticationSession` in Core that we use across all modules since we have deviated and have 3 different implementations in SEPA vs PayPal vs PaymentFlow. Even if we don't move it into Core we likely can/should use the same implementation across the modules unless there is a compelling reason not to.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@
       * `.fetchConfigurationFailed` = 7
       * `.missingRedirectURL` = 8
       * `.missingReturnURL` = 9
+      * `.webSessionError` = 10
```
scannillo, line 33: This CHANGELOG entry needs to go under `unreleased`. Here, it was added under the 6.0.0-beta3 release, which already went out. The changes you're making in this PR apply to the next unreleased version, which will be tagged upon release.

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
```diff
@@ -59,10 +59,12 @@ import BraintreeDataCollector
     public weak var localPaymentFlowDelegate: BTLocalPaymentRequestDelegate?
     
     // MARK: - Internal Properties
-    
+    var paymentFlowAPIClient: BTAPIClient? {
```
scannillo, line 62: Nitpick
```suggestion
    var apiClient: BTAPIClient? {
```

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
```diff
@@ -167,7 +174,7 @@ extension BTLocalPaymentRequest: BTPaymentFlowRequestDelegate {
             apiClient.post("v1/local_payments/create", parameters: params) { body, response, error in
                 if let error {
                     if (error as NSError).code == BTCoreConstants.networkConnectionLostCode {
-                        apiClient.sendAnalyticsEvent("ios.local-payment-methods.network-connection.failure")
+                        apiClient.sendAnalyticsEvent(BTPaymentFlowAnalytics.paymentNetworkConnectionLost)
```
scannillo, line 177: This event is missing from our event spreadsheet. Can you add it?

Also, it is missing from SEPA & 3DS as well.

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
scannillo: ➕ 1

scannillo: Victoria - we don't want to send a tokenization fail event for the bool status of the ASWeb. That is because we are not calling the merchant's completion in that case. I don't think that bool value is reliable enough.

There is a discrepancy with the PayPal implementation [here](https://github.com/braintree/braintree_ios/blob/9cb3aa0b50b027e63f93a8abc4cda6492d897b1a/Sources/BraintreePayPal/BTPayPalClient.swift#L357). This PayPal code linked is incorrect and needs to be updated.

Not calling a merchant completion nor sending a tokenize failed event for the `asWeb.start()` bool result matches our [v5 implementation](https://github.com/braintree/braintree_ios/blob/8f34d3f1deada3559ad287d41d0fb7c65de2c9b3/Sources/BraintreePayPal/BTPayPalDriver.m#L299-L305).

KunJeongPark: The reason I made separate implementation here was that the PayPal implementation did not have separate completion for` sessionDidDisplay` and `sessionDidComplete`. At the time I did implementation for PayPal, I was not aware that we were conforming to PPCP way of doing this.

 I didn't think it would matter since if not failed here with merchant completion here, it will be caught in the completion as it is with the implementation without sending merchant completion fail here. I do agree that we want consistency but I think either way, failure to display returns to merchant app with fail eventually.

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
KunJeongPark: So we do want this for all of the modules? I was following the spreadsheet for other modules.

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
KunJeongPark: Here are questions I have remaining: 
1> Do we want to send networkConnectionLost analytics for all the modules? Sammy, I think for one of the modules, we decide not to and I had a lapse for LocalPayment module and included it.
2> Should I make a separate ticket for not returning to merchant completion right away with sessionDidDisplay false in PayPal module.?
3> Do we make a separate ticket for unified websession module?
Do we want to include refactoring PayPal module to use this wrapper class for ASWebSession instead of using ASWebSession directly?

jaxdesmarais: > 1> Do we want to send networkConnectionLost analytics for all the modules? Sammy, I think for one of the modules, we decide not to and I had a lapse for LocalPayment module and included it.

I think this is up to us to decide. The merchant that requested this to be added only uses PayPal and Venmo through us. It may make sense to add it to the missing modules in a follow up PR as it looks like we have it in about half the modules now.

> 2> Should I make a separate ticket for not returning to merchant completion right away with sessionDidDisplay false in PayPal module?

We should make a follow up PR as a whole to align our usage of ASWebAuthenticationSession as a whole. All of our modules depend on Core so I think it makes sense to move the implementation of our WebAuthenticationSession into there in a follow up PR.

> 3> Do we make a separate ticket for unified websession module?
Do we want to include refactoring PayPal module to use this wrapper class for ASWebSession instead of using ASWebSession directly?

Yes, this should be in a follow up PR and not in this PR.

KunJeongPark: Thank you Jax. So should those be two separate PR's - one for not returning to merchant fail completion after failure to display in PayPal module, and one for refactoring PayPal modulate make unified WebSession class?

KunJeongPark: And networkConnectionLost, if we decide to include, we should have another PR to add them all in.

jaxdesmarais: > Thank you Jax. So should those be two separate PR's - one for not returning to merchant fail completion after failure to display in PayPal module, and one for refactoring PayPal module to use wrapper class for web session and make unified WebSession class for all three modules?

Yep, we should have a follow up PR to align our usage and implementation of ASWebAuthenticationSession.

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
scannillo: I updated the spreadsheet to include networkLost events for all modules!

Sources/BraintreePaymentFlow/BTPaymentFlowError.swift
```diff
@@ -82,6 +87,8 @@ enum BTPaymentFlowError: Error, CustomNSError, LocalizedError {
             return "Failed to complete payment flow due to missing redirectURL."
         case .missingReturnURL:
             return "An error occured completing the payment authorization flow. The ASWebAuthenticationSession returned a nil URL."
+        case .webSessionError(let error):
+                   return "ASWebAuthenticationSession failed with \(error.localizedDescription)"
```
scannillo, line 91:
```suggestion
               return "ASWebAuthenticationSession failed with \(error.localizedDescription)"
```

UnitTests/BraintreePaymentFlowTests/BTPaymentFlowAnalytics_Tests.swift
```diff
@@ -0,0 +1,20 @@
...
+        XCTAssertEqual(BTPaymentFlowAnalytics.paymentCanceled, "local-payment:start-payment:browser-login:canceled")
+
+        XCTAssertEqual(BTPaymentFlowAnalytics.browserPresentationSucceeded, "local-payment:start-payment:browser-presentation:succeeded")
+        XCTAssertEqual(BTPaymentFlowAnalytics.browserPresentationFailed, "local-payment:start-payment:browser-presentation:failed")
+        XCTAssertEqual(BTPaymentFlowAnalytics.browserLoginAlertCanceled, "local-payment:start-payment:browser-login:alert-canceled")
+        XCTAssertEqual(BTPaymentFlowAnalytics.browserLoginFailed, "local-payment:start-payment:browser-login:failed")
+        XCTAssertEqual(BTPaymentFlowAnalytics.paymentNetworkConnectionLost, "local-payment:start-payment:network-connection:failed")
+    }
+}
+
```
scannillo, line 19:
```suggestion
```

scannillo (approved):

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
```diff
@@ -469,7 +469,8 @@ class BTThreeDSecureClient_Tests: XCTestCase {
         client.startPaymentFlow(threeDSecureRequest) { _, _ in }
 
         waitForExpectations(timeout: 4)
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.start-payment.selected"))
+        // TODO: Change to appropriate message after BTThreeDSecure analytics PR is merged
+        //XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("three-d-secure:start-payment:started"))
```
jaxdesmarais, line 473: We should make sure to address these TODOs and commented out tests

Sources/BraintreePaymentFlow/BTLocalPaymentRequest.swift
KunJeongPark: Here is the Jira ticket for adding networkConnectionLost event for all modules: https://paypal.atlassian.net/browse/DTBTSDK-2711

Sources/BraintreePaymentFlow/BTPaymentFlowClient.swift
jaxdesmarais: Resolving: ticket for unified ASWeb logic is DTBTSDK-2712

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
KunJeongPark: I will take care of this in the 3DS analytic PR after merging in the updated fpti-feature there.

PR #1016, opened 2023-04-19
Updating PayPalCheckout to 0.112.1

jonathajones:
### Summary of changes

- This PR bumps the version of PayPalCheckout from `0.110.0` to `0.112.1`. This version would be the version we use for the GA within Braintree.
- Updates include:
  - The AddCard flow includes address inputs
  - Support for adding a card natively if the user has no available FI
  - Increments minimum xcode version to `14.1`
  - Increments Swift version to `5.7.1`
  - Increments minimum supported iOS version to `13`
  - Layout fixes 
  - Misc. bugfixes

I will add more detailed notes to the changelog if we need - the only merchant / user facing change included here would be the add card changes. 

### Checklist

- [X] Added a changelog entry

- @jonathajones


scannillo:
`master` is our V6 beta, while `5.x` is our branch for v5 stable releases, so we'll want this PR to target `5.x` instead.

jonathajones:
Reopening this PR to target Braintree V6 release with Native Checkout GA

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## Unreleased
```
scannillo, line 3:
```suggestion
## unreleased
```

scannillo:
Why is this marked as "do not release"?

Package.swift
```diff
@@ -90,8 +90,8 @@ let package = Package(
         ),
         .binaryTarget(
             name: "PayPalCheckout",
-            url: "https://github.com/paypal/paypalcheckout-ios/releases/download/0.110.0/PayPalCheckout.xcframework.zip",
-            checksum: "e9895c202b090a7bde5c47685e96aef68b6f334e3f3798a79dd49b32c81fe130"
+            url: "https://github.com/paypal/paypalcheckout-ios/releases/download/0.112.1/PayPalCheckout.xcframework.zip",
+            checksum: "b1da7584ccb2964e3bddab1871e7f4b4121b4639f7de91b3988371a7e2fce546"
```
jcnoriega, line 94: Im updating MXO to the latest version in PPCP. Whats the diference between the way its done here than what we are doing in PPCP? [Here is the diff](https://github.com/paypal/iOS-SDK/pull/142/files#diff-f913940c58e8744a2af1c68b909bb6383e49007e6c5a12fb03104a9006ae677eR38).

jcnoriega (approved):

Package.swift
jonathajones: Good question - i think adding it as a dependency would be preferable, but it may be treated as a dependency for the entire package (rather than a dependency for just one target). @scannillo may have more info.

jaxdesmarais: Yeah, so when we link it as a dependency (in PPCP) vs a binary target (in BT) we end up pulling it in for everything. So for example when we link the library as a dependency in BT, if a merchant was using just Card for example, it would still link NXO even though it wasn't being used by that module.

Likely, we want to take a similar approach on PPCP that we opted to take on BT and move to a binary target, but we can discuss this further as a team.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPalNativeCheckout
+  * Update NativeCheckout version from 0.110.0 to 0.112.1
```
jaxdesmarais, line 5: Do we think it's worth adding a note about the `BraintreePayPalNativeCheckout` now being GA? We will also need to update the public dev docs as they note that is in beta currently.

jaxdesmarais (commented):
Right now on the BT side we target swift `5.7`, I see in the PR description NXO requires `5.7.1`. Would would happen to merchants using `5.7.0` (supported by BT) but attempting to import NXO? Is there a specific change in the minor version that NXO needed?

jonathajones:
> Right now on the BT side we target swift `5.7`, I see in the PR description NXO requires `5.7.1`. Would would happen to merchants using `5.7.0` (supported by BT) but attempting to import NXO? Is there a specific change in the minor version that NXO needed?

All app store submissions are required to be done with Xcode 14.1 as of April 25 (https://developer.apple.com/news/?id=jd9wcyov) - Swift 5.7.1 is the swift version compiled with Xcode 14.1.

So i think the requirements should be updated, as merchants couldn't use an Xcode version lower than 14.1

scannillo:
> Pro tip, don't take a sabbatical if you don't want to miss Apple announcements. 🙈 I can make a ticket on our end for this (assuming @scannillo hasn't already). Still catching up so it's very possible this was already done.

Already done! https://github.com/braintree/braintree_ios/pull/1018

scannillo:
@jonathajones - are we 🟢 ✅ OK to merge/release this yet?

jonathajones:
> are we 🟢 ✅ OK to merge/release this yet?

Not yet - I'll need to update this release with another version. I'll reopen when it's ready.

PR #1024, opened 2023-04-28
Remove `BTAnalyticsMetadata`

scannillo:
JIRA DTBTSDK-2683

### Summary of changes

- Remove sending **all** analytics metadata in every gateway request
   - Previously we were sending things like `isSimulator`, `iosPackageManager` version etc. 
   - These things are only relevant for our analytic events, and not useful to the GW
- Mirror `_meta` params sent in  [JS](https://github.com/braintree/braintree-web/blob/34d75482492e352062d5dbf95d8d0b21a60eb9c3/test/payment-request/unit/external/payment-request.js#L832) and [Android](https://github.com/braintree/braintree_android/blob/d5015fea5ae9a205fbcdd36eb3651a43fea63c39/BraintreeCore/src/main/java/com/braintreepayments/api/MetadataBuilder.kt#L64-L68)
    - Add `version` and `platform` keys (which used to live in `BTAnalyticsMetadata`)

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


Sources/BraintreeCore/Analytics/BTAnalyticsMetadata.swift
```diff
@@ -1,89 +0,0 @@
...
-        }
-        return dropInVersion
-    }
-
-    // MARK: - Construct Metadata
-
-    static var metadata: [String: Any] {
-        [
-            "platform": platform,
-            "platformVersion": platformVersion,
```
sshropshire, line 74: q: does this remove OS version tracking?

sshropshire (approved):

Sources/BraintreeCore/Analytics/BTAnalyticsMetadata.swift
scannillo: Client OS is still sent [here](https://github.com/braintree/braintree_ios/blob/67cb8db8fd26f0fc294c5e0c41ba0326166c8c79/Sources/BraintreeCore/Analytics/FPTIBatchData.swift#L107), in the analytics metadata.

We were sending our "AnalyticsMetadata" in every single API request to the gateway, for example tokenizing a card. We were sending things like, `isSimulator`, `iosPackageManager`, etc - things not relevant to tokenization. I am gonna guess that it started off that way since a few things did overlap with analytics metadata values, like platform and SDK version.

Let me know if that makes sense.

sshropshire: Got it yeah I was confused because it looks like a pure deletion in the diff. I wasn't aware it was duplication.

Sources/BraintreeCore/BTClientMetadata.swift
```diff
@@ -37,7 +37,9 @@ import Foundation
         [
             "integration": self.integrationString,
             "source": self.sourceString,
-            "sessionId": self.sessionID
+            "sessionId": self.sessionID,
+            "platform": "iOS",
```
KunJeongPark, line 41: Yay! I was worried where that was going to come from.

KunJeongPark (approved):

PR #1010, opened 2023-04-13
braintreeCore audit analytics

KunJeongPark:
Summary of changes
These changes are going into the fpti-feature branch which will get merged once we audit all modules and make the switch to the FPTI endpoints

Update BraintreeCore analytic events to prepare switch to FPT
-Deleted queueAnalyticsEvent events and tests

Checklist
~- [ ] Added a changelog entry~

Authors
@KunJeongPark

PR #1025, opened 2023-05-01
ThreeDSecure analytics audit

KunJeongPark:
Summary of changes

Update BTThreeDSecure analytic events to prepare switch to FPT
Update tests

Authors
@KunJeongPark

### DIFF
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureAnalytics.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureAnalytics.swift
@@ -1,0 +1,15 @@
+import Foundation
+
+class BTThreeDSecureAnalytics {
+    
+    // MARK: Conversion Events
+    
+    static let verifyStarted = "3ds:verify:started"
+    static let verifyFailed = "3ds:verify:failed"
+    
+    // MARK: Others
+    static let challengeRequired = "3ds:verify:challenge-required"
+    static let challengeSucceeded = "3ds:verify:challenge.succeeded"
+    static let challengeFailed = "3ds:verify:challenge.failed"
+    static let jwtAuthSucceeded = "3ds:verify:authenticate-jwt:succeeded"
+}
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
@@ -57,7 +57,7 @@ class BTThreeDSecureAuthenticateJWT {
                 // so that merchants can transact with the lookup nonce if desired.
                 threeDSecureResult.tokenizedCard = lookupResult?.tokenizedCard
             }
-
+            apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.jwtAuthSucceeded)
             completion(threeDSecureResult, nil)
             return
         }
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
@@ -29,7 +29,7 @@ import BraintreeCore
     ///   - request: A BTThreeDSecureRequest request.
     ///   - completionBlock: This completion will be invoked exactly once when the 3DS flow is complete or an error occurs.
     public func startPaymentFlow(_ request: BTThreeDSecureRequest, completion: @escaping (BTThreeDSecureResult?, Error?) -> Void) {
-        apiClient.sendAnalyticsEvent("ios.three-d-secure.start-payment.selected")
+        apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.verifyStarted)
         
         self.request = request
         self.merchantCompletion = completion
@@ -235,13 +235,10 @@ import BraintreeCore
     }
         
     private func start(request: BTThreeDSecureRequest, configuration: BTConfiguration) {
-        apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.started")
-                
         performThreeDSecureLookup(request) { lookupResult, error in
             DispatchQueue.main.async {
                 guard let lookupResult, error == nil else {
-                    self.apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.failed")
-                    self.apiClient.sendAnalyticsEvent("ios.three-d-secure.start-payment.failed")
+                    self.apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.verifyFailed)
                     self.merchantCompletion?(nil, error)
                     return
                 }
@@ -251,7 +248,9 @@ import BraintreeCore
 
                 self.request?.threeDSecureRequestDelegate?.onLookupComplete(request, lookupResult: lookupResult) {
                     let requiresUserAuthentication = lookupResult.lookup?.requiresUserAuthentication ?? false
-                    self.apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.challenge-presented.\(self.stringFor(requiresUserAuthentication))")
+                    if requiresUserAuthentication {
+                        self.apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.challengeRequired)
+                    }
                     self.process(lookupResult: lookupResult, configuration: configuration)
                 }
             }
@@ -260,6 +259,8 @@ import BraintreeCore
     
     private func process(lookupResult: BTThreeDSecureResult, configuration: BTConfiguration) {
         if lookupResult.lookup?.requiresUserAuthentication == false || lookupResult.lookup == nil {
+            // looup succeeded
+            // verify succeeded
             merchantCompletion?(lookupResult, nil)
             return
         }
@@ -272,11 +273,12 @@ import BraintreeCore
     private func performV2Authentication(with lookupResult: BTThreeDSecureResult) {
         threeDSecureV2Provider?.process(lookupResult: lookupResult) { result, error in
             guard let result else {
-                self.apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.failed")
+                self.apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.verifyFailed)
                 self.merchantCompletion?(nil, error)
                 return
             }
 
+            self.apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.challengeSucceeded)
             self.logThreeDSecureCompletedAnalytics(forResult: lookupResult, apiClient: self.apiClient)
             self.merchantCompletion?(result, error)
         }
--- a/UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureAnalytics_Tests.swift
+++ b/UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureAnalytics_Tests.swift
@@ -1,0 +1,15 @@
+import XCTest
+@testable import BraintreeThreeDSecure
+
+final class BTThreeDSecureAnalytics_Tests: XCTestCase {
+    
+    func test_ThreeDSecureAnalyticsEvents_sendsExpectedEventNames() {
+        XCTAssertEqual(BTThreeDSecureAnalytics.verifyStarted, "3ds:verify:started")
+        XCTAssertEqual(BTThreeDSecureAnalytics.verifyFailed, "3ds:verify:failed")
+        XCTAssertEqual(BTThreeDSecureAnalytics.challengeRequired, "3ds:verify:challenge-required")
+        XCTAssertEqual(BTThreeDSecureAnalytics.challengeSucceeded, "3ds:verify:challenge.succeeded")
+        XCTAssertEqual(BTThreeDSecureAnalytics.challengeFailed, "3ds:verify:challenge.failed")
+        XCTAssertEqual(BTThreeDSecureAnalytics.jwtAuthSucceeded, "3ds:verify:authenticate-jwt:succeeded")
+        
+    }
+}
--- a/UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
+++ b/UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
@@ -469,13 +469,10 @@ class BTThreeDSecureClient_Tests: XCTestCase {
         client.startPaymentFlow(threeDSecureRequest) { _, _ in }
 
         waitForExpectations(timeout: 4)
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.start-payment.selected"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.initialized"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.verification-flow.started"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.verification-flow.3ds-version.1.0"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.verification-flow.challenge-presented.true"))
+        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTThreeDSecureAnalytics.verifyStarted))
+        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTThreeDSecureAnalytics.challengeRequired))
     }
-
+    
     func testStartPayment_failure_sendsAnalyticsEvents() {
         let client = BTThreeDSecureClient(apiClient: mockAPIClient)
         mockAPIClient.cannedConfigurationResponseBody = BTJSON(value: [
@@ -491,11 +488,8 @@ class BTThreeDSecureClient_Tests: XCTestCase {
         }
 
         waitForExpectations(timeout: 4)
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.start-payment.selected"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.initialized"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.verification-flow.started"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.verification-flow.failed"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.start-payment.failed"))
+        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTThreeDSecureAnalytics.verifyStarted))
+        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTThreeDSecureAnalytics.verifyFailed))
     }
 
     // MARK: - prepareLookup
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -19,6 +19,8 @@
 		3BC0E09429E46C58009217D6 /* BTSEPADirectAnalytics_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BC0E09329E46C58009217D6 /* BTSEPADirectAnalytics_Tests.swift */; };
 		3BF7FFC829D123F9005CA469 /* BTPayPalNativeCheckoutAnalytics.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BF7FFC729D123F9005CA469 /* BTPayPalNativeCheckoutAnalytics.swift */; };
 		3BF7FFCC29D1D4D0005CA469 /* BTPayPalNativeCheckoutAnalytics_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BF7FFC929D1D480005CA469 /* BTPayPalNativeCheckoutAnalytics_Tests.swift */; };
+		3BEB03C529FD55CA001133D5 /* BTThreeDSecureAnalytics.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BEB03C429FD55CA001133D5 /* BTThreeDSecureAnalytics.swift */; };
+		3BEB03C829FD5716001133D5 /* BTThreeDSecureAnalytics_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BEB03C629FD5669001133D5 /* BTThreeDSecureAnalytics_Tests.swift */; };
 		3DAB5933286B40E2003A7BC5 /* BTPayPalNativeCheckoutRequest_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3DAB5932286B40E2003A7BC5 /* BTPayPalNativeCheckoutRequest_Tests.swift */; };
 		3DAB5937286B9C6E003A7BC5 /* BTPayPalNativeHermesResponse_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3DAB5936286B9C6E003A7BC5 /* BTPayPalNativeHermesResponse_Tests.swift */; };
 		3DAB5939286BA3D0003A7BC5 /* BTPayPalNativeTokenizationClient_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3DAB5938286BA3D0003A7BC5 /* BTPayPalNativeTokenizationClient_Tests.swift */; };
@@ -605,6 +607,8 @@
 		3BC0E09329E46C58009217D6 /* BTSEPADirectAnalytics_Tests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTSEPADirectAnalytics_Tests.swift; sourceTree = "<group>"; };
 		3BF7FFC729D123F9005CA469 /* BTPayPalNativeCheckoutAnalytics.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalNativeCheckoutAnalytics.swift; sourceTree = "<group>"; };
 		3BF7FFC929D1D480005CA469 /* BTPayPalNativeCheckoutAnalytics_Tests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalNativeCheckoutAnalytics_Tests.swift; sourceTree = "<group>"; };
+		3BEB03C429FD55CA001133D5 /* BTThreeDSecureAnalytics.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTThreeDSecureAnalytics.swift; sourceTree = "<group>"; };
+		3BEB03C629FD5669001133D5 /* BTThreeDSecureAnalytics_Tests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTThreeDSecureAnalytics_Tests.swift; sourceTree = "<group>"; };
 		3DAB5932286B40E2003A7BC5 /* BTPayPalNativeCheckoutRequest_Tests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalNativeCheckoutRequest_Tests.swift; sourceTree = "<group>"; };
 		3DAB5936286B9C6E003A7BC5 /* BTPayPalNativeHermesResponse_Tests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalNativeHermesResponse_Tests.swift; sourceTree = "<group>"; };
 		3DAB5938286BA3D0003A7BC5 /* BTPayPalNativeTokenizationClient_Tests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalNativeTokenizationClient_Tests.swift; sourceTree = "<group>"; };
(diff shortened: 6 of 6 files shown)


Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -260,6 +259,8 @@ import BraintreeCore
     
     private func process(lookupResult: BTThreeDSecureResult, configuration: BTConfiguration) {
         if lookupResult.lookup?.requiresUserAuthentication == false || lookupResult.lookup == nil {
```
KunJeongPark, line 261: I didn't understand the `lookupResult.lookup == nil` condition. Is this considered the same as acsURL being nil? This `lookupResult.lookup == nil` is not an error in this flow?

jaxdesmarais: This is just saying if the `lookupResult.lookup == nil` we didn't get back a 3DS lookup object so can continue as if we aren't requiring 3D Secure.

If we look at the 2 call sites for this method:
```swift
private func start(request: BTThreeDSecureRequest, configuration: BTConfiguration) {
    ...
    guard let lookupResult, error == nil else { ... }
    ...
    self.process(lookupResult: lookupResult, configuration: configuration)
}
```
Here we would always have a `lookupResult.lookup` since we are doing a guard on that property to check that it's not nil.

The second call site for this method is here:
```swift
public func initializeChallenge(
        lookupResponse: String,
        request: BTThreeDSecureRequest,
        completion: @escaping (BTThreeDSecureResult?, Error?) -> Void
    ) {
    ...
    let jsonResponse = BTJSON(data: dataResponse)
    let lookupResult = BTThreeDSecureResult(json: jsonResponse)
    ...
    self.process(lookupResult: lookupResult, configuration: configuration)
}
```
In this case (server side 3DS calls for large merchants) there could be an instance where when creating the `BTThreeDSecureResult` from the JSON data passed from the merchant it wouldn't contain the `lookupResult.lookup` property and thus hit the check `lookupResult.lookup == nil` in this method.

Sources/BraintreeThreeDSecure/BTThreeDSecureAnalytics.swift
```diff
@@ -0,0 +1,18 @@
...
+    static let verifySucceeded = "3ds:verify:succeeded"
+    static let verifyFailed = "3ds:verify:failed"
+    
+    // MARK: Others
+    static let challengeRequired = "3ds:verify:challenge-required"
+    static let challengeSucceeded = "3ds:verify:challenge.succeeded"
+    static let challengeFailed = "3ds:verify:challenge.failed"
+    static let jwtAuthSucceeded = "3ds:verify:authenticate-jwt:succeeded"
+    
+    static let lookupSucceeded = "3ds:verify:lookup:succeeded"
```
jaxdesmarais, line 17:
```suggestion
    // MARK: Others
    
    static let challengeRequired = "3ds:verify:challenge-required"
    static let challengeSucceeded = "3ds:verify:challenge.succeeded"
    static let challengeFailed = "3ds:verify:challenge.failed"
    static let jwtAuthSucceeded = "3ds:verify:authenticate-jwt:succeeded"
    static let lookupSucceeded = "3ds:verify:lookup:succeeded"
```

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
KunJeongPark: What I was wondering was is it an expected behavior that lookup object is nil? Would that be something intentional? If not, would it be an error? I see that we set requiresAuthentication field to the acsURL being nil. Above, we are not doing guard against lookupResult.lookup but lookupResult itself.

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
```diff
@@ -469,13 +469,60 @@ class BTThreeDSecureClient_Tests: XCTestCase {
...
+        mockAPIClient.cannedResponseBody = BTJSON(value: responseBody)
+
+        let client = BTThreeDSecureClient(apiClient: mockAPIClient)
+        client.startPaymentFlow(threeDSecureRequest) { _, _ in }
+
+        waitForExpectations(timeout: 4)
+        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTThreeDSecureAnalytics.verifyStarted))
+        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTThreeDSecureAnalytics.lookupSucceeded))
+        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTThreeDSecureAnalytics.verifySucceeded))
+        XCTAssertFalse(mockAPIClient.postedAnalyticsEvents.contains(BTThreeDSecureAnalytics.challengeRequired))
```
KunJeongPark, line 523: To get end events in success case , I think we need to mock BTThreeDSecureV2Provider?

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
```diff
@@ -469,13 +469,60 @@ class BTThreeDSecureClient_Tests: XCTestCase {
         client.startPaymentFlow(threeDSecureRequest) { _, _ in }
 
         waitForExpectations(timeout: 4)
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.start-payment.selected"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.initialized"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.verification-flow.started"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.verification-flow.3ds-version.1.0"))
-        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("ios.three-d-secure.verification-flow.challenge-presented.true"))
+        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTThreeDSecureAnalytics.verifyStarted))
+        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTThreeDSecureAnalytics.challengeRequired))
```
KunJeongPark, line 473: To get end events in success case with authentication required , I think we need to mock BTThreeDSecureV2Provider?

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
jaxdesmarais: > Above, we are not doing guard against lookupResult.lookup but lookupResult itself.

Ah, yep you are right - see still getting my brain back. 😄 

> What I was wondering was is it an expected behavior that lookup object is nil? Would that be something intentional? If not, would it be an error? I see that we set requiresAuthentication field to the acsURL being nil.

It can be `nil` without causing issues. It would just mean that we wouldn't attempt a 3D Secure challenge. But there are any number of reasons that we may not get this object back and as such treat it the same way we treat not requiring 3D Secure. It would not be an error, it just means we don't attempt or launch a 3D Secure challenge.

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
KunJeongPark, line 473: To test end events here, we'd need to create a mock of BTThreeDSecureV2Provider.

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
KunJeongPark: I'm just glad that you are back, Jax. We missed you. I am a bit confused. In the OR condition in `process` function, that block returns success. Shouldn't success here be intentional, such as lookup with no acsURL?

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -88,17 +90,20 @@ import BraintreeCore
         self.request = request
         
         guard apiClient.clientToken != nil else {
+            self.apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.verifyFailed)
```
KunJeongPark, line 93: I wasn't sure about the flow in the case merchant calls this function directly. In the flow where the private function `prepareLookup` gets called form `startPaymentflow` function, we send `verifyFailed` analytic messages from completion handler of the private` prepareLookup` function.

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -272,26 +279,20 @@ import BraintreeCore
     private func performV2Authentication(with lookupResult: BTThreeDSecureResult) {
         threeDSecureV2Provider?.process(lookupResult: lookupResult) { result, error in
             guard let result else {
-                self.apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.failed")
+                if let nsError = error as NSError?, nsError.code == BTThreeDSecureError.canceled.errorCode {
+                    self.apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.verifyCanceled)
+                } else {
```
KunJeongPark, line 284: Both cancel and error cases get sent as an error from ThreDSecureV2Provider's process function, so instead of sending all the messages separately, I just distinguished the cases here.

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
```diff
@@ -118,11 +116,24 @@ extension BTThreeDSecureV2Provider: CardinalValidationDelegate {
...
+        case .failure:
+            apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.challengeFailed)
+            BTThreeDSecureAuthenticateJWT.authenticate(
+                jwt: serverJWT,
+                withAPIClient: apiClient,
+                forResult: lookupResult,
+                completion: completionHandler
+            )
+        case .noAction:
             BTThreeDSecureAuthenticateJWT.authenticate(
```
KunJeongPark, line 137: What is this case? Is it failure or success?

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
KunJeongPark: Should "cancel" from `ThreeDSecureV2Provider` return` challengeFailed`? I think no.  Those are sent from BTThreeDSecureProvider class's `CardinalValidationDelegate` function.

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -235,23 +241,20 @@ import BraintreeCore
...
         
     private func start(request: BTThreeDSecureRequest, configuration: BTConfiguration) {
-        apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.started")
-                
         performThreeDSecureLookup(request) { lookupResult, error in
             DispatchQueue.main.async {
                 guard let lookupResult, error == nil else {
-                    self.apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.failed")
-                    self.apiClient.sendAnalyticsEvent("ios.three-d-secure.start-payment.failed")
+                    self.apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.lookupFailed)
```
KunJeongPark, line 247: This is a duplication. I will remove in next commit.

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
```diff
@@ -118,11 +116,24 @@ extension BTThreeDSecureV2Provider: CardinalValidationDelegate {
...
+        case .success:
+            apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.challengeSucceeded)
+            BTThreeDSecureAuthenticateJWT.authenticate(
+                jwt: serverJWT,
+                withAPIClient: apiClient,
+                forResult: lookupResult,
+                completion: completionHandler
+            )
+        case .failure:
+            apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.challengeFailed)
```
KunJeongPark, line 129: Is ".failure" from Cardinal considered "challengeFailed". The` challengeFailed` analytic description in our Excel doc says "Fired when the Cardinal SDK returns an error in completing the 3DS challenge". Both are `challengeFailed`, right?

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
KunJeongPark: I think we need to make MockCardinalSession and change BTThreeDSecureVwProvider to take a protocol, CardinalSessionProtocol. I tried to run unit tests that require cardinal session to complete but tests quit out on continueWith function.

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
jaxdesmarais: Based on the above docs I listed digging into the website link it says the following for this status:
```
The API calls to Centinel API were completed and there is no further actionable items to complete. This can indicate that the card holder is not enrolled in 3-D Secure or it could indicate a validation error was encountered at Centinel Core. It is recommended that the rest of the response is reviewed to determine what has occurred.
```
It sounds like this could be a success or a failure which is... confusing at best. For the sake of analytics I wonder if we need another event for this since it doesn't seem to fit nicely anywhere. What do you think @KunJeongPark and @scannillo?

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
KunJeongPark: I think it's same pattern in PayPal except we deal with ASWebAuthentication cancelLogin.

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
KunJeongPark: I thought checking card holder is enrolled or not is done at `perform3DSLookup`, which is before getting to this point. 
 "validation error" means user's info is not right or some glitch happened to error out during validation?

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
KunJeongPark: So for analytics, we don't distinguish service level error and error based on user's input of credentials?

KunJeongPark: Thank you for looking those up. I will do that first in the future.

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
```diff
@@ -12,15 +12,14 @@ class BTThreeDSecureAuthenticateJWT {
...
-
         guard let nonce = lookupResult?.tokenizedCard?.nonce else {
-            apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.upgrade-payment-method.errored")
+            apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.jwtAuthFailed)
             completion(nil, BTThreeDSecureError.failedAuthentication("Tokenized card nonce is required."))
             return
         }
 
         guard let urlSafeNonce = nonce.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) else {
+            apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.jwtAuthFailed)
```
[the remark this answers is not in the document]
KunJeongPark: the completion handler from the client (threeDSecureV2Provider?.process) handles general events.

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
```diff
@@ -32,32 +31,28 @@ class BTThreeDSecureAuthenticateJWT {
             parameters: requestParameters
         ) { body, _, error in
             if let error = error as NSError? {
-                if error.code == BTCoreConstants.networkConnectionLostCode {
-                    apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.network-connection.failure")
-                }
-
-                apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.upgrade-payment-method.errored")
+                apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.jwtAuthFailed)
```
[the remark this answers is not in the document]
KunJeongPark: same answer as above.

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -272,26 +275,20 @@ import BraintreeCore
...
+                    self.apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.verifyCanceled)
+                } else {
+                    self.apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.verifyFailed)
+                }
                 self.merchantCompletion?(nil, error)
                 return
             }
 
-            self.logThreeDSecureCompletedAnalytics(forResult: lookupResult, apiClient: self.apiClient)
+            self.apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.verifySucceeded)
```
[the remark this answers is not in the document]
KunJeongPark: I think the guard statement above should be `guard let result, error = nil else {`

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
```diff
@@ -12,15 +12,14 @@ class BTThreeDSecureAuthenticateJWT {
         forResult lookupResult: BTThreeDSecureResult?,
         completion: @escaping (BTThreeDSecureResult?, Error?) -> Void
     ) {
-        apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.upgrade-payment-method.started")
-
         guard let nonce = lookupResult?.tokenizedCard?.nonce else {
-            apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.upgrade-payment-method.errored")
+            apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.jwtAuthFailed)
```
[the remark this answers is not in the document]
scannillo: ➕ 1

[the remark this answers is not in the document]
scannillo: I re-raised this concern [here](https://github.com/braintree/braintree_ios/pull/994#discussion_r1160073989), but I think we should move to the `notify` pattern like PPCP. It's less error-prone.

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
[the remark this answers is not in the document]
scannillo: I think this is confusing - because now our modules aren't acting the same way. We want to be as consistent as we can across modules.

See my other comment, but I do think in a single follow up PR, we should restructure to go back to the `notify` pattern like in PPCP.

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
scannillo: I wouldn't change any business logic for the purpose of analytics. [We should keep it like we had it](https://github.com/braintree/braintree_ios/blob/5.x/Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.m#L104-L113), and continue to call `authenticate` for .success, .noAction, and .failure

scannillo: The `3ds:verify:challenge:failed` event should be fired [here](https://github.com/braintree/braintree_ios/blob/02234df74c9837637165c668b44f879fbb945f4c/Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift#L128-L141).

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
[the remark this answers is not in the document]
KunJeongPark: The only thing I can think of that might be problem in doing that is that we don't really have a connect right now from client test to for AuthenticateJWT right now.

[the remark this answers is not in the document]
KunJeongPark: I guess I can test the verifySuccess here.

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
scannillo: Let's just treat it as we were before, which is a "challenge success". [Android does the same](https://github.com/braintree/braintree_android/blob/6b0c5a60ad05a89771d23052f79f690848cf818a/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java#L507-L509), proceeds with the `/authenticate_from_jwt`

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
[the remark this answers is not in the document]
KunJeongPark: I will make a ticket for refactor to notify pattern but general event is sent from completion handler in client as of now.

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
jaxdesmarais: Yes, since this guard calls `completion(nil, BTThreeDSecureError.configuration("A client token must be used for ThreeDSecure integrations."))` right after this we can handle this the same way we handle any other completion with an error in this file.

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
jaxdesmarais: ➕ that makes sense to me!

KunJeongPark: I think we didn't send `challenge:failed` for error other than cancel https://github.com/braintree/braintree_ios/blob/9cb3aa0b50b027e63f93a8abc4cda6492d897b1a/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift#L190

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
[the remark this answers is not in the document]
KunJeongPark: I made a ticket here: https://paypal.atlassian.net/browse/DTBTSDK-2694

scannillo:
Re-commenting [this note ](https://github.com/braintree/braintree_ios/pull/1010#issuecomment-1512050181)about using the PR's markdown template

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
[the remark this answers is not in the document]
scannillo: 👏  Thanks for making a ticket! However, there are a few clarifications and links that tickets needs:

- That work will be across all modules in the SDK, not just 3DS. This is to have a unified analytics pattern across the entire SDK, making it less error-prone and easier to read.
- The ticket needs links to the relevant PR conversions that we had about using the `notify` pattern.
- The ticket needs acceptance criteria, or something to define what "done" looks like.

We want there to be enough detail that anyone could pick it up and have all the information to understand the root issue.

Would you mind adding those updates to the ticket?

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureAnalytics_Tests.swift
```diff
@@ -0,0 +1,23 @@
...
+        XCTAssertEqual(BTThreeDSecureAnalytics.challengeRequired, "3ds:verify:lookup:challenge-required")
+        XCTAssertEqual(BTThreeDSecureAnalytics.lookupSucceeded, "3ds:verify:lookup:succeeded")
+        XCTAssertEqual(BTThreeDSecureAnalytics.lookupFailed, "3ds:verify:lookup:failed")
+        
+        XCTAssertEqual(BTThreeDSecureAnalytics.challengeSucceeded, "3ds:verify:challenge.succeeded")
+        XCTAssertEqual(BTThreeDSecureAnalytics.challengeFailed, "3ds:verify:challenge.failed")
+        
+        XCTAssertEqual(BTThreeDSecureAnalytics.jwtAuthSucceeded, "3ds:verify:authenticate-jwt:succeeded")
+        XCTAssertEqual(BTThreeDSecureAnalytics.jwtAuthFailed, "3ds:verify:authenticate-jwt:failed")
+
```
scannillo, line 21:
```suggestion
```

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
scannillo: @KunJeongPark  - Making sure [this comment](https://github.com/braintree/braintree_ios/pull/1025/files#r1184279390) gets addressed. We want to revert to the original business logic.

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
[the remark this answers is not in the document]
jaxdesmarais: ➕ to everything Sammy said! We can also link to how it's done in PPCP if that's the pattern we want to follow.

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
KunJeongPark: you mean sending challenge fail for error, timeout block? It's done. we are calling authenticate for all the cases, I just separated them out to send different analytic messages.

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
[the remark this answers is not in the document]
KunJeongPark: To threeDSecureV2Provider and to BTThreeDSecureAuthenticateJWT classes, we are not passing merchant completion but a completion that discerns cancel vs other errors and then calls merchant completion. 

The way we have handled it was to just send error or success to that completion from both classes. If we want to handle the errors and analytic messages in the same way, cleanest way is to refactor to pass the merchant completion handler  to both classes.

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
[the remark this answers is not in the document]
KunJeongPark: I've made changes in the ticket.

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -350,10 +349,6 @@ import BraintreeCore
                 parameters: requestParameters
             ) { body, _, error in
                 if let error = error as NSError? {
-                    if error.code == BTCoreConstants.networkConnectionLostCode {
```
jaxdesmarais, line 353: Do we no longer want to send the `networkConnectionLost` event here?

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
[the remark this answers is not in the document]
KunJeongPark: It is a bit tricky with unit testing the general success after authentication since it would require a lot of mock response from Cardinal. I've manually tested general success case from authentication but tough with general fail from authentication. I can look into maybe using Proxyman or some other way to test general fail.

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
KunJeongPark: We have a ticket for [networkConnectionLost](https://paypal.atlassian.net/browse/DTBTSDK-2711). We realized that we were missing them so we added them to the Excel sheet listing all events. I can just add it for this module here.

jaxdesmarais: That makes sense. Yeah I think for the modules we have PRs for we can make sure they are added. Then in the followup worry about just the missing ones that we didn't catch the first time.

Sources/BraintreeThreeDSecure/BTThreeDSecureAuthenticateJWT.swift
```diff
@@ -32,32 +31,28 @@ class BTThreeDSecureAuthenticateJWT {
             parameters: requestParameters
         ) { body, _, error in
             if let error = error as NSError? {
-                if error.code == BTCoreConstants.networkConnectionLostCode {
-                    apiClient.sendAnalyticsEvent("ios.three-d-secure.verification-flow.network-connection.failure")
```
jaxdesmarais, line 36: Let's add this network connection error code back in too

jaxdesmarais (approved):
Looks like a couple Unit Tests are failing - not sure if CI is just being flakey or if it's worth looking into. In either case this looks good as soon as they are green. Thanks so much for addressing feedback, comments, and really digging into some of these nitty gritty flows!

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
scannillo: 👋 @KunJeongPark my [comments](https://github.com/braintree/braintree_ios/pull/1025/files#r1184279390) here still haven't been addressed.

scannillo (changes requested):

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
scannillo: @KunJeongPark I would consolidate where possible, so:

```
        case .success, .noAction:
            apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.challengeSucceeded)
            BTThreeDSecureAuthenticateJWT.authenticate(
                jwt: serverJWT,
                withAPIClient: apiClient,
                forResult: lookupResult,
                completion: completionHandler
            )
        case .failure:
            apiClient.sendAnalyticsEvent(BTThreeDSecureAnalytics.challengeFailed)
            BTThreeDSecureAuthenticateJWT.authenticate(
                jwt: serverJWT,
                withAPIClient: apiClient,
                forResult: lookupResult,
                completion: completionHandler
            )
```

Otherwise, LGTM!

scannillo: We could also go a step further to avoid duplicating the authenticate() call with an `if` inside the triple case block. Not sure which would be the most readable.

scannillo (approved):

PR #1027, opened 2023-05-03
Update Test Apps to Add Basic Client Creation

jaxdesmarais:
### Summary of changes

- Based on the [SPM bug](https://paypal.sharepoint.com/:w:/r/sites/BTSDK/_layouts/15/Doc.aspx?sourcedoc=%7B006D11FE-7148-456F-B191-B9E77CC67A71%7D&file=iOS%20v6%20Beta%20SPM%20Bug.docx&action=default&mobileredirect=true) we found earlier in the conversion process, we realized that we have no way to "smoke test" these integrations as our Sample Apps don't do much
    - This adds the basic structure of creating all of our clients (which would have caught the bug)
    - This also acts as a bit more robust sample app vs what we had previously with no implementation
    - In the future we can decide if it makes sense to add more to these sample apps

### Checklist

- @jaxdesmarais

### DIFF
--- a/SampleApps/CarthageTest/CarthageTest/ViewController.swift
+++ b/SampleApps/CarthageTest/CarthageTest/ViewController.swift
@@ -8,6 +8,8 @@ import BraintreePaymentFlow
 import BraintreePayPal
 import BraintreeThreeDSecure
 import BraintreeVenmo
+import PayPalCheckout
+import BraintreePayPalNativeCheckout
 import BraintreeSEPADirectDebit
 
 class ViewController: UIViewController {
--- a/SampleApps/CarthageTest/CarthageTest/ViewController.swift
+++ b/SampleApps/CarthageTest/CarthageTest/ViewController.swift
@@ -8,7 +8,6 @@ import BraintreePaymentFlow
 import BraintreePayPal
 import BraintreeThreeDSecure
 import BraintreeVenmo
-import PayPalCheckout
 import BraintreePayPalNativeCheckout
 import BraintreeSEPADirectDebit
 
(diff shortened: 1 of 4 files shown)


scannillo (approved):

SampleApps/CarthageTest/CarthageTest/ViewController.swift
```diff
@@ -13,14 +13,18 @@ import BraintreeSEPADirectDebit
...
-        // Dispose of any resources that can be recreated.
+        let amexClient = BTAmericanExpressClient(apiClient: apiClient)
+        let applePayClient = BTApplePayClient(apiClient: apiClient)
+        let cardClient = BTCardClient(apiClient: apiClient)
+        let dataCollector = BTDataCollector(apiClient: apiClient)
+        let paymentFlowClient = BTPaymentFlowClient(apiClient: apiClient)
+        let payPalClient = BTPayPalClient(apiClient: apiClient)
+        let threeDSecureClient = BTThreeDSecureClient(apiClient: apiClient)
+        let venmoClient = BTVenmoClient(apiClient: apiClient)
+        let payPalNativeCheckoutClient = BTPayPalNativeCheckoutClient(apiClient: apiClient)
```
scannillo, line 26: Looks like we need to import the PayPalNativeCheckout module

jaxdesmarais: Yep - guess the other benefit of this is finding things we forgot about a while ago! 🙈 We don't have the framework linked at all for `BraintreePayPalNativeCheckout`.

PR #1028, opened 2023-05-03
sepa analytics alert cancel bug fix

KunJeongPark:
Summary of changes
- reset boolean after notification trigger between web sessions for accurate alert cancel vs browser cancel for accuracy on multiple sessions.

Checklist
~- [] Added a changelog entry~

Authors
@KunJeongPark


jaxdesmarais (commented):
One thing to note, I think it's important that we leave the PR description formatting consistent across all of our PRs using the markdown provided when opening a PR. In addition to making the PRs consistent it also makes it easier to see what has/hasn't been done on any given PR. The only lines we should need to delete are the lines about localizations (before `### Summary of changes`) but the rest of the PR markdown format should remain as is.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -18,7 +18,6 @@ import BraintreeCore
    
     // MARK: - Private Properties
 
-    private var isAuthenticationSessionStarted: Bool = false
```
jaxdesmarais, line 21: Would you mind going a bit more into the details of the bug? I am not sure I am following since `BTPayPalClient` uses both `isAuthenticationSessionStarted` and `returnedToAppAfterPermissionAlert`. Since they are both using `ASWebAuthenticationSession` I would expect for these parameters and their implementation to match.

Looking here in BTPayPalClient: https://github.com/braintree/braintree_ios/blob/02234df74c9837637165c668b44f879fbb945f4c/Sources/BraintreePayPal/BTPayPalClient.swift#L324-L341

KunJeongPark: We don't need the `isAuthenticationSessionStarted` here since we are resetting the `returnedToAppAfterPermissionAlert` before starting each webSession. This was not the source of the bug. The bug was from not resetting `returednToAppAfterPermissionAlert` in before start of the session.

KunJeongPark: In PayPal module, that boolean was used to send different analytic messages on browser failure or success but here we don't need that boolean since we send in completion handler `sessionDidDisplay` for the webSession class.

KunJeongPark: This is the flow for alert cancel vs browser cancel here:
Flow for alert cancel => 
  #0. sessionDidDisplay: didDisplay
  #1.  alert view appears, I hit cancel
  #2. Web session is done, handleWebAuthSessionResult => !returnedToAppAfterPermission in completion handler for web session result
  #3. Hits notification

Flow for browser cancel => 
#0. sessionDidDisplay: didDisplay
#1 hit continue on alert button
#2 hit notification
#3 I hit cancel on browser
#4 hit handleWebAuthenticationSessionResult  (outside of !returnedToAppAfterPermissionAlert)

If sessionDidDisplay is false, it wouldn't enter into cancel check section of the completion handler.

KunJeongPark: I guess if there are other events that can trigger this notification during webSession so there is a small chance for alert view cancel, unexpected results.

jaxdesmarais: Gotcha - so the issue is that we introduced a bug in the [SEPA Analytics audit PR](https://github.com/braintree/braintree_ios/pull/1004), correct? Just switching between `master` and the `fpti-feature` branch to fully understand what got introduced, when it happened, and where.

To confirm `PayPal` and `PaymentFlow` are working as expected, correct? It looks like in `PayPal` we are setting `        returnedToAppAfterPermissionAlert = false` before calling `start`. Does doing the same here allow us to hit the cancel we expect and resolve the bug?

jaxdesmarais: I am more trying to think of this from the perspective of your follow up PR to align the use of `ASWebAuthenticationSession` what will be easiest to reason about and ensure that we don't reintroduce this or a similar bug.

KunJeongPark: Yes, I discovered this bug while working on `PaymentFlow`. PayPal module is working as expected, you just don't need to set the returnedToAPPAfterPermissionAlert to isAuthenticaitonSessionStarted with the notification since it is same as setting it to true for our purposes.

KunJeongPark: Yes, I discovered this bug while working on `PaymentFlow`. PayPal module is working as expected, you just don't need to set the returnedToAPPAfterPermissionAlert to isAuthenticaitonSessionStarted with the notification since it is same as setting it to true for our purposes.

jaxdesmarais: > I also just deleted the unnecessary step of assigning returnedToAppAfterPermissionAlert to isAuthenticationSessionStarted which stays true within a web session once it starts but that was not the source of the bug.
>
> Yes, I discovered this bug while working on PaymentFlow. PayPal module is working as expected, you just don't need to set the returnedToAPPAfterPermissionAlert to isAuthenticaitonSessionStarted with the notification since it is same as setting it to true for our purposes.

I am not sure I follow this. It would be set to false if we background/terminate the app before `start` is called, right? At least in PaymentFlow and PayPal it would. So now with this change in SEPA we are always setting it to true even if the result of `start` is false?

KunJeongPark: It's because we only go into cancel block of checking `returnedToAppAfterPermissionAlert`  if the start was true. `isAuthenticationStarted` is not reset to false after the session is kicked off, so when we are checking for this value, it will always be true.

KunJeongPark: I did find that variable name confusing but I opt for keeping it for now since we have the same name in other modules. Maybe in the refactor ticket to use single wrapper class for ASWebAuthenticationSession, we can rename the variable for all of them.

jaxdesmarais: Cool - could you add a line item to the ASWeb ticket for us to revisit the parameter name `returnedToAppAfterPermissionAlert`?

jaxdesmarais (approved):

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
KunJeongPark: Good idea

sshropshire (approved):

PR #1033, opened 2023-05-08
paypal analytics changelog and error return fix

KunJeongPark:
### Summary of changes

- fix changelog entries on fpti-feature  to be under "unreleased" section 
- fix return to merchant completion on `sessionDidDisplay` false, to be handled in completion handler
- delete unused `WebSessionFailedToLaunch` error

### Checklist

- [x ] Added a changelog entry

### Authors
@KunJeongPark


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -355,7 +355,6 @@ import BraintreeDataCollector
         } else {
             apiClient.sendAnalyticsEvent(BTPayPalAnalytics.browserPresentationFailed)
             apiClient.sendAnalyticsEvent(BTPayPalAnalytics.tokenizeFailed)
```
scannillo, line 357:
```suggestion
```

CHANGELOG.md
```diff
@@ -77,6 +77,10 @@
         * Add error `.invalidAPIClient` = 6
         * Add error `.cannotCastRequest`= 7
         * Add error `.jsonSerializationFailure` = 8
+  * BraintreePayPal
+    * Add `BTPaymentFlowError` case
+        * Add error `.webSessionError` = 8
+
```
jaxdesmarais, line 83: We want to move this under the `unreleased` section since it was added in the `fpti-feature` branch and that branch hasn't been merged in or released yet.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -355,7 +355,6 @@ import BraintreeDataCollector
...
-            completion(nil, BTPayPalError.webSessionFailedToLaunch)
```
jaxdesmarais, line 358: Wouldn't this now mean we are returning from this without a completion? We still want to send an error to the merchant here.

KunJeongPark: I think @scannillo didn't want to return to merchant app at this point but for it to be handled in web session completion handler as an error.

CHANGELOG.md
```diff
@@ -79,7 +82,7 @@
         * Add error `.canceled` = 5
         * Add error `.invalidAPIClient` = 6
         * Add error `.cannotCastRequest`= 7
-        * Add error `.jsonSerializationFailure` = 8
+        * Add error `.jsonSerializationFailure` = 8
```
jaxdesmarais, line 85: We can probably revert this line

Sources/BraintreePayPal/BTPayPalError.swift
```diff
@@ -27,9 +27,6 @@ enum BTPayPalError: Error, CustomNSError, LocalizedError {
     /// Unable to create BTPayPalAccountNonce
     case failedToCreateNonce
     
-    /// Unable to launch ASWebAuthenticationSession
-    case webSessionFailedToLaunch
```
jaxdesmarais, line 31: We should add that we removed this error case to the CHANGELOG

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: Gotcha, so this would now be returned in `completion(nil, BTPayPalError.webSessionError(error))` correct?

Sources/BraintreePayPal/BTPayPalError.swift
scannillo: I believe it was added in `fpti-feature` only and then removed here, so it doesn't need a CHANGELOG. @KunJeongPark can confirm

scannillo (approved):

Sources/BraintreePayPal/BTPayPalError.swift
jaxdesmarais: Ah, cool - in that case... disregard 🙈

CHANGELOG.md
KunJeongPark: fixed in commit ba406a4e001c6139be262b107668a8048acce0c2

jaxdesmarais (approved):

PR #1034, opened 2023-05-09
Add `@_documentation` attributes to hide public docs

scannillo:
### Background

- [This](https://forums.swift.org/t/nodoc-attribute-for-hiding-symbols-from-the-symbol-graph/59227?page=3) Swift Forums thread advocated for an equivalent of `nodoc` that [Jazzy offers](https://github.com/realm/jazzy#controlling-what-is-documented), but for Xcode's built in documentation compiler, DocC.
- Swift 5.8 (Xcode 14.3) introduces the underscored attribute [`@_documentation(visibility: ...)` ](https://github.com/apple/swift/blob/main/docs/ReferenceGuides/UnderscoredAttributes.md#_documentationvisibility-)
    - _This can be used to treat an otherwise-public symbol as being internal or private for the purposes of documentation, to hide it from public docs._

### Changes

- Add a `documentation` underscored attribute everywhere we use `nodoc`
- Remove redundant `nodoc`s where appropriate

### Next steps

- We need to keep `nodoc` until we ditch Jazzy for DocC to generate our hosted reference docs
- This exercise of visualizing the docs in Xcode's documentation UI reveals several places where we should be hiding docs (or maybe even reducing access level) but aren't (for example the `typealias` in the screenshots below). 
    - I'll leave this as a follow-up, and I think would be a good task to pair on.

### Example

Notice the unnecessary methods on `BTAPIClient` visible to merchants using Xcode to build documentation (Product > Build Documentation). This change removes them from docs visibility.

#### Before
Screenshot 2023-05-09 at 8 05 08 AM

#### After
Screenshot 2023-05-09 at 8 07 25 AM

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


sshropshire (approved):
🌊

jaxdesmarais (approved):

PR #1043, opened 2023-05-25
Move Analytics to Notify Pattern and Send Error Details to FPTI

jaxdesmarais:
### Summary of changes

- Rename `SEPADirectDebitError` to `BTSEPADirectDebitError` to match other error cases
    - Update source code with name change
    - Update tests with name change
- Add notify pattern on all `Client`s for sending analytics
- Remove all `network-connection:failed` events
- Add optional `errorDescription` to `sendAnalyticsEvent`
- Add `errorDescription` to event - we have confirmed the error description is sent through to FPTI as expected 🥳 
- Remove `BTCoreConstants.networkConnectionLostCode`
- Add `BTCardError.fetchConfigurationFailed` error
- Add `BTSEPADirectDebitError.failedToCreateNonce` error
- Remove all `networkConnectionLost` tests

### Checklist

- [x] Added a changelog entry

- @scannillo @jaxdesmarais


sshropshire (approved):

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -27,10 +27,12 @@ struct FPTIBatchData: Codable {
     /// Encapsulates a single event by it's name and timestamp.
     struct Event: Codable {
         
+        let errorDescription: String?
         let eventName: String
         let timestamp: String
         
         enum CodingKeys: String, CodingKey {
+            case errorDescription = "error_desc"
```
scannillo, line 35: I added this to our FPTI tags excel doc 👍

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -48,17 +48,21 @@ class BTAnalyticsService: Equatable {
     ///  Events are queued and sent in batches to the analytics service, based on the status of the app and
     ///  the number of queued events. After exiting this method, there is no guarantee that the event has been sent.
     /// - Parameter eventName: String representing the event name
-    func sendAnalyticsEvent(_ eventName: String) {
+    func sendAnalyticsEvent(_ eventName: String, errorDescription: String? = nil) {
```
scannillo, line 51: We can look at this later, but this method looks like it is only called within the tests. This is a little odd to me since that means we're basically never checking for the flush threshold before sending an API request with events.

scannillo (approved):
🦸‍♂️

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
jaxdesmarais: Yeah, I just jumped into v5 to poke around at this and we were doing the exact same thing there. We have the method:
```
- (void)sendAnalyticsEvent:(NSString *)eventKind {
    dispatch_async(dispatch_get_main_queue(), ^{
        [self enqueueEvent:eventKind];
        [self checkFlushThreshold];
    });
}
```
in Obj-C and the only callers are our tests. We should certainly look into this and remove it (if possible) or move the logic we want into the other flush method. We can 🏈 for now but agree we should revisit.

jaxdesmarais: Ah, actually it looks like the caller was these events that used to exist in APIClient `[self queueAnalyticsEvent:@"ios.started.client-key"];`. Which I guess makes sense for how that event was tied to everything. Now that we removed the events in APIClient, we can likely clean these methods up entirely.

PR #1046, opened 2023-05-31
Check For 3DSv1 in Lookup

jaxdesmarais:
### Summary of changes

- Add else block for if we get 3DSv1 back in the lookup response
- Add new unit test

### Checklist

- @scannillo @jaxdesmarais

### DIFF
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
@@ -281,7 +281,6 @@ import BraintreeCore
                 with: BTThreeDSecureError.configuration("3D Secure v1 is deprecated and no longer supported. See https://developer.paypal.com/braintree/docs/guides/3d-secure/client-side for more information."),
                 completion: merchantCompletion
             )
-
         }
     }
     
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
@@ -270,12 +270,18 @@ import BraintreeCore
     
     private func process(lookupResult: BTThreeDSecureResult, configuration: BTConfiguration) {
         if lookupResult.lookup?.requiresUserAuthentication == false || lookupResult.lookup == nil {
-            self.notifySuccess(with: lookupResult, completion: merchantCompletion)
+            notifySuccess(with: lookupResult, completion: merchantCompletion)
             return
         }
 
         if lookupResult.lookup?.isThreeDSecureVersion2 == true {
             performV2Authentication(with: lookupResult)
+        } else {
+            notifyFailure(
+                with: BTThreeDSecureError.configuration("3D Secure v1 is deprecated and no longer supported. See https://developer.paypal.com/braintree/docs/guides/3d-secure/client-side for more information."),
+                completion: merchantCompletion
+            )
+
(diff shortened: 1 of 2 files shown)


sshropshire (approved):

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -270,12 +270,18 @@ import BraintreeCore
     
     private func process(lookupResult: BTThreeDSecureResult, configuration: BTConfiguration) {
         if lookupResult.lookup?.requiresUserAuthentication == false || lookupResult.lookup == nil {
-            self.notifySuccess(with: lookupResult, completion: merchantCompletion)
+            notifySuccess(with: lookupResult, completion: merchantCompletion)
             return
         }
 
         if lookupResult.lookup?.isThreeDSecureVersion2 == true {
```
scannillo, line 277: Realizing, don't we want to return an error in _any case_ of 3DS v1 mentioned in the lookup? In that case, we'd want to move this logic to the top of the function. 

```
    private func process(lookupResult: BTThreeDSecureResult, configuration: BTConfiguration) {
        if lookupResult.lookup?.isThreeDSecureVersion2 == false {
            notifyFailure(
                with: BTThreeDSecureError.configuration("3D Secure v1 is deprecated and no longer supported. See https://developer.paypal.com/braintree/docs/guides/3d-secure/client-side for more information."),
                completion: merchantCompletion
            )
            return
        }
        
        if lookupResult.lookup?.requiresUserAuthentication == false || lookupResult.lookup == nil {
            notifySuccess(with: lookupResult, completion: merchantCompletion)
            return
        }
    }
```

If that _is_ the behavior we want, we'll have to update [Android](https://github.com/braintree/braintree_android/blob/87a71a0ea2189abc038dfe990cddc0fb549fdb21/ThreeDSecure/src/main/java/com/braintreepayments/api/ThreeDSecureClient.java#L394-L412) as well.

jaxdesmarais: 🤔 that is a good question. I _think_ how it is now is correct, it would mean that if we don't have any lookup information/challenge information/liability shift info or if we don't require user authentication (determined by the presence of the `acsUrl`) we just want to return the lookup result regardless of the 3DS version. These would be cases like if the card was ineligible for 3D Secure. I think in those cases, since 3D Secure liability would not be shifted it's up to the merchant to determine if they want to accept liability for those transactions so the 3D Secure version is just another piece of information for them to make that determination.

scannillo: 👍 That makes sense to me

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -270,12 +270,18 @@ import BraintreeCore
...
             performV2Authentication(with: lookupResult)
+        } else {
+            notifyFailure(
+                with: BTThreeDSecureError.configuration("3D Secure v1 is deprecated and no longer supported. See https://developer.paypal.com/braintree/docs/guides/3d-secure/client-side for more information."),
+                completion: merchantCompletion
+            )
+
```
scannillo, line 284:
```suggestion
```

scannillo (approved):

PR #1049, opened 2023-06-07
Add Deallocated Client Error Cases

jaxdesmarais:
### Summary of changes

- Rename `BTPayPalNativeError` to `BTPayPalNativeCheckoutError` to match other error enums
- Add `BTPayPalNativeCheckoutError.deallocated` = 11
- Add `BTAmericanExpressError.deallocated` = 2
- Add `BTAPIClientError.deallocated` = 2
- Add `BTHTTPError.deallocated` = 12
- Add `BTPayPalError.deallocated` = 9
- Add `BTThreeDSecureError.deallocated` = 8
- Add `BTSEPADirectDebitError.deallocated` = 8

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais


Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
```diff
@@ -30,7 +30,10 @@ import BraintreeCore
         apiClient.sendAnalyticsEvent(BTAmericanExpressAnalytics.started)
 
         apiClient.get("v1/payment_methods/amex_rewards_balance", parameters: parameters) { [weak self] body, response, error in
-            guard let self = self else { return }
+            guard let self else {
+                self?.notifyFailure(with: BTAmericanExpressError.deallocated, completion: completion)
```
sshropshire, line 34: I believe this will no-op since `self` is always `nil` in the else block.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -288,7 +294,11 @@ import Foundation
     @objc(POST:parameters:httpType:completion:)
     public func post(_ path: String, parameters: [String: Any]? = nil, httpType: BTAPIClientHTTPService, completion: @escaping RequestCompletion) {
         fetchOrReturnRemoteConfiguration { [weak self] configuration, error in
-            guard let self else { return }
+            guard let self else {
+                completion(nil, nil, BTAPIClientError.deallocated)
+
```
scannillo, line 299:
```suggestion
```

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
jaxdesmarais: Yeah, you are right - just tested this. I'll update where we use an optional self to pass the completion and analytics event in the else instead vs using notify. Good catch.

scannillo: +1

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
 
+## unreleased
+* BraintreePayPalNativeCheckout
+  * Rename `BTPayPalNativeError` to `BTPayPalNativeCheckoutError`
+  * Add `BTPayPalNativeCheckoutError.deallocated` = 11
+* BraintreeAmericanExpress
+  * Add `BTAmericanExpressError.deallocated` = 2
+* BraintreeCore
+  * Add `BTAPIClientError.deallocated` = 2
+  * Add `BTHTTPError.deallocated` = 12
```
KunJeongPark, line 11: Hello, can you tell me how the clients get deallocated during the session? Was this observed in specific modules, cases?

Sources/BraintreeCore/BTAPIClient.swift
jaxdesmarais: Updated as part of this change: https://github.com/braintree/braintree_ios/pull/1049/commits/1fadabdbc626829f4dbb6e6decf9fb59066acf2a

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
jaxdesmarais: Updated these here: https://github.com/braintree/braintree_ios/pull/1049/commits/1fadabdbc626829f4dbb6e6decf9fb59066acf2a

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
```diff
@@ -30,7 +30,10 @@ import BraintreeCore
         apiClient.sendAnalyticsEvent(BTAmericanExpressAnalytics.started)
 
         apiClient.get("v1/payment_methods/amex_rewards_balance", parameters: parameters) { [weak self] body, response, error in
-            guard let self = self else { return }
+            guard let self else {
+                completion(nil, BTAmericanExpressError.deallocated)
```
jaxdesmarais, line 34: We can't send analytics for these because using `apiClient` would rely on using `self`

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
```diff
@@ -146,7 +146,9 @@ import BraintreeDataCollector
         requestParameters["_meta"] = metadataParameters
 
         apiClient.post("/v1/payment_methods/paypal_accounts", parameters: requestParameters) { [weak self] body, response, error in
-            guard let self else { return }
+            guard let self else {
+                return
```
jaxdesmarais, line 150: How do we want to handle the cases in this class? These methods don't have a completion and instead rely on `merchantCompletion` which relies on self, which we won't have access to. Should we log something to the console in this case?

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+* BraintreePayPal
+  * Add `BTPayPalError.deallocated` = 9
+* BraintreeThreeDSecure
+  * Add `BTThreeDSecureError.deallocated` = 8
+* BraintreeSEPADirectDebit
+  * Add `BTSEPADirectDebitError.deallocated` = 8
```
scannillo, line 17: Super nitpick / idea - the error codes throughout the CHANGELOG entries for V6 betas get a little unruly/hard to read IMO.

Maybe we could amend the previous entries to remove all error code specific ones, and then instead summarize them in a single table that we include in the V6 GA release notes?

jaxdesmarais: I like that idea. Agree right now the changes to the errors is quite hard to reason about. I created internal ticket DTBTSDK-2788 for this work.

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
scannillo: I'm down for a console log, but also don't feel strongly if it just no-ops. We could leave a TODO to refactor these methods to pass the completion like the other clients do

scannillo (approved):

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -147,7 +147,10 @@ import Foundation
         let parameters: [String: Any] = ["configVersion": "3"]
 
         configurationHTTP?.get(configPath, parameters: parameters, shouldCache: true) { [weak self] body, response, error in
-            guard let self else { return }
+            guard let self else {
+                completion(nil, BTAPIClientError.deallocated)
+                return
+            }
```
scannillo, line 154: We shouldn't need the explicit `self`s below (ex: `self.apiHTTP` and `self.http`)

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
jaxdesmarais: Updated: https://github.com/braintree/braintree_ios/pull/1049/commits/46f409ef668c8c37343aad61b0903bddaf6fb067

Sources/BraintreeCore/BTAPIClient.swift
jaxdesmarais: Good call - we were still using self _a lot_ cleaned up here: https://github.com/braintree/braintree_ios/pull/1049/commits/ff0e5256cea5c1a4fac9f5bcfaa12a3468ebc47f

PR #1052, opened 2023-06-09
Audit Public Methods and TODOs

jaxdesmarais:
### Summary of changes

- Remove `BTMutableClientMetadata` - this is no longer needed since the entire SDK is in Swift
- Remove `sourceString` and `integrationString` in favor of using the `stringValue` on the enums directly
- Add `nodoc` and `@_documentation(visibility: private)` to classes/methods/properties where missing
- Update access levels in a few places

### Checklist

### Authors

- @jaxdesmarais


Sources/BraintreeApplePay/BTApplePayAnalytics.swift
```diff
@@ -1,6 +1,5 @@
 import Foundation
 
-// NEXT_MAJOR_VERSION: Move this class to an enum once the entire SDK is in Swift
```
jaxdesmarais, line 3: Removing this comment, ticket DTBTSDK-2772 exists for this task. Don't think we need a note in 1 of the classes since this will need to be updated in all of the modules.

Sources/BraintreeCore/BTAPIPinnedCertificates.swift
```diff
@@ -8,9 +8,9 @@ import Foundation
 ///
 /// Do not update this file directly. If new certificates are needed, update the `certificates` directory
 /// in `braintree-ios-certificate-generation` and re-run the certificate generation script.
-@objcMembers public class BTAPIPinnedCertificates: NSObject {
+class BTAPIPinnedCertificates {
```
jaxdesmarais, line 11: This class is only used in Core. Checked the Drop-in and it's not being used there so safe to make it internal.

Sources/BraintreeCore/BTClientMetadata.swift
```diff
@@ -11,56 +13,45 @@ import Foundation
...
     
-    /// String representation of the integration
-    public var integrationString: String {
-        integration.stringValue
-    }
-    
-    /// String representation of the source
-    public var sourceString: String {
-        source.stringValue
-    }
```
jaxdesmarais, line 33: Since these were public anyways, it made sense to remove them and expose the `stringValue` directly.

Sources/BraintreeCore/BTClientMetadata.swift
```diff
@@ -11,56 +13,45 @@ import Foundation
...
-        result.source = self.source
-        result.sessionID = self.sessionID
-        return result
-    }
-    
+
+    /// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     /// Creates a copy of `BTClientMetadata`
-    @objc(copyWithZone:)
+    @_documentation(visibility: private)
```
jaxdesmarais, line 54: We only use `NSCopying` for tests. Do we find these tests necessary at all? If not, we should just remove this conformance since it requires public access modifiers.

Sources/BraintreeCore/BTCoreConstants.swift
```diff
@@ -1,29 +1,39 @@
...
+    /// :nodoc: This property is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    @_documentation(visibility: private)
     public static var braintreeSDKVersion: String = "6.0.0-beta4"
 
-    public static var apiVersion: String = "2016-10-07"
-    
-    public static var graphQLVersion: String = "2018-03-06"
-    
+    /// :nodoc: This property is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    @_documentation(visibility: private)
```
jaxdesmarais, line 12: Do we need this for each property or is it covered by the class one?

Sources/BraintreeApplePay/BTConfiguration+ApplePay.swift
```diff
@@ -7,7 +7,7 @@ import BraintreeCore
 
 extension BTConfiguration {
 
-    ///  :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    /// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     /// Indicates whether Apple Pay is enabled for your merchant account.
     @objc public var isApplePayEnabled: Bool {
```
scannillo, line 12: Do these also need `@_documentation(visibility: private)`?

Sources/BraintreeCore/BTClientMetadata.swift
scannillo: I'd also vote to remove, assuming we can still write a meaningful test - maybe comparing `parameters` value directly vs using `NSCopy`

Sources/BraintreeCore/BTCoreConstants.swift
```diff
@@ -1,29 +1,39 @@
...
     public static let callbackURLScheme: String = "sdk.ios.braintree"
 
+    static var apiVersion: String = "2016-10-07"
+    
+    static var graphQLVersion: String = "2018-03-06"
+
     // MARK: - BTHTTPError Constants
     // NEXT_MAJOR_VERSION (v7): When the entire SDK is in Swift we will likely want to move these properties into the BTHTTPError enum
     // and make it public. We cannot do that currently since the enum is used in Obj-C modules and tests and you cannot expose enums
     // with associated values to Obj-C (only Int enums).
```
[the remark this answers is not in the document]
scannillo: Which enum do you mean?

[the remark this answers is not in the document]
jaxdesmarais: `BTHTTPErrors`. That is also where they used to live publicly in Obj-C: https://github.com/braintree/braintree_ios/blob/5.x/Sources/BraintreeCore/Public/BraintreeCore/BTHTTPErrors.h

So we could expose them again there vs the constants file if we wanted. If we don't want to expose that enum we should remove this comment.

Sources/BraintreeCore/BTClientMetadata.swift
jaxdesmarais: Yeah in both `BTClientMetadata_Tests` and `BTThreeDSecurePostalAddress_Tests` we test the parameter values directly in the other tests in that file. Added tests in `BTPostalAddress_Tests` to do the same. Updated in this commit: https://github.com/braintree/braintree_ios/pull/1052/commits/ecf8c936e14d944f426fb8316f0bf4dbb4347ec4

Sources/BraintreeCore/BTCoreConstants.swift
jaxdesmarais: Yep, you are right, only needed at the class level. Updated in this commit https://github.com/braintree/braintree_ios/pull/1052/commits/90bd774f53b6ca5fc774abe6e86e7dd4e9027035 and double checked the compiled docs.

Sources/BraintreeCore/BTClientMetadata.swift
```diff
@@ -11,61 +13,39 @@ import Foundation
 /// the sessionID should remain constant. To achieve this, users of this class
 /// should use `mutableCopy` to create a new copy based on the existing session
```
sshropshire, line 14: We should get rid of these docs that mention making a `mutableCopy` since we no longer implement `NSMutableCopying` here.

Sources/BraintreeCore/BTCoreConstants.swift
```diff
@@ -1,29 +1,39 @@
...
-    
-    public static var graphQLVersion: String = "2018-03-06"
-    
+    /// :nodoc: This property is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    @_documentation(visibility: private)
     public static let callbackURLScheme: String = "sdk.ios.braintree"
 
+    static var apiVersion: String = "2016-10-07"
+    
+    static var graphQLVersion: String = "2018-03-06"
```
sshropshire, line 15:
```suggestion
    static let graphQLVersion: String = "2018-03-06"
```

Sources/BraintreeApplePay/BTConfiguration+ApplePay.swift
jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/1052/commits/e80b3733fbe9e0adadfab35fc5800c87bee46f79

sshropshire (approved):
🌊

scannillo (approved):

Sources/BraintreeCore/BTClientMetadata.swift
jaxdesmarais: Good call, done: https://github.com/braintree/braintree_ios/pull/1052/commits/5fdb7ac2347176bff1f2adc82917e548f49a1aa0

Sources/BraintreeCore/BTCoreConstants.swift
jaxdesmarais: Updated both here: https://github.com/braintree/braintree_ios/pull/1052/commits/9836cf081b9fefc1aea3175ab8f0a0cde4b2217e

PR #1059, opened 2023-06-16
Add `.spi.yml` for Swift Package Index docs

scannillo:
### Summary

Braintree iOS is currently accessible via the Swift Package Index (https://swiftpackageindex.com/braintree/braintree_ios). This PR enhances our Swift Package Index experience by supporting it's auto-hosting DocC feature.

### Changes
- Add a `.spi.yml` file
    - [See SPI docs](https://blog.swiftpackageindex.com/posts/auto-generating-auto-hosting-and-auto-updating-docc-documentation/) on how this auto-hosted, auto-generated docs works. It's really neat! We could even have this replace our hosted reference docs entirely 👀. 

### Result

Once merged into `master` our SPI landing site will add this button:
Screenshot 2023-06-16 at 10 45 42 AM

The hosted docs & new button won't be visible until merged.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -2,4 +2,4 @@ version: 1
 builder:
   configs:
     - platform: ios
-      documentation_targets: [Braintree]
+      documentation_targets: [BraintreeAmericanExpress, BraintreeApplePay, BraintreeCard, BraintreeCore, BraintreeDataCollector, BraintreeLocalPayment, BraintreePayPal, BraintreePayPalNativeCheckout, BraintreeSEPADirectDebit, BraintreeThreeDSecure, BraintreeVenmo]
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,0 +1,5 @@
+version: 1
+builder:
+  configs:
+    - platform: ios
+      documentation_targets: [Braintree]


scannillo:
There seems to be a limitation with DocC supporting SDKs with multiple target frameworks (see Issue - https://github.com/apple/swift-docc/issues/255). 

I don't mind merging, seeing how SPI handles this, and then reverting if its borked.

.spi.yml
```diff
@@ -0,0 +1,5 @@
+version: 1
+builder:
+  configs:
+    - platform: ios
+      documentation_targets: [Braintree]
```
jaxdesmarais, line 5: Looking at some examples on github it seems like we may need to list each target independently? I was looking at [swift collections](https://github.com/apple/swift-collections/blob/939cfd25234472b4dc91c3caeab304d15bca9a73/.spi.yml#L4) as an example. Though it looks like maybe this just works? Not totally sure. This is super neat though!

scannillo: Good find! I was thinking this too, but I didn't find a good example of an SDK that used multiple targets. Looking at [Swift Collections Package.swift](https://github.com/apple/swift-collections/blob/939cfd25234472b4dc91c3caeab304d15bca9a73/Package.swift), this looks similar to how our project is separated out. Will update

scannillo: Done - f264f9c5c3da9a6fe10315bfa90086da1324f76e

jaxdesmarais (approved):

adam-rocska:
@scannillo did it work? Just saw your comment it’ll get reverted if it doesn’t work, and kept if it works ¯\\_(ツ)_/¯

scannillo:
Hi @adam-rocska, this did indeed work! You can navigate here (https://swiftpackageindex.com/braintree/braintree_ios) and click `Documentation` to see the docs that SPI is building on our behalf.

We would love feedback as well to your experience using those docs versus our self-hosted [reference docs](https://braintree.github.io/braintree_ios/current/).

adam-rocska:
@scannillo  H*l* s*t! THANKS! ❤️

PR #1060, opened 2023-06-21
Fix MXO User Action Bug

jaxdesmarais:
### Summary of changes

- Remove decorate method from the PayPal flow and instead pass the `user_action` into the `experience_profile`
    - Not sure the historical context on passing this as a query item, but there is no regression on web when passing this in the `experience_profile` instead

### Visuals
This is the result of setting `request.userAction` to `payNow` in both the web and MXO checkout flows with the changes in this PR:

| Web | MXO |
|-----|-----|
|pay-now-web|pay-now-mxo|

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPalNativeCheckout (BETA)
+  * Fix bug where setting `userAction` does not update button as expected
+
 ## 6.0.0 (2023-06-20)
 * The Braintree SDK is now written in Swift
 * Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@
 * BraintreePayPalNativeCheckout (BETA)
   * Fix bug where setting `userAction` does not update button as expected
 
+## 6.1.0 (2023-06-22)
+* BraintreeVenmo
+  * Allow merchants to collect enriched customer data if enabled in the Braintree Control Panel
+  * Add the following properties to `BTVenmoRequest`
+    * `collectCustomerBillingAddress`
+    * `collectCustomerShippingAddress`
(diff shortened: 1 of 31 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPalNativeCheckout (BETA)
```
scannillo, line 4: When are we planning to remove the (BETA) tag? On the next NXO release?

jaxdesmarais: The plan is to remove the beta tag once we have the GA version in the SDK! It sounds like there may be a fix version before the GA but we can ask today in our meeting with them.

scannillo:
Looks like a few tests in PayPalClient_Tests, BTPayPalRequest_Tests, and BTPayPalNativeCheckoutRequest_Tests will need to be updated to remove assertions for `user_action` in the URL, and then to add assertions for it under the `experience_profile` in the POST

scannillo:
I made this sample branch with an alternative implementation we can do that doesn't require passing around the `userAction` param. See https://github.com/braintree/braintree_ios/compare/update-user-action...update-user-action-2 👀

jaxdesmarais:
> I made this sample branch with an alternative implementation we can do that doesn't require passing around the userAction param. See https://github.com/braintree/braintree_ios/compare/update-user-action...update-user-action-2 👀
>
> ^ Since Swift Dictionaries are value types, this basically makes a copy of what already exists in experience_profile, appends what we need to it, then re-assigns it.

Nice! I dig this a lot more, certainly didn't feel great to have the other flows know about this param. Updated here: https://github.com/braintree/braintree_ios/pull/1060/commits/3c5846d3ee6a104540bd8892e05b017c868cc16c

scannillo (approved):

PR #1062, opened 2023-06-26
Demo App Updates

jaxdesmarais:
Spent some open dev time making updates to the demo app. It still certainly needs some love but tackled some low hanging fruit.

### Summary of changes

- The demo app now supports dark mode for all feature view controllers
- Removed `nonceStringCompletionBlock` from the demo app - was just used for Obj-C + Swift interop, now that the SDK is fully Swift we don't need this anymore
- Remove `BraintreeDemoCardTokenizationViewController.xib` and instead use the card form that's also used for 3DS - this will allow us to make card form updates in 1 place vs 2 in the future
- Update `BraintreeDemoPayPalNativeCheckoutViewController` to resolve ambiguous layout warnings

### Visuals
Before and after of the views that were the biggest offenses in dark mode - all views have been checked to ensure they still look nice in both light/dark mode.
| Before | After |
|-----|-----|
|bad-card|good-card|
|bad-3ds|good-3ds|
|bad-apple-pay|goo-apple-pay|

### Checklist

### Authors

- @jaxdesmarais


scannillo (approved):
🎸

PR #1031, opened 2023-05-05
Bump to Xcode 14.3 & Swift 5.8

scannillo:
### Summary of changes

- Bump to Xcode 14.3 and Swift 5.8 for our next major version, V6

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

PR #1067, opened 2023-06-30
Fix Flakey Tests

jaxdesmarais:
As outlined in [this article](https://www.avanderlee.com/debugging/flaky-tests-test-repetitions/#communication-with-external-services) a lot of our previously failing tests rely on external services (the gateway or Cardinal). The best course of action in these cases is often to bump the timeouts.

### Observations
There are 3 main sections of tests that were failing:
1. UI tests - specifically tests relying on fetching a client token - waiting for this API call sometimes takes a few seconds so if we want to continue to have UI tests for Amex and 3DS we will need to rely on bumping timeouts if these tests become flakey. These tests also take an average of ~15 minutes to run, we should explore if snapshot testing can reduce the overall UI test time as we only have 22 UI tests.
2. Integration tests - specifically tests relying on API calls - the integration tests that were failing previously were API calls to `fetchOrReturnRemoteConfig` - similarly if we deem these tests necessary the only solution is to bump timeouts since the tests rely on an external service.
3. Cardinal Unit tests - these tests tend to rely on actual networking requests and/or calls to `cardinalSession.setup` - since this is an external service bumping timeouts is the first line of defense. That said we should also explore mocking out the `setup` step from Cardinal to see if there is a more reliable way to test these cases vs relying on actual networking requests.

### Considerations
Things we can/should explore in a future PR and open questions:
- Consider using self-hosted runners, cached dependencies, and cached build folders in GitHub actions - without these the runners will always build this from scratch
    - would this also be able to help our carthage build times if we are able to cache some portions of the SDK?
- Consider mocking Cardinal and/or reducing our Cardinal UI tests to 1 happy path and 1 sad path - we need to determine what and how extensive our UI tests need to be as this could reduce our longest running test
- Can we avoid fetching a client token for our UI tests - while this won't have a significant impact as it's one of the fastest parts of these test, it would reduce the test times where used by a couple of seconds

### Summary of changes

- Update Unit, Integration, and UI tests to fix some flakiness that we have observed

### Testing
In order to test these changes I ran the full suite of tests 10 times on this PR. Here are the results of those runs:
- Run 1: all tests passed
- Run 2: UI test `testThreeDSecurePaymentFlowV2_challengeFlow_andFails` failed - all others passed
- Run 3: all tests passed
- Run 4: all tests passed
- Run 5: UI test `testThreeDSecurePaymentFlowV2_challengeFlow_andFails` failed - all others passed
- Run 6: Unit tests `BTThreeDSecureClient_Tests.testPrepareLookup_getsJsonString()` and `BTThreeDSecureClient_Tests.testStartPayment_whenNoBodyReturned_returnsAnError()` failed - all others passed
- Run 7: all tests passed
- Run 8: UI test `testThreeDSecurePaymentFlowV2_challengeFlow_andFails` failed - all others passed
- Run 9: all tests passed
- Run 10: all tests passed

**Notes:**
- On run 6 UI tests just stopped - this is a runner issue not a test issue.
- Based on the above commit https://github.com/braintree/braintree_ios/pull/1067/commits/4f2af5e6c90115c44c4b8230364cca982bb14dd0 was pushed to address the failures from the 10 runs. Again mainly bumping timeouts based on lack of 3DS mocks.

The overall results in an increase in reliability of these tests but we should continue to explore the problems outlined above to see if we can build out more robust solutions in a follow up PR.

### Checklist

### Authors

- @jaxdesmarais


Demo/UI Tests/American Express UI Tests/AmericanExpress_UITests.swift
```diff
@@ -15,19 +15,25 @@ class AmericanExpress_UITests: XCTestCase {
     }
 
     func testValidCard_receivesRewardsBalance() {
+        waitForElementToBeHittable(app.buttons["Valid card"])
```
jaxdesmarais, line 18: The Amex test need to be run with a client token (flow does not work with tokenization key). The tests were timing out due to not waiting long enough for the button to appear. Using `waitForElementToBeHittable` allows us to wait for the client token to be generated from the server and the buttons to appear. A sleep has been added to tests where not present for consistency.

Demo/UI Tests/ThreeDSecure UI Tests/ThreeDSecure_V2_UITests.swift
```diff
@@ -13,19 +13,21 @@ class ThreeDSecure_V2_UITests: XCTestCase {
         app.launchArguments.append("-Integration:BraintreeDemoThreeDSecurePaymentFlowViewController")
         app.launch()
 
-        _ = app.cardNumberTextField.waitForExistence(timeout: 10)
+        waitForElementToAppear(app.cardNumberTextField)
```
jaxdesmarais, line 16: The 3DS test need to be run with a client token (flow does not work with tokenization key). The tests were timing out due to not waiting long enough for the button to appear. Using `waitForElementToAppear ` allows us to wait for the client token to be generated from the server and the buttons to appear. A sleep has been added to tests where not present for consistency.

Demo/UI Tests/ThreeDSecure UI Tests/ThreeDSecure_UITests_Extensions.swift
```diff
@@ -68,5 +68,5 @@ internal extension XCUIApplication {
 }
 
 internal extension TimeInterval {
-    static let threeDSecureTimeout = 20.0
+    static let threeDSecureTimeout = 30.0
```
jaxdesmarais, line 71: Some of the UI tests were timing out while waiting for elements in the Cardinal window - bumping timeouts as this UI element is fully out of our control

IntegrationTests/Braintree-API-Integration-Specs/BTAPIClient_IntegrationTests.m
```diff
@@ -16,7 +16,7 @@ - (void)testFetchConfiguration_withTokenizationKey_returnsTheConfiguration {
         [expectation fulfill];
     }];
 
-    [self waitForExpectationsWithTimeout:5 handler:nil];
+    [self waitForExpectationsWithTimeout:10 handler:nil];
```
jaxdesmarais, line 19: These tests are for a network call to `fetchOrReturnRemoteConfiguration` - since we are making a network call in this test, we can bump the timeout to allow for the networking later to return a result.

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
```diff
@@ -276,7 +276,7 @@ class BTThreeDSecureClient_Tests: XCTestCase {
             expectation.fulfill()
         }
 
-        waitForExpectations(timeout: 4)
+        waitForExpectations(timeout: 10)
```
jaxdesmarais, line 279: Similar to the UI tests, many of these tests are waiting on a result from the cardinal `setup` call. In the future, we should consider mocking the Cardinal API call (which may prove to be difficult). A number of our 3DS tests rely on making actual network requests either in our SDK or the Cardinal SDK. We should consider revisiting these test as a whole to determine what can be mocked in this class.

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
```diff
@@ -350,30 +348,55 @@ class BTThreeDSecureClient_Tests: XCTestCase {
...
+                    "status": "authenticate_successful",
+                ] as [String: Any],
+                "type": "CreditCard",
+            ] as [String: Any],
+            "success": true,
+            "threeDSecureInfo":     [
+                "liabilityShiftPossible": true,
+                "liabilityShifted": true,
+            ],
+            "lookup": [
```
jaxdesmarais, line 387: This specific test was looking for the existence of this `lookup` object. Mocking this result allows us to speed up and hopefully hit this timeout less frequently.

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
```diff
@@ -253,8 +253,29 @@ class BTThreeDSecureClient_Tests: XCTestCase {
...
+            XCTAssertEqual(error?.localizedDescription, "BTThreeDSecureRequest amount can not be nil or NaN.")
+            expectation.fulfill()
+        }
+
+        waitForExpectations(timeout: 2)
+        XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTThreeDSecureAnalytics.verifyFailed))
+    }
     
-    func testStartPayment_returnsError_whenAmountIsMissing() {
+    func testStartPayment_whenNoBodyReturned_returnsAnError() {
```
jaxdesmarais, line 278: This was actually testing `noBodyReturned` and not missing amount, updated the test name and added the test `testStartPaymentFlow_whenAmountIsNil_throwsError` to test what this test originally said it was testing.

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
scannillo: ➕ 1 - our unit tests should not make any network requests / should work fine without internet

Demo/UI Tests/ThreeDSecure UI Tests/ThreeDSecure_UITests_Extensions.swift
scannillo: The weird thing is that our UI and Integration test were relatively stable for the longest time.

These issues seem to be correlated to [our bump to Xcode 14.3 and Mac OS 13](https://github.com/braintree/braintree_ios/pull/1031/files) ~2 months ago. We should dig in GH Actions Issues to see if other users are experience similar issues.

Demo/UI Tests/ThreeDSecure UI Tests/ThreeDSecure_V2_UITests.swift
scannillo: We have some hard coded client tokens for our integration tests ([see code](https://github.com/braintree/braintree_ios/blob/803a41d3ad64b0792e755c37400947154911eefb/IntegrationTests/BraintreePayPal_IntegrationTests.swift#L16-L39)). We can see if these are usable in the UI tests as well.

sshropshire (approved):

sshropshire:
@jaxdesmarais the Cardinal ones do fail on Android from time to time. I wonder if there's a clean way to retry the failed tests until they all pass. Not ideal, but the other option is to work with Cardinal to see why they fail–there may be legitimate bugs.

scannillo:
I'm not able to understand why our logging for the UnitTests does not show failure reasons, while UI and IntegrationTests do since their setup in the `tests.yml` file is the exact same. My only guess here is since the UnitTests scheme is a combo of several test targets combined, while UI and Integration schemes are standalone.

We should try out moving to fastlane - see if the built in logging there is any better (also would love some better timing analysis & to get rid of the xcbeautify Pod 😄)

Demo/UI Tests/ThreeDSecure UI Tests/ThreeDSecure_UITests_Extensions.swift
jaxdesmarais: The integration tests were stable but the UI tests have been flakey for a while - though they have certainly gotten worse. I looked through the runner GitHub issues and was unable to find something specific to this Xcode + macOS version combo, there are similar reports in earlier macOS and Xcode versions but ours were more consistent on those older versions.

IntegrationTests/Braintree-API-Integration-Specs/BTAPIClient_IntegrationTests.m
jaxdesmarais: Yeah, agree it shouldn't take that long - I do think we should seriously consider exploring self hosted runners. This is a bandaid fix but the issue ultimately seems to lie somewhere in the GitHub action runner resources.

Demo/UI Tests/ThreeDSecure UI Tests/ThreeDSecure_V2_UITests.swift
jaxdesmarais: That's a good call - added a launch argument to use a hardcoded client token in UI tests here: https://github.com/braintree/braintree_ios/pull/1067/commits/1e482626fbedd946ad7fe069918d337842d43995

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
jaxdesmarais: I created a new branch off of this PR to mock the class where the setup call is made - if we like this approach to mock the setup step, I can merge it into this branch: [mock-cardinal-setup-call](https://github.com/braintree/braintree_ios/compare/fix-flakey-tests...mock-cardinal-setup-call) - I was able to bump almost all of the timeouts back down.

IntegrationTests/Braintree-API-Integration-Specs/BTAPIClient_IntegrationTests.m
scannillo: Yeh - we should look into other iOS specific CI offerings like BitRise and CircleCI. I think a custom solution would be a ton of work and most companies have entire teams dedicated to that infrastructure. I would love to learn more about what resources Venice & Venmo use 👀

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
jaxdesmarais: Merged these changes into the branch

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
```diff
@@ -281,12 +287,17 @@ class BTThreeDSecureClient_Tests: XCTestCase {
...
-            "assetsUrl": "http://assets.example.com"
-        ] as [String: Any])
-        let client = BTThreeDSecureClient(apiClient: mockAPIClient)
+        let mockThreeDSecureV2Provider = MockThreeDSecureV2Provider(
+            configuration: BTConfiguration(json: mockConfiguration),
+            apiClient: mockAPIClient,
+            request: threeDSecureRequest,
+            completion: mockCompletion
+        )
+
```
KunJeongPark, line 299: Is the purpose of this `mockThreeDSecureV2Provider` to avoid having to make `apiClient.fetchOrReturnRemoteConfiguration` call?

scannillo: Good question - it's to prevent[ this call](https://github.com/braintree/braintree_ios/blob/9576e337797013bc2874f8b75f6e21989a21549c/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift#L233-L242) (which eventually calls Cardinal's [setup()](https://github.com/braintree/braintree_ios/blob/9576e337797013bc2874f8b75f6e21989a21549c/Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift#L48-L54). Setup does API calls we want to prevent for UnitTests).

jaxdesmarais: Yeah you are right, we are essentially forcing the tests not to care about when `threeDSecureV2Provider = BTThreeDSecureV2Provider()` is called. We can create a `MockThreeDSecureClient` class though and override `prepareLookup` where the above is called to get around this. Just tested this locally, will push the commit up shortly. Good catch!

jaxdesmarais: Updated here: https://github.com/braintree/braintree_ios/pull/1067/commits/e5215687f8e211b5696a841770b7122b0fb92d88

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
```diff
@@ -253,20 +258,37 @@ class BTThreeDSecureClient_Tests: XCTestCase {
...
         let expectation = expectation(description: "willCallCompletion")
+        mockAPIClient.cannedConfigurationResponseBody = mockConfiguration
 
-        mockAPIClient.cannedConfigurationResponseBody = BTJSON(value: [
-            "threeDSecure": ["cardinalAuthenticationJWT": "FAKE_JWT"],
-            "assetsUrl": "http://assets.example.com"
-        ] as [String: Any])
-        let client = BTThreeDSecureClient(apiClient: mockAPIClient)
-
+        let client = MockThreeDSecureClient(apiClient: mockAPIClient)
```
scannillo, line 291: @jaxdesmarais A general rule of thumb - you don't want to mock the system/subject under test that you're testing. So in this file, the SUT is `BTThreeDSecureClient`. If we do this, then we aren't actually testing `BTThreeDSecureClient`.

I think the cleanest approach here is to mock only the dependencies of `BTThreeDSecureClient` that need to be: for example to avoid network calls, DBs, or other external dependencies.

Check out[ this branch](https://github.com/braintree/braintree_ios/compare/fix-flakey-tests...mock-cardinal-session) which mocks `CardinalSession` directly via a protocol.

jaxdesmarais: That makes sense. Yeah I wasn't sure where the line was with mocking the SUT in this case. But mocking `CardinalSession` via a protocol makes sense to me. I'll merge in that branch.

Agree we can punt on adding tests for `BTThreeDSecureV2Provider` to get us unblocked for now on test failures.

jaxdesmarais: Merged in here: https://github.com/braintree/braintree_ios/pull/1067/commits/faf81dd64913c08f6ae29371d4cb6e13c11fa02e

scannillo: Sweet! Yeh I went with `CardinalSessionTestable` for the protocol name. We should pick a naming convention for these kind of protocols, stick to it, and add it to our style guide. We have a few in PPCP too. I think some are just `<THING>Protocol`. I like `Testable` 🤷

scannillo (approved):
🔥

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureClient_Tests.swift
jaxdesmarais: Would `Testable` only be for protocols used for testing/mocking? I like that naming for those purposes but feels a little odd for things that aren't used for mocking/testing. If that makes sense. Agree we should have a standard though just not sure if we need 1 for testing/mocking and 1 for everything else/merchant facing things/etc

PR #1073, opened 2023-07-11
Update `Package.resolved` with Correct NXO Version

jaxdesmarais:
### Summary of changes

- The [docs step of the release](https://github.com/braintree/braintree_ios/actions/runs/5512469326/jobs/10049340682) failed yesterday with the error `Your local changes to the following files would be overwritten by checkout: Braintree.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved`
    - Remove `xcshareddata` from `gitignore` so updates to NXO result in the `Package.resolved` file updating as expected - `xcshareddata` contains target configurations that we want to persist across all users of the target and should be checked in
    - Add missing `IDEWorkspaceChecks.plist`
    - Bump versions in `Package.resolved`

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Braintree.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
+++ b/Braintree.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
+	<true/>
+</dict>
+</plist>
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,5 @@ SampleApps/CarthageTest/Cartfile.resolved
 .swiftpm
 .build
 SampleApps/SPMTest/SPMTest.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
-Braintree.xcworkspace/xcshareddata/swiftpm/
 
 *.pid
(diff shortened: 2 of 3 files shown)


Braintree.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jaxdesmarais, line 1: This file was generated automatically when updating the `Package.resolved`. Looking at [this StackOverflow thread](https://stackoverflow.com/questions/50367896/ideworkspacechecks-plist-file-suddenly-appear-after-updated-xcode) it seems like it's necessary. From the Apple release notes:

> Xcode 9.3 adds a new IDEWorkspaceChecks.plist file to a workspace's shared data, to store the state of necessary workspace checks. Committing this file to source control will prevent unnecessary rerunning of those checks for each user opening the workspace.

sshropshire (approved):

Braintree.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
...
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
```
scannillo, line 5: We dropped 32bit device support a long time ago (as did Apple). Since we don't need a check for that anymore, do we need this file?

scannillo (approved):

Braintree.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
jaxdesmarais: It seems like we still need it since it was generated automagically. We have one in [PPCP as well](https://github.com/paypal/iOS-SDK/blob/2b39110e75b5b8c4c79f16b76a6e5e010230f8ec/PayPal.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist#L5) so I am guessing if we don't include it it'll be generated and need to be ignored with each commit.

scannillo: Gotcha - thanks for checking!

PR #1075, opened 2023-07-12
Fix Venmo Dictionary Construction

jaxdesmarais:
### Summary of changes

- Fix bug where tokenizations failed by only adding transaction details dictionary when present (fixes #1074)
- Fix typo in docstrings

### Checklist

- [x] Added a changelog entry

### Authors

- @scannillo @jaxdesmarais

PR #974, opened 2023-03-15
Add address, amount and line item fields to Venmo createPaymentContext

khushboo18:
Thank you for your contribution to Braintree. 

> Before submitting this PR, note we cannot accept language translation PRs. We support the same languages that are supported by PayPal, and have a dedicated localization team to provide the translations. If there is an error in a specific translation, you may open an issue and we will escalate it to the localization team.

### Summary of changes

Adds the following fields to BTVenmoRequest to be used in createPaymentContext:

2 new address collection fields -

collectCustomerBillingAddress
collectCustomerShippingAddress

and 6 new fields for transaction amount and line items -

lineItems
subTotalAmount
discountAmount
taxAmount
shippingAmount
totalAmount

Note: Should not be merged until [this GraphQL PR](https://github.braintreeps.com/braintree/cosmos-apps/pull/8614) is merged.

### JIRA
https://paypal.atlassian.net/browse/DTBTVENMO-940

### Checklist

- [x] Added a changelog entry

- khushboo18

PR #1076, opened 2023-07-13
Clarify docstrings for BTVenmoRequest line items

scannillo:
### Summary of changes

- Follow up PR to #1075 
- Call out in docstrings that if you set one of the new params introduced in https://github.com/braintree/braintree_ios/pull/974
    - If not, the tokenization API results in a failure
    - Apparently this is a known behavior of the API (confirmed by @khushboo18)

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeVenmo/BTVenmoRequest.swift
+++ b/Sources/BraintreeVenmo/BTVenmoRequest.swift
@@ -43,29 +43,43 @@ import Foundation
     public var displayName: String?
     
     /// Whether the customer's billing address should be collected and displayed on the Venmo paysheet.
-    /// Defaults to `false`
+    /// Defaults to `false`.
+    ///
+    /// If this value is set, `totalAmount` must also be set.
     public var collectCustomerBillingAddress: Bool = false
     
     /// Whether the customer's shipping address should be collected and displayed on the Venmo paysheet.
-    /// Defaults to `false`
+    /// Defaults to `false`.
+    ///
+    /// If this value is set, `totalAmount` must also be set.
     public var collectCustomerShippingAddress: Bool = false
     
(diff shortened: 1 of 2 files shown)


Sources/BraintreeVenmo/BTVenmoRequest.swift
```diff
@@ -43,29 +43,43 @@ import Foundation
...
+    ///
+    /// If this value is set, `totalAmount` must also be set.
     public var collectCustomerBillingAddress: Bool = false
     
     /// Whether the customer's shipping address should be collected and displayed on the Venmo paysheet.
-    /// Defaults to `false`
+    /// Defaults to `false`.
+    ///
+    /// If this value is set, `totalAmount` must also be set.
     public var collectCustomerShippingAddress: Bool = false
```
scannillo, line 55: @khushboo18 - can you confirm that `totalAmount` is required when `collectCustomerBillingAddress` and `collectCustomerShippingAddress` are set?

I am seeing the same error for these params, but just want to confirm this is also expected.

khushboo18: `totalAmount` is only _required_ when `transactionDetails` are set

khushboo18: Merchants can set `collectCustomerBillingAddress` and `collectCustomerShippingAddress` without having to pass `totalAmount` as long as no other amounts or line items are being sent

scannillo: Got it - addressed! 7406aff3b7be8f51b1b50d618c8af4dd89c96efc

khushboo18 (approved):
🌮 LGTM! 
Thanks for making this change

jaxdesmarais (approved):
👻

PR #1077, opened 2023-07-13
Add `BTAppContextSwitcher.setReturnURLScheme` func missing from v5 conversion

scannillo:
### Summary

We got an [inbound](https://github.com/braintree/braintree_ios/discussions/1070) about this function missing when using our V6 SDK from a Swift app

### Changes
- [Add `setReturnURLScheme` func](https://github.com/braintree/braintree_ios/blob/752a71f2381dbef641c1865c38ce2d88779013b4/Sources/BraintreeCore/Public/BraintreeCore/BTAppContextSwitcher.h#L52-L60), like we had in V5
- Cleanup tests to use this setter over accessing `sharedInstance` directly

### Discussion

What used to be static in V5 (which is all of the public methods on [BTAppContextSwitcher](https://github.com/braintree/braintree_ios/blob/5.x/Sources/BraintreeCore/Public/BraintreeCore/BTAppContextSwitcher.h)) are class methods in V6. So currently in V6, merchants need to call these functions like this: `BTAppContextSwitcher.sharedInstance().handleOpenURL(context: UIOpenURLContext)` or `BTAppContextSwitcher.sharedInstance.handleOpenURL(URL)`.

I am guessing this was not intentional, since it isn't capture in our migration guide?

IMO merchants should not have access to `sharedInstance` and all methods on `BTAppContextSwitcher` should be static.

If folks agree, we'll have to decide how we want to proceed. A few options:
- Treat this as a "bug fix" and hide `sharedInstance` & convert methods back to static (most generous)
- Deprecate class methods & `sharedInstance` property + add new static methods that merchants should use

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeCore/BTAppContextSwitcher.swift
+++ b/Sources/BraintreeCore/BTAppContextSwitcher.swift
@@ -42,6 +42,14 @@ import UIKit
         return false
     }
     
+    ///  Sets the return URL scheme for your app.
+    ///
+    ///  This must be configured if your app integrates a payment option that may switch to either ASWebAuthenticationSession or to another app to finish the payment authorization workflow.
+    /// - Parameter scheme: The return URL scheme
+    public static func setReturnURLScheme(_ scheme: String) {
+        sharedInstance.returnURLScheme = scheme
+    }
+    
     /// Registers a class `Type` that can handle a return from app context switch with a static method.
     /// - Parameter client: A class `Type` that implements `BTAppContextSwitchClient`, the methods of which will be invoked statically on the class.
     @objc(registerAppContextSwitchClient:)
--- a/UnitTests/BraintreeCoreTests/BTAppContextSwitcher_Tests.swift
+++ b/UnitTests/BraintreeCoreTests/BTAppContextSwitcher_Tests.swift
@@ -4,11 +4,6 @@ import XCTest
 class BTAppContextSwitcher_Tests: XCTestCase {
     var appSwitch = BTAppContextSwitcher.sharedInstance
 
-    override func setUp() {
-        super.setUp()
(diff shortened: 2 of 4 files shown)


scannillo:
👋 @jaxdesmarais thanks for the background here!

I get where you're coming from in wanting it to be transparent. IMO - I don't see what value the merchant gets from knowing it's a singleton. It feels weird to expose `sharedInstance` at all since there is nothing a merchant should be doing with it.

Also, I don't think we even need a singleton in this file. This could just be a "utility" class with all static methods & properties ([see Stack Overflow](https://stackoverflow.com/questions/37806982/difference-between-static-function-and-singleton-class-in-swift)).

That's just my two cents though! So going forward we can:
- Leave everything as-is
- Add back the static helper methods + "deprecate" the sharedInstance class methods
- Leave a note for V7 to remove the singleton

Whatever we decide we should just be sure to update the migration guide & docs. Thoughts?

jaxdesmarais:
@scannillo -

I am fine with any of those options! I don't feel super strongly in favor/opposed to any of them, though the 1st and 3rd would be the least breaking. But even option 2 we can set up as a non breaking warning. Are you leaning one way vs the other?

scannillo:
> I am fine with any of those options! I don't feel super strongly in favor/opposed to any of them, though the 1st and 3rd would be the least breaking. But even option 2 we can set up as a non breaking warning. Are you leaning one way vs the other?

Hm - I lean toward option 2 + a note to remove the singleton in V7. It just feels a _little_ cleaner to me. I'll make those changes to this PR and we can see how it feels.

scannillo:
Closing due to low priority

PR #1083, opened 2023-07-31
Send `live` over `production` in FPTI payload

scannillo:
### Summary
While creating analytics dashboards, we realized that the BT & PPCP SDKs differ in the values we are sending for the `merchant_sdk_env` key to FPTI. BT Mobile SDKs are sending: production & sandbox; PPCP mobile SDKs are sending live & sandbox. 

To easily look at our analytics cross products, we want to keep things the same as much as possible.

### Changes

- Send `live` instead of `production` in the FPTI payload
- This makes more sense to me than sending `production` from the PPCP SDKs for a few reasons
    - In FPTI, "live" is the accepted name for the FPTI Data Service's "live" environment
    - Alignment with existing PP products already on FPTI (PP JS, NXO, etc)

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeCore/BTConfiguration.swift
+++ b/Sources/BraintreeCore/BTConfiguration.swift
@@ -14,6 +14,14 @@ import Foundation
     public var environment: String? {
         json?["environment"].asString()
     }
+    
+    /// The environment name sent to PayPal's FPTI analytics service
+    var fptiEnvironment: String? {
+        if environment == "production" {
+            return "live"
+        }
+        return environment
+    }
 
     /// :nodoc: This initalizer is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     ///  Used to initialize a `BTConfiguration`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## unreleased
 * BraintreeVenmo
   * Add additional error parsing for Venmo errors
+* BraintreeCore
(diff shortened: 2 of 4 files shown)


sshropshire (approved):

Sources/BraintreeCore/BTConfiguration.swift
```diff
@@ -14,6 +14,14 @@ import Foundation
     public var environment: String? {
         json?["environment"].asString()
     }
+    
+    /// The environment name sent to PayPal's FPTI analytics service
+    var fptiEnvironment: String? {
+        if environment == "production" {
+            return "live"
+        }
+        return environment
```
jaxdesmarais, line 23:
```suggestion
        environment == "production" ? "live" : environment
```

jaxdesmarais (approved):
We need to ensure that this change is documented somewhere for product/analytics team/us. Since this SDK is already GA we will need to track both before and after this change to get accurate results in our dashboards vs only tracking this new value since any merchant below a certain version in production will use the "old" environment.

scannillo:
> We need to ensure that this change is documented somewhere for product/analytics team/us. Since this SDK is already GA we will need to track both before and after this change to get accurate results in our dashboards vs only tracking this new value since any merchant below a certain version in production will use the "old" environment.

For sure, this is a good call! I added a CHANGELOG entry, so we can have record of this change.

But I also don't think this warrants much communication at this point. The analytics team helping with our dashboards are already accounting for both names (live & production). Also these early V6 versions (6.4.0 and below) are seeing really low traffic, so I think it will be in our later versions that a merchant actually go live with V6. At that point, we can probably just ignore any `production` marked events.

Issue #1080, opened 2023-07-19
Make errors public


mikezaslavskiy:
### Is your feature request related to a problem? Please describe.

In 6.x.x release the errors become internal with a reference to use error codes instead from [this](https://github.com/braintree/braintree_ios/blob/main/SDK_ERROR_CODES.md) list. Meaning that developers who are using the SDK will probably need to compare the error codes. However, this might be a bit unsafe since the developers have no control over the internal processes and modifications to error codes and can't be sure that the error code won't change in the future, potentially causing an outage.

### Describe the solution you'd like.

Make error codes public

jaxdesmarais:
Hello @mikezaslavskiy -

Similar to the [locale code issue](https://github.com/braintree/braintree_ios/issues/1079) we cannot expose the errors publicly as we support both Swift and Objective-C. Objective-C enums require a raw `Int` type. 

Could you share a bit more about how you're looking to use the error codes in the error enum? Our intent is to follow semantic versioning with our error codes, but we'd like to understand your use case a bit better to ensure we are coming up with the best solution.

mikezaslavskiy:
So in our codebase, we will need to define the errors we need to handle ourselves and map them to the error codes provided in the documentation. However, the biggest fear here is for some reason error codes will change under the hood - our error-handling logic might be broken. How stable will be the error codes provided in the documentation? As far as I understand you don't plan to change them? If you will - will it be considered a breaking change?

jaxdesmarais:
Hey @mikezaslavskiy -

I know this is quite an old issue but just wanted to let you know we moved forward with a decision to make our errors public (and equatable). This change was released in [version 6.11.0](https://github.com/braintree/braintree_ios/releases/tag/6.11.0) and should make it easier to compare errors without relying on the code documentation if desired. 🚀

PR #1087, opened 2023-08-01
Add BTVenmoError.canceled case

### LINKED ISSUE #1085: BTVenmoClient `tokenize` async method leaks memory and does not return/throw an error when the user cancels the authorisation.
### Braintree SDK Version

6.4.0

### Environment

Both

### Xcode Version

Xcode 14.3.1

### OS Version & Device

iOS 16.2

### Integration type

Swfit Package Manager

### Development Processor

Intel

### Describe the bug

Given the sample provided below when the user cancels the authorisation within the Venmo app the `tokenize` method call does not return an error or at least something that indicates the cancelation.

```
venmoClient = BTVenmoClient(apiClient: btApiClient)
let request = BTVenmoRequest(paymentMethodUsage: .multiUse)
do {
    let nonceAccount = try await venmoClient.tokenize(request)
    // send nonceAccount.nonce
} catch let error {
    // handle error here
}
```

### To reproduce

1. Create an `BTVenmoRequest`
2. Call the `tokenize` async method
3. Cancel the authorisation from Venmo app.
4. Check the return nonce / or error

### Expected behavior

I would expect to get an error code that would indicate the cancelation.

### Screenshots

Screenshot 2023-08-01 at 17 09 58
Screenshot 2023-08-01 at 17 15 30

scannillo:
### Summary of changes

Fixes #1085. Previously, we returned a completion of `(nil, nil)` when the Venmo flow is canceled by the user. This results in a no-op for merchants using our async-await function wrappers.

This PR aligns with how the SEPA & PayPal modules handle cancelation cases, by returning an error to the merchant (either via the completion, or throwing if using async-await).

- Add a `.canceled` error case to `BTVenmoError`
- Clarify docstring
- Update SDK_ERROR_CODES.md

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo


scannillo:
CI is having a field day

Demo/Application/Features/Venmo - Custom Button/BraintreeDemoVenmoViewController.swift
```diff
@@ -80,7 +80,11 @@ class BraintreeDemoVenmoViewController: BraintreeDemoPaymentButtonBaseViewContro
                 progressBlock("Got a nonce 💎!")
                 completionBlock(venmoAccount)
             } catch {
-                progressBlock(error.localizedDescription)
+                if (error as NSError).code == 10 {
```
sshropshire, line 83: nit: I know we can reference the error codes markdown, but this code would read better if there was a static constant used in place of `10`.

Sources/BraintreeVenmo/BTVenmoError.swift
```diff
@@ -57,6 +57,9 @@ enum BTVenmoError: Error, CustomNSError, LocalizedError {
 
     /// 9. Enriched Customer Data is disabled
     case enrichedCustomerDataDisabled
+    
+    /// 10.  The Venmo flow was canceled by the user
+    case canceled
```
sshropshire, line 62: q: is this a breaking change?

sshropshire (approved):

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## unreleased
 * BraintreeVenmo
   * Add additional error parsing for Venmo errors
+  * Throw cancelation specific error for `BTVenmoClient.tokenize()` (fixes #1085)
```
jaxdesmarais, line 6: Should we call out more specifically that we no longer return `nil, nil` in the completion for cancel? It's possible merchants have logic built around that expectation so may be good to call it out more explicitly.

jaxdesmarais (approved):

Demo/Application/Features/Venmo - Custom Button/BraintreeDemoVenmoViewController.swift
scannillo: That's a good call. And I think it should be easier for a merchant to check against our error codes (a merchant asked about it in #1080) versus needing them to check our SDK_ERROR_CODES.md file list.

Something like : `if error.code == BTVenmoError.cancel.code { // do things }` would be nice to expose for merchants.

@jaxdesmarais What do you think? Should we ticket this work?

Sources/BraintreeVenmo/BTVenmoError.swift
scannillo: Nope, since we are only adding an error code & not changing existing ones we didn't think this was breaking

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -54,7 +54,8 @@ import BraintreeCore
     /// - Parameters:
     ///   - request: A Venmo request.
     ///   - completion: This completion will be invoked when app switch is complete or an error occurs. On success, you will receive
-    ///   an instance of `BTVenmoAccountNonce`; on failure, an error; on user cancellation, you will receive `nil` for both parameters.
+    ///   an instance of `BTVenmoAccountNonce`; on failure or user cancelation you will receive an error.
+    ///   If the user cancels out of the flow, the error code will be `.canceled`.
     @objc(tokenizeWithVenmoRequest:completion:)
```
scannillo, line 59: The docs for the async method also needs this update!

Demo/Application/Features/Venmo - Custom Button/BraintreeDemoVenmoViewController.swift
jaxdesmarais: Yeah, I think I would still lean towards exposing the errors as a whole publicly for merchants using Swift. Since many of our errors use associated types, we can't expose those to Obj-C publicly as only `Int` type enums can be exposed. 

If we wanted to expose them publicly for Obj-C merchants we would need to move the error codes into their own enum without associated types and construct the error messages at the callsite or in a method where we can include the details we are currently including with the associated type when constructing the `localizedDescription`.

jaxdesmarais:
> CI is having a field day

Should be fixed now! See PR: https://github.com/braintree/braintree_ios/pull/1089

CHANGELOG.md
scannillo: Done!

Demo/Application/Features/Venmo - Custom Button/BraintreeDemoVenmoViewController.swift
scannillo: Yeh - there are ways to do it without having to move the error creation to the callsite, though it makes our internal `<Feature>Error.swift` files a little uglier. Something like this:

```
enum BTAmericanExpressError: Int, Error, CustomNSError, LocalizedError {

    case unknown

    case noRewardsData

    case deallocated

    static var errorDomain: String {
        "com.braintreepayments.BTAmericanExpressErrorDomain"
    }
    
    var errorCode: Int {
        switch self {
        case .unknown:
            return BTAmericanExpressErrorCode.unknown.rawValue
        case .noRewardsData:
            return BTAmericanExpressErrorCode.noRewardsData.rawValue
        case .deallocated:
            return BTAmericanExpressErrorCode.deallocated.rawValue
        }
    }
    
    var errorDescription: String? {
        switch self {
        case .unknown:
            return "An unknown error occurred. Please contact support."

        case .noRewardsData:
            return "No American Express Rewards data was returned. Please contact support."

        case .deallocated:
            return "BTAmericanExpressClient has been deallocated."
        }
    }
}


/// Public for merchants to reference (accessible in both Swift & ObjC)
@objc public enum BTAmericanExpressErrorCode: Int {
    
    case unknown = 0
    case noRewardsData = 1
    case deallocated = 2
}
```

There might even be a slicker way to do that switch statement, but something like this would solve the issue

jaxdesmarais: Ah, yeah I hadn't thought about just extracting the code portion.

Yeah, I guess the question in that case would be the merchant experience. With the above the callsite would look something like:
```swift
if (error as? NSError)?.code == BTAmericanExpressErrorCode.noRewardsData.rawValue {
    // do something for this specific error
}
```
vs making the enum public the callsite could look something like:
```swift
switch error as? BTAmericanExpressError {
case . noRewardsData:
    // do something for this specific error
default:
    // do something else for other errors
}
```
I think you could technically use a switch for the first but it'd be something like `switch ((error as? NSError)?.code as? BTJSONErrorCode)`. It really comes down to how we want to support Obj-C vs Swift first in the future, which is up to us! We can certainly play around with some other options and see what feels best or engage the merchant that opened the issue to see what their ideal world would be.

scannillo: Thats a great point. I added this convo to our "iOS Minor Version Improvements" doc!

PR #1089, opened 2023-08-04
Update Test Workflow

jaxdesmarais:
### Summary of changes

- CI started 💩 itself out of nowhere - did some research and found [this issue on the GitHub actions runner](https://github.com/actions/runner-images/issues/8023)
    - Seems like `macos-13` is now defaulting to iOS 17... which is still in beta so buggy AF
    - We can hardcode our OS in `xcodebuild` as indicated in that thread [here](https://github.com/actions/runner-images/issues/8023#issuecomment-1662478605) - hopefully in a later Xcode beta/GA this will be resolved and we can go back to letting CI do it's normal thing if we want 

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Install CocoaPod dependencies
         run: pod install
       - name: Run Unit Tests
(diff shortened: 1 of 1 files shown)


hollabaq86 (approved):
fun times 🙃 , ship it!

sshropshire (approved):

PR #1093, opened 2023-08-25
Convert `BraintreeDemoIdealViewController` to Swift

jaxdesmarais:
### Summary of changes

- Convert `BraintreeDemoIdealViewController` to Swift

### Future Demo App Improvements
- Remove `BraintreeDemo` from all the class names - it does not seem necessary
- Consider flattening the feature view controllers - right now they are all in their own directories which doesn't seem necessary especially when there is 1 file in each directory in the case of Swift
- Convert the rest of the Demo ViewControllers
- Combine PayPal Web ViewControllers into 1 ViewController similar to NXO
- ... I am sure we will find more ways to make it better 😄 

### Checklist

### Authors

- @jaxdesmarais


scannillo (approved):
I like those improvement ideas!

PR #1094, opened 2023-08-25
Convert `BraintreeDemoBTDataCollectorViewController` to Swift

jaxdesmarais:
### Summary of changes

- Convert `BraintreeDemoBTDataCollectorViewController` to Swift

### Checklist

### Authors

- @jaxdesmarais


Demo/Application/Base/Base View Controllers/BraintreeDemoPaymentButtonBaseViewController.m
```diff
@@ -23,7 +23,8 @@ - (void)viewDidLoad {
     [self.view addSubview:self.paymentButton];
 
     [NSLayoutConstraint activateConstraints:@[
-        [self.paymentButton.centerXAnchor constraintEqualToAnchor:self.view.centerXAnchor],
+        [self.paymentButton.leadingAnchor constraintEqualToAnchor:self.view.leadingAnchor],
+        [self.paymentButton.trailingAnchor constraintEqualToAnchor:self.view.trailingAnchor],
```
jaxdesmarais, line 27: Not having a leading and trailing anchor means that the text will expand forever. In the case of printing device data, we don't want to happen. Previously `BraintreeDemoBTDataCollectorViewController` subclassed `BraintreeDemoBaseViewController` so this was a non issue. That doesn't make a ton of sense to have some ViewControllers subclass one view and others subclass a different one so updated for consistency.

Demo/Application/Features/Fraud Protection - BTDataCollector/BraintreeDemoBTDataCollectorViewController.swift
```diff
@@ -0,0 +1,88 @@
+import UIKit
+import CoreLocation
+import BraintreeDataCollector
+import BraintreeCore
+
+class BraintreeDemoBTDataCollectorViewController: BraintreeDemoPaymentButtonBaseViewController {
+
+    let locationManager = CLLocationManager()
+    var dataLabel = UILabel()
```
sshropshire, line 9: take or leave: Lazy instantiation would allow for `dataLabel` to be non-nil. The initial `UILabel` created here ends up being overwritten in `createPaymentButton()`.

sshropshire (approved):

Demo/Application/Features/Fraud Protection - BTDataCollector/BraintreeDemoBTDataCollectorViewController.swift
jaxdesmarais: Using a `lazy var` would still require the property to have an initializer. Since we need access to this property in `tappedCollect` to update the text I think it makes sense as is to create an empty label them override it later.

Demo/Application/Features/Fraud Protection - BTDataCollector/BraintreeDemoBTDataCollectorViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+        switch locationStatus {
+        case .notDetermined:
+            locationManager.requestWhenInUseAuthorization()
+
+        default:
+            progressBlock("Location authorization requested previously. Update authorization in Settings app.")
+        }
+    }
+
+    private func createLocationPermissionButton() {
```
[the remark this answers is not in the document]
jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/1094/commits/af8a0e06d82a9cc7d0276de5fd9bf02402907851

Did we used to require merchants to do this? I've never known why it existed in this view controller.

PR #1096, opened 2023-08-28
Remove NXO Package Dependency from Demo App

jaxdesmarais:
### Summary of changes

- We had NXO as a package dependency for the Demo app - it doesn't seem like it should be needed directly in our Demo app so we can likely remove it
- The NXO flow still works as expected in the demo app with this change

### Checklist

### Authors

- @jaxdesmarais


scannillo (approved):
😍

sarahkoop (approved):

PR #1100, opened 2023-09-14
Fix Bug where `fetchPaymentMethodNonces` Return Incorrect Type

### LINKED ISSUE #1099: fetchPaymentMethodNonces in BTAPIClient return wrong nonce types
### Braintree SDK Version

6.6.0

### Environment

Sandbox

### Xcode Version

Xcode 14.3.1

### OS Version & Device

iOS 16.4

### Integration type

CocoaPods

### Development Processor

Intel

### Describe the bug

The fetchPaymentMethodNonces method in BTAPIClient always returns the nonce in the "Unknown" type.

Root cause: After the SDK migrated to Swift. The jsonParsingBlocks in BTPaymentMethodNonceParser is not get registered with different types.  

### To reproduce

1. Call fetchPaymentMethodNonces in BTAPIClient when there are saved nonce.
2. Debug the nonce values. (Check the screenshot)

### Expected behavior

1. The SDK should behave as the Objective-C version
2. The nonce type should be the correct type.

### Screenshots

CleanShot 2023-09-14 at 09 07 13@2x

jaxdesmarais:
### Summary of changes

- We were never parsing out the type from the nonce as expected
    - In Objective-C we did this in the load method, we can instead just parse it out from the JSON
- Fixes #1099 

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais


Sources/BraintreeCore/BTPaymentMethodNonceParser.swift
```diff
@@ -53,18 +53,80 @@ import Foundation
...
         }
 
-        if json?["nonce"].isString != false {
+        if json?["nonce"].isString == false {
+            return nil
+        }
+
+        let type = json?["type"].asString()
+
+        if type == "CreditCard", let cardType = json?["details"]["cardType"].asString() {
```
jaxdesmarais, line 66: The below logic mirrors the behavior of iOS v5 with regards to the type. Though this is with the constraints of not being able to cast to unknown nonce types in Swift. The load method used previously was a very hacky workaround and does not map to Swift - the method (`load`) in Obj-C is prohibited in Swift. Currently the JS SDK has the method `fetchPaymentMethodNonces` in Core while Android has this method in the Drop-in. We should consider where this method should live and align across all 3 platforms.

UnitTests/BraintreePayPalTests/BTPayPalAccountNonce_Tests.swift
```diff
@@ -1,33 +1,9 @@
 import XCTest
 @testable import BraintreePayPal
 
-class BTPaymentMethodNonceParser_PayPal_Tests: XCTestCase {
-    func testSharedParser_whenTypeIsPayPal_returnsPayPalAccountNonce() {
```
jaxdesmarais, line 5: Moved the parser specific test `BTPaymentMethodNonceParser_Tests.swift`

UnitTests/BraintreePayPalTests/BTPayPalAccountNonce_Tests.swift
```diff
@@ -1,33 +1,9 @@
...
-            )
-        )
-
-        XCTAssertEqual(payPalAccountNonce?.nonce, "a-nonce")
-        XCTAssertEqual(payPalAccountNonce?.type, "PayPal")
-        XCTAssertEqual(payPalAccountNonce?.email, "jane.doe@example.com")
-        XCTAssertTrue(payPalAccountNonce!.isDefault)
-        XCTAssertNil(payPalAccountNonce?.creditFinancing)
-    }
+final class BTPayPalAccountNonce_Tests: XCTestCase {
```
jaxdesmarais, line 4: Renamed this class and the tests since they were not actually testing the shared parser and the test names were not accurate.

Sources/BraintreeCore/BTPaymentMethodNonceParser.swift
```diff
@@ -53,18 +53,80 @@ import Foundation
...
+        } else if type == "PayPalAccount" {
+            return BTPaymentMethodNonce(
+                nonce: json?["nonce"].asString() ?? "",
+                type: "PayPal",
+                isDefault: json?["default"].isTrue ?? false
+            )
+        } else if type == "VenmoAccount" {
+            return BTPaymentMethodNonce(
+                nonce: json?["nonce"].asString() ?? "",
+                type: "Venmo",
```
scannillo, line 87: It looks like we previously had used `VenmoAccount` ([see V5 load method](https://github.com/braintree/braintree_ios/blob/75a48494b48dc83204ced72a0cee68057a8e7170/Sources/BraintreeVenmo/BTVenmoDriver.m#L49-L51))

Sources/BraintreeCore/BTPaymentMethodNonceParser.swift
```diff
@@ -53,18 +53,80 @@ import Foundation
...
+        } else if type == "ApplePayCard" {
+            return BTPaymentMethodNonce(
+                nonce: json?["nonce"].asString() ?? "",
+                type: json?["details"]["cardType"].asString() ?? "ApplePayCard",
+                isDefault: json?["default"].isTrue ?? false
+            )
+        } else if type == "PayPalAccount" {
+            return BTPaymentMethodNonce(
+                nonce: json?["nonce"].asString() ?? "",
+                type: "PayPal",
```
scannillo, line 81: Same comment here, re using `PayPalAccount` in [V5's load method](https://github.com/braintree/braintree_ios/blob/75a48494b48dc83204ced72a0cee68057a8e7170/Sources/BraintreePayPal/BTPayPalDriver.m#L79-L81)

```suggestion
                type: "PayPalAccount",
```

Sources/BraintreeCore/BTPaymentMethodNonceParser.swift
jaxdesmarais: That is just for the register type piece but under the hood the type was passed as `Venmo` - see [this unit test](https://github.com/braintree/braintree_ios/blob/75a48494b48dc83204ced72a0cee68057a8e7170/UnitTests/BraintreeVenmoTests/BTPaymentMethodNonceParser_Venmo_Tests.swift#L20) in v5. I did test this in the demo app by calling this method to verify that the types would be `Venmo` and `PayPal` in v5.

Sources/BraintreeCore/BTPaymentMethodNonceParser.swift
jaxdesmarais: Same as Venmo, under the hood we pass the type as PayPal per [this test](https://github.com/braintree/braintree_ios/blob/75a48494b48dc83204ced72a0cee68057a8e7170/UnitTests/BraintreePayPalTests/BTPaymentMethodNonceParser_PayPal_Tests.swift#L22) and not what's passed into the register method. Verified in the demo app in v5 as well.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -203,15 +203,20 @@ import Foundation
...
     /// Must be using client token with a customer ID specified.
     ///  - Parameter completion: Callback that returns either an array of payment method nonces or an error
+    ///  - Note: Only the top level `BTPaymentMethodNonce` type is returned, fetching any additional details will need to be done on the server
     public func fetchPaymentMethodNonces(_ completion: @escaping ([BTPaymentMethodNonce]?, Error?) -> Void) {
         fetchPaymentMethodNonces(false, completion: completion)
     }
 
+    // NEXT_MAJOR_VERSION: this should move into the Drop-in for parity with Android
+    // This will also allow us to return the types directly which we were doing in the +load method
+    // previously in Obj-C - this is not available in Swift
```
sshropshire, line 213: We've actually gotten multiple requests to move this back into core. We initially moved it to DropIn to get rid of circular dependencies, and we're planning to implement it the correct way in `v5`.

jaxdesmarais: Ah interesting... we may have to rethink what we want to return to merchants if we move it back to core. Not sure if Android is capable of what we were doing in v5. We were using the `load` method to dynamically register types that don't exist in core (`BTCardNonce`/`PayPalNonce`/'VenmoNonce`/`ApplePayNonce`) at compile time. We don't have the ability to do that hacky workaround in Swift and importing those modules into Core would cause a circular dependency. How are y'all planning to correctly implement it in v5?

Sources/BraintreeCore/BTPaymentMethodNonceParser.swift
scannillo: Got it, thanks for confirming Jax!

Sources/BraintreeCore/BTAPIClient.swift
sshropshire: Yeah it was a hack in Android `v3` as well. We used `compileOnly` in our Gradle files to basically create weak references in the dependency tree. That allowed Gradle to build successfully, but it caused some issues in the long run.

By "correct" I'm thinking "in a non-hacky way." It's tough because there will probably be tradeoffs. An enum feels like the right way to me, but I'd be up to do some cross-platform pairing on this since we need a solution on Android too.

scannillo (approved):

PR #1108, opened 2023-10-05
Send PayPal `correlation_id` in analytics payload

scannillo:
### Summary

Currently, we cannot tie our analytics to a PayPal correlationID (aka clientMetadataID) because we are not sending the value in our analytics batch payload. We recently had a Slack inbound where this data would be extremely helpful for handoff to the underlying PP Web or PP NXO team to do further investigation on the specific PayPal transaction.

### Changes
- Update `FPTIBatchData` to send `correlation_id` at the event level (I verified that I can see these values populate in FPTI event explorer)
- Update `BTAnalyticsService` to accept an optional `correlationID` param
- Send already-existant `clientMetadataID` in BTPayPalWebClient
- Collect `clientMetadataID` & send in analytics for BTPayPalNativeCheckoutClient

### Feedback/Questions
- We don't have a testing infrastructure for BTPayPalNativeCheckoutClient set up right now, which I left a TODO for. It will be a handful of work, since MXO is hard to test (ticket DTBTSDK-3076)
- We don't have testing in place for our analytics (ticket DTBTSDK-2441)
- We _could_ block this PR until the above ^ are taken care of, if folks feel strongly? What do you think?

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Update `component` from `btmobilesdk` to `braintreeclientsdk` for PayPal's analytics service (FPTI)
 * BraintreeCore
   * Fix bug where `type` was always returned as `Unknown` in `fetchPaymentMethodNonces` (fixes #1099)
+* Send `correlation_id`, when possible, in PayPal analytic events
 
 ## 6.6.0 (2023-08-22)
 * BraintreePayPalNativeCheckout
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
-* Send `tenant_name` in `event_params` to PayPal's analytics service (FPTI)
-* Update `component` from `btmobilesdk` to `braintreeclientsdk` for PayPal's analytics service (FPTI)
 * BraintreeCore
   * Fix bug where `type` was always returned as `Unknown` in `fetchPaymentMethodNonces` (fixes #1099)
-* Send `correlation_id`, when possible, in PayPal analytic events
+  * Analytics
+    * Send `tenant_name` in `event_params` to PayPal's analytics service (FPTI)
+    * Update `component` from `btmobilesdk` to `braintreeclientsdk` for PayPal's analytics service (FPTI)
+    * Send `correlation_id`, when possible, in PayPal analytic events
 
 ## 6.6.0 (2023-08-22)
 * BraintreePayPalNativeCheckout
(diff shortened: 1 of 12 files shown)


Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
```diff
@@ -131,21 +136,22 @@ import PayPalCheckout
...
+                        self.clientMetadataID = approval.data.correlationIDs.riskCorrelationID
 
                         tokenize(approval: approval, request: request, completion: completion)
                     },
                     onCancel: {
                         self.notifyCancel(completion: completion)
                     },
                     onError: { error in
+                        self.clientMetadataID = error.correlationIDs.riskCorrelationID
                         self.notifyFailure(with: BTPayPalNativeCheckoutError.checkoutSDKFailed, completion: completion)
```
jaxdesmarais, line 155: Since we get an `error` in the completion here, should we add it to the `checkoutSDKFailed` error description? It could be helpful for the MXO team to know what error their SDK returned in this callback.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Update `component` from `btmobilesdk` to `braintreeclientsdk` for PayPal's analytics service (FPTI)
 * BraintreeCore
   * Fix bug where `type` was always returned as `Unknown` in `fetchPaymentMethodNonces` (fixes #1099)
+* Send `correlation_id`, when possible, in PayPal analytic events
```
jaxdesmarais, line 8: Do we want to group all of the FPTI entries together? I had put the other ones above since it applied to all modules not just core - but I don't feel super strongly where they live. Maybe a `Analytics Changes` header or something similar makes sense?

jaxdesmarais:
> We could block this PR until the above ^ are taken care of, if folks feel strongly? What do you think?

I personally don't think we need to wait on these. I feel like MXO being easily testable is a long ways off. With the analytics ticket maybe we can prioritize that in the next few sprints? I feel like that one may be a bit easier to tackle sooner but still not a blocker.

scannillo:
> I personally don't think we need to wait on these. I feel like MXO being easily testable is a long ways off. With the analytics ticket maybe we can prioritize that in the next few sprints? I feel like that one may be a bit easier to tackle sooner but still not a blocker.

Cool, yeh I'm down for not blocking this!

The testing story on MXO won't change, realistically. So we'll have to do something like PPCP with a wrapper protocol. It's possible, but a pain :/. We could also explore if OCMock makes things easier too (I captured this in the ticket).

CHANGELOG.md
scannillo: Done!

Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutClient.swift
scannillo: Great idea, done!

scannillo:
UI Test failures are unrelated :/

sarahkoop (approved):

PR #1111, opened 2023-10-06
[DO NOT REVIEW] Update GH Runner to Apple Silicon

scannillo:
WIP - PR in draft to test CI

### Summary of changes

- 

### Checklist

- [ ] Added a changelog entry

-

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,23 +5,23 @@ concurrency:
   cancel-in-progress: true
 jobs:
   cocoapods:
-    name: CocoaPods (Xcode 14.3)
-    runs-on: macOS-13
+    name: CocoaPods (Xcode 15.0)
+    runs-on: macOS-13-arm64
     steps:
       - name: Check out repository
         uses: actions/checkout@v3
         with:
           ref: ${{ github.event.pull_request.head.ref }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
-      - name: Use Xcode 14.3
-        run: sudo xcode-select -switch /Applications/Xcode_14.3.app
+      - name: Use Xcode 15.0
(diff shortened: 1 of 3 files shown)


scannillo:
Sad - using `macOS-13-arm64` essentially always keeps our jobs in the `Queued` state

scannillo:
@jaxdesmarais I updated the description here, curious if you have thoughts on if we should merge this at all?

jaxdesmarais:
My vote would be to make this change. Not sure how long the beta is but would hopefully give us a few months of data to point to for budget. Though it may be more sad if the budget doesn't get approved but hey at least we'd get some bliss of fast runners! Is that what you are thinking as well @scannillo?

I do wonder how we know it went out of beta. Do they just stop working until we pay? Do we get a bill that we have to pay either way for the time we didn't know it was GA?

scannillo:
> My vote would be to make this change. Not sure how long the beta is but would hopefully give us a few months of data to point to for budget. Though it may be more sad if the budget doesn't get approved but hey at least we'd get some bliss of fast runners! Is that what you are thinking as well @scannillo?
> I do wonder how we know it went out of beta. Do they just stop working until we pay? Do we get a bill that we have to pay either way for the time we didn't know it was GA?

Yeh I would also like to just try it out. One thing we should be aware of is getting timing data on our old PRs now, since they don't last in the GH Actions history forver.

In terms of what happens when it's out of beta, I assume something like [when I tried](https://github.com/braintree/braintree_ios/pull/1111#issuecomment-1750438211) `macOS-13-arm64`. Also, there's this new section of the `Actions` tab which I imagine would update to require some billing details.

Screenshot 2023-10-09 at 12 53 54 PM

jaxdesmarais (approved):
🚤

PR #1112, opened 2023-10-06
Convert Demo AppDelegate & SceneDelegate to Swift

scannillo:
### Summary of changes

- Convert AppDelegate & SceneDelegate to Swift

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


Demo/Application/Base/Settings/Settings.bundle/Root.plist
```diff
@@ -103,7 +103,6 @@
 			<array>
 				<integer>0</integer>
 				<integer>1</integer>
-				<integer>2</integer>
```
scannillo, line 106: This was left over from when we had a 3rd auth type, PayPalIDToken. There should only be 2 entries here (for client token & tokenization key)

Demo/Application/Base/AppDelegate.swift
```diff
@@ -0,0 +1,82 @@
...
+            var defaultsToRegister: [String: Any] = [:]
+            preferences.forEach { prefSpecification in
+                print(prefSpecification)
+                if let key = prefSpecification["Key"] as? String, prefSpecification.keys.contains("DefaultValue") {
+                    defaultsToRegister[key] = prefSpecification["DefaultValue"]
+                }
+            }
+            
+            UserDefaults.standard.register(defaults: defaultsToRegister)
+        }
```
scannillo, line 68: I think we can delete these lines. I'm not sure really what they do

Demo/Application/Base/AppDelegate.swift
```diff
@@ -0,0 +1,81 @@
+import Foundation
+
+@objcMembers class AppDelegate: UIResponder, UIApplicationDelegate {
```
jaxdesmarais, line 3: Do we need `@objcMembers` here?

Demo/Application/Base/AppDelegate.swift
```diff
@@ -0,0 +1,81 @@
...
+        
+    let BraintreeDemoAppDelegatePaymentsURLScheme = "com.braintreepayments.Demo.payments"
```
jaxdesmarais, line 5:
```suggestion
    let returnURLScheme = "com.braintreepayments.Demo.payments"
```

Demo/Application/Base/AppDelegate.swift
```diff
@@ -0,0 +1,81 @@
...
+                UserDefaults.standard.setValue(testIntegration, forKey: "BraintreeDemoSettingsAuthorizationOverride")
+            }
+        }
+        
+        if ProcessInfo.processInfo.arguments.contains("-ClientTokenVersion2") {
+            UserDefaults.standard.set("2", forKey: "BraintreeDemoSettingsClientTokenVersionDefaultsKey")
+        } else if ProcessInfo.processInfo.arguments.contains("-ClientTokenVersion3") {
+            UserDefaults.standard.set("3", forKey: "BraintreeDemoSettingsClientTokenVersionDefaultsKey")
+        }
+        // End checking for testing arguments
```
jaxdesmarais, line 49: Could we extract setting up the testing arguments into it's own method? We can also probably extract some of these into constants:
```swift
let arguments = ProcessInfo.processInfo.arguments
let userDefaults = UserDefaults.standard
```

Demo/Application/Base/AppDelegate.swift
jaxdesmarais: My assumption was it saves things like selecting client token or the demo type between sessions to user defaults but that may be totally incorrect

Demo/Application/Base/AppDelegate.swift
```diff
@@ -0,0 +1,81 @@
...
+    // MARK: - UISceneSession lifecycle
+    
+    func application(
+        _ application: UIApplication,
+        configurationForConnecting connectingSceneSession: UISceneSession,
+        options: UIScene.ConnectionOptions
+    ) -> UISceneConfiguration {
+        // Called when a new scene session is being created.
+        // Use this method to select a configuration to create the new scene with.
+        return UISceneConfiguration(name: "Default Configuration", sessionRole: connectingSceneSession.role)
```
jaxdesmarais, line 79:
```suggestion
        UISceneConfiguration(name: "Default Configuration", sessionRole: connectingSceneSession.role)
```

Demo/Application/Base/SceneDelegate.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+
+@objcMembers class SceneDelegate: UIResponder, UIWindowSceneDelegate {
```
jaxdesmarais, line 3: Same question here on if we need `@objcMembers`

Demo/Application/Base/AppDelegate.swift
scannillo: We need the class name to be available to ObjC to reference [here](https://github.com/braintree/braintree_ios/blob/c393c47b99c23357cd4250a4fd81a5a9444b6996/Demo/Application/Supporting%20Files/main.m#L8). Though, we can just use `@objc` instead of `@objcMembers`

Demo/Application/Base/AppDelegate.swift
scannillo: Addressed in dc311d2bb8860d88472b04f886b40b4358ac1c75

Demo/Application/Base/SceneDelegate.swift
scannillo: Not needed! Addressed in dc311d2bb8860d88472b04f886b40b4358ac1c75

Demo/Application/Base/AppDelegate.swift
scannillo: Addressed in dc311d2bb8860d88472b04f886b40b4358ac1c75

Demo/Application/Base/AppDelegate.swift
jaxdesmarais: Yeah it's a bit wonky for sure. I wonder if we get some of this for free if we move away from `InAppSettingsKit`

Demo/Application/Supporting Files/main.m
```diff
@@ -1,10 +1,10 @@
 #import <UIKit/UIKit.h>
 
-#import "BraintreeDemoAppDelegate.h"
+#import "Demo-Swift.h"
 
 int main(int argc, char * argv[])
 {
     @autoreleasepool {
-        return UIApplicationMain(argc, argv, nil, NSStringFromClass([BraintreeDemoAppDelegate class]));
+        return UIApplicationMain(argc, argv, nil, NSStringFromClass([AppDelegate class]));
```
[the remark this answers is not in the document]
scannillo: Great suggestion, applied!

jaxdesmarais (approved):
🚀

sarahkoop (approved):

PR #1114, opened 2023-10-10
[DO NOT REVIEW] Remove xcbeautify Pod

scannillo:
### Summary of changes

- Use `xcpretty` [already installed in GH Actions macOS image ](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#xcode-support-tools)versus pulling in `xcbeautify` pod. 
- (These libraries are used to format xcodebuild test output)

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Install CocoaPod dependencies
         run: pod install
       - name: Run Unit Tests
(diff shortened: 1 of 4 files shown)


scannillo:
xcpretty does not support parallel test execution :(

scannillo:
We could move to installing `xcbeautify` via homebrew in our CI commands only (versus having it locally with our Pods), but I don't see the value in it at this time since we still use Pods for other test dependencies.

PR #1119, opened 2023-10-25
[5.x] Magnes 5.4.1

jaxdesmarais:
### Summary of changes

- Update 5.x branch with Magnes 5.4.1 security fix (stage URLs removed)
- This has been tested on Xcode 14.1 to verify backwards compatibility - since Magnes does not use any Swift 5.8 specific features we are covered by Swift's library evolution to be compatible with lower Swift/Xcode versions.

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais


jaxdesmarais:
We will need to get a new version of PPRiskMagnes built with Xcode 14.1 + Swift 5.7. I've reached out to them internally to request this. Once we hear back we will need to redeploy assets with the legacy build for Carthage merchants as well as update this PR with the XCFramework that was built with Xcode 14.1.

jaxdesmarais:
> Pending your xcode downgrade confirmation change LGTM

Already verified this morning (forgot to update the comment and just updated the PR description) so we should be g2g. Everything compiles fine on Xcode 14.1 with Swift 5.7. 🚀

scannillo (approved):

PR #1123, opened 2023-10-31
Update V6_MIGRATION Guide

jaxdesmarais:
### Summary of changes

- Add `BTAppContextSwitcher` method name changes to Venmo section

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/V6_MIGRATION.md
+++ b/V6_MIGRATION.md
@@ -42,6 +42,11 @@ The possible values for `BTVenmoPaymentMethodUsage` include:
 
 `BTVenmoClient.isiOSAppAvailableForAppSwitch()` has been renamed to `BTVenmoClient.isVenmoAppInstalled()`
 
+The following `BTAppContextSwitcher` methods have been renamed:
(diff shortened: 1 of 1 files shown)


jaxdesmarais:
@scannillo looking through the slack thread in the ticket it mentions the Local Payment snippets being wrong. I think this is expected as that section mentioned the old methods no longer being needed. Here is the snippet in question:
```
We have renamed the module `BraintreePaymentFlow` to `BraintreeLocalPayment`
We have replaced `SFAuthenticationSession` with `ASWebAuthenticationSession` in the Local Payment Method flow. With this change, you no longer need to:
  * Register a URL Scheme or set a return URL via the `BTAppContextSwitcher.setReturnURLScheme()` method
  * Handle app context switching via the `BTAppContextSwitcher.handleOpenURL(context: UIOpenURLContext)` or `BTAppContextSwitcher.handleOpenURL(URL)`
```
I oped to add the code changes to the Venmo section instead but let me know if you think something in the LPM section should also be updated. Imo to seems odd to replace the v5 code with v6 code snippets only to say it no longer is needed for LPMs.

sshropshire (approved):

scannillo (approved):
Thanks for tackling this one!

PR #1129, opened 2023-11-02
Deprecate `BTCardRequest`

scannillo:
### Summary of changes

- I noticed `BTCardRequest` is unused
- Added deprecation warning & NEXT_MAJOR_VERSION note

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo


jaxdesmarais (approved):

sarahkoop (approved):
This is going to deviate from Android which uses `CardRequest` but I think that's ok

scannillo:
> @sarahkoop This is going to deviate from Android which uses CardRequest but I think that's ok

It looks like Android also uses just `Card` type as the input param to [tokenize()](https://github.com/braintree/braintree_android/blob/96e7a0c3e65910a9a8fa957365b924d9a0a9bf9f/Card/src/main/java/com/braintreepayments/api/CardClient.java#L49). I also don't see a `CardRequest` type when I search in Android either.

I was surprised by this honestly, I thought we aligned all the `tokenize()` methods in iOS V6 to take in `<PaymentMethod>Request` types.

sarahkoop:
> > @sarahkoop This is going to deviate from Android which uses CardRequest but I think that's ok
> 
> It looks like Android also uses just `Card` type as the input param to [tokenize()](https://github.com/braintree/braintree_android/blob/96e7a0c3e65910a9a8fa957365b924d9a0a9bf9f/Card/src/main/java/com/braintreepayments/api/CardClient.java#L49). I also don't see a `CardRequest` type when I search in Android either.
> 
> I was surprised by this honestly, I thought we aligned all the `tokenize()` methods in iOS V6 to take in `<PaymentMethod>Request` types.

Ah nevermind - I was mistaken you're right!

PR #1130, opened 2023-11-02
Deprecate `cardAddChallenge` & add internal property

scannillo:
Follow up to PR #1122 & [this comment](https://github.com/braintree/braintree_ios/pull/1122#discussion_r1376494256)

### Summary of changes

- Deprecate `BTThreeDSecure.cardAddChallenge`
- Create internal variable for `cardAddChallenge` to use within the SDK
    - Since we cannot silence deprecation warnings in Swift, Jax's PR above revealed that referencing deprecated methods & properties internally will surface errors to merchants
    - This works around that limitation by only accessing the _internal_ variable for `cardAddChallenge` throughout the SDK, while still deprecating the merchant-facing public property
    - Since the internal property is not deprecated, our SDK will not trigger warnings

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
@@ -58,8 +58,14 @@ import BraintreeCore
     /// If set to `.notRequested` the authentication challenge will not be requested from the issuer.
     /// If set to `.unspecified`, when the amount is 0, the authentication challenge will be requested from the issuer.
     /// If set to `.unspecified`, when the amount is greater than 0, the authentication challenge will not be requested from the issuer.
-    public var cardAddChallenge: BTThreeDSecureCardAddChallenge = .unspecified
+    @available(*, deprecated, message: "Use `cardAddChallengeRequested` instead")
+    public var cardAddChallenge: BTThreeDSecureCardAddChallenge {
+        get { return _cardAddChallenge }
+        set { _cardAddChallenge = newValue }
+    }
 
+    var _cardAddChallenge: BTThreeDSecureCardAddChallenge = .unspecified
+    
     /// Optional.  An authentication created using this flag should only be used for vaulting operations (creation of customers' credit cards or payment methods) and not for creating transactions.
     /// If set to `true`, a card-add challenge will be requested from the issuer.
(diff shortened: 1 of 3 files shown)


Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -58,8 +58,16 @@ import BraintreeCore
     /// If set to `.notRequested` the authentication challenge will not be requested from the issuer.
     /// If set to `.unspecified`, when the amount is 0, the authentication challenge will be requested from the issuer.
     /// If set to `.unspecified`, when the amount is greater than 0, the authentication challenge will not be requested from the issuer.
-    public var cardAddChallenge: BTThreeDSecureCardAddChallenge = .unspecified
-
+    @available(*, deprecated, message: "Use `cardAddChallengeRequested` instead")
```
jaxdesmarais, line 61: I think we can use:
```swift
@available(*, deprecated, renamed: "cardAddChallengeRequested" message: "whatever we want for the message here")
```
I believe this way lets you use `Fix` in Xcode.

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -58,8 +58,16 @@ import BraintreeCore
...
+    public var cardAddChallenge: BTThreeDSecureCardAddChallenge {
+        get { return _cardAddChallenge }
```
jaxdesmarais, line 63:
```suggestion
        get { _cardAddChallenge }
```

jaxdesmarais (approved):

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
scannillo: Ooo that's awesome, fixed!

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
scannillo: Fixed!

sarahkoop (approved):

PR #1132, opened 2023-11-08
Fix Bug where some `BTThreeDSecureRequest` Default Values Returned Errors Unexpectedly

jaxdesmarais:
### Summary of changes

- Fix bug where defaults for `BTThreeDSecureRequest.accountType`, `BTThreeDSecureRequest.requestedExemptionType`, and `BTThreeDSecureRequest.dfReferenceID` were improperly returning an error if not passed into the request
- The `Any` in these cases for the dictionary should be optional to allow for nil values (behavior in v5) - currently we were defaulting to an empty string in the `nil` case which is invalid
- Add UnitTest for this case in the future

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
@@ -380,7 +380,7 @@ import BraintreeCore
 
             self.apiClient.post(
                 "v1/payment_methods/\(urlSafeNonce)/three_d_secure/lookup",
-                parameters: requestParameters
+                parameters: requestParameters as [String: Any] 
             ) { body, _, error in
                 if let error = error as NSError? {
                     // Provide more context for card validation error when status code 422


Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -339,15 +339,15 @@ import BraintreeCore
...
 
-            var requestParameters: [String: Any] = [
-                "amount": request.amount ?? 0,
+            var requestParameters: [String: Any?] = [
+                "amount": request.amount,
                 "customer": customer,
                 "requestedThreeDSecureVersion": "2",
-                "dfReferenceId": request.dfReferenceID ?? "",
-                "accountType": request.accountType.stringValue ?? "",
+                "dfReferenceId": request.dfReferenceID,
```
sarahkoop, line 346: Curious why we want to pass nil rather than just not including the param in the request if it's null? Is the behavior the same either way?

jaxdesmarais: Yeah the behavior would be the same - we could do either technically, seems like v5 did a `nil` check then passed the key/value if not nil ([code here](https://github.com/braintree/braintree_ios/blob/45ec8f92a1803c91d36a55fc381ba41dcb9538c0/Sources/BraintreeThreeDSecure/BTPaymentFlowDriver%2BThreeDSecure.m#L57-L59)). I'm fine with either option and this is _slightly_ cleaner than a bunch of `if/else` below but if folks felt strongly we could extract these out.

It's also worth noting when we do `compactMapValues` below we strip out any nil key/values so either option has the same effect. From a playground:
```swift
let myDictionary: [String: Any?] = [
    "string1": "hasValue",
    "string2": nil,
    "int1": 0
]

print(myDictionary)
print(myDictionary.compactMapValues { $0 })
```

The above would return with `compactMapValues` stripping out nil values:
```swift
["string2": nil, "string1": Optional("hasValue"), "int1": Optional(0)] // print(myDictionary)
["int1": 0, "string1": "hasValue"] // print(myDictionary.compactMapValues { $0 })
```

scannillo (approved):

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
sarahkoop: Ah cool thanks for the explanation!

sarahkoop (approved):

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
jaxdesmarais: No problem! `compactMapValues` is pretty cool and was definitely a benefit of the Swift move. 🥳

PR #1134, opened 2023-11-13
Add Cardinal UI Type and Render Types

jaxdesmarais:
### Summary of changes

- Add `uiType` to `BTThreeDSecureRequest`
- Add `renderTypes` to `BTThreeDSecureRequest`
- Set default parameters in Demo app ([from Cardinal docs](https://cardinaldocs.atlassian.net/wiki/spaces/CMSDK/pages/2005696619/Setting+up+CardinalMobileSDK+-+iOS#Available-Configurations--%5BinlineExtension%5D))
- Add unit tests
- Android PR for this change: https://github.com/braintree/braintree_android/pull/738 - we finally heard back from the merchants that these worked as expected, so exposing them on iOS

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureRenderTypes.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureRenderTypes.swift
@@ -1,7 +1,7 @@
 import Foundation
 
 /// Render types that the device supports for displaying specific challenge user interfaces within the 3D Secure challenge.
-@objcMembers public class BTThreeDSecureRenderTypes: NSObject {
+@objcMembers public class BTThreeDSecureRenderType: NSObject {
 
     public typealias StringValue = String
 
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureRenderTypes.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureRenderTypes.swift
@@ -1,0 +1,22 @@
+import Foundation
+
+/// Render types that the device supports for displaying specific challenge user interfaces within the 3D Secure challenge.
+@objcMembers public class BTThreeDSecureRenderTypes: NSObject {
+
+    public typealias StringValue = String
+
+    /// OTP
+    public static let otp: StringValue = "CardinalSessionRenderTypeOTP"
+
+    /// HTML
+    public static let html: StringValue = "CardinalSessionRenderTypeHTML"
+
+    /// Single select
+    public static let singleSelect: StringValue = "CardinalSessionRenderTypeSingleSelect"
+
+    /// Multi Select
+    public static let multiSelect: StringValue = "CardinalSessionRenderTypeMultiSelect"
+
+    /// OOB
+    public static let oob: StringValue = "CardinalSessionRenderTypeOOB"
+}
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
@@ -86,7 +86,7 @@ import BraintreeCore
     ///
     /// - Note: When using `BTThreeDSecureUIType.both` or `BTThreeDSecureUIType.html`, all `BTThreeDSecureRenderType` options must be set.
     /// When using `BTThreeDSecureUIType.native`, all `BTThreeDSecureRenderType` options except `BTThreeDSecureRenderType.html` must be set.
-    public var renderTypes: [BTThreeDSecureRenderTypes.StringValue]?
+    public var renderTypes: [BTThreeDSecureRenderType.StringValue]?
 
     /// A delegate for receiving information about the ThreeDSecure payment flow.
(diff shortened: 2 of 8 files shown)


Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -77,6 +77,17 @@ import BraintreeCore
...
+    /// Optional. Sets all UI types that the device supports for displaying specific challenge user interfaces in the 3D Secure challenge.
+    ///
+    /// Defaults to `.both`
+    public var uiType: BTThreeDSecureUIType = .both
+
+    /// Optional. List of all the render types that the device supports for displaying specific challenge user interfaces within the 3D Secure challenge.
+    ///
+    /// - Note: When using `BTThreeDSecureUIType.both` or `BTThreeDSecureUIType.html`, all `BTThreeDSecureRenderType` options must be set.
+    /// When using `BTThreeDSecureUIType.native`, all `BTThreeDSecureRenderType` options except `BTThreeDSecureRenderType.html` must be set.
+    public var renderTypes: [BTThreeDSecureRenderType.StringValue]?
```
jaxdesmarais, line 89: Arrays of `Int` enums do not exist in Obj-C so we cannot use the cleaner version (array of enums) here. 😢 This seems to be the consensus on the best practice in these cases. Open to other ideas if folks find something that works better as this is certainly not the most Swift-y approach. But working in the confines of Obj-C it seems to be our best bet.

sarahkoop: I don't have any better ideas for this but curious if we are planning to remove obj-c support in the next major version?

sarahkoop (approved):

sshropshire (approved):

Sources/BraintreeThreeDSecure/BTThreeDSecureRenderTypes.swift
```diff
@@ -0,0 +1,22 @@
...
+    public static let html: StringValue = "CardinalSessionRenderTypeHTML"
+
+    /// Single select
+    public static let singleSelect: StringValue = "CardinalSessionRenderTypeSingleSelect"
+
+    /// Multi Select
+    public static let multiSelect: StringValue = "CardinalSessionRenderTypeMultiSelect"
+
+    /// OOB
+    public static let oob: StringValue = "CardinalSessionRenderTypeOOB"
```
sshropshire, line 21: Would the `@objc` annotation work for us ([stack post](https://stackoverflow.com/a/28436181))?

jaxdesmarais: That works for the enum but that's not the issue, the issue is that arrays of `Int` enums don't exist in Obj-C and enums can only be bridged to `Int` types to be interoperable with Obj-C. A Swift array is bridged to Obj-C's `NSArray`, and an Obj-C array should contain only pointers, but an enum of `Int` doesn't contain pointers so cannot be bridged.

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
jaxdesmarais: Currently there isn't a plan to remove Obj-C support in the next major version but we could consider discussing it further as an option! Or if Apple announces an official deprecation policy at some point it'd be a much easier choice.

Sources/BraintreeThreeDSecure/BTThreeDSecureRenderTypes.swift
sshropshire: Ah got it yeah. That makes sense. Swift does have `OptionSet` as an `NS_OPTIONS` equivalent, but I can't tell if option sets are accessible from Obj-C.

jaxdesmarais: 🤔 we may be able to make that work - it'll need to be a class instead of a struct since structs don't exist in Obj-c, but I think that may be okay. I'll poke around at it a bit.

jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/1134/commits/5911c551bbc829b3002f5f98941f1a248c97f988. Great suggestion, that's way cleaner 🔥

jaxdesmarais: Also if anyone is interested this is how the implementation will look for Obj-C merchants:
```objective-c
BTThreeDSecureRequest *request = [[BTThreeDSecureRequest alloc] init];
request.renderTypes = @[BTThreeDSecureRenderType.oob, BTThreeDSecureRenderType.html, BTThreeDSecureRenderType.singleSelect, BTThreeDSecureRenderType.multiSelect, BTThreeDSecureRenderType.otp];
```

PR #1141, opened 2023-12-02
Remove `BTAPIClientHTTPType.braintreeAPI` enum case

scannillo:
### Summary of changes

- There were no API requests in the SDK using the `BTAPIClientHTTPType.braintreeAPI` enum case
- I double checked that DropIn isn't using this enum case, as well

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


jaxdesmarais (approved):
🚀  can we remove the entire `BTAPIHTTP` file as well? Seems like this was logic left over from when we wanted to be able to use access tokens as an auth type aka PPCP version 2? 3? 🙈

PR #932, opened 2023-01-19
Remove Preferred Payments

mattwylder:
### Summary of changes

- Removed `BTPreferredPaymentMethods` and `BTPreferredPaymentMethodResult` as well as accompanying tests, and references in the Demo app.

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais @mattwylder

PR #1142, opened 2023-12-04
Add `BTShopperInsights` public API shell

scannillo:
**JIRA DTBTSDK-3173** 

### Summary of changes

- Add public API shell for the Payment Insights feature
    - `BTShopperInsightsClient`
    - `BTShopperInsightsResult`
- Add template unit test file for ` BTShopperInsightsClient_Tests`

### Note

This feature is basically a V2/re-do of PreferredPaymentMethods (see [removal PR](https://github.com/braintree/braintree_ios/pull/932)). Though this feature will leverage a different underlying API to perform the eligibility checks.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo @MaxHastingsPP @saperi22


Sources/BraintreeCore/BTPaymentInsightsClient.swift
```diff
@@ -0,0 +1,28 @@
...
+        self.apiClient = apiClient
+    }
+    
+    /// This method confirmes if the customer is a user of PayPal services using their email and phone number.
+    /// - Parameters:
+    ///   - email: The buyer's email address
+    ///   - phone: The buyer's phone number
+    /// - Returns: A `BTPaymentInsightsResult` instance
+    /// - Note: This feature is in beta. It's public API may change in future releases.
+    public func getRecommendedPaymentMethods(email: String, phone: String) async throws -> BTPaymentInsightsResult {
```
scannillo, line 23: Apparently we need to collect the phone's "countryCode" & "nationalNumber" separately. We should probably move to a `BTPaymentInsightsRequest` object as the input param instead

cc: @MaxHastingsPP @saperi22

scannillo: Done!

Sources/BraintreeCore/BTShopperInsightsClient.swift
```diff
@@ -0,0 +1,28 @@
...
+        self.apiClient = apiClient
+    }
+    
+    /// This method confirms if the customer is a user of PayPal services using their email and phone number.
+    /// - Parameters:
+    ///   - email: The buyer's email address
+    ///   - phone: The buyer's phone number
+    /// - Returns: A `BTShopperInsightsResult` instance
+    /// - Note: This feature is in beta. It's public API may change in future releases.
+    public func getRecommendedPaymentMethods(request: BTShopperInsightsRequest) async throws -> BTShopperInsightsResult {
```
jaxdesmarais, line 23: Do we want a completion method for this as well similar to the other clients or just this one?

scannillo: I intentionally left it out to use this as a case for gauging merchant feedback on if we need completion handler syntax. Figured let's start with less & add based on need?

Also this feature is going out as a beta first so we can (and will) iterate. What do you think?

Sources/BraintreeCore/BTShopperInsightsRequest.swift
```diff
@@ -0,0 +1,26 @@
...
+@objcMembers public class BTShopperInsightsRequest: NSObject {
+    
+    /// The buyer's email address.
+    public let email: String
+    
+    /// The buyer's country code prefix to the national telephone number. An identifier for a specific country.
+    public let phoneCountryCode: String
+    
+    /// The buyer's national phone number.
+    public let phoneNationalNumber: String
```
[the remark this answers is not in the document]
scannillo: Good question - since we can't test the actual API yet, I'm not sure how it will treat special characters. It _should_ be able to handle the case with characters + digits fine, as well as only digits. The examples they've given us show only as digits.
 
We can document it as only numbers now to be safe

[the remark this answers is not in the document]
scannillo: Done in 5973db06a76bc98e8dd013a4630ba5a58c3bbd26!

Sources/BraintreeCore/BTShopperInsightsRequest.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+/// Buyer data required to use the Shopper Insights feature.
+/// - Note: This feature is in beta. It's public API may change in future releases.
+@objcMembers public class BTShopperInsightsRequest: NSObject {
+    
+    /// The buyer's email address.
+    public let email: String
+    
+    /// The buyer's country code prefix to the national telephone number. An identifier for a specific country.
```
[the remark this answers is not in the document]
scannillo: Done in 5973db06a76bc98e8dd013a4630ba5a58c3bbd26!

Sources/BraintreeCore/BTShopperInsightsClient.swift
jaxdesmarais: I think that makes sense - will be a good way to gauge things for potentially removing completion syntax in v7. Let's do it!

Sources/BraintreeCore/BTShopperInsightsRequest.swift
```diff
@@ -0,0 +1,28 @@
...
+    public let phoneCountryCode: String
+    
+    /// The buyer's national phone number. Must not contain special characters.
+    public let phoneNationalNumber: String
+    
+    /// Initialize a `BTShopperInsightsRequest`
+    /// - Parameters:
+    ///   - email: The buyer's email address.
+    ///   - phoneCountryCode: The buyer's country code prefix to the national telephone number. An identifier for a specific country.
+    ///   - phoneNationalNumber: The buyer's national phone number.
```
jaxdesmarais, line 21: nit to add the same docstring here:
```suggestion
    ///   - phoneCountryCode: The buyer's country code prefix to the national telephone number. An identifier for a specific country. Must not contain special characters.
    ///   - phoneNationalNumber: The buyer's national phone number. Must not contain special characters.
```

Sources/BraintreeCore/BTShopperInsightsRequest.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+/// Buyer data required to use the Shopper Insights feature.
+/// - Note: This feature is in beta. It's public API may change or be removed in future releases.
+@objcMembers public class BTShopperInsightsRequest: NSObject {
```
jaxdesmarais, line 5: General q based on [this conversation](https://github.com/braintree/braintree_ios/pull/1142/files#r1414559278). I know we've also talked about moving away from the dot syntax and using the init only in v7. Do we have any interest in doing a similar trial of that here and making these properties internal and only the init public? The init is currently internal.

jaxdesmarais (approved):

UnitTests/BraintreeCoreTests/BTShopperInsightsClient_Tests.swift
```diff
@@ -0,0 +1,27 @@
...
+    func testGetRecommendedPaymentMethods_returnsDefaultRecommendations() async {
+        let request = BTShopperInsightsRequest(
+            email: "fake-email",
+            phoneCountryCode: "fake-country-code",
+            phoneNationalNumber: "fake-national-phone"
+        )
+        let result = try? await sut.getRecommendedPaymentMethods(request: request)
+        
+        XCTAssertNotNil(result!.isPayPalRecommended)
+        XCTAssertNotNil(result!.isPayPalRecommended)
```
MaxHastingsPP, line 25: Should be `isVenmoRecommended` ?

Sources/BraintreeCore/BTShopperInsightsRequest.swift
```diff
@@ -0,0 +1,28 @@
...
+    /// Initialize a `BTShopperInsightsRequest`
+    /// - Parameters:
+    ///   - email: The buyer's email address.
+    ///   - phoneCountryCode: The buyer's country code prefix to the national telephone number. An identifier for a specific country.
+    ///   - phoneNationalNumber: The buyer's national phone number.
+    /// - Note: This feature is in beta. It's public API may change or be removed in future releases.
+    init(email: String, phoneCountryCode: String, phoneNationalNumber: String) {
+        self.email = email
+        self.phoneCountryCode = phoneCountryCode
+        self.phoneNationalNumber = phoneNationalNumber
```
sshropshire, line 26: random: Are all fields required? Or does the insight become more accurate when more data is provided.

scannillo: Yeh, they're all required at the moment

Sources/BraintreeCore/BTShopperInsightsClient.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+
+///  Use `BTShopperInsightsClient` to optimize your checkout experience by prioritizing the customer’s preferred payment methods in your UI.
+///  By customizing each customer’s checkout experience, you can improve conversion, increase sales/repeat buys and boost user retention/loyalty.
+/// - Note: This feature is in beta. It's public API may change or be removed in future releases.
+@objcMembers class BTShopperInsightsClient: NSObject {
```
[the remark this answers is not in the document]
scannillo: These are great questions - I honestly am very down to not add Objective-C support to this feature. We could always add it later if we see demand for it. Also worth noting - this feature will go out as an "alpha" so we have the room to iterate on it, break the API, etc.

Done in 0884b7835516edddce143676ef302679ce747831. Will loop @jaxdesmarais in on this one though to confirm.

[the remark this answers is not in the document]
scannillo: > Actually I see now that this feature lives in core - what is the reason for not having a separate module for this?

We could make this a separate module. I was approaching this from a path of least resistance approach for now as we target "alpha". We might save us the headache from creating a documenting an entire new module, if the feature doesn't get adopted. And should be relatively simple to move the files into a module if we do want to go that route later on. Open to folks thoughts, though!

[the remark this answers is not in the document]
scannillo: Let's do it!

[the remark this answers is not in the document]
jaxdesmarais: Should we also drop the BT prefix? Or do we want to leave that consistent until v7?

Sources/BraintreeCore/BTShopperInsightsResult.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+/// A summary of the buyer's recommended payment methods.
+/// - Note: This feature is in beta. It's public API may change or be removed in future releases.
+public struct BTShopperInsightsResult {
+    
+    /// If true, display the PayPal button with high priority.
+    public var isPayPalRecommended = false
+    
+    /// If true dislpay the Venmo button with high priority.
```
sarahkoop, line 10:
```suggestion
    /// If true, dislpay the Venmo button with high priority.
```

Sources/BraintreeCore/BTShopperInsightsClient.swift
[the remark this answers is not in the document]
scannillo: Another great question - maybe let's leave the prefix until we're ready to do that overhaul holistically?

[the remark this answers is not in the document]
jaxdesmarais: That makes sense to me! I agree it could be confusing to have some things prefixed and some not so the consistency would be great then we can say goodbye to it in v7. 🚀

scannillo:
UI test failures are un-related & failing flakily on main

PR #1145, opened 2023-12-05
Move Analytics to Caseless Enum

jaxdesmarais:
### Summary of changes

When we initially added the analytics classes we still had some modules in Obj-C. Caseless string enums are not a thing in Obj-C so we opted to use classes for the time being. Now that the SDK is fully Swift we can use caseless enums for these constants. These are safer because they cannot be constructed and do not contain initializers. This also tends to be preferred for storing constant values (Apple has also established this pattern for namespacing in newer frameworks such as Combine).

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeAmericanExpress/BTAmericanExpressAnalytics.swift
+++ b/Sources/BraintreeAmericanExpress/BTAmericanExpressAnalytics.swift
@@ -1,7 +1,7 @@
 import Foundation
 
-class BTAmericanExpressAnalytics {
-    
+enum BTAmericanExpressAnalytics {
+
     // MARK: - Get Rewards Balance Events
     
     static let started = "amex:rewards-balance:started"
--- a/Sources/BraintreeApplePay/BTApplePayAnalytics.swift
+++ b/Sources/BraintreeApplePay/BTApplePayAnalytics.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-class BTApplePayAnalytics {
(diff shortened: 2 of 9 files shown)


Sources/BraintreeAmericanExpress/BTAmericanExpressAnalytics.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
-class BTAmericanExpressAnalytics {
-    
+enum BTAmericanExpressAnalytics {
+
     // MARK: - Get Rewards Balance Events
     
     static let started = "amex:rewards-balance:started"
```
sarahkoop, line 7: Is there a preference for `static let` vs `case` since these are enums now?

jaxdesmarais: Yeah the preference for constants is a static let in an enum, if we used case at the callsite it gets a bit more verbose as well and a bit more unclear that we are using a constant. Here is an example of how we'd call the caseless enum with a constant vs a standard enum:
```swift
enum MyEnum {
    static let name = "Name"
}

enum MyEnumTwo: String {
    case name = "Name"
}

print(MyEnum.name)
print(MyEnumTwo.name.rawValue)
```

sarahkoop (approved):

sshropshire (approved):

PR #1147, opened 2023-12-06
Create PayPal Messaging Module

jaxdesmarais:
### Summary of changes

- Create `BraintreePayPalMessaging` module shell
    - `BraintreePayPalMessaging` directory with empty file
    - `BraintreePayPalMessagingTests` with empty file and `Info/plist`
- Update `release.yml` workflow to publish reference docs once merged
- Add scheme for `BraintreePayPalMessaging`
- Add `BraintreePayPalMessagingTests` to app scheme for running tests
- Add `BraintreePayPalMessaging` to `Braintree.podspec`, `Package.swift`
    - Note: the Messages SDK is not currently published for Carthage (we have raised this as a blocker to that team)
    - Will update with the Cartfile updates in a future PR
- 

This work is going into the `paypal-messaging-feature` branch

### Checklist

### Authors

- @agedd @jaxdesmarais

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -36,6 +36,10 @@ let package = Package(
             targets: ["BraintreePayPal", "PPRiskMagnes"]
         ),
         .library(
+            name: "BraintreePayPalMessaging",
+            targets: ["BraintreePayPalMessaging"]
+        ),
+        .library(
             name: "BraintreePayPalNativeCheckout",
             targets: ["BraintreePayPalNativeCheckout"]
         ),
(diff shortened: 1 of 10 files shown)


Package.swift
```diff
@@ -35,6 +35,10 @@ let package = Package(
             name: "BraintreePayPal",
             targets: ["BraintreePayPal", "PPRiskMagnes"]
         ),
+        .library(
+            name: "BraintreePayPalMessaging",
```
sarahkoop, line 39: Is now the time to create new modules without the `Braintree` prefix? Or do we want to drop all the prefixes in v7 instead? Either way is fine by me just wanted to call it out

jaxdesmarais: The Braintree prefix isn't an Obj-C thing and I believe will always be required on iOS. It's used because CocoaPods requires a namespace. ex: merchants using this module with CocoaPods would do `pod 'Braintree/PayPalMessaging'`. @scannillo to keep me honest.

Now the `BT` prefix on classes/enums/etc is another question entirely. I'm not sure if we want to wait for v7 for that change or make it sooner ([I asked that question yesterday here](https://github.com/braintree/braintree_ios/pull/1142#discussion_r1418029329))

sshropshire (approved):

Package.swift
```diff
@@ -83,6 +87,15 @@ let package = Package(
             name: "BraintreePayPal",
             dependencies: ["BraintreeCore", "BraintreeDataCollector"]
         ),
+        .target(
+            name: "BraintreePayPalMessaging",
+            dependencies: ["BraintreeCore"]
```
scannillo, line 92:
```suggestion
            dependencies: ["BraintreeCore", "PayPalMessages"]
```

Package.swift
scannillo: Yeh, it's really just for "parity" with CocoaPods naming which requires the `Braintree` prefix. It is up to us, though!

scannillo (approved):

PR #1148, opened 2023-12-06
Add `URLOpener` protocol for mocking `UIApplication.shared`

scannillo:
### Summary of changes

- Move to protocol & extension to mock `UIApplication.shared`
    - This let's us remove the logic that lived in `BTVenmoClient` source code which only existed for mocking
- Update `FakeApplication` to conform to new protocol

### References

http://swiftyjimmy.com/mock-uiapplication-swift/

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeVenmo/BTVenmoClient.swift
+++ b/Sources/BraintreeVenmo/BTVenmoClient.swift
@@ -15,8 +15,7 @@ import BraintreeCore
     var apiClient: BTAPIClient
 
     /// Defaults to `UIApplication.shared`, but exposed for unit tests to inject test doubles
-    /// to prevent calls to openURL. Its type is `id` and not `UIApplication` because trying to subclass
-    /// UIApplication is not possible, since it enforces that only one instance can ever exist
+    /// to prevent calls to openURL. Subclassing UIApplication is not possible, since it enforces that only one instance can ever exist.
     var application: URLOpener = UIApplication.shared
 
     /// Defaults to `Bundle.main`, but exposed for unit tests to inject test doubles to stub values in infoDictionary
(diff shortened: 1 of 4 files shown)


Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -221,25 +220,17 @@ import BraintreeCore
...
 
     /// Switches to the App Store to download the Venmo application.
     @objc public func openVenmoAppPageInAppStore() {
         apiClient.sendAnalyticsEvent("ios.pay-with-venmo.app-store.invoked")
-        if let _ = application as? UIApplication {
-            UIApplication.shared.open(appStoreURL)
-        } else {
-            application.open(appStoreURL)
-        }
+        application.open(appStoreURL, options: [:], completionHandler: nil)
```
jaxdesmarais, line 233: Take it or leave it, but if we want to provide defaults like `open` in `UIApplication` we can provide it in a protocol extension to `URLOpener`:
```swift
public extension URLOpener {

    func open(_ url: URL, options: [UIApplication.OpenExternalURLOptionsKey: Any] = [:], completionHandler completion: ((Bool) -> Void)? = nil) {
        open(url, options: options, completionHandler: completion)
    }
}
```

Would allow us to update this and the call to open in `performAppSwitch` to:
```swift
application.open(appStoreURL) // openVenmoAppPageInAppStore
application.open(appSwitchURL) { success in // performAppSwitch
```

sshropshire (approved):

Sources/BraintreeVenmo/BTVenmoClient.swift
scannillo: Yeh this crossed my mind - but I thought the purpose of the protocol extension was confusing to read/deduce. I don't mind the passing of the empty dict and nil completion, but will defer to what folks think is more clear!

jaxdesmarais: I like the cleaner callsite but I 100% just think that's a personal preference. Lol. I am fine with either so will leave it up to you to decide!

PR #1149, opened 2023-12-07
Allow email OR phone to init `BTShopperInsightsRequest`

scannillo:
### Summary of changes

- Add multiple inits to `BTShopperInsightsRequest` to allow for merchants to pass the following combinations:
   1. email only
   2. phone only
   3. email & phone
- Cleanup docstrings & remove redundancies

Note: Tests for this will come once we start constructing the POST json body.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeCore/BTShopperInsightsRequest.swift
+++ b/Sources/BraintreeCore/BTShopperInsightsRequest.swift
@@ -4,25 +4,52 @@ import Foundation
 /// - Note: This feature is in beta. It's public API may change or be removed in future releases.
 public struct BTShopperInsightsRequest {
     
+    // MARK: - Private Properties
+    
     /// The buyer's email address.
-    private let email: String
+    private var email: String?
     
-    /// The buyer's country code prefix to the national telephone number. An identifier for a specific country.
-    /// Must not contain special characters.
-    private let phoneCountryCode: String
+    /// The buyer's phone number details.
+    private var phone: Phone?
     
-    /// The buyer's national phone number. Must not contain special characters.
-    private let phoneNationalNumber: String
+    // MARK: - Initializers
+
+    /// Initialize a `BTShopperInsightsRequest`
+    /// - Parameters:
+    ///   - email: The buyer's email address.
+    ///   - phone: The buyer's phone number details.
+    /// - Note: This feature is in beta. It's public API may change or be removed in future releases.
+    public init(email: String, phone: Phone) {
+        self.email = email
+        self.phone = phone
+    }
     
     /// Initialize a `BTShopperInsightsRequest`
     /// - Parameters:
     ///   - email: The buyer's email address.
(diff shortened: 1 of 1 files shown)


scannillo:
UI test failures unrelated. Created a JIRA to investigate DTBTSDK-3253

Sources/BraintreeCore/BTShopperInsightsRequest.swift
```diff
@@ -4,25 +4,55 @@ import Foundation
...
+        self.phone = phone
+    }
+    
+    // MARK: - Data Types
+    
+    /// Buyer's phone number details.
+    public struct Phone {
+        
+        private let phoneCountryCode: String
+        private let phoneNationalNumber: String
```
sshropshire, line 46:
```suggestion
        private let countryCode: String
        private let nationalNumber: String
```
nit: To make property accesses like `phone.countryCode` and `phone.nationalNumber` read cleanly.

sshropshire (approved):

Sources/BraintreeCore/BTShopperInsightsRequest.swift
```diff
@@ -4,25 +4,55 @@ import Foundation
...
+    ///   - phone: The buyer's phone number details.
+    /// - Note: This feature is in beta. It's public API may change or be removed in future releases.
+    public init(phone: Phone) {
+        self.phone = phone
+    }
+    
+    // MARK: - Data Types
+    
+    /// Buyer's phone number details.
+    public struct Phone {
```
jaxdesmarais, line 43: Do we want this struct to live inside the `BTShopperInsightsRequest` struct or outside of it? I think typically we've leaned towards outside and this seems to be what Apple follows with [PKPaymentRequest](https://developer.apple.com/documentation/passkit/pkpaymentrequest) as an example as well.

scannillo: Addressed in 7b84ca4255bb3d56280432f306908741693d0816!

Sources/BraintreeCore/BTShopperInsightsRequest.swift
```diff
@@ -4,25 +4,54 @@ import Foundation
...
 public struct BTShopperInsightsRequest {
     
-    /// The buyer's email address.
-    private let email: String
+    // MARK: - Private Properties
     
-    /// The buyer's country code prefix to the national telephone number. An identifier for a specific country.
-    /// Must not contain special characters.
-    private let phoneCountryCode: String
+    private var email: String?
```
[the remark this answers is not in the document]
scannillo: Hm, maybe let's leave them private until we have to? I'm not sure that we will necessarily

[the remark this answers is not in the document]
scannillo: This type will end up being encoded for the POST to PaymentReady API

PR #1150, opened 2023-12-08
Add SPM workspace dir to gitignore

scannillo:
Screenshot 2023-12-08 at 12 07 37 PM

### Summary of changes

- We don't check in this file and it's always a pain when switching b/w branches. 
- Added it to the gitignore

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,5 +34,6 @@ SampleApps/CarthageTest/Cartfile.resolved
 .swiftpm
 .build
 SampleApps/SPMTest/SPMTest.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+Braintree.xcworkspace/xcshareddata/swiftpm/
 
 *.pid


.gitignore
```diff
@@ -34,5 +34,6 @@ SampleApps/CarthageTest/Cartfile.resolved
 .swiftpm
 .build
 SampleApps/SPMTest/SPMTest.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+Braintree.xcworkspace/xcshareddata/swiftpm/
```
jaxdesmarais, line 37: 🤔 I don't know if we actually _do_ want to be commit-ing this file when we bump the MXO version. Seems like based on [this PR](https://github.com/braintree/braintree_ios/pull/1073) we may run into the doc steps failing if we add it back to the gitignore.

scannillo: Hm, I think they're 2 different files:

1. Braintree.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
    - _Is_ checked into the repo on `main` and is not in the gitignore (seen in the PR you linked).
3. Braintree.xcworkspace/xcshareddata/swiftpm/Package.resolved
    - _Is not_ checked into the repo.

jaxdesmarais: My memory is vague for sure so I don't _entirely_ remember why this line was removed from the `.gitignore` and if we should have been checking that file in. 🤷

jaxdesmarais (approved):

.gitignore
scannillo: The file I added to the `gitignore` (#2 in the list above) is currently not checked in, so this PR isn't removing/modifying any files.

The files are the exact same, and I _think_ we have 2 because of how our workspace is set up and that we work off of an `xcworkspace` vs a `xcodeproj`. `[appName].xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` looks to be the standard place for that file to live, which is the one we have checked in.

sarahkoop (approved):

scannillo:
UI test failures unrelated (DTBTCLIENTSDK-3253)

PR #1153, opened 2023-12-11
Paypal Messaging Implementation

jaxdesmarais:
This work is going into to `paypal-messaging-feature`

### Summary of changes

- Add `BTPayPalMessagingRequest`, `BTPayPalMessagingColor`, `BTPayPalMessagingLogoType`, `BTPayPalMessagingOfferType`, `BTPayPalMessagingPlacement`, `BTPayPalMessagingTextAlignment`, and `BTPayPalMessagingDelegate`
- Add `BTPayPalMessagingView` as the main entry point for merchants

Follow up PRs will include:
- Testing
- Analytics
- Demo App

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -250,6 +250,7 @@
 		BEA0F91E2B23728000C21EFA /* BTPayPalMessagingDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEA0F91D2B23728000C21EFA /* BTPayPalMessagingDelegate.swift */; };
 		BEA0F9202B23731700C21EFA /* BTPayPalMessagingLogoType.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEA0F91F2B23731700C21EFA /* BTPayPalMessagingLogoType.swift */; };
 		BEA0F9222B23736500C21EFA /* BTPayPalMessagingOfferType.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEA0F9212B23736500C21EFA /* BTPayPalMessagingOfferType.swift */; };
+		BEA0F9242B2373A600C21EFA /* BTPayPalMessagingPlacement.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEA0F9232B2373A600C21EFA /* BTPayPalMessagingPlacement.swift */; };
 		BEB9BF532A26872B00A3673E /* BTWebAuthenticationSessionClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEB9BF522A26872B00A3673E /* BTWebAuthenticationSessionClient.swift */; };
 		BEBC222728D25BB400D83186 /* Helpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 80DBE69423A931A600373230 /* Helpers.swift */; };
 		BEBC6E4B29258FD4004E25A0 /* BraintreeCore.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 570B93AC285397520041BAFE /* BraintreeCore.framework */; };
@@ -893,6 +894,7 @@
 		BEA0F91D2B23728000C21EFA /* BTPayPalMessagingDelegate.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalMessagingDelegate.swift; sourceTree = "<group>"; };
 		BEA0F91F2B23731700C21EFA /* BTPayPalMessagingLogoType.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalMessagingLogoType.swift; sourceTree = "<group>"; };
 		BEA0F9212B23736500C21EFA /* BTPayPalMessagingOfferType.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalMessagingOfferType.swift; sourceTree = "<group>"; };
+		BEA0F9232B2373A600C21EFA /* BTPayPalMessagingPlacement.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalMessagingPlacement.swift; sourceTree = "<group>"; };
 		BEB9BF522A26872B00A3673E /* BTWebAuthenticationSessionClient.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTWebAuthenticationSessionClient.swift; sourceTree = "<group>"; };
 		BEBC6E5D2927CF59004E25A0 /* Braintree.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = Braintree.h; sourceTree = "<group>"; };
 		BEBC6F252937A510004E25A0 /* BTClientMetadata_Tests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTClientMetadata_Tests.swift; sourceTree = "<group>"; };
@@ -1765,6 +1767,7 @@
 				BEA0F91D2B23728000C21EFA /* BTPayPalMessagingDelegate.swift */,
 				BEA0F91F2B23731700C21EFA /* BTPayPalMessagingLogoType.swift */,
 				BEA0F9212B23736500C21EFA /* BTPayPalMessagingOfferType.swift */,
+				BEA0F9232B2373A600C21EFA /* BTPayPalMessagingPlacement.swift */,
 			);
 			path = BraintreePayPalMessaging;
 			sourceTree = "<group>";
@@ -3232,6 +3235,7 @@
 			isa = PBXSourcesBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				BEA0F9242B2373A600C21EFA /* BTPayPalMessagingPlacement.swift in Sources */,
 				BEA0F9222B23736500C21EFA /* BTPayPalMessagingOfferType.swift in Sources */,
 				BEA0F9202B23731700C21EFA /* BTPayPalMessagingLogoType.swift in Sources */,
 				BEA0F91E2B23728000C21EFA /* BTPayPalMessagingDelegate.swift in Sources */,
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -1768,6 +1768,7 @@
 		BECC91942B20DF7F00A2EE42 /* BraintreePayPalMessaging */ = {
 			isa = PBXGroup;
 			children = (
+				BECC91962B20E01F00A2EE42 /* BTPayPalMessagingClient.swift */,
 				BEA0F9162B235DDA00C21EFA /* BTPayPalMessagingColor.swift */,
 				BEA0F91D2B23728000C21EFA /* BTPayPalMessagingDelegate.swift */,
 				BEA0F9292B23768600C21EFA /* BTPayPalMessagingError.swift */,
@@ -1776,7 +1777,6 @@
 				BEA0F9232B2373A600C21EFA /* BTPayPalMessagingPlacement.swift */,
 				BEA0F9272B23741900C21EFA /* BTPayPalMessagingRequest.swift */,
 				BEA0F9252B2373E400C21EFA /* BTPayPalMessagingTextAlignment.swift */,
-				BECC91962B20E01F00A2EE42 /* BTPayPalMessagingClient.swift */,
 			);
 			path = BraintreePayPalMessaging;
 			sourceTree = "<group>";
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -275,7 +275,7 @@
 		BECC91702B20DDFF00A2EE42 /* BraintreeCore.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 570B93AC285397520041BAFE /* BraintreeCore.framework */; };
 		BECC91872B20DE7C00A2EE42 /* BraintreeTestShared.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = A903E1A624F9D34000C314E1 /* BraintreeTestShared.framework */; };
 		BECC918F2B20DE9900A2EE42 /* BraintreePayPalMessaging.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = BECC91762B20DDFF00A2EE42 /* BraintreePayPalMessaging.framework */; platformFilter = ios; };
-		BECC91972B20E01F00A2EE42 /* BTPayPalMessagingView.swift in Sources */ = {isa = PBXBuildFile; fileRef = BECC91962B20E01F00A2EE42 /* BTPayPalMessagingView.swift */; };
+		BECC91972B20E01F00A2EE42 /* BTPayPalMessagingClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = BECC91962B20E01F00A2EE42 /* BTPayPalMessagingClient.swift */; };
 		BECC91992B20E1E400A2EE42 /* BTPayPalMessagingView_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BECC91982B20E1E400A2EE42 /* BTPayPalMessagingView_Tests.swift */; };
 		BED00CAC28A4504E00D74AEC /* BraintreeCore.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 570B93AC285397520041BAFE /* BraintreeCore.framework */; };
 		BED00CAE28A5419900D74AEC /* BTBinData.swift in Sources */ = {isa = PBXBuildFile; fileRef = BED00CAD28A5419900D74AEC /* BTBinData.swift */; };
@@ -919,7 +919,7 @@
 		BECBA0E52AEABC99002518AC /* BTCardClient_IntegrationTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTCardClient_IntegrationTests.swift; sourceTree = "<group>"; };
 		BECC91762B20DDFF00A2EE42 /* BraintreePayPalMessaging.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; includeInIndex = 0; path = BraintreePayPalMessaging.framework; sourceTree = BUILT_PRODUCTS_DIR; };
 		BECC918D2B20DE7C00A2EE42 /* BraintreePayPalMessagingTests.xctest */ = {isa = PBXFileReference; explicitFileType = wrapper.cfbundle; includeInIndex = 0; path = BraintreePayPalMessagingTests.xctest; sourceTree = BUILT_PRODUCTS_DIR; };
-		BECC91962B20E01F00A2EE42 /* BTPayPalMessagingView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalMessagingView.swift; sourceTree = "<group>"; };
+		BECC91962B20E01F00A2EE42 /* BTPayPalMessagingClient.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalMessagingClient.swift; sourceTree = "<group>"; };
 		BECC91982B20E1E400A2EE42 /* BTPayPalMessagingView_Tests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalMessagingView_Tests.swift; sourceTree = "<group>"; };
 		BECC919C2B20E87D00A2EE42 /* Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = Info.plist; sourceTree = "<group>"; };
 		BED00CAD28A5419900D74AEC /* BTBinData.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTBinData.swift; sourceTree = "<group>"; };
@@ -1776,7 +1776,7 @@
 				BEA0F9232B2373A600C21EFA /* BTPayPalMessagingPlacement.swift */,
 				BEA0F9272B23741900C21EFA /* BTPayPalMessagingRequest.swift */,
 				BEA0F9252B2373E400C21EFA /* BTPayPalMessagingTextAlignment.swift */,
-				BECC91962B20E01F00A2EE42 /* BTPayPalMessagingView.swift */,
+				BECC91962B20E01F00A2EE42 /* BTPayPalMessagingClient.swift */,
 			);
 			path = BraintreePayPalMessaging;
 			sourceTree = "<group>";
@@ -3250,7 +3250,7 @@
 				BEA0F9262B2373E400C21EFA /* BTPayPalMessagingTextAlignment.swift in Sources */,
 				BEA0F91E2B23728000C21EFA /* BTPayPalMessagingDelegate.swift in Sources */,
 				BEA0F9282B23741900C21EFA /* BTPayPalMessagingRequest.swift in Sources */,
-				BECC91972B20E01F00A2EE42 /* BTPayPalMessagingView.swift in Sources */,
+				BECC91972B20E01F00A2EE42 /* BTPayPalMessagingClient.swift in Sources */,
 				BEA0F92A2B23768600C21EFA /* BTPayPalMessagingError.swift in Sources */,
 				BEA0F9172B235DDA00C21EFA /* BTPayPalMessagingColor.swift in Sources */,
 			);
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -244,6 +244,9 @@
 		BE9EC0982899CF040022EC63 /* BTAPIPinnedCertificates.swift in Sources */ = {isa = PBXBuildFile; fileRef = BE9EC0972899CF040022EC63 /* BTAPIPinnedCertificates.swift */; };
 		BE9FB82B2898324C00D6FE2F /* BTPaymentMethodNonce.swift in Sources */ = {isa = PBXBuildFile; fileRef = BE9FB82A2898324C00D6FE2F /* BTPaymentMethodNonce.swift */; };
 		BE9FB82D28984ADE00D6FE2F /* BTPaymentMethodNonceParser.swift in Sources */ = {isa = PBXBuildFile; fileRef = BE9FB82C28984ADE00D6FE2F /* BTPaymentMethodNonceParser.swift */; };
+		BEA0F9172B235DDA00C21EFA /* BTPayPalMessagingColor.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEA0F9162B235DDA00C21EFA /* BTPayPalMessagingColor.swift */; };
+		BEA0F91A2B235E7A00C21EFA /* PayPalMessages in Frameworks */ = {isa = PBXBuildFile; productRef = BEA0F9192B235E7A00C21EFA /* PayPalMessages */; };
+		BEA0F91C2B235E9700C21EFA /* PayPalMessages in Frameworks */ = {isa = PBXBuildFile; productRef = BEA0F91B2B235E9700C21EFA /* PayPalMessages */; };
 		BEB9BF532A26872B00A3673E /* BTWebAuthenticationSessionClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = BEB9BF522A26872B00A3673E /* BTWebAuthenticationSessionClient.swift */; };
 		BEBC222728D25BB400D83186 /* Helpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 80DBE69423A931A600373230 /* Helpers.swift */; };
 		BEBC6E4B29258FD4004E25A0 /* BraintreeCore.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 570B93AC285397520041BAFE /* BraintreeCore.framework */; };
@@ -883,6 +886,7 @@
 		BE9EC0972899CF040022EC63 /* BTAPIPinnedCertificates.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTAPIPinnedCertificates.swift; sourceTree = "<group>"; };
 		BE9FB82A2898324C00D6FE2F /* BTPaymentMethodNonce.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPaymentMethodNonce.swift; sourceTree = "<group>"; };
 		BE9FB82C28984ADE00D6FE2F /* BTPaymentMethodNonceParser.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPaymentMethodNonceParser.swift; sourceTree = "<group>"; };
+		BEA0F9162B235DDA00C21EFA /* BTPayPalMessagingColor.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTPayPalMessagingColor.swift; sourceTree = "<group>"; };
 		BEB9BF522A26872B00A3673E /* BTWebAuthenticationSessionClient.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTWebAuthenticationSessionClient.swift; sourceTree = "<group>"; };
 		BEBC6E5D2927CF59004E25A0 /* Braintree.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = Braintree.h; sourceTree = "<group>"; };
 		BEBC6F252937A510004E25A0 /* BTClientMetadata_Tests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTClientMetadata_Tests.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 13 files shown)


Braintree.xcodeproj/project.pbxproj
```diff
@@ -900,9 +918,7 @@
 		BECA56C929C3F0A80098EC3C /* BTThreeDSecureV2UICustomization.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTThreeDSecureV2UICustomization.swift; sourceTree = "<group>"; };
 		BECBA0E52AEABC99002518AC /* BTCardClient_IntegrationTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = BTCardClient_IntegrationTests.swift; sourceTree = "<group>"; };
 		BECC91762B20DDFF00A2EE42 /* BraintreePayPalMessaging.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; includeInIndex = 0; path = BraintreePayPalMessaging.framework; sourceTree = BUILT_PRODUCTS_DIR; };
-		BECC91772B20DE0000A2EE42 /* BraintreeVenmo copy-Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; name = "BraintreeVenmo copy-Info.plist"; path = "/Users/jdesmarais/Desktop/PayPal/braintree_ios/BraintreeVenmo copy-Info.plist"; sourceTree = "<absolute>"; };
```
jaxdesmarais, line 903: These files got accidentally committed when duplicating the target - removing them

Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
```diff
@@ -1,10 +1,149 @@
...
+    /// - Parameter apiClient: The Braintree API client
+    public init(apiClient: BTAPIClient) {
+        self.apiClient = apiClient
+
+        super.init(frame: .zero)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
```
jaxdesmarais, line 31: Required to subclass `UIView`

Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
```diff
@@ -1,10 +1,149 @@
...
+            ])
+
+            return
+        }
+    }
+}
+
+// MARK: - UIViewRepresentable protocol conformance
+
+public extension BTPayPalMessagingView {
```
jaxdesmarais, line 90: This is our implementation for SwiftUI merchants. Have tested in the spike branch that this will work as expected for Swift UI merchants. Here is what the implementation looks like:
```swift
struct PayPalCreditMessagingSwiftUIView: View {

    let apiClient = BTAPIClient(authorization: "MY_AUTHORIZATION")!
    let request = BTPayPalMessagingRequest(
        amount: 2.0,
        offerType: .payLaterLongTerm,
        buyerCountry: "US",
        logoType: .primary,
        textAlignment: .center
    )

    var body: some View {
        BTPayPalMessagingView.Representable(apiClient: apiClient, request: request, delegate: messagingDelegate)
    }

    private var messagingDelegate: PayPalMessagingSwiftUIDelegate {
        PayPalMessagingSwiftUIDelegate()
    }
}

class PayPalMessagingSwiftUIDelegate: BTPayPalMessagingDelegate {

    func didSelect(_ payPalCreditMessageView: BTPayPalMessagingView) {
        print("didSelect fired")
    }

    func willApply(_ payPalCreditMessageView: BTPayPalMessagingView) {
        print("willApply fired")
    }

    func willAppear(_ payPalCreditMessageView: BTPayPalMessagingView) {
        print("willAppear fired")
    }

    func didAppear(_ payPalCreditMessageView: BTPayPalMessagingView) {
        print("didAppear fired")
    }

    func onError(_ payPalCreditMessageView: BTPayPalMessagingView, error: Error) {
        print("onError fired with \(error.localizedDescription)")
    }
}
```

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -0,0 +1,149 @@
...
+
+    /// Creates a `BTPayPalMessagingClient` toto be displayed to promote offers such as Pay Later and PayPal Credit to customers.
+    /// - Parameter request: an optional `BTPayPalMessagingRequest`
+    public func createView(_ request: BTPayPalMessagingRequest = BTPayPalMessagingRequest()) {
+        apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
+            if let error {
+                self.delegate?.onError(self, error: error)
+                return
+            }
+            guard let configuration else {
```
jaxdesmarais, line 43:
```suggestion
            }

            guard let configuration else {
```

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -0,0 +1,149 @@
...
+        super.init(frame: .zero)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Public Method
+
+    /// Creates a `BTPayPalMessagingClient` toto be displayed to promote offers such as Pay Later and PayPal Credit to customers.
```
scannillo, line 35:
```suggestion
    /// Creates a `BTPayPalMessagingClient` to be displayed to promote offers such as Pay Later and PayPal Credit to customers.
```

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -0,0 +1,150 @@
...
+            let messageView = PayPalMessageView(config: messageConfig, stateDelegate: self, eventDelegate: self)
+            messageView.translatesAutoresizingMaskIntoConstraints = false
+            self.addSubview(messageView)
+
+            NSLayoutConstraint.activate([
+                messageView.leadingAnchor.constraint(equalTo: self.leadingAnchor),
+                messageView.trailingAnchor.constraint(equalTo: self.trailingAnchor),
+                messageView.topAnchor.constraint(equalTo: self.topAnchor),
+                messageView.bottomAnchor.constraint(equalTo: self.bottomAnchor)
+            ])
```
scannillo, line 82: Do we need to have `BTPayPalMessagingClient` conform to `UIView`? What about having a method that returns the merchant a `UIView`?

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -0,0 +1,149 @@
...
+    var apiClient: BTAPIClient
+
+    // MARK: - Initializers
+
+    ///  Initializes a PayPal Messaging client.
+    /// - Parameter apiClient: The Braintree API client
+    public init(apiClient: BTAPIClient) {
+        self.apiClient = apiClient
+
+        super.init(frame: .zero)
```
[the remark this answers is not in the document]
jaxdesmarais: `apiClient` needs to be initialized before calling the `super.init`. There is a build error switching the order.

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
jaxdesmarais: If we want to expose this for merchants using SwiftUI, we need to be bound to a `View` type to have access to the corresponding SwiftUI `Representable` (either `UIViewRepresentable` or `UIViewControllerRepresentable` most commonly). If we don't want to support SwiftUI, we can just have a method that returns a `UIView`, but if we do it's necessary.

Sources/BraintreePayPalMessaging/BTPayPalMessagingRequest.swift
```diff
@@ -0,0 +1,44 @@
...
+    ///   - logoType: Logo type option for a PayPal Message. Defaults to `.inline`
+    ///   - textAlignment: Text alignment option for a PayPal Message. Defaults to `.right`
+    ///   - color: Text and logo color option for a PayPal Message. Defaults to `.black`
+    public init(
+        amount: Double? = nil,
+        placement: BTPayPalMessagingPlacement? = nil,
+        offerType: BTPayPalMessagingOfferType? = nil,
+        buyerCountry: String? = nil,
+        logoType: BTPayPalMessagingLogoType = .inline,
+        textAlignment: BTPayPalMessagingTextAlignment = .right,
```
sarahkoop, line 32: Should the default be left aligned?

Sources/BraintreePayPalMessaging/BTPayPalMessagingOfferType.swift
```diff
@@ -0,0 +1,33 @@
...
+public enum BTPayPalMessagingOfferType {
+
+   /// Pay Later short term installment
+   case payLaterShortTerm
+
+   /// Pay Later long term installments
+   case payLaterLongTerm
+
+   /// Pay Later deferred payment
+   case payLaterPayIn1
```
sarahkoop, line 15:
```suggestion
   case payLaterPayInOne
```
Take it or leave it - seeing numbers in enum cases feels wrong

Sources/BraintreePayPalMessaging/BTPayPalMessagingRequest.swift
jaxdesmarais: I based these on the defaults in the `PayPalMessages`, but not entirely sure why they went right instead of left. Technically we can do whatever we want, so if we prefer left we can do that.

Sources/BraintreePayPalMessaging/BTPayPalMessagingOfferType.swift
jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/1153/commits/b0f635ea91decac7f88552ff291bbd4473d6e34a

Sources/BraintreePayPalMessaging/BTPayPalMessagingRequest.swift
sarahkoop: If that's the default in PayPalMessages it's fine with me 🤷

sarahkoop (approved):

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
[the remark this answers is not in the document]
hollabaq86: 🤔 does it make sense to add a comment about this since calling `super.init` first is a common engineering practice?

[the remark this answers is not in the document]
hollabaq86: Or, is it a code smell that `apiClient` needs to be set first?

Sources/BraintreePayPalMessaging/BTPayPalMessagingRequest.swift
hollabaq86: let's get confirmation from PP friends which is preferred default. I want to make sure we have parity across PP and BT

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -0,0 +1,150 @@
...
+                amount: request.amount,
+                placement: request.placement?.placementRawValue,
+                offerType: request.offerType?.offerTypeRawValue
+            )
+
+            messageData.buyerCountry = request.buyerCountry
+
+            let messageConfig = PayPalMessageConfig(
+                data: messageData,
+                style: PayPalMessageStyle(
```
hollabaq86, line 66: Question, and this may be something we bring to PP or product folks to get confirmation.

Do we need to be concerned about merchants opting logo types and colors that are NOT meant to be together, per our brand guidelines? 

Alternatively, are there logo and color combinations that absolutely would not pass an accessibility check?

Asking because I'm curious if we may want to surface a developer error for these edge case situations.

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -0,0 +1,150 @@
...
+                    logoType: request.logoType.logoTypeRawValue,
+                    color: request.color.messageColorRawValue,
+                    textAlignment: request.textAlignment.textAlignmentRawValue
+                )
+            )
+
+            let messageView = PayPalMessageView(config: messageConfig, stateDelegate: self, eventDelegate: self)
+            messageView.translatesAutoresizingMaskIntoConstraints = false
```
hollabaq86, line 74: shouldn't this be `AutoResizing` instead of `Autoresizing`?

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -0,0 +1,150 @@
...
+
+// MARK: - PayPalMessageViewEventDelegate and PayPalMessageViewStateDelegate protocol conformance
+
+extension BTPayPalMessagingClient: PayPalMessageViewEventDelegate, PayPalMessageViewStateDelegate {
+
+    public func onClick(_ paypalMessageView: PayPalMessages.PayPalMessageView) {
+        delegate?.didSelect(self)
+    }
+
+    public func onApply(_ paypalMessageView: PayPalMessages.PayPalMessageView) {
```
hollabaq86, line 135: I love this pattern ❤️

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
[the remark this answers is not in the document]
jaxdesmarais: It's not common in Swift. All the non optional properties (in this case `apiClient`) of a class need to be initialized before being accessed or calling the super init. I don't think it needs to be documented nor does it signal code smell. It's part of the languages safety as a whole: https://medium.com/@dineshk1389/swift-why-super-init-is-called-after-setting-all-self-properties-d6827e9f4eb2

Sources/BraintreePayPalMessaging/BTPayPalMessagingLogoType.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+import PayPalMessages
+
+/// Logo type option for a PayPal Message
+/// - Note: This module is in beta. It's public API may change or be removed in future releases.
+public enum BTPayPalMessagingLogoType {
+
+    /// PayPal logo positioned inline within the message
```
hollabaq86, line 8: we'll eventually want to include a link to docs showing what these different logos look like

Sources/BraintreePayPalMessaging/BTPayPalMessagingRequest.swift
jaxdesmarais: `right` is their default, see: https://github.com/paypal/paypal-messages-ios/blob/6904fdcc27b51c3dc0e05356423b618d584a925b/Sources/PayPalMessages/Config/PayPalMessageConfig.swift#L77

Sources/BraintreePayPalMessaging/BTPayPalMessagingLogoType.swift
```diff
@@ -0,0 +1,33 @@
...
+    case inline
+
+    /// Primary logo including both the PayPal monogram and logo
+    case primary
+
+    /// Alternative logo including just the PayPal monogram
+    case alternative
+
+    /// "PayPal" as bold text inline with the message
+    case none
```
hollabaq86, line 18: instead of `none`, are we ok with the term `default`,  or `minimal`? "None" seems misleading in that there _is_ going to be a logo, it's just really simple.

Maybe this is something we check in with Product for their thoughts

Sources/BraintreePayPalMessaging/BTPayPalMessagingOfferType.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+import PayPalMessages
+
+/// Preferred message offer to display
+/// - Note: This module is in beta. It's public API may change or be removed in future releases.
+public enum BTPayPalMessagingOfferType {
+
+   /// Pay Later short term installment
```
hollabaq86, line 8: We're eventually going to want docs that clearly explain the difference between short term and long term that we reference or copy/paste the definitions into our comments

Sources/BraintreePayPalMessaging/BTPayPalMessagingOfferType.swift
```diff
@@ -0,0 +1,33 @@
...
+   case payPalCreditNoInterest
+
+   var offerTypeRawValue: PayPalMessageOfferType {
+       switch self {
+       case .payLaterShortTerm:
+           return .payLaterShortTerm
+       case .payLaterLongTerm:
+           return .payLaterLongTerm
+       case .payLaterPayInOne:
+           return .payLaterPayIn1
```
hollabaq86, line 27: does this need to be changed to `payLaterPayIn1`?

hollabaq86 (commented):
Great work implementing this. Almost all of my Qs are from an outsider's perspective so apologies in advance if they may come from left field or are something an iOS dev would immediately know.

One other Q I'd like us to take a moment and consider

- do we think we'll want an analytics event showing button logo and color selections? 

This doesn't have to be included in this PR, but it's something I want to make sure we get requirements from Product on so we can address it in the future.

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
jaxdesmarais: > Do we need to be concerned about merchants opting logo types and colors that are NOT meant to be together, per our brand guidelines?

Yep the colors are black, white, monochrome, and greyscale and all already approved to go with the logos currently.

> Alternatively, are there logo and color combinations that absolutely would not pass an accessibility check?

PayPal Messages if working with the accessibility team on accessibility across the board. Imo, if there are "bad" combos the messaging SDK should surface as an error instead of our SDK.

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
jaxdesmarais: This is an Apple property 😄 : https://developer.apple.com/documentation/uikit/uiview/1622572-translatesautoresizingmaskintoco

Sources/BraintreePayPalMessaging/BTPayPalMessagingLogoType.swift
jaxdesmarais: Agree - we can have the Messages team include that when we get to writing docs for this module.

Sources/BraintreePayPalMessaging/BTPayPalMessagingLogoType.swift
jaxdesmarais: `default` is a protected case name in Swift for enums. I am good with `minimal`, `simple`, or `basic`. Do you have a preference?

Sources/BraintreePayPalMessaging/BTPayPalMessagingOfferType.swift
jaxdesmarais: Agree, [JS doesn't have the same options for offer type](https://developer.paypal.com/docs/checkout/pay-later/us/integrate/customize-messages/) so just went with what the Messages team had documented for now

Sources/BraintreePayPalMessaging/BTPayPalMessagingOfferType.swift
jaxdesmarais: Intentionally changed our naming to `one` based on this discussion: https://github.com/braintree/braintree_ios/pull/1153#discussion_r1424484439. But this is returning the raw type in the messaging SDK which uses `1` so they will differ.

jaxdesmarais:
@hollabaq86 
> do we think we'll want an analytics event showing button logo and color selections?

The Messaging SDK will have analytics on this. There is an analytics PR up here: #1161 but with iOS v6 we have generally opted to keep our analytics quite simple so we can get a good read on conversion. In this case, the conversion question will answer:
1. Are merchants using this module
2. Did it result in a success or failure

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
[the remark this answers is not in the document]
scannillo: Ah, great article! It's reversed in Objective-C and I mixed them up. Thank you clarifying 👏 

Resolving this thread.

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Public Method
+
+    /// Creates a `BTPayPalMessagingClient` to be displayed to promote offers such as Pay Later and PayPal Credit to customers.
+    /// - Parameter request: an optional `BTPayPalMessagingRequest`
```
scannillo, line 36: This is confusing to me since this method doesn't return anything.

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
hollabaq86: > Imo, if there are "bad" combos the messaging SDK should surface as an error instead of our SDK.

Agreed, it's always been our practice to be messengers and let the API/Payment Method/downstream component to be the source of truth on validation decisions. Let's check in with our friends in PayPal Messaging about this consideration and move on.

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
hollabaq86: 🙃 welp, we can disregard this q then haha

Sources/BraintreePayPalMessaging/BTPayPalMessagingLogoType.swift
hollabaq86: I like `simple`! Let's go with that

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -0,0 +1,150 @@
...
+            let messageView = PayPalMessageView(config: messageConfig, stateDelegate: self, eventDelegate: self)
+            messageView.translatesAutoresizingMaskIntoConstraints = false
+            self.addSubview(messageView)
+
+            NSLayoutConstraint.activate([
+                messageView.leadingAnchor.constraint(equalTo: self.leadingAnchor),
+                messageView.trailingAnchor.constraint(equalTo: self.trailingAnchor),
+                messageView.topAnchor.constraint(equalTo: self.topAnchor),
+                messageView.bottomAnchor.constraint(equalTo: self.bottomAnchor)
+            ])
```
scannillo, line 82: Here we're doing UI work off the main thread (in the callback from `fetchOrReturnRemoteConfiguration`) - I guess since this view isn't yet set onscreen (which the merchant would have to do), this doesn't cause an issue? Just wanted to call this out.

Sources/BraintreePayPalMessaging/BTPayPalMessagingOfferType.swift
hollabaq86: 😬 ok, we're going to want to get JS in parity with us. It looks like JS is also using the description `none` for the simple logo, so we may want to take to JS and PP Messaging folks about that choice, too.

I'll circle this feedback to Product, we can resolve this convo

Sources/BraintreePayPalMessaging/BTPayPalMessagingOfferType.swift
hollabaq86: awww I see. Thought it was a hard limitation, thanks for confirming!

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
scannillo: Another thing - these operations happen async since we're blocked by the config fetch. With the current design, a merchant needs to wait for this to `createView()` call to complete the necessary operations before displaying it onscreen.

I think we need to provide a hook into the completion of this method, or else a merchant could proceed before those constraints are set and nothing would load.

scannillo: Another thing, can we make it more clear in our docstrings where a merchant has to hook into the result of this `createView()` call (`BTPayPalMessagingDelegate`).

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
scannillo: I get that something has to conform to `UIView`, but I think calling that a `Client` is confusing and goes against our patterns for the other modules. Also in the current design, if they don't call `createView()` nothing works, right? I think that is what's tripping me up about the naming here. Could we re-name Client to View?

Another idea - do we need `createView()` to be a separate method call? What about initializing the `BTPayPalMessagingView/Client` with both an apiClient & an optional request.

Sources/BraintreePayPalMessaging/BTPayPalMessagingLogoType.swift
jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/1153/commits/afb0e5af39ad353985e8807618d43ee56b4a95a5

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
jaxdesmarais: > I guess since this view isn't yet set onscreen (which the merchant would have to do), this doesn't cause an issue? Just wanted to call this out.

Correct, this will adjust view layout based on provided constraints once the view created and does not need to be called on the main thread.

jaxdesmarais: > Another thing, can we make it more clear in our docstrings where a merchant has to hook into the result of this createView() call (BTPayPalMessagingDelegate).

Good call, updated: https://github.com/braintree/braintree_ios/pull/1153/commits/8463a60b5ed3ff8dd651c4788ac5455e7cbf4c1d

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
jaxdesmarais: > Could we re-name Client to View?

It was changed from view to client based on this feedback: https://github.com/braintree/braintree_android/pull/847#discussion_r1422591121. I'm not opposed to changing it back to view but we should align across platforms. cc: @sarahkoop since you originally provided this feedback in the Android PR.

> Another idea - do we need createView() to be a separate method call? What about initializing the BTPayPalMessagingView/Client with both an apiClient & an optional request.

We could move to this pattern if that's preferred. I'm not totally sure how this module could change in the future so it feels a bit less restrictive to have it as a method within the class and also more closely aligns with every module that exists today, but we are also listing this as a beta so we _do_ have the flexibility to change things if we don't like it. What do y'all think?

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
jaxdesmarais: Updated: https://github.com/braintree/braintree_ios/pull/1153/commits/a9c78b44c62adce131a7a4b450be5c201007b1cd

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
sshropshire: qq: will we instruct merchants to display a loader while the messaging view is configured?

jaxdesmarais: We can if we think it'd be helpful, but I think it's pretty common to leave undocumented and up to the merchant to handle any loading screen before the view is displayed.

sshropshire: I wonder if the PayPal messaging team could show a [content placeholder](https://www.codepicky.com/content-placeholders/) by default. Then merchants wouldn't have to wait to call `addSubview()`.

jaxdesmarais: I'm not sure I follow. We aren't rendering HTML/CSS here we are rendering their raw assets from the native SDK. This is also not the merchant demo this is our implementation and we need to add their subview + add constraints to our view so merchants can actually render these views. If we don't do this the view would never display since we wrap their SDK. We don't have to wait to call it nor will merchants as seen in #1160.

scannillo: I think what Steven means is that currently the time in between the merchant calls `createView()` & `didAppear()` gets fired, just a blank box is shown. I ran this on a device with simulated poor network quality too, and there's potential for that state to last a few seconds.

This would be a bigger change - but it might make more sense to only return the `UIView` instance if we know in fact that it's content is ready to be displayed. So like an async function/init that returns a `UIView`, or throws an error.

jaxdesmarais: Ah, gotcha - we should raise this to the messaging team. Imo it makes sense to add some sort of loading indicator to their SDK since they are responsible for the network calls being made before rendering the view. They could add network calls on their end that could impact us knowing when something is ready to display or not so makes since for them to handle that logic end to end. What do you think?

scannillo: Back to[ this comment](https://github.com/braintree/braintree_ios/pull/1153/files#r1430229794) regarding naming with the current pattern, I think the `Client` name could work, if it didn't conform to `UIView`. For example, if it had a method within it that returned the `PayPalMessageView` and the merchant set the delegate on that instead.

Having one `Client` conform to UIView that is rendered on screen, and _requires_ a method call (`createView()`) in addition to it's init for it properly function, is different than the rest of our `Cilent`s.

scannillo: Back to your Demo PR, [doing this](https://github.com/braintree/braintree_ios/blob/1fd50d04042da40fa0f06a7bb5ca024de90af2e9/Demo/Application/Features/PayPalMessagingViewController.swift#L22-L30) `view.addSubview(payPalMessagingClient)` reads odd to me

Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
scannillo: For a future ticket, but maybe we can add a separate Demo feature that is only in SwiftUI? Would be nice to also have reference to this in our demo somewhere

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
scannillo: > This is a blocker for supporting SwiftUI easily.

Right, we would need to introduce _another_ class type, that does conform to `View`, if we kept the `Client` name and `Client` didn't conform to View.

scannillo: I think we should mostly leave your PR as-is but revert the rename of `BTPayPalMessagingClient` --> `BTPayPalMessagingView`. Maybe Sarah's initial comment was not considering that this class will conform to View, since in that Android PR the implementation isn't shown there?

Also ➕ 1 to your comment re: bringing up the loading/blank state to the messaging team. It's beyond our scope!

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -0,0 +1,151 @@
...
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Public Method
+
+    /// Creates a view to be displayed to promote offers such as Pay Later and PayPal Credit to customers.
+    /// - Parameter request: an optional `BTPayPalMessagingRequest`
+    /// - Note: use `BTPayPalMessagingDelegate` protocol to receive notifications for events
+    public func createView(_ request: BTPayPalMessagingRequest = BTPayPalMessagingRequest()) {
```
scannillo, line 38: Another nit/idea - this name could be something that signals it is an asynchronous operation? Would `loadView` be more clear? Or `requestView()`?

Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
jaxdesmarais: Agree - it'd be great to have both in the same demo, but shouldn't be too hard for a future PR.

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
jaxdesmarais: > I think we should mostly leave your PR as-is but revert the rename of BTPayPalMessagingClient --> BTPayPalMessagingView.

Yeah I think sometimes too seeing it written out in a demo app context helps as well. We also don't have any modules like this currently so figuring out some uncharted territory. Done: https://github.com/braintree/braintree_ios/pull/1153/commits/13c898d482cea8396811efeff73571947d908a73

> Also ➕ 1 to your comment re: bringing up the loading/blank state to the messaging team. It's beyond our scope!

Already raised it in our slack channel with their team. 😄

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
jaxdesmarais: 🤔 `requestView(request)` reads a bit weird to me. We could also consider moving the request into the `init` and making this method internal. I know we threw around that idea earlier but could be one way to clear that up. Could also use `start` which is more similar to how we launch UI based flows on PPCP, so `BTPayPalMessageView.start(request)`?

scannillo (approved):
Thanks for all of the thorough responses to my many comments 😆 

Also like you mentioned, this is in beta so we have the power to make breaking changes if we need!

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPalMessaging (BETA)
+  * Add `BTPayPalMessagingRequest`, `BTPayPalMessagingColor`, `BTPayPalMessagingLogoType`, `BTPayPalMessagingOfferType`, `BTPayPalMessagingPlacement`, `BTPayPalMessagingTextAlignment`, and `BTPayPalMessagingDelegate`
+  * Add `BTPayPalMessagingView` to display PayPal messages to promote offers such as Pay Later and PayPal Credit to customers.
+    * To get started create a `BTPayPalMessagingView` and call `createView(_:) with a `BTPayPalMessagingRequest`
```
jaxdesmarais, line 7:
```suggestion
    * To get started create a `BTPayPalMessagingView` and call `start(_:)` with a `BTPayPalMessagingRequest`
```

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
jaxdesmarais: irl-ed and decided to go with `start`: https://github.com/braintree/braintree_ios/pull/1153/commits/322008fb14f0b7ed6ca2f342b3332edd85568afa

PR #1159, opened 2023-12-12
Make SDK Errors Public

### LINKED ISSUE #1152: Better Error Handling
### Is your feature request related to a problem? Please describe.

Hello! 

I use BrainTree for 3D Secure Actions. Unfortunately, when user cancels the flow on `Secure Checkout` window.

Simulator Screenshot - iPhone 15 - 2023-12-11 at 10 42 47

I get `Error Domain=com.braintreepayments.BTThreeDSecureFlowErrorDomain Code=5 "(null)"`.

I'd expec one of `BTThreeDSecureError` to be thrown - i.e. `BTThreeDSecureError.cancelled`.

Not sure if there is anything wrong I'm doing. Here is the code - card is tokenized by our server.

```
        let btRequest = BTThreeDSecureRequest(payload: verifyCardPayload)
        let btClient = BTAPIClient(authorization: action.data.clientToken)!
        btThreeDSecureClient = BTThreeDSecureClient(apiClient: btClient)
        btRequest.threeDSecureRequestDelegate = self
        btThreeDSecureClient?.startPaymentFlow(btRequest) { result, error in
            if let error {
                print(error)
            } else if let nonce = result?.tokenizedCard?.nonce {
                print(nonce)
            } else {
                print(result)
            }
        }
```

Where `verifyCardPayload` is:

```
  ▿ amount : Optional<String>
    - some : "0.01"
  ▿ nonce : Optional<String>
    - some : "tokencc_xyz"
  ▿ bin : Optional<String>
    - some : "123456"
  ▿ email : Optional<String>
    - some : "a@b.com"
  - dfReferenceId : nil
  - exemptionRequested : false
  - challengeRequested : false
  - billingAddress : nil
  - additionalInformation : nil
```

Let me know if you need any more information. I use version `6.10.0`

### Describe the solution you'd like.

_No response_

jaxdesmarais:
### Summary of changes

- Update all SDK errors to be public (fixes #1152 and #1080)
- Update all errors to conform to Equatable
    - This allows for a better merchant experience as they won't need to cast to `NSError` and can do error comparison in a more Swift-y way (see Demo app ViewController updates as an example)

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* Update all SDK errors to be public and Equatable (fixes #1152 and #1080)
+
 ## 6.10.0 (2023-11-17)
 * BraintreePayPalNativeCheckout
   * Update PayPalCheckout from 1.1.0 to 1.2.0.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
-* Update all SDK errors to be public and Equatable (fixes #1152 and #1080)
+* Update all SDK errors to be public and [Equatable](https://developer.apple.com/documentation/swift/equatable) (fixes #1152 and #1080)
 * BraintreeThreeDSecure
   * Fix bug where `BTThreeDSecureClient.initializeChallenge()` callback wasn't properly invoked (fixes #1154)
 
--- a/Demo/Application/Features/VenmoViewController.swift
+++ b/Demo/Application/Features/VenmoViewController.swift
@@ -66,7 +66,7 @@ class VenmoViewController: PaymentButtonBaseViewController {
                 progressBlock("Got a nonce 💎!")
                 completionBlock(venmoAccount)
             } catch {
-                if (error as NSError).code == 10 {
+                if error as? BTVenmoError == .canceled {
                     progressBlock("Canceled 🔰")
                 } else {
                     progressBlock(error.localizedDescription)
--- a/Sources/BraintreeCard/BTCardError.swift
+++ b/Sources/BraintreeCard/BTCardError.swift
@@ -1,7 +1,7 @@
 import Foundation
 
 // Error codes associated with cards
-enum BTCardError: Error, CustomNSError, LocalizedError {
+public enum BTCardError: Error, CustomNSError, LocalizedError, Equatable {
 
     /// 0. Unknown error
     case unknown
@@ -18,11 +18,11 @@ enum BTCardError: Error, CustomNSError, LocalizedError {
     /// 4. Failed to fetch Braintree configuration
     case fetchConfigurationFailed
 
-    static var errorDomain: String {
+    public static var errorDomain: String {
         "com.braintreepayments.BTCardClientErrorDomain"
     }
 
-    var errorCode: Int {
+    public var errorCode: Int {
         switch self {
         case .unknown:
             return 0
@@ -37,7 +37,7 @@ enum BTCardError: Error, CustomNSError, LocalizedError {
         }
     }
 
-    var errorUserInfo: [String: Any] {
+    public var errorUserInfo: [String: Any] {
         switch self {
         case .unknown:
             return [NSLocalizedDescriptionKey: "An unknown error occurred. Please contact support."]
@@ -50,5 +50,11 @@ enum BTCardError: Error, CustomNSError, LocalizedError {
         case .fetchConfigurationFailed:
             return [NSLocalizedDescriptionKey: "Failed to fetch Braintree configuration."]
         }
+    }
+
+    // MARK: - Equatable Conformance
+
+    public static func == (lhs: BTCardError, rhs: BTCardError) -> Bool {
+        lhs.errorCode == rhs.errorCode
     }
 }
--- a/Sources/BraintreeCore/Analytics/BTAnalyticsServiceError.swift
+++ b/Sources/BraintreeCore/Analytics/BTAnalyticsServiceError.swift
@@ -1,20 +1,20 @@
 import Foundation
 
 ///  Error codes associated with a API Client.
-enum BTAnalyticsServiceError: Int, Error, CustomNSError, LocalizedError {
+public enum BTAnalyticsServiceError: Int, Error, CustomNSError, LocalizedError, Equatable {
 
     /// 0. Invalid API client
     case invalidAPIClient
 
-    static var errorDomain: String {
+    public static var errorDomain: String {
         "com.braintreepayments.BTAnalyticsServiceErrorDomain"
     }
 
-    var errorCode: Int {
+    public var errorCode: Int {
         rawValue
     }
 
-    var errorDescription: String? {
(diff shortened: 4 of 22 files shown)


Sources/BraintreeCard/BTCardError.swift
```diff
@@ -51,4 +51,10 @@ enum BTCardError: Error, CustomNSError, LocalizedError {
             return [NSLocalizedDescriptionKey: "Failed to fetch Braintree configuration."]
         }
     }
+
+    // MARK: - Equatable Conformance
+
+    public static func == (lhs: BTCardError, rhs: BTCardError) -> Bool {
+        lhs.errorCode == rhs.errorCode
+    }
```
jaxdesmarais, line 59: This is required for any of the enums with associated types. It it's able to resolve the equality of the error codes automatically.

Demo/Application/Features/VenmoViewController.swift
```diff
@@ -66,7 +66,7 @@ class VenmoViewController: PaymentButtonBaseViewController {
                 progressBlock("Got a nonce 💎!")
                 completionBlock(venmoAccount)
             } catch {
-                if (error as NSError).code == 10 {
+                if error as? BTVenmoError == .canceled {
```
jaxdesmarais, line 69: 🌟

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
```diff
@@ -141,18 +129,9 @@ extension BTThreeDSecureV2Provider: CardinalValidationDelegate {
...
-                completion: completionHandler
-            )
+            completionHandler(nil, NSError(domain: BTThreeDSecureError.errorDomain, code: errorCode, userInfo: userInfo))
         case .cancel:
-            notifyError(
-                withDomain: BTThreeDSecureError.errorDomain,
-                errorCode: BTThreeDSecureError.canceled.errorCode,
-                completion: completionHandler
-            )
+            completionHandler(nil, BTThreeDSecureError.canceled)
```
jaxdesmarais, line 134: This refactor pointed out a flaw in this logic. As we weren't _actually_ sending the `canceled` error here and instead constructing an error similar to cancel, merchants would need to check for both the code and the raw `canceled` error. Instead above we can construct the `NSError` for the error case, but should have been sending our cancel error here. Without this updated logic we actually don't fire our cancel analytics event in `BTThreeDSecureClient`.

scannillo: Good catch! I actually think that resolves this ticket too (DTBTSDK-3276). I also noticed we don't have a proper unit test for the cancel case from cardinal, which I included in that ticket. Maybe we update that ticket to be unit test specific since your PR here fixes the bug?

Sources/BraintreeCore/Analytics/BTAnalyticsServiceError.swift
```diff
@@ -1,20 +1,20 @@
...
 
 ///  Error codes associated with a API Client.
-enum BTAnalyticsServiceError: Int, Error, CustomNSError, LocalizedError {
+public enum BTAnalyticsServiceError: Int, Error, CustomNSError, LocalizedError, Equatable {
 
     /// 0. Invalid API client
     case invalidAPIClient
 
-    static var errorDomain: String {
+    public static var errorDomain: String {
```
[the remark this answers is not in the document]
jaxdesmarais: Yeah - it needs to be public to satisfy protocol requirements since we are making the enum as a whole public.

Demo/Application/Features/VenmoViewController.swift
scannillo: This is slick!

scannillo: 🏈 Punt, but curious your thoughts.

For V7, we could get to something like this `if error == .canceled { ... }` if the error type we return is BT SDK specific. Is that what you're thinking as the V7 goal?

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
jaxdesmarais: Ah - nice! Yeah this should resolve that issue. I think we can make the ticket you created test specific and include this fix here. Is it worth calling out as a separate CHANGELOG point?

Demo/Application/Features/VenmoViewController.swift
jaxdesmarais: Yeah - I think that would be dope if we return strongly typed errors. The only case it _could_ be a little weird is we would need to wrap any errors from core. I think in most cases we do that anyways but I know there are a few instances where we return a core error directly. Though I don't think it should be a blocker for strongly typed errors.

sarahkoop (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* Update all SDK errors to be public and Equatable (fixes #1152 and #1080)
```
hollabaq86, line 4:
```suggestion
* Update all SDK errors to be public and [Equatable](https://developer.apple.com/documentation/swift/equatable) (fixes #1152 and #1080)
```

hollabaq86: suggestion to add some breadcrumbs for newer devs

Sources/BraintreeCore/Analytics/BTAnalyticsServiceError.swift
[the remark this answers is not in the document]
hollabaq86: we need a taylor swift emoji for "swifty" things 😄

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
hollabaq86: is there any chance we can include the test coverage work in this PR, since we're making the change now? I'm really leery about skipping tests for another day...

hollabaq86 (approved):
Love this work! Small tweak to CHANGELOG and other general thoughts.

If we're changing code, we should incorporate test coverage into those changes, too. We absolutely can and should take the time to do the right thing:TM:. I have no problem having higher level conversations if we get pushback about this.

Sources/BraintreeThreeDSecure/BTThreeDSecureV2Provider.swift
jaxdesmarais: I poked around at testing this and it actually looks like _nothing_ in this method is tested. Mainly because it involves testing `CardinalValidationDelegate` which uses a class with no inits `CardinalSession`. I can't figure out a good way to test this closed class `CardinalSession` which is required for a mock of this class. @scannillo not sure if you have any other ideas on testing this but imo it seems like it'd be a really large initiative outside of the scope of this PR.

jaxdesmarais:
Leaving testing for a future PR as it involves wrapping a final Cardinal class without an init. None of the Cardinal delegate is tested at all so figuring out how/if we test it should be a separate workstream.

PR #1161, opened 2023-12-14
Paypal Messaging Analytics

jaxdesmarais:
This work is going into the `paypal-messaging-feature` branch

### Summary of changes

- Add `started`, `failed`, and `succeeded` events for PayPal Messaging feature

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreePayPalMessaging/BTPayPalMessagingAnalytics.swift
+++ b/Sources/BraintreePayPalMessaging/BTPayPalMessagingAnalytics.swift
@@ -1,0 +1,8 @@
+import Foundation
+
+enum BTPayPalMessagingAnalytics {
+
+    static let started = "paypal-messaging:create-view:started"
+    static let failed = "paypal-messaging:create-view:failed"
+    static let succeeded = "paypal-messaging:create-view:succeeded"
+}
--- a/Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
+++ b/Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
@@ -35,19 +35,20 @@ public class BTPayPalMessagingClient: UIView {
     /// Creates a `BTPayPalMessagingClient` to be displayed to promote offers such as Pay Later and PayPal Credit to customers.
     /// - Parameter request: an optional `BTPayPalMessagingRequest`
     public func createView(_ request: BTPayPalMessagingRequest = BTPayPalMessagingRequest()) {
+        apiClient.sendAnalyticsEvent(BTPayPalMessagingAnalytics.started)
         apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
             if let error {
-                self.delegate?.onError(self, error: error)
+                self.notifyFailure(with: error)
                 return
             }
 
             guard let configuration else {
-                self.delegate?.onError(self, error: BTPayPalMessagingError.fetchConfigurationFailed)
+                self.notifyFailure(with: BTPayPalMessagingError.fetchConfigurationFailed)
                 return
             }
 
             guard let clientID = configuration.json?["paypal"]["clientId"].asString() else {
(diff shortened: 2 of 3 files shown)


Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -141,10 +142,12 @@ extension BTPayPalMessagingClient: PayPalMessageViewEventDelegate, PayPalMessage
...
 
     public func onSuccess(_ paypalMessageView: PayPalMessages.PayPalMessageView) {
+        apiClient.sendAnalyticsEvent(BTPayPalMessagingAnalytics.succeeded)
         delegate?.didAppear(self)
     }
 
     public func onError(_ paypalMessageView: PayPalMessages.PayPalMessageView, error: PayPalMessages.PayPalMessageError) {
-        delegate?.onError(self, error: error)
+        apiClient.sendAnalyticsEvent(BTPayPalMessagingAnalytics.failed, errorDescription: error.localizedDescription)
+        self.delegate?.onError(self, error: error)
```
jaxdesmarais, line 151:
```suggestion
        delegate?.onError(self, error: error)
```

Sources/BraintreePayPalMessaging/BTPayPalMessagingClient.swift
```diff
@@ -141,10 +142,12 @@ extension BTPayPalMessagingClient: PayPalMessageViewEventDelegate, PayPalMessage
     }
 
     public func onSuccess(_ paypalMessageView: PayPalMessages.PayPalMessageView) {
+        apiClient.sendAnalyticsEvent(BTPayPalMessagingAnalytics.succeeded)
```
jaxdesmarais, line 145: Typically we have used the `notify` pattern in other PRs. Open to using that pattern here if folks prefer but to me it seemed odd to extract out the delegate call below and the analytics event for just success and error while keeping the rest of the delegate calls above when we aren't sending analytics.

Sources/BraintreePayPalMessaging/BTPayPalMessagingAnalytics.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+enum BTPayPalMessagingAnalytics {
+
+    static let started = "paypal-messaging:create-view:started"
+    static let failed = "paypal-messaging:create-view:failed"
+    static let succeeded = "paypal-messaging:create-view:succeeded"
+}
```
jaxdesmarais, line 8: The PayPal Messages team collects their own analytics so we can decide what we want analytics for. I opted to keep things super simple for now, but if we feel that other events are important to capture we can add them. We can additionally add more events in the future if we decide that we are not getting enough data from these.

sarahkoop (approved):

hollabaq86 (commented):
Agree, let's keep it simple and build out from there.

Do we have test coverage for events being sent?

Sources/BraintreePayPalMessaging/BTPayPalMessagingAnalytics.swift
scannillo: Can we add these to our events excel doc + description?

scannillo: Maybe with a note that they're "to be released"?

jaxdesmarais: Added!

jaxdesmarais:
> Do we have test coverage for events being sent?

There is a follow up ticket as a whole for testing this feature. Since this is going into a feature branch we wanted to keep the PRs small/digestible. But there will be tests - that will be the next PR once the current 3 are merged in. 😄

scannillo (approved):

hollabaq86:
@jaxdesmarais ahhhh I see. Carry on!

sshropshire (approved):

PR #1170, opened 2024-01-10
Add PayPalCheckout to Demo App - Required to Run on Device

jaxdesmarais:
### Summary of changes

- Right now the Demo app can't run on device because we don't have PayPal Checkout pulled in as a linked framework - the demo app builds as expected on simulator
- Have confirmed with this change that the Demo app no longer crashes when running on device

I spent some time exploring alternatives (using SPM or Cocoapods) for the demo app. The major downside to not using a "direct/carthage" integration is that we would need to be tied to a branch name or version vs linking the library directly like we do today. This would mean when adding features we'd need to change the branch to add the feature to our Demo app, then change it back to main/a specific version post merge in a separate PR.

IMO having to switch the branch around regularly seems like a pain as well as something that could easily get missed and our Demo app would be building on an old SDK version for some time if we forget to open the subsequent PR. The "best" path forward is to continue to link PayPal Checkout via SPM in the demo app as we did previously. The obvious con is that we have removed and added this again multiple times because it is hard to tell running on simulator that it's necessary.

Open question: is there a better way to add "building on device" to our development process or is there a way to automate a on device build via CI in the future? This could potentially eliminate this issue but would be a separate exploration on feasibility.

### Checklist

### Authors

- @jaxdesmarais


scannillo (approved):
➕1 - would love to incorporate on-device builds into our CI.

We used to have a BrowserStack subscription we can follow up on. Also I've heard good things about Bitrise & Appium.

richherrera (approved):

PR #1171, opened 2024-01-10
[ShopperInsights] Add EligiblePayments API Call

scannillo:
### Summary of changes

- Update `BTAPIClient` & `BTHTTP` to support API calls to paypal.com APIs
- Add `BTEligiblePaymentsRequest` struct to house POST body for `v2/payments/find-eligible-methods` API call
- Add relevant unit tests
- Leave TODO for parsing result of API call

### Notes
- API call still expected to 403 right now ⚠️ 
- Adding support for hitting a PayPal API reveals how tightly coupled our networking implementation is to BT API specifics
    - A few highlights:
       - `BTHTTP` & `BTGraphQLHTTP` have a lot of overlap & duplication (3 different `URLSession` configurations, 2 different `URLRequest` creation sites)
       - No service-agnostic networking interface
    - When time permits, I think there's room to rework things similar to how we did in a [series of PPCP PRs](https://github.com/paypal/paypal-ios/pulls?q=is%3Apr+networking+label%3Anetworking-refactor)
    - Especially if we want to make BraintreeCore the potential unified dependency for networking & analytics b/w PPCP & BT 👀 

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -181,7 +181,7 @@ import Foundation
                 }
                 
                 if payPalHTTP == nil {
-                    let paypalBaseURL: URL? = paypalAPIURL(forEnvironment: configuration?.environment ?? "")
+                    let paypalBaseURL: URL? = payPalAPIURL(forEnvironment: configuration?.environment ?? "")
                     
                     if let clientToken, let paypalBaseURL {
                         payPalHTTP = BTHTTP(url: paypalBaseURL, authorizationFingerprint: clientToken.authorizationFingerprint)
@@ -481,7 +481,7 @@ import Foundation
         return components.url
     }
     
-    func paypalAPIURL(forEnvironment environment: String) -> URL? {
+    func payPalAPIURL(forEnvironment environment: String) -> URL? {
         if environment.caseInsensitiveCompare("sandbox") == .orderedSame ||
             environment.caseInsensitiveCompare("development") == .orderedSame {
             return URL(string: "https://api-m.sandbox.paypal.com")
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -26,7 +26,7 @@ import Foundation
 
     var http: BTHTTP?
     var graphQLHTTP: BTGraphQLHTTP?
-    var paypalHTTP: BTHTTP?
+    var payPalHTTP: BTHTTP?
     
     var session: URLSession {
         let configurationQueue: OperationQueue = OperationQueue()
@@ -180,11 +180,11 @@ import Foundation
                     }
                 }
                 
-                if paypalHTTP == nil {
+                if payPalHTTP == nil {
                     let paypalBaseURL: URL? = paypalAPIURL(forEnvironment: configuration?.environment ?? "")
                     
                     if let clientToken, let paypalBaseURL {
-                        paypalHTTP = BTHTTP(url: paypalBaseURL, authorizationFingerprint: clientToken.authorizationFingerprint)
+                        payPalHTTP = BTHTTP(url: paypalBaseURL, authorizationFingerprint: clientToken.authorizationFingerprint)
                     }
                 }
             }
@@ -366,7 +366,7 @@ import Foundation
             return parameters?.merging(["_meta": metadata.parameters]) { $1 }
         case .graphQLAPI:
             return parameters?.merging(["clientSdkMetadata": metadata.parameters]) { $1 }
-        case .paypalAPI:
+        case .payPalAPI:
             return parameters
         }
     }
@@ -496,8 +496,8 @@ import Foundation
             return http
         case .graphQLAPI:
             return graphQLHTTP
-        case .paypalAPI:
-            return paypalHTTP
+        case .payPalAPI:
+            return payPalHTTP
         }
     }
 }
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -26,7 +26,8 @@ import Foundation
 
     var http: BTHTTP?
     var graphQLHTTP: BTGraphQLHTTP?
-
+    var paypalHTTP: BTHTTP?
+    
     var session: URLSession {
         let configurationQueue: OperationQueue = OperationQueue()
         configurationQueue.name = "com.braintreepayments.BTAPIClient"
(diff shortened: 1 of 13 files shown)


UnitTests/BraintreeShopperInsightsTests/BTShopperInsightsClient_Tests.swift
```diff
@@ -41,6 +43,37 @@ class BTShopperInsightsClient_Tests: XCTestCase {
...
+        
+        XCTAssertEqual(mockAPIClient.lastPOSTPath, "/v2/payments/find-eligible-methods")
+        XCTAssertEqual(mockAPIClient.lastPOSTAPIClientHTTPType, .paypalAPI)
+        
+        guard let lastPostParameters = mockAPIClient.lastPOSTParameters else {
+            XCTFail()
+            return
+        }
+        
+        let customer = lastPostParameters["customer"] as! [String: Any]
```
scannillo, line 57: 🏈 Would love for us to implement a cleaner way to verify JSON structure across our SDKs test suite

Sources/BraintreeShopperInsights/BTEligiblePaymentsRequest.swift
```diff
@@ -0,0 +1,76 @@
...
+            
+            enum CodingKeys: String, CodingKey {
+                case merchantID = "merchant_id"
+            }
+        }
+    }
+    
+    struct Preferences: Encodable {
+        let includeAccountDetails = true
+        let includeVaultTokens = true
```
MaxHastingsPP, line 52: I plan to remove `include_vault_tokens` on my PR as you mentioned it is not needed. And from testing I still got back Venmo and PayPal recommendations.

UnitTests/BraintreeShopperInsightsTests/BTShopperInsightsClient_Tests.swift
saperi22: On Android, we use https://github.com/skyscreamer/JSONassert.
Maybe you could try using https://github.com/iwill/generic-json-swift on iOS?

Sources/BraintreeShopperInsights/BTEligiblePaymentsRequest.swift
scannillo: Removed in 35cf4f54323c87ea86ba9cab1f0ea5b58cc1bbe5!

UnitTests/BraintreeShopperInsightsTests/BTShopperInsightsClient_Tests.swift
scannillo: 👀 Thanks for those resources! Instead of doing it in this PR, I was thinking we could explore some options and once we land on one we like update all of our JSON related tests to use it. I can throw this in the backlog

Sources/BraintreeCore/URL+IsPayPal.swift
```diff
@@ -4,6 +4,6 @@ extension URL {
     
     /// Used to determine if a the URL is a paypal.com domain in order to format API requests accordingly
```
agedd, line 5: small nit:
```suggestion
    /// Used to determine if the URL is a paypal.com domain in order to format API requests accordingly
```

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -26,7 +26,8 @@ import Foundation
 
     var http: BTHTTP?
     var graphQLHTTP: BTGraphQLHTTP?
-
+    var paypalHTTP: BTHTTP?
```
jaxdesmarais, line 29: super nit, but I think at one point we decided to case PayPal as `payPal` across the SDK. Do we want to make that update across this PR?

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -178,6 +179,14 @@ import Foundation
                         graphQLHTTP = BTGraphQLHTTP(url: graphQLBaseURL, tokenizationKey: tokenizationKey)
                     }
                 }
+                
+                if paypalHTTP == nil {
+                    let paypalBaseURL: URL? = paypalAPIURL(forEnvironment: configuration?.environment ?? "")
```
jaxdesmarais, line 184: Do we want to make this method return a non-optional URL? Based on this method it seems like the default is the prod URL. Based on the defaulting behavior it may also make sense to make the environment optional instead of passing an empty string here if we don't get one from the config. Thoughts?

scannillo:
Marking this as "Blocked" since there is talks that the BT GW will wrap this API call, versus us having to hit the PP API directly. There is also talks about moving this feature entirely to the server-side.

Sources/BraintreeCore/BTAPIClient.swift
scannillo: Ah - good ❓. I was following the existing pattern that we have for the GraphQL API implementation ([see here](https://github.com/braintree/braintree_ios/blob/c90a487da1b93fb0790c1fd6cca61267fb90b42b/Sources/BraintreeCore/BTAPIClient.swift#L173)).

I agree though, it's definitely funky.

Wrote this in the PR notes a bit, but this PR revealed how tightly coupled our networking layer is to BT APIs. Even the GraphQL stuff feels very tacked-on. I think we could benefit from a re-design, but wanted to 🏈 on it.

Sources/BraintreeCore/BTAPIClient.swift
scannillo: Good catch, fixed in d944ed6fb987650ac4984f24d0dcfd9ccb6a56a2

scannillo:
Removing blocked label, as the API team should be updating this SDK to work with default test values in sandbox tomorrow.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -469,13 +480,24 @@ import Foundation
         components.path = "/graphql"
         return components.url
     }
+    
+    func paypalAPIURL(forEnvironment environment: String) -> URL? {
```
jaxdesmarais, line 484:
```suggestion
    func payPalAPIURL(forEnvironment environment: String) -> URL? {
```

UnitTests/BraintreeCoreTests/BTAPIClient_Tests.swift
```diff
@@ -585,4 +585,20 @@ class BTAPIClient_Tests: XCTestCase {
...
+        let apiClientDev = BTAPIClient(authorization: "development_tokenization_key")
+        let baseURLDev: URL? = apiClientDev?.paypalAPIURL(forEnvironment: "development")
+        XCTAssertEqual(baseURLDev?.absoluteString, "https://api-m.sandbox.paypal.com")
+    }
+    
+    func testPayPalBaseURLForEnvironment_returnsProductionURL_asDefault() {
+        let apiClient = BTAPIClient(authorization: "development_tokenization_key")
+        let baseURL: URL? = apiClient?.paypalAPIURL(forEnvironment: "unknown")
+        XCTAssertEqual(baseURL?.absoluteString, "https://api-m.paypal.com")
+    }
```
jaxdesmarais, line 603: Do we also want a test for passing production as the env returns the production URL? Since it's the default may be covered above but may also be helpful to have a test for. What do you think?

Sources/BraintreeCore/BTAPIClient.swift
scannillo: Good catch, fixed in 335b594f5f61bfec610f8897c3fa96aa708c92f1

UnitTests/BraintreeCoreTests/BTAPIClient_Tests.swift
scannillo: Yes! Addressed in 335b594f5f61bfec610f8897c3fa96aa708c92f1

MaxHastingsPP (approved):
lgtm

jaxdesmarais (approved):
🚀

PR #1173, opened 2024-01-12
Update SEPA Requests to `Encodable`

jaxdesmarais:
### Summary of changes

- Add encodable types for `post` bodies used in the SEPA Direct Debit flow:
    - Add `SEPADebitAccountsRequest` encodable type
    - Add `SEPADebitRequest` encodable type
- The SEPA flow has been tested in the Demo app to ensure the dictionaries are still constructed as expected
- Unit tests for these types are in `SEPADirectDebitAPI_Tests`

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
+++ b/Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
@@ -1,16 +1,16 @@
 import Foundation
 
 /// The POST body for `v1/payment_methods/sepa_debit_accounts`
-struct BTSEPADebitAccountsRequest: Encodable {
+struct SEPADebitAccountsRequest: Encodable {
 
-    private let createMandateResult: BTCreateMandateResult
+    private let createMandateResult: SEPADebitAccountRequest
 
     enum CodingKeys: String, CodingKey {
         case createMandateResult = "sepa_debit_account"
     }
 
     init(createMandateResult: CreateMandateResult) {
-        self.createMandateResult = BTCreateMandateResult(
+        self.createMandateResult = SEPADebitAccountRequest(
             last4: createMandateResult.ibanLastFour,
             merchantOrPartnerCustomerID: createMandateResult.customerID,
             bankReferenceToken: createMandateResult.bankReferenceToken,
@@ -18,7 +18,7 @@ struct BTSEPADebitAccountsRequest: Encodable {
         )
     }
 
-    struct BTCreateMandateResult: Encodable {
+    struct SEPADebitAccountRequest: Encodable {
 
         let last4: String?
         let merchantOrPartnerCustomerID: String?
--- a/Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
+++ b/Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
@@ -9,15 +9,6 @@ struct SEPADebitAccountsRequest: Encodable {
         case createMandateResult = "sepa_debit_account"
     }
 
-    init(createMandateResult: CreateMandateResult) {
-        self.createMandateResult = SEPADebitAccountRequest(
-            last4: createMandateResult.ibanLastFour,
-            merchantOrPartnerCustomerID: createMandateResult.customerID,
-            bankReferenceToken: createMandateResult.bankReferenceToken,
-            mandateType: createMandateResult.mandateType
-        )
-    }
-
     struct SEPADebitAccountRequest: Encodable {
 
         let last4: String?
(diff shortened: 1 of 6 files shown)


Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+
+/// The POST body for `v1/payment_methods/sepa_debit_accounts`
+struct SEPADebitAccountsRequest: Encodable {
```
scannillo, line 4: ❓ - do we want to create a `/Models` directory within each `Sources/<PaymentMethod>` directory? Specifically to house these new classes specific to encoding (and maybe eventually to house ones specific to decoding if we go that route).

Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
```diff
@@ -0,0 +1,35 @@
...
+
+    private let createMandateResult: SEPADebitAccountRequest
+
+    enum CodingKeys: String, CodingKey {
+        case createMandateResult = "sepa_debit_account"
+    }
+
+    init(createMandateResult: CreateMandateResult) {
```
scannillo, line 12: Nit - I would move the init to the bottom of this class.

Maybe these new Encodable structs is a good candidate for us to add to our style guide how we should organize these, MARKs, etc

Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
jaxdesmarais: Yeah, I think that would be helpful. I was thinking that as I was adding this. Should we do that as a follow up PR for all of the models we've created so far? I think it changes the source paths for the package managers (at least Cocoapods). Also down to do that for SEPA in this PR and just a follow up for the other 2. Thoughts?

scannillo: I like the idea of having those be 1 PR! But your call

Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
jaxdesmarais: I had followed the pattern we established in [BTApplePaymentTokensRequest](https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreeApplePay/BTApplePaymentTokensRequest.swift) where the nested struct is under the `init`. Though I do think having structs within structs could be making our preferred style more convoluted. Thoughts on removing the nesting?

Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
```diff
@@ -0,0 +1,35 @@
...
+            mandateType: createMandateResult.mandateType
+        )
+    }
+
+    struct SEPADebitAccountRequest: Encodable {
+
+        let last4: String?
+        let merchantOrPartnerCustomerID: String?
+        let bankReferenceToken: String?
+        let mandateType: String?
```
agedd, line 26: nit(entirely optional): maybe we can add short in-line docs to describe each property within these structs

agedd (approved):
lgtm! just left one small comment

Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
jaxdesmarais: Cool - agree we can 🏈 on it in this PR to establish a pattern in a single PR moving forward

Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
jaxdesmarais: > Yeh, looking back I think having all the struct definitions be near the property that uses it & pushing the init down is more readable vs having them intertwined. What do you think?

Yeah I am down if we feel like it makes things more readable! I pushed a commit here with the change: https://github.com/braintree/braintree_ios/pull/1173/commits/d479ab91ac3697cc7b8c32461d0250eb416470ab

Also agree on style guiding whatever we decide (and finding a home for the style guide).

> In terms of nested structs, I think they sometimes help readability to know when something is defined within the context of something else. For example with the ApplePaymentToken one, that name is pretty generic, but namespacing it the <Request>.ApplePaymentToken helps reduce name collisions.

That makes sense with naming collisions, I forgot about that aspect. Especially with something like `card` that could be a situation where a merchant has something similar by the same name.

warmkesselj (approved):

Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift
scannillo: I added a backlog ticket for updating our style guide - DTBTSDK-3373

scannillo (approved):

PR #1175, opened 2024-01-17
Drop -m suffix from `api.paypal.com` URL

scannillo:
### Background

- The `-m` suffix to `api.paypal.com` was added by suggestion from Randhir after an office hours session
   - When following up for more information on what this `-m` modifier means or does, we have not been able to get information from the API teams or Confluence
   - The FPTI API docs (go/fpti) also have no mention of this URL
- Aligns w/ Arachne & PPCP SDKs, which don't append `-m`
- Prepare for PaymentReady API which only works without the `-m` suffix

### Changes

- Move PayPal API strings into shared constants file
- Remove `-m` suffix from PayPal API URLs

### Notes

- Verified that events sent still appear in FPTI tables

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


scannillo:
Going to give the API teams 1-2 more days to raise concern with this before marking as "Ready for Review", but highly unlikely they will.

jaxdesmarais (approved):
🔥

sarahkoop (approved):

agedd (approved):
🚀

scannillo:
Several folks who work on PayPal APIs never heard of `-m`, so going 🟢 🚥 with removing it

sshropshire:
> Several folks who work on PayPal APIs never heard of `-m`, so going 🟢 🚥 with removing it

This is a good call! It might actually improve latency `api-m.paypal.com` appears to be an alias for `api.paypal.com`. Better to go for the target domain directly 🙌 .

PR #1177, opened 2024-01-22
[ShopperInsights] Add Analytics Events

agedd:
### Summary of changes

- Added 3 SDK-triggered analytics events to track Payment Eligible API flows (aligned across JS/Android as well)
    - shopper-insights:get-recommended-payments:started
    - shopper-insights:get-recommended-payments:failed
    - shopper-insights:get-recommended-payments:succeeded
- Added appropriate unit tests 

Related Ticket: [DTBTSDK-3408](https://paypal.atlassian.net/browse/DTBTSDK-3408)

### Checklist

- ~Added a changelog entry~

- @scannillo @agedd


Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -45,7 +49,14 @@ public class BTShopperInsightsClient {
...
                 httpType: .payPalAPI
-            ) { json, response, error in
+            ) { json, _, error in
+                Task {
+                    if let error {
+                        self.notifyFailure(with: error)
+                        throw error
+                    }
+                }
+                self.notifySuccess(with: BTShopperInsightsResult())
```
scannillo, line 59: I know this is just a default case until we start messing with the API default responses, but in this case we want to make sure we fire an analytic + give the merchant back a result

```
let result = BTShopperInsightsResult()
self.notifySuccess(with: BTShopperInsightsResult())
return result
```

scannillo: This _might_ run into some async-await error though, we'll have to mess with the Task maybe.

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -31,8 +31,12 @@ public class BTShopperInsightsClient {
     /// - Returns: A `BTShopperInsightsResult` instance
     /// - Note: This feature is in beta. It's public API may change or be removed in future releases.
     public func getRecommendedPaymentMethods(request: BTShopperInsightsRequest) async throws -> BTShopperInsightsResult {
+        apiClient.sendAnalyticsEvent(BTShopperInsightsAnalytics.recommendedPaymentsStarted)
+        
         if isVenmoAppInstalled() && isPayPalAppInstalled() {
-            return BTShopperInsightsResult(isPayPalRecommended: true, isVenmoRecommended: true)
+            let result = BTShopperInsightsResult(isPayPalRecommended: true, isVenmoRecommended: true)
+            notifySuccess(with: result)
+            return result
```
scannillo, line 39: For folks reviewing, the other pattern Alekhya & I discussed looked something like:

```
// Callsite
return notifySuccess(with: result)

// Method definition
func notifySuccess(with result: BTShopperInsightsResult) -> BTShopperInsightsResult {
    apiClient.sendAnalyticEvent(...)
    return result
}
```

scannillo: Since we're using an async-await method here, we can't follow the exact pattern we've established for analytic helper methods which pass a completion around (see [here](https://github.com/braintree/braintree_ios/blob/f11dccb5b9f9f7ee67a8de8f60262ff290a4604d/Sources/BraintreeCard/BTCardClient.swift#L206-L220)).

What option do folks find more clear?

UnitTests/BraintreeShopperInsightsTests/BTShopperInsightsClient_Tests.swift
```diff
@@ -28,6 +28,8 @@ class BTShopperInsightsClient_Tests: XCTestCase {
         
         XCTAssertNotNil(result!.isPayPalRecommended)
         XCTAssertNotNil(result!.isVenmoRecommended)
+        XCTAssertEqual(mockAPIClient.postedAnalyticsEvents[mockAPIClient.postedAnalyticsEvents.count-2], "shopper-insights:get-recommended-payments:started")
```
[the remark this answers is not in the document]
agedd: yep it is, addressed in 849b17c014f735ea87a99d3947dc6f8f24b6112f

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
warmkesselj: I like the above proposed in this comment thread.

jaxdesmarais: I assume we can make a similar thow-ing method for notify failure as well, but yeah I like what you proposed Sammy and think it's most similar to what we do today.

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -31,8 +31,12 @@ public class BTShopperInsightsClient {
     /// - Returns: A `BTShopperInsightsResult` instance
     /// - Note: This feature is in beta. It's public API may change or be removed in future releases.
     public func getRecommendedPaymentMethods(request: BTShopperInsightsRequest) async throws -> BTShopperInsightsResult {
+        apiClient.sendAnalyticsEvent(BTShopperInsightsAnalytics.recommendedPaymentsStarted)
```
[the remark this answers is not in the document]
agedd: yeah that's a great callout - im fine with either but feel more inclined to leave as is since it seems the other clients (i.e. `BTCardClient`) define analytic helper methods for only success, failure and cancel and for start events, they fire it like above.

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
agedd: ahhh gotcha, addressed in a29756160275030f202b97edc5f4f2bc276d920b

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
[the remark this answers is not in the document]
jaxdesmarais: We only call started 1 time per flow so I think we can continue to leave it without extracting it into its own method as @agedd stated.

Braintree.xcodeproj/project.pbxproj
```diff
@@ -2967,6 +2970,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				80B59A132B27C7FC004C2FA3 /* BTShopperInsightsClient_Tests.swift in Sources */,
+				62970D102B5AF2C000BAB584 /* BTShopperInsightsAnalytics_Tests.swift in Sources */,
```
jaxdesmarais, line 2973: Super nitpick, but we try to always sort the files by name. Can we bump this above the client tests? Should be able to right click the directory then click "sort by name"

agedd: yep absolutely, addressed in 5c16b60b274c998158563c2f6a4080e2aa991891

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
agedd: thanks for the suggestions y'all, addressed in 5c16b60b274c998158563c2f6a4080e2aa991891

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -88,4 +100,15 @@ public class BTShopperInsightsClient {
...
+    // MARK: - Analytics Helper Methods
+    
+    private func notifySuccess(with result: BTShopperInsightsResult) -> BTShopperInsightsResult {
+        apiClient.sendAnalyticsEvent(BTShopperInsightsAnalytics.recommendedPaymentsSucceeded)
+        return result
+    }
+    
+    private func notifyFailure(with error: Error) {
+        apiClient.sendAnalyticsEvent(BTShopperInsightsAnalytics.recommendedPaymentsFailed, errorDescription: error.localizedDescription)
+    }
```
jaxdesmarais, line 113:
```suggestion
    private func notifyFailure(with error: Error) throws {
        apiClient.sendAnalyticsEvent(BTShopperInsightsAnalytics.recommendedPaymentsFailed, errorDescription: error.localizedDescription)
        throw error
    }
```

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -45,7 +48,16 @@ public class BTShopperInsightsClient {
...
-            ) { json, response, error in
+            ) { json, _, error in
+                Task {
+                    if let error {
+                        self.notifyFailure(with: error)
+                        throw error
+                    }
+                    let result = BTShopperInsightsResult()
+                    self.notifySuccess(with: result)
+                    return result
```
jaxdesmarais, line 59:
```suggestion
                    return self.notifySuccess(with: result)
```

jaxdesmarais (approved):
👏

scannillo (approved):

PR #1178, opened 2024-01-22
[ShopperInsights] Add Email & Phone Text Fields to Demo App

agedd:
### Summary of changes

- This PR adds 3 text fields to the ShopperInsights Demo Feature - email, country code, national number. 
- The entered values will be used to construct a `BTShopperInsightsRequest` and depending upon the result of the payment eligibility call, the eligible payment method buttons will be enabled. As of now, this functionality is disabled since the API is yet to be handled (in a separate future PR).
- Related Ticket: DTBTSDK-3375

### Visual
shopper-insights-demo

### Checklist

- ~Added a changelog entry~

- @agedd


Demo/Application/Features/Helpers/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,54 @@
...
+//  UIViewController+Extensions.swift
+//  Demo
+//
+//  Created by Alekhya Geddam on 1/19/24.
+//  Copyright © 2024 braintree. All rights reserved.
+//
+
+import UIKit
+
+extension UIViewController {
```
jaxdesmarais, line 11: Would it be possible to combine both the label and text field together in a stack view? Then we can create 1 object with everything we need in the view controller. Thinking something like this could work but you'll probably need to make some adjustments:
```swift
    func textFieldWithLabel(labelText: String, placeholder: String? = nil, textFieldText: String? = nil) -> UIView {
        let label = UILabel()
        label.font = .boldSystemFont(ofSize: 15)
        label.text = labelText
        label.lineBreakMode = .byWordWrapping
        label.numberOfLines = 0

        let textField = UITextField()
        textField.translatesAutoresizingMaskIntoConstraints = false
        textField.borderStyle = .roundedRect
        textField.clearButtonMode = .never
        textField.placeholder = placeholder
        textField.text = textFieldText

        let stackView = UIStackView(arrangedSubviews: [label, textField])
        stackView.axis = .vertical
        stackView.alignment = .center
        stackView.spacing = 10
        stackView.distribution = .fillEqually
        stackView.translatesAutoresizingMaskIntoConstraints = false

        return stackView
    }
```

What do you think?

agedd: yea i like the idea! the only concern i'd have tho is how to read the `text` value of the textfield if it's embedded within a subview - assuming  `var emailView = textFieldWithLabel(labelText: "Email", placeholder: "Email")`, i think it'd be something like this: `emailView.subviews.compactMap{ $0 as? UITextField }.first?.text` which reads a bit odd to me. wdyt?

Demo/Application/Features/Helpers/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,54 @@
...
+        textField.borderStyle = .roundedRect
+        textField.placeholder = placeholder
+        textField.text = text
+        textField.clearButtonMode = clearButton
+        textField.autocapitalizationType = .none
+        textField.autocorrectionType = .no
+        return textField
+    }
+    
+    func button(
```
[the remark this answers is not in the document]
agedd: ahh gotcha, addressed in 6a9ce07a7aef5ee60161aa0622bd11dea0a6bb2c

Demo/Application/Features/Helpers/UIViewController+Extensions.swift
jaxdesmarais: 🤔 yeah that could get tricky for sure especially with multiple views using the same subview. We could also take the approach of `BTCardFormView` of having a dedicated class for this as a whole. That would allow us to have properties for each text field we need without needing to do a bunch of weird stuff under the hood. We don't have any other text fields in the SDK so we have some flexibility.

agedd: ahh yep! created a separate class to handle the text field + label creation here 5ff7ab017c44668bf8f4c3a018448ec1c721ff68

Demo/Application/Features/Helpers/UIViewController+Extensions.swift
```diff
@@ -1,29 +0,0 @@
-import UIKit
-
-extension UIViewController {
```
scannillo: Weird this file is showing as deleted when it doesn't exist on the `payment-insights-feature` branch

scannillo (approved):

vradopp (approved):

PR #1180, opened 2024-01-24
[Shopper Insights] Fix return flow bug

scannillo:
### Summary 

- When I ran our demo app, I realized there was an issue with [this logic](https://github.com/braintree/braintree_ios/blob/2d931f498e795d3b5f843a3182c6e73614a035c9/Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift#L52-L62) where line 62 would always execute before the `Task` had a chance to complete & even if you properly returned in all conditions within the `Task` the compiler still requires an exit out of the function outside of the `Task`
- The cleanest approach (from the `BTShopperInsightsClient` perspective) is to stick with all async-await syntax

### Changes
- Add an `async` version of `BTAPIClient.post()`
- Update `BTShopperInsightsClient` to use new `post()` async method
- Add unit tests
    - Added failing tests first to confirm (TDD for _life_)

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -347,6 +347,31 @@ import Foundation
     }
 
     /// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    ///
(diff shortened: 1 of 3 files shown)


jaxdesmarais (approved):

agedd (approved):
lgtm ~ thanks for these changes 🚀

PR #1181, opened 2024-01-25
Add `paypal_context_id` FPTI Tag

jaxdesmarais:
### Summary of changes

- Pass `paypal_context_id` to FPTI when available:
    - PayPal Web: passing either the `token` (checkout) or `ba_token` (vault) as the `paypal_context_id`
    - PayPal Native: passing the `orderID` as the `paypal_context_id`
    - Local Payments: passing the `paymentToken` (which is an order ID) as the `paypal_context_id`
    - If we get an empty string back we will not pass an empty string to FPTI
- I have verified that we are passing this value as expected in the above flows to FPTI
- Add Unit Tests

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeCore/Analytics/FPTIBatchData.swift
+++ b/Sources/BraintreeCore/Analytics/FPTIBatchData.swift
@@ -97,7 +97,9 @@ struct FPTIBatchData: Codable {
         let merchantAppVersion: String = Bundle.main.infoDictionary?[kCFBundleVersionKey as String] as? String ?? "N/A"
 
         let merchantID: String?
-        
+
+        let payPalContextID: String?
+
         let platform = "iOS"
 
         let sessionID: String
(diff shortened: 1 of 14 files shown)


sarahkoop (approved):

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -97,7 +97,9 @@ struct FPTIBatchData: Codable {
         let merchantAppVersion: String = Bundle.main.infoDictionary?[kCFBundleVersionKey as String] as? String ?? "N/A"
 
         let merchantID: String?
-        
+
+        let payPalContextID: String?
```
scannillo, line 101: Take it or leave it comment - adding a docstring to this property with some of the description for the different values this can represent; (ecToken, orderID, etc).

scannillo: Might help a new person understanding what this represents, or future us

scannillo (approved):

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
jaxdesmarais: Good call! I also added a docstring to each of the clients describing what the value is for that particular client: https://github.com/braintree/braintree_ios/pull/1181/commits/e7c3fe0b513a861d6899cbab6aeee05ba077b3ae

PR #1183, opened 2024-01-26
[DRAFT] Remove Response from HTTP Requests

jaxdesmarais:
### Summary of changes

- Remove `HTTPURLResponse` from all HTTP requests
- Update Tests

#### Additional Details:
This was only used in 1 place in the entire codebase when checking the `statusCode` of a `configuration` object. This code was added ~9 years ago and has just always existed as a check in the SDK despite this being the only place it's used. 

Instead, we can check that both body and error are not nil. This is in line with what Android does as well ([code here](https://github.com/braintree/braintree_android/blob/13140a52d349533d1285d7c0a0257852c3f1cd57/BraintreeCore/src/main/java/com/braintreepayments/api/ConfigurationLoader.kt#L35-L48)). 

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -6,7 +6,7 @@ import Foundation
 
     /// :nodoc: This typealias is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     @_documentation(visibility: private)
-    public typealias RequestCompletion = (BTJSON?, HTTPURLResponse?, Error?) -> Void
+    public typealias RequestCompletion = (BTJSON?, Error?) -> Void
 
     // MARK: - Public Properties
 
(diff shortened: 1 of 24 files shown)


Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -144,39 +144,41 @@ import Foundation
...
             guard let self else {
                 completion(nil, BTAPIClientError.deallocated)
                 return
             }
 
-            if error != nil {
+            if let error {
                 completion(nil, error)
                 return
-            } else if response?.statusCode != 200 {
```
jaxdesmarais, line 156: Now we can just check that body is not nil, which matches the [Android implementation](https://github.com/braintree/braintree_android/blob/13140a52d349533d1285d7c0a0257852c3f1cd57/BraintreeCore/src/main/java/com/braintreepayments/api/ConfigurationLoader.kt#L35-L48). Seems like we were checking the status as a way to get a read on the body not being nil, instead we can just check the body (which we weren't doing before).

jaxdesmarais: Actually it seems like we are doing something slightly more complicated further up the chain so this isn't working the same way for us. Digging into it now.

jaxdesmarais: So this is bizarre `fetchPaymentMethodNonces` makes a `get` call, that subsequently calls `fetchOrReturnRemoteConfig` and the `get` called from `fetchPaymentMethodNonces` relies on body and error both being nil, but the response being 200. 

Our tests may be wrong for this but that seems odd since it's a large number of tests. Based on our tests (the ones failing in this PR) it seems like we actually expect to essentially return a `BTConfiguration` with `nil` JSON when making the call to `fetchPaymentMethodNonces` but in all other cases we expect either `body` or `error` to be non-nil.

If someone wants to tackle unwrapping this while I am on PTO feel free, otherwise I'll pick this back up next open dev.

jaxdesmarais:
Closing this PR for now - I've got a spike going locally to move this to a result object, but that's a larger undertaking. Will re-open a PR once that work is in a reviewable state.

PR #1184, opened 2024-01-29
[DTBTSDK 3415] Parse Shopper Insights API Response

agedd:
### Summary of changes

- Adds the parsing logic and returns the proper `BTShopperInsightsResult` to the merchant 
- Adds related unit test

### Checklist

- ~Added a changelog entry~

- @agedd


Demo/Application/Features/ShopperInsightsViewController.swift
```diff
@@ -31,10 +31,10 @@ class ShopperInsightsViewController: PaymentButtonBaseViewController {
         self.progressBlock("Fetching shopper insights...")
         
         let request = BTShopperInsightsRequest(
-            email: "my-email@gmail.com",
+            email: "PR1_merchantname@personal.example.com",
             phone: Phone(
                 countryCode: "1",
-                nationalNumber: "1234567"
+                nationalNumber: "4082321001"
```
agedd, line 37: note: added some dummy data for testing purposes but eventually these values will be replaced by the values entered in the shopper-insights demo feature

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -41,21 +41,37 @@ public class BTShopperInsightsClient {
             let postParameters = BTEligiblePaymentsRequest(
                 email: request.email,
                 phone: request.phone,
-                merchantID: "TODO-merchant-id-type"
+                merchantID: "MXSJ4F5BADVNS"
```
agedd, line 44: note: adding a sample merchant id for now, but still need to figure how to fetch this value (tbd)

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -41,21 +41,37 @@ public class BTShopperInsightsClient {
...
             )
             
             do {
-                let (_, _) = try await apiClient.post("/v2/payments/find-eligible-methods", parameters: postParameters, httpType: .payPalAPI)
-                
-                // TODO: - Handle API Response. DTBTSDK-3388
-                let result = BTShopperInsightsResult()
+                let (json,_) = try await apiClient.post("/v2/payments/find-eligible-methods", parameters: postParameters, httpType: .payPalAPI)
+                let eligibleMethodsJSON: BTJSON = json?["eligible_methods"] ?? BTJSON()
```
scannillo, line 49: Also, similar to my comment [here](https://github.com/braintree/braintree_android/pull/894/files#r1471496304) in Android, we should align on what we do when the API response doesn't error (200), but the response body isn't what we expect. 

I think we should surface a new merchant error in that case, similar to [BTSEPADirectDebitError.noBodyReturned](https://github.com/braintree/braintree_ios/blob/075220827e2df164c8230a8ed43b81dba7a402d3/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitError.swift#L24-L25). We could even add more granular error cases if we see fit

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -41,21 +41,37 @@ public class BTShopperInsightsClient {
...
                 throw self.notifyFailure(with: error)
             }
         }
     }
     
+    /// This method determines whether a payment source is recommended
+    /// - Parameters:
+    ///    - paymentMethodDetail: a `BTEligiblePaymentMethodDetails` containing the payment source's information
+    /// - Returns: `true` if both `eligibleInPPNetwork` and `recommended` are enabled, otherwise returns false.
+    /// - Note: This feature is in beta. It's public API may change or be removed in future releases.        -
```
scannillo, line 66: 👏 This is so awesome and so well written! Ideally all of our functions are documented as well as this

Nit - since it's private we can remove that "beta" note.
```suggestion
    /// This method determines whether a payment source is recommended
    /// - Parameters:
    ///    - paymentMethodDetail: a `BTEligiblePaymentMethodDetails` containing the payment source's information
    /// - Returns: `true` if both `eligibleInPPNetwork` and `recommended` are enabled, otherwise returns false.
```

Sources/BraintreeShopperInsights/BTShopperInsightsError.swift
```diff
@@ -0,0 +1,22 @@
...
+    }
+    
+    public var errorCode: Int {
+        rawValue
+    }
+    
+    public var errorDescription: String? {
+        switch self {
+        case .emptyBodyReturned:
+            "An empty body was returned from the Payments Ready API during the request."
```
scannillo, line 19: Nit for internal consistency w/ API name
```suggestion
            "An empty body was returned from the Eligible Payments API during the request."
```

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -41,21 +41,39 @@ public class BTShopperInsightsClient {
...
-                merchantID: "TODO-merchant-id-type"
+                merchantID: "MXSJ4F5BADVNS"
             )
             
             do {
-                let (_, _) = try await apiClient.post("/v2/payments/find-eligible-methods", parameters: postParameters, httpType: .payPalAPI)
-                
-                // TODO: - Handle API Response. DTBTSDK-3388
-                let result = BTShopperInsightsResult()
+                let (json,_) = try await apiClient.post("/v2/payments/find-eligible-methods", parameters: postParameters, httpType: .payPalAPI)
```
scannillo, line 48:
```suggestion
                let (json, _) = try await apiClient.post("/v2/payments/find-eligible-methods", parameters: postParameters, httpType: .payPalAPI)
```

warmkesselj (approved):

PR #1186, opened 2024-02-06
[ShopperInsights] Remove app installed checks

scannillo:
### Summary of changes

- Product decision wants us to remove the app installed checks and always call the EligiblePayments API

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
+++ b/Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
@@ -33,32 +33,27 @@ public class BTShopperInsightsClient {
     public func getRecommendedPaymentMethods(request: BTShopperInsightsRequest) async throws -> BTShopperInsightsResult {
         apiClient.sendAnalyticsEvent(BTShopperInsightsAnalytics.recommendedPaymentsStarted)
         
-        if isVenmoAppInstalled() && isPayPalAppInstalled() {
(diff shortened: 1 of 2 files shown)


agedd (approved):
🚀

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -33,32 +33,27 @@ public class BTShopperInsightsClient {
...
+        let postParameters = BTEligiblePaymentsRequest(
+            email: request.email,
+            phone: request.phone,
+            merchantID: "MXSJ4F5BADVNS"
+        )
+        
+        do {
+            let (json, _) = try await apiClient.post("/v2/payments/find-eligible-methods", parameters: postParameters, httpType: .payPalAPI)
+            guard let eligibleMethodsJSON = json?["eligible_methods"].asDictionary(),
+                  eligibleMethodsJSON.count != 0 else {
```
jaxdesmarais, line 46: I think we can use `!eligibleMethodsJSON.isEmpty` here instead if we want

jaxdesmarais (approved):
👻

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
scannillo: Hm - since `eligibleMethodsJSON` is an `NSDictionary`, we run into `Value of type 'NSDictionary' has no member 'isEmpty'`

PR #1187, opened 2024-02-07
[DTBTSDK-3403] Send PayPal-Client-Metadata-Id to PaymentReady API Call

agedd:
### Summary of changes

- Adds the `PayPal-Client-Metadata-Id` to the Shopper Insights API Call
- Adds appropriate unit test to ensure `PayPal-Client-Metadata-Id` is properly set to the `sessionID` 
- Adds the optional `headers` parameter to all relevant HTTP POST methods within various files
- Adds `lastPOSTAdditionalHeaders` property to `MockAPIClient` for testing purposes

### Checklist

- ~Added a changelog entry~

- @scannillo

### DIFF
--- a/Sources/BraintreeCore/BTHTTP.swift
+++ b/Sources/BraintreeCore/BTHTTP.swift
@@ -114,14 +114,14 @@ class BTHTTP: NSObject, NSCopying, URLSessionDelegate {
     }
 
     // TODO: - Remove when all POST bodies use Codable, instead of BTJSON/raw dictionaries
-    func post(_ path: String, parameters: [String: Any]? = nil, completion: @escaping RequestCompletion) {
+    func post(_ path: String, parameters: [String: Any]? = nil, headers: [String: String]? = nil, completion: @escaping RequestCompletion) {
         httpRequest(method: "POST", path: path, parameters: parameters, completion: completion)
     }
     
-    func post(_ path: String, parameters: Encodable, completion: @escaping RequestCompletion) {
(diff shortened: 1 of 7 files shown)


Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -283,6 +287,10 @@ class BTHTTP: NSObject, NSCopying, URLSessionDelegate {
                 headers["Client-Key"] = key
             }
         }
+        
+        if let additionalHeaders {
+            headers = headers.merging(additionalHeaders) { (_, newHeader) in newHeader}
```
jaxdesmarais, line 292: We should be able to use the following syntax here:
```suggestion
            headers = headers.merging(additionalHeaders) { $1 }
```

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -179,6 +179,7 @@ class BTHTTP: NSObject, NSCopying, URLSessionDelegate {
         method: String,
         path: String,
         parameters: [String: Any]? = [:],
+        headers: [String: String]? = [:],
```
jaxdesmarais, line 182: It looks like in other places we are defaulting to `nil`. Is there a reason we want to default to an empty dictionary for this method?

agedd: ahh good catch! yea there's no particular reason - updated it to `nil` instead

jaxdesmarais (approved):

scannillo (approved):

PR #1188, opened 2024-02-12
Accept email in `BTPayPalVaultRequest`

scannillo:
### Summary of changes

- Add `payerEmail` property to `BTPayPalRequest`
   - Format request accordingly for POST to `v1/paypal_hermes/setup_billing_agreement`
   - This is supposed to optimize the user's PwPP web checkout experience by pre-populating their email in the flow when possible, though I was unable to see any differences when testing in sand & prod

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalVaultRequest.swift
+++ b/Sources/BraintreePayPal/BTPayPalVaultRequest.swift
@@ -11,6 +11,9 @@ import BraintreeCore
 
     /// Optional: Offers PayPal Credit if the customer qualifies. Defaults to `false`.
     public var offerCredit: Bool
+    
+    /// Optional: User email to initiate a quicker authentication flow in cases where the user has a PayPal Account with the same email.
+    public var payerEmail: String?
 
     // MARK: - Initializer
 
@@ -30,8 +33,12 @@ import BraintreeCore
         let baseParameters = super.parameters(with: configuration)
         var vaultParameters: [String: Any] = ["offer_paypal_credit": offerCredit]
 
-        if billingAgreementDescription != nil {
+        if let billingAgreementDescription {
             vaultParameters["description"] = billingAgreementDescription
+        }
+        
+        if let payerEmail {
+            vaultParameters["payer_email"] = payerEmail
         }
 
         if let shippingAddressOverride {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## unreleased
 * Send `paypal_context_id` in `batch_params` to PayPal's analytics service (FPTI) when available
+* Add `BTPayPalRequest.payerEmail` optional property
 
 ## 6.12.0 (2024-01-18)
 * BraintreePayPal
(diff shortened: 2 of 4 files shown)


Sources/BraintreePayPal/BTPayPalVaultRequest.swift
```diff
@@ -11,6 +11,9 @@ import BraintreeCore
 
     /// Optional: Offers PayPal Credit if the customer qualifies. Defaults to `false`.
     public var offerCredit: Bool
+    
+    /// Optional: User email to initiate a quicker authentication flow in cases where the user has a PayPal Account with the same email.
+    public var payerEmail: String?
```
scannillo, line 16: We call this `userAuthenticationEmail` for the Native offering ([here](https://github.com/braintree/braintree_ios/blob/075220827e2df164c8230a8ed43b81dba7a402d3/Sources/BraintreePayPalNativeCheckout/BTPayPalNativeCheckoutRequest.swift#L16-L17)).

Do folks have a naming preference here?

jaxdesmarais: I feel like `userAuthenticationEmail` could make more sense vs `payerEmail`. The latter doesn't indicate as clearly that the email is used to attempt to authenticate quicker, imo.

Though looking through other modules, we don't tend to prefix payer/user/customer, for example `BTPayPalAccountNonce` [here](https://github.com/braintree/braintree_ios/blob/075220827e2df164c8230a8ed43b81dba7a402d3/Sources/BraintreePayPal/BTPayPalAccountNonce.swift#L10-L20) and `BTLocalPaymentRequest` [here](https://github.com/braintree/braintree_ios/blob/075220827e2df164c8230a8ed43b81dba7a402d3/Sources/BraintreeLocalPayment/BTLocalPaymentRequest.swift#L41-L51). So we _could_ go with just `authenticationEmail` potentially. What do you think?

tdchow: I think matching the naming from the Native PayPal module makes sense.

I feel having `payer` or `buyer` adds some clarity in the param name, but also good to align with the other properties we have that have it omitted.

scannillo: Updated to match the NXO module!

scannillo:
The `testSendAnalyticsEvent_whenNumberOfQueuedEventsMeetsThreshold_sendsAnalyticsEvent` test failure on CI seems unrelated. I am also getting this failure locally on `main`.

scannillo:
Waiting on final ✅ of approval to release this. E2E testing team wants to verify behavior first.

jaxdesmarais (approved):
🚀

agedd (approved):
lgtm! 🚀

scannillo:
Updating the base branch to target the ShopperInsights feature work. Passing this email will be part of the E2E testing teams verification before release.

PR #1193, opened 2024-02-22
Venmo Universal Links

jaxdesmarais:
### Summary of changes

- Add `BTVenmoRequest.fallbackToWeb`
    - If set to true customers will fallback to a web based Venmo flow if the Venmo app is not installed
- This feature has been added to the Demo app, select "Venmo Universal Links" in the Venmo Integration to test this flow
- Add Unit Tests

### Checklist

- [x] Added a changelog entry

- @jaxdesmarais


Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -220,7 +228,7 @@ import BraintreeCore
 
     /// Returns true if the proper Venmo app is installed and configured correctly, returns false otherwise.
     @objc public func isVenmoAppInstalled() -> Bool {
-        guard let appSwitchURL = BTVenmoAppSwitchRedirectURL().baseAppSwitchURL else {
+        guard let appSwitchURL = BTVenmoAppSwitchRedirectURL.baseAppSwitchURL else {
```
sshropshire, line 231: random: curious if we could leverage [#URL macros](https://www.avanderlee.com/swift/macros/#what-are-swift-macros) in iOS? It's new I'm not sure if there are minimum iOS version requirements, but it'd be cool if we could eliminate `if let` for the static URLs.

Sources/BraintreeVenmo/BTVenmoAppSwitchRedirectURL.swift
```diff
@@ -60,36 +46,48 @@ struct BTVenmoAppSwitchRedirectURL {
...
-        }
+    // MARK: - Internal Methods
+
+    func universalLinksURL() -> URL {
+        let universalLinkURL = URL(string: "https://venmo.com/go/checkout")!
+
+        var urlComponent = URLComponents(url: universalLinkURL, resolvingAgainstBaseURL: false)!
+        urlComponent.percentEncodedQuery = BTURLUtils.queryString(from: queryParameters as NSDictionary)
+
+        return urlComponent.url!
```
scannillo, line 71: Do we want to leave these force unwraps in? I remember adding these quick for spiking.

Though nothing in here will ever dynamically change, prob best practice to avoid

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -37,6 +40,10 @@ import BraintreeCore
     /// We require a static reference of the client to call `handleReturnURL` and return to the app.
     static var venmoClient: BTVenmoClient? = nil
 
+    /// Used for linking events from the client to server side request
+    /// In the PayPal flow this will be either an EC token or a Billing Agreement token
+    private var payPalContextID: String? = nil
```
scannillo, line 45: Maybe more clear to name this `paymentContextID` instead, and store it directly from line 176

Sources/BraintreeVenmo/BTVenmoAppSwitchRedirectURL.swift
jaxdesmarais: That's a good call, updated here: https://github.com/braintree/braintree_ios/pull/1193/commits/bb18d40c4543a606e03f1bffcac4913b6be97e50. I opted to unwrap it with a guard at the callsite before passing it into `startVenmoFlow`, but we could make the URL optional in `startVenmoFlow` and unwrap it there. I thought it made sense to do earlier in the flow to indicate which URL was invalid vs in `startVenmoFlow` when we may not know which of the two URLs is causing the error. Not opposed to moving it though if you have a strong preference.

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: I thought it may make it more clear why we are storing it by calling it `payPalContextID`, but am fine with either name. We aren't able to store it until after it's initialized by the guard, though. It's a bit goofy for sure but we'd get a compiler error. Thoughts?

Sources/BraintreeVenmo/BTVenmoAppSwitchRedirectURL.swift
scannillo: I like it!

UnitTests/BraintreeVenmoTests/BTVenmoRequest_Tests.swift
```diff
@@ -18,4 +18,21 @@ class BTVenmoRequest_Tests: XCTestCase {
...
+        XCTAssertEqual(request.fallbackToWeb, true)
+    }
+
+    func testFallbackToWeb_whenFalse_setsValueAsFalse() {
+        let request = BTVenmoRequest(paymentMethodUsage: .singleUse)
+        request.fallbackToWeb = false
+        XCTAssertEqual(request.fallbackToWeb, false)
+    }
+
+    func testFallbackToWeb_whenNotPassed_defaultsValueAsFalse() {
```
scannillo, line 34: Same comment here re: only needing the default test

Sources/BraintreeVenmo/BTVenmoAppSwitchRedirectURL.swift
```diff
@@ -60,36 +46,48 @@ struct BTVenmoAppSwitchRedirectURL {
...
+
+        var urlComponent = URLComponents(url: universalLinkURL, resolvingAgainstBaseURL: false)!
+        urlComponent.percentEncodedQuery = BTURLUtils.queryString(from: queryParameters as NSDictionary)
+
+        return urlComponent.url
+    }
 
-        var components = appSwitchBaseURLComponents()
-        components.percentEncodedQuery = BTURLUtils.queryString(from: appSwitchParameters as NSDictionary)
+    func appSwitchURL() -> URL? {
```
scannillo, line 74: Nit - can we call this `urlSchemeURL()` or something similar? "app switch" encompasses both linking types. Was a little confused when reading the tests real quick

scannillo:
Making a note that we'll want to also pass the `paymentContextID` for Android as well. We should prob make a separate ticket for that, since the work to pass `paypal_context_id` in our analytics payload isn't complete yet.

Sources/BraintreeVenmo/BTVenmoAppSwitchRedirectURL.swift
jaxdesmarais: For sure! Updated: https://github.com/braintree/braintree_ios/pull/1193/commits/9ca06b4307c684f323edcdad3421d21119f58796

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -370,7 +399,8 @@ import BraintreeCore
             throw BTVenmoError.disabled
         }
 
-        if !isVenmoAppInstalled() {
+
+        if !shouldFallback && !isVenmoAppInstalled() {
```
scannillo, line 403: Nit - it's a little cleaner IMO to pass the `shouldFallback` bool as a param to `verifyAppSwitch()` versus storing it as a property

jaxdesmarais:
We have completed our test cases for this flow based on current versions of the Venmo app. The Venmo team will be completing backwards compatibility testing off of this branch. Once their tests are complete as well we should be able to remove the "do not merge" tag.

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: That's a great callout, updated: https://github.com/braintree/braintree_ios/pull/1193/commits/d02955e587213989e0d059df2dd3834a461bbe6b

scannillo (approved):
🚀

jaxdesmarais:
Backwards compatibility testing is complete and the E2E testing team has verified these changes. 🥳

PR #1195, opened 2024-02-27
[DTBTSDK-3544] Remove merchant_id from the Payment Ready API Request

agedd:
### Summary of changes

- This PR removes the `merchantID` parameter from the Payment Ready API Request as this value is now determined by the auth fingerprint.

### Checklist

-~Added a changelog entry~

- @agedd

### DIFF
--- a/Sources/BraintreeShopperInsights/BTEligiblePaymentsRequest.swift
+++ b/Sources/BraintreeShopperInsights/BTEligiblePaymentsRequest.swift
@@ -25,7 +25,7 @@ struct BTEligiblePaymentsRequest: Encodable {
             case phone = "phone"
         }
     }
-    
+
     struct PurchaseUnit: Encodable {
         let amount = Amount()
         
--- a/Sources/BraintreeShopperInsights/BTEligiblePaymentsRequest.swift
+++ b/Sources/BraintreeShopperInsights/BTEligiblePaymentsRequest.swift
@@ -5,8 +5,8 @@ import PassKit
 struct BTEligiblePaymentsRequest: Encodable {
     
     private let customer: Customer
-    private let purchaseUnits = [PurchaseUnit()]
-    private let preferences = Preferences()
+    private let purchaseUnits: [PurchaseUnit]
+    private let preferences: Preferences
     
     enum CodingKeys: String, CodingKey {
         case customer = "customer"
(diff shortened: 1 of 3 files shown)


tdchow (approved):
Looks good! Thanks for making the change!

scannillo (approved):

Sources/BraintreeShopperInsights/BTEligiblePaymentsRequest.swift
```diff
@@ -5,7 +5,7 @@ import PassKit
 struct BTEligiblePaymentsRequest: Encodable {
     
     private let customer: Customer
-    private let purchaseUnits: [PurchaseUnit]
+    private let purchaseUnits = [PurchaseUnit()]
```
jaxdesmarais, line 8: Nit but I think we generally prefer this style for setting up the empty array:
```suggestion
    private let purchaseUnits: [PurchaseUnit] = []
```

agedd: ahhh i see, in this case was wanting to initialize and fill the array with the `PurchaseUnit` object inline since it is no longer initialized with the `merchantID` as [before](https://github.com/braintree/braintree_ios/blob/d57b773c54cd95bc3d73786baabe6059ee1cfc46/Sources/BraintreeShopperInsights/BTEligiblePaymentsRequest.swift#L72)

jaxdesmarais: Ah, gotcha. We can also consider moving that initialization down to the `init` and keep these constants consistent:
```swift
private let purchaseUnits: [PurchaseUnit]
private let preferences: Preferences
...
init(email: String?, phone: Phone?) {
    self.customer = Customer(email: email, phone: phone)
    self.preferences = Preferences()
    self.purchaseUnits = [PurchaseUnit()]
}
``` 
Up to you though, this isn't a blocker so I'll leave it up to you to decide what's best! 😄

agedd: i like that! updated [here](465e89dfc0a47eb4fe796ac3f2afcb943ccd2edd)

jaxdesmarais (approved):

PR #1198, opened 2024-02-29
Send `link_type` FPTI Tag

jaxdesmarais:
### Summary of changes

- Send `link_type` in `event_params` to PayPal's analytics service (FPTI) - this will help us determine if users are going through the universal link or deeplink flow
- Verified via FPTI explorer we are sending this new tag as expected
- Updated `UnitTests`

#### Notes for Android
- In Android v4 we send `venmo.app-links.<STATUS>` events - these are noted in a new tab on the FPTI Events excel sheet
- Ticket DTBTSDK-3596 covers the Android v5 workstream for adding this param

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais


Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -94,6 +94,9 @@ struct FPTIBatchData: Codable {
             #endif
         }()
 
+        /// The type of link the SDK will be handling, currently deeplink or universal
+        let linkType: String?
```
scannillo, line 98: Take it or leave it - using an enum here to enforce we only want these 2 possible string values to be sent for this param. Also would move the string logic for what this field could be out of VenmoClient

scannillo (approved):

agedd (commented):
lgtm! 🚀

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -122,6 +125,7 @@ struct FPTIBatchData: Codable {
             case packageManager = "ios_package_manager"
             case integrationType = "api_integration_type"
             case isSimulator = "is_simulator"
+            case linkType = "link_type"
```
scannillo, line 128: Should this be a batch param? Or instead event level?

scannillo: Similar Q for the paypalContextID. Thinking about it - these are both specific to a unique Venmo or PayPal XO flow

jaxdesmarais: These parameters are only available at this level in FPTI and I have verified that the parameters are sending as expected in FPTI explorer.

scannillo: Are you saying it's not possible to send this data within `event_params`? I assumed this wouldn't be an issue.

Let's take "paypalContextID" for example. That value isn't a "batch" param, since it is specific to each individual event. In the example below, we'll only send 1 paypalContextID when each event actually requires it's own (for example attempting PP multiple times, or switching over to Venmo).

```
{
    "events": [{
        "batch_params": {
            "app_id": "com.braintree.uber",
            "app_name": "Uber",
            "c_sdk_ver": "5.12.0",
            "client_os": "iOS 13.7",
            "comp": "test-sdk",
            "device_manufacturer": "Apple",
            "event_source": "mobile-native",
            "ios_package_manager": "Swift Package Manager",
            "is_simulator": true,
            "mapv": "7.11.0",
            "mobile_device_model": "iPhone11,8",
            "platform": "iOS",
            "session_id": "172B29EC22C84E9884029448B4B95A71",
            "tenant_name": "paypal",
            // Braintree only
            "merchant_id": "123",
            "tokenization_key": "123",
            "auth_fingerprint": "123",
            // PPCP only
            "order_id": "123",
            "partner_client_id": "123"
        },
        "event_params": [   
            {
                "event_name": "venmo:tokenize:started",
                "t": "1666816838453"
            },
            {
                "event_name": "venmo:tokenize:canceled",
                "t": "1666816838453"
            },
            {
                "event_name": "paypal:tokenize:started",
                "t": "1666816838453"
            },
            {
                "event_name": "paypal:tokenize:canceled",
                "t": "1666816838453"
            },
            {
                "event_name": "paypal:tokenize:started",
                "t": "1666816838453"
            },
            {
                "event_name": "paypal:tokenize:succeeded",
                "t": "1666816838453"
            },
            {
                "event_name": "venmo:tokenize:started",
                "t": "1666816838453"
            },
        ]
    }]
}
```

jaxdesmarais: > Are you saying it's not possible to send this data within event_params? I assumed this wouldn't be an issue.

Maybe I've been going about this wrong but I consulted where other teams are passing these params today to determine the level FPTI expected them to be passed at under the assumption that FPTI needed certain tags as certain levels. If this is an incorrect assumption we can switch where they are passed for sure.

> That value isn't a "batch" param, since it is specific to each individual event. In the example below, we'll only send 1 paypalContextID when each event actually requires it's own

This is not true, we will send a new value each time we fetch and set a new context ID on the client. Here is a screenshot from FPTI where I clicked in and out of PayPal, canceled, and then completed the flow, each has a unique context ID:
Screenshot 2024-03-06 at 3 43 02 PM
Happy to swap this around, I could have been operating under the wrong assumptions about FPTI this entire time! 🙈 🙃

jaxdesmarais: We should also figure out why we are sending events multiple times, but that appears to be separate from this issue.

scannillo: Yeh, we can look into those multiple event sends another time.

Similar to how we have [correlationID as an event param](https://github.com/braintree/braintree_ios/blob/8e631aac435210357524374fc73632bb97895187/Sources/BraintreeCore/Analytics/FPTIBatchData.swift#L28-L43), I would think paypalContextID also should go there. If it's in the batch, we don't know that we're properly sending it every time since we don't know when our batch threshold will be met.

Down to chat IRL as well!

jaxdesmarais: Ah, interesting find - I assume we want to update that at some point to actually batch the data. What you are saying makes sense, I think I was confused because they _are_ sending as expected but it looks like that's probably a bug and not a feature. 😄 

I've updated where we send `link_type` here: https://github.com/braintree/braintree_ios/pull/1198/commits/219f283b470c27f661a94ed461cb481ec18e788e

Once this PR is merged I will open a PR to move the level that `paypalContextID` is passed at since it's out of scope for this PR.

UnitTests/BraintreeCoreTests/Analytics/FPTIBatchData_Tests.swift
```diff
@@ -9,7 +9,7 @@ final class FPTIBatchData_Tests: XCTestCase {
     let batchMetadata = FPTIBatchData.Metadata(
         authorizationFingerprint: "fake-auth",
         environment: "fake-env",
-        integrationType: "fake-integration-type",
+        integrationType: "fake-integration-type",
```
jaxdesmarais, line 12:
```suggestion
        integrationType: "fake-integration-type",
```

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
scannillo: I agree! We should remove that batch method if we never plan to use it, or implement proper batching. TBH if we do want proper batching we have a lot of work to do. There is no logic for sending the queued analytics if the app goes into the background or out of memory. Android has a more advanced implementation w/ a local DB.

Maybe it's fine we don't batch since we're sending events less frequently on iOS?

PR #1205, opened 2024-03-06
Universal Links Demo

jaxdesmarais:
### Summary of changes

- Add "Universal Link Flow" button to PayPal Web demo app - this button with open a page that will allow us to return to the app on tapping "open"
- This PR is a proof of concept that our demo app can be set up for universal linking, follow up PRs to this feature branch will build on this work

### Demo
Simulator Screen Recording - iPhone 15 - 2024-03-05 at 09 05 53

#### Notes
- We are currently using Fly.io for hosting this proof of concept, I have hosted the [server side code here](https://github.com/jaxdesmarais/fly-server-demo) for the time being, as this feature develops we can consider moving this to a shared space or Heroku depending on if the free tier of Fly.io is sufficient for our usecase
- This is part of the `paypal-app-switch-feature` work and will be built upon as this feature develops

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Demo/Application/Base/AppDelegate.swift
+++ b/Demo/Application/Base/AppDelegate.swift
@@ -18,7 +18,7 @@ import BraintreeCore
     }
 
     func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool {
-        return true
+        true
     }
 
     func registerDefaultsFromSettings() {
--- a/Demo/Application/Base/AppDelegate.swift
+++ b/Demo/Application/Base/AppDelegate.swift
@@ -17,10 +17,6 @@ import BraintreeCore
         return true
     }
 
-    func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool {
-        true
-    }
-
     func registerDefaultsFromSettings() {
         if processInfoArgs.contains("-EnvironmentSandbox") {
             userDefaults.set(BraintreeDemoEnvironment.sandbox.rawValue, forKey: BraintreeDemoSettings.EnvironmentDefaultsKey)
--- a/Demo/Application/Base/AppDelegate.swift
+++ b/Demo/Application/Base/AppDelegate.swift
@@ -16,7 +16,11 @@ import BraintreeCore
         
         return true
     }
-    
+
+    func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool {
+        return true
+    }
+
     func registerDefaultsFromSettings() {
         if processInfoArgs.contains("-EnvironmentSandbox") {
             userDefaults.set(BraintreeDemoEnvironment.sandbox.rawValue, forKey: BraintreeDemoSettings.EnvironmentDefaultsKey)
--- a/Demo/Demo.entitlements
+++ b/Demo/Demo.entitlements
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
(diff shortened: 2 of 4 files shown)


Demo/Application/Base/AppDelegate.swift
```diff
@@ -16,7 +16,11 @@ import BraintreeCore
         
         return true
     }
-    
+
+    func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool {
```
scannillo, line 20: Why is this necessary? Running it on my simulator it never gets triggered

scannillo:
When switching back to the Demo app, neither `UIApplicationDelegate.application(_:didFinishLaunchingWithOptions:)` nor `UISceneDelegate.scene(_:openURLContexts:)` gets triggered. This is a problem since our SDK will need to receive the returnURL from the universal link switch into the Demo app.

Maybe there's an additional piece we need to configure 🤔

Demo/Application/Base/AppDelegate.swift
jaxdesmarais: Ah, good catch. We need to use the scene delegate `continue` instead ([indicated here](https://developer.apple.com/documentation/xcode/supporting-universal-links-in-your-app)). We won't hit `openURLContexts` ever with our current setup since out app is currently always running running with hitting the URL directly from our demo app. Updated here: https://github.com/braintree/braintree_ios/pull/1205/commits/59f76e05ad9c8cf4f401010451d01aa5c8fe8ea4

scannillo: Gotcha! So a merchant will need to provide us a hook from [scene(_:continue:)](https://developer.apple.com/documentation/uikit/uiscenedelegate/3238056-scene)
 if using SceneDelegates _or_ [application(_:continue:restorationHandler:)](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623072-application) if using AppDelegate only.
 
We'll also need to find out the SwiftUI equivalent (similar to our SwiftUI entry [here](https://developer.paypal.com/braintree/docs/guides/venmo/client-side/ios/v6#handle-app-context-switching)). We could 🏈 punt on the SwiftUI one though, if we want to make a backlog ticket?

jaxdesmarais: Yep, exactly! I think we can punt on the Swift UI portion for now until we have the implementation hammered out a bit more for the feature.

scannillo (approved):

scannillo:
Non-blocking ❓, but I'm curious what App/SceneDelegate method gets invoked if the merchant app is in the "not running" state - for example if a user kills the merchant app after they've switched to PayPal

Demo/Demo.entitlements
```diff
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>com.apple.developer.associated-domains</key>
+	<array>
+		<string>applinks:braintree-ios-demo.fly.dev</string>
```
agedd, line 7: just curious, why do we add this here

agedd (approved):
this is pretty cool!! 🚀

Demo/Demo.entitlements
jaxdesmarais: Great question - so this is adding a new entitlement for associated domains to our demo app. [Associated domains](https://developer.apple.com/documentation/xcode/supporting-associated-domains) establish a secure association between domains and our app. It essentially establishes a secure connection between the server we hosted our app association file at and our demo app while also establishing the permissions of that connection (in our case we are just allowing app links/universal links). The app association file on the server and associated domain entitlement have to match exactly to work as expected.

Demo/Application/Base/AppDelegate.swift
jaxdesmarais: I added this to our docs ticket (DTBTSDK-3609) to ensure we add these methods for both UIKit and SwiftUI. If you think this needs a separate ticket happy to break it out but made sense as an addition to our docs story.

PR #1127, opened 2023-11-02
[5.x] Update 5.4.1 Magnes Pt 2

jaxdesmarais:
### Summary of changes

- The versions somehow got goofed in the last PR, no idea what version of Magnes we linked since they don't contain the version anywhere, this is actually 5.4.1 with staging removed
- Can be verified by pulling down this branch and running the following:
    - Navigate to `Frameworks/XCFrameworks/PPRiskMagnes.xcframework/ios-arm64_x86_64-simulator/PPRiskMagnes.framework` and `Frameworks/XCFrameworks/PPRiskMagnes.xcframework/ios-arm64/PPRiskMagnes.framework`
    - Run `strings PPRiskMagnes | grep "stage"`
    - Nothing should be returned
- The command `strings PPRiskMagnes | grep "stage"` can also be run for the version linked for Carthage at `https://assets.braintreegateway.com/mobile/ios/carthage-frameworks/pp-risk-magnes/PPRiskMagnes.json` for version `5.4.1-stage-removed`

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

PR #1209, opened 2024-03-07
[main] Amend v5.24.0 CHANGELOG entry

scannillo:
### Summary of changes

- We were investigating our Magnes versions and realized that in v5.24.0 of Braintree iOS we switched to a static version of Magnes
   - Technically switching b/w static & dynamic is a breaking change, but since it's v5 and been out for awhile we'll just leave it :) 
- This PR clarifies the CHANGELOG entry for future readers

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo @agedd

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 ## 5.24.0 (2023-10-30)
 * BraintreePayPalDataCollector
   * Update PPRiskMagnes with 5.4.1 - staging removed (fixes #1107)
+  * This version of the PPRiskMagnes framework is static
 
 ## 6.8.0 (2023-10-24)
 * BraintreeDataCollector
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 ## 5.24.1 (2023-11-17)
 * BraintreePayPalDataCollector
   * Update previously incorrect version of PPRiskMagnes 5.4.1 with staging removed 
+  * This version of the PPRiskMagnes framework is dynamic
 
 ## 6.9.0 (2023-11-16)
 * BraintreeThreeDSecure


CHANGELOG.md
```diff
@@ -39,6 +39,7 @@
 ## 5.24.0 (2023-10-30)
 * BraintreePayPalDataCollector
   * Update PPRiskMagnes with 5.4.1 - staging removed (fixes #1107)
+  * This version of the PPRiskMagnes framework is static
```
jaxdesmarais, line 42: I am not sure this is correct, v5 uses `5.4.1-stage-removed-xcode14-1` ([Cartfile](https://github.com/braintree/braintree_ios/blob/5.x/Cartfile)). Downloading that version from [the assets](https://assets.braintreegateway.com/mobile/ios/carthage-frameworks/pp-risk-magnes/PPRiskMagnes.json) and running `file PPRiskMagnes` on both the simulator and non simulator files produces the expected dynamically linked outputs:
 Screenshot 2024-03-07 at 10 30 04 AM

For static frameworks we would expect to see `current ar archive` instead of `Mach-O 64-bit dynamically linked shared library arm64` correct?

scannillo: 👀 Good catch! OK, so it looks like 5.24.0 switched to static, and then 5.24.1 switches back to dynamic.

I checked this by checking out the tag version in the framework version local to the repo:
Screenshot 2024-03-07 at 4 57 33 PM

jaxdesmarais: Yeah exactly - iirc there was an issue with the version uploaded where it wasn't _actually_ the correct magnes version at all (I think we brought this up to magnes as a reason that we should not host their SDK for them). Here is the PR: https://github.com/braintree/braintree_ios/pull/1127. But yes, `5.24.1` is correct, and we have no idea what magnes version was in `5.24.0` because they don't include version numbers anywhere.

scannillo: I updated the PR to call out the specific difference b/w the 2 versions!

jaxdesmarais (approved):

KunJeongPark (approved):

PR #1210, opened 2024-03-07
Fix bug where AnalyticsServices fires same event multiple times

scannillo:
### Summary 

- This PR came out of [this conversation](https://github.com/braintree/braintree_ios/pull/1198#discussion_r1515426571) in a previous PR
- We found a bug where every analytic event was being sent multiple times
    - The reason why was because our [sessionsQueue](https://github.com/braintree/braintree_ios/blob/8e631aac435210357524374fc73632bb97895187/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift#L19) was never being wiped after events were already sent out in an API request
- This PR takes a path-of-least-resistance approach to fixing this duplicate events bug
   - It uses the existing batch POST formatting, but only sends a single event per API request
   - Previously we were still sending an API request every time `sendAnalyticsEvent()` was called, but with this change, the payload won't contain duplicate events
   - Our batching logic needs a dedicated revamp (handling app backgrounding, memory wipe, etc) - JIRA **DTBTSDK-3629** created

### Changes

- Remove unused `flushThreshold` & flush logic (this was never called within the SDK)
- Remove improperly managed `DispatchQueue` for holding events. This was never getting cleared, thus sending every event multiple times. We can re-add this when we have more time to properly implement the queue's maintenance and clearance.
- Move to async-await style for `BTAnalyticsService.sendAnalyticsEvent()` method

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo @jaxdesmarais

### DIFF
--- a/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
+++ b/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
@@ -84,7 +84,11 @@ class BTAnalyticsService: Equatable {
             }
 
             let postParameters = self.createAnalyticsEvent(config: configuration, sessionID: self.apiClient.metadata.sessionID, event: event)
-            self.http?.post("v1/tracking/batch/events", parameters: postParameters, completion: { _,_,_ in })
+            self.http?.post("v1/tracking/batch/events", parameters: postParameters, completion: { _,_, error in
+                if let error {
+                    NSLog("[Braintree SDK] Failed to send analytics: \(error.localizedDescription)")
+                }
+            })
         }
     }
 
--- a/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
+++ b/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
@@ -63,6 +63,7 @@ class BTAnalyticsService: Equatable {
                 
         apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
             guard let configuration, error == nil else {
+                NSLog("[Braintree SDK] Failed to send analytics")
                 return
             }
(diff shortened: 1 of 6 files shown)


Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -38,53 +16,53 @@ class BTAnalyticsService: Equatable {
...
+            correlationID: correlationID,
+            errorDescription: errorDescription,
+            eventName: eventName,
+            timestamp: String(timestampInMilliseconds)
+        )
+                
         apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
             guard let configuration, error == nil else {
-                if let error {
-                    completion(error)
```
scannillo, line 86: This completion was always ignored, so we removed it entirely

jaxdesmarais (approved):
Love how this bug also got rid of our recently problematic tests. 😂

agedd (approved):
looks great and super clean! 🚀

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
richherrera: Curious why this is happening, thoughts?

jaxdesmarais: `BTAPIClient` was always no-oping the result of this completion. I don't think we would ever want to return an error to the merchant and stop the flow if there is an issue with analytics since that's just for us, so this is safe for us to remove since we just discarded the result.

PR #1212, opened 2024-03-08
[Open Dev] Provide Default View Controller

jaxdesmarais:
### Summary of changes

Right now if the `FeatureViewController` cannot be found, a message is displayed "Demo not available", by providing a default as `PayPalWebCheckoutViewController` we can load a View Controller instead of returning `nil` without a super clear message to the user. This is mainly helpful for first time installs, when nothing is stored in `UserDefaults` yet or when switching between feature branches where the demo exists on a branch but not on main.

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Demo/Application/Base/ContainmentViewController.swift
+++ b/Demo/Application/Base/ContainmentViewController.swift
@@ -188,7 +188,7 @@ class ContainmentViewController: UIViewController {
 
     private func instantiateViewController(with authorization: String) -> BaseViewController? {
         guard let integrationName = UserDefaults.standard.string(forKey: "BraintreeDemoSettingsIntegration") else {
-            return nil
+            return PayPalWebCheckoutViewController(authorization: authorization)
         }
 
         switch integrationName {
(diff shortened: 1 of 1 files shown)


Demo/Application/Base/ContainmentViewController.swift
```diff
@@ -188,7 +188,7 @@ class ContainmentViewController: UIViewController {
 
     private func instantiateViewController(with authorization: String) -> BaseViewController? {
         guard let integrationName = UserDefaults.standard.string(forKey: "BraintreeDemoSettingsIntegration") else {
-            return nil
+            return PayPalWebCheckoutViewController(authorization: authorization)
```
jaxdesmarais, line 191: PayPal Web made the most sense as the default to me, but if someone feels strongly that we should default to one of the other features, happy to do so!

scannillo (approved):

sshropshire (approved):

PR #1214, opened 2024-03-12
[QL] Add `BTAppContextSwitcher.universalLink`

jaxdesmarais:
This work is going into the `paypal-app-switch-feature` branch

### Summary of changes

- Add `BTAppContextSwitcher.sharedInstance.universalLink` property
- Parse path in return URL `SceneDelegate` method

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 * BraintreeCore
-  * Added property `BTAppContextSwitcher.sharedInstance.universalLink` for the PayPal app switch flow
+  * Add property `BTAppContextSwitcher.sharedInstance.universalLink` for the PayPal app switch flow
 
 ## 6.13.0 (2024-03-12)
 * BraintreeVenmo
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreeCore
+  * Added property `BTAppContextSwitcher.sharedInstance.universalLink` for the PayPal app switch flow
+
 ## 6.13.0 (2024-03-12)
 * BraintreeVenmo
   * Add `isFinalAmount` to `BTVenmoRequest`
--- a/Demo/Application/Base/SceneDelegate.swift
+++ b/Demo/Application/Base/SceneDelegate.swift
@@ -21,7 +21,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         URLContexts.forEach { urlContext in
             let url = urlContext.url
             if url.scheme?.localizedCaseInsensitiveCompare("com.braintreepayments.Demo.payments") == .orderedSame {
-                _ = BTAppContextSwitcher.sharedInstance.handleOpenURL(context: urlContext)
+                BTAppContextSwitcher.sharedInstance.handleOpenURL(context: urlContext)
             }
         }
     }
@@ -29,7 +29,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     func scene(_ scene: UIScene, continue userActivity: NSUserActivity) {
         if let returnURL = userActivity.webpageURL, returnURL.path == "/braintree-payments" {
             print("Returned to Demo app via universal link: \(returnURL)")
-            _ = BTAppContextSwitcher.sharedInstance.handleOpen(returnURL)
+            BTAppContextSwitcher.sharedInstance.handleOpen(returnURL)
         }
     }
 }
--- a/Demo/Application/Base/SceneDelegate.swift
+++ b/Demo/Application/Base/SceneDelegate.swift
@@ -27,9 +27,9 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     }
 
     func scene(_ scene: UIScene, continue userActivity: NSUserActivity) {
-        if let returnURL = userActivity.webpageURL {
-            // TODO: implementation - pass full URL to BT SDK
+        if let returnURL = userActivity.webpageURL, returnURL.path == "/braintree-payments" {
             print("Returned to Demo app via universal link: \(returnURL)")
+            _ = BTAppContextSwitcher.sharedInstance.handleOpen(returnURL)
         }
     }
 }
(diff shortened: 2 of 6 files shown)


Sources/BraintreeCore/BTAppContextSwitcher.swift
```diff
@@ -13,7 +13,10 @@ import UIKit
     /// The URL scheme to return to this app after switching to another app or opening a SFSafariViewController.
     /// This URL scheme must be registered as a URL Type in the app's info.plist, and it must start with the app's bundle ID.
     public var returnURLScheme: String = ""
-    
+
+    /// The URL to use for the PayPal universal link flow. Must contain the path `braintree-payments`.
```
jaxdesmarais, line 17: Do we want to require this specific path or another one for this feature?

UnitTests/BraintreeCoreTests/BTAppContextSwitcher_Tests.swift
```diff
@@ -80,13 +80,19 @@ class BTAppContextSwitcher_Tests: XCTestCase {
...
         let handled = BTAppContextSwitcher.sharedInstance.handleOpenURL(context: mockURLContext)
         XCTAssertFalse(handled)
     }
 
+    // MARK: - universalLink Tests
+
+    func testSetUniversalLink() {
+        BTAppContextSwitcher.sharedInstance.universalLink = "https://fake.com"
+        XCTAssertEqual(appSwitch.universalLink, "https://fake.com")
+    }
```
jaxdesmarais, line 95: This is a similar test that exists for the return URL - there is not a ton we can test until we add the protocol conformance to the client. It seems like most of the return URL tests exist on the client. Do we want any other tests in this file?

Sources/BraintreeCore/BTAppContextSwitcher.swift
scannillo: I was thinking similar to[ URL schemes](https://developer.paypal.com/braintree/docs/guides/venmo/client-side/ios/v6#register-a-url-type), we want them to include some path specific to BT payments, but it could be whatever they want

scannillo: I don't think we should validate that it contains `braintree-payments`. They might have an existing path they want to use for their checkout flow

jaxdesmarais: Sounds great - I can update this to the following if that sounds good:
```suggestion
    /// The URL to use for the PayPal universal link flow. Must be dedicated to Braintree app switch returns.
```

Sources/BraintreeCore/BTAppContextSwitcher.swift
```diff
@@ -13,7 +13,10 @@ import UIKit
...
+    public var universalLink: String = ""
```
jaxdesmarais, line 18: Do we want a CHANGELOG entry for this?

Sources/BraintreeCore/BTAppContextSwitcher.swift
scannillo: Love it!

scannillo (approved):

Sources/BraintreeCore/BTAppContextSwitcher.swift
scannillo, line 17: Nitpick - should we call this the PayPal app switch flow? Probably doesn't matter too much.

Also do we want to call out at the `BTAppContextSwitcher` level that setting the universalLink is only for PP and the urlScheme is only for Venmo?

Sources/BraintreeCore/BTAppContextSwitcher.swift
scannillo: Yes!

Sources/BraintreeCore/BTAppContextSwitcher.swift
jaxdesmarais: That's a great call - I don't think it can hurt to call it out. Updated here: https://github.com/braintree/braintree_ios/pull/1214/commits/22ab53a399841f4a0825c728527f4a3156a3c97f

Sources/BraintreeCore/BTAppContextSwitcher.swift
jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/1214/commits/69332b848acb3ba53e5fb519c9192ada6e9e67be

sshropshire (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreeCore
+  * Added property `BTAppContextSwitcher.sharedInstance.universalLink` for the PayPal app switch flow
```
```suggestion
  * Add property `BTAppContextSwitcher.sharedInstance.universalLink` for the PayPal app switch flow
```

Demo/Application/Base/SceneDelegate.swift
```diff
@@ -27,9 +27,9 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     }
 
     func scene(_ scene: UIScene, continue userActivity: NSUserActivity) {
-        if let returnURL = userActivity.webpageURL {
-            // TODO: implementation - pass full URL to BT SDK
+        if let returnURL = userActivity.webpageURL, returnURL.path == "/braintree-payments" {
             print("Returned to Demo app via universal link: \(returnURL)")
+            _ = BTAppContextSwitcher.sharedInstance.handleOpen(returnURL)
```
richherrera, line 32: Wondering to know if we can use `@discardableResult`

richherrera (approved):

Demo/Application/Base/SceneDelegate.swift
jaxdesmarais: 🤔 I am down - historically we haven't used it but I just tested and it doesn't seem to be a breaking change for either Obj-C or Swift interface, so no reason not to. Also tested to ensure that merchants that may _want_ to capture the result will still be able to. As long as CI is happy lets go for it! Updated here: https://github.com/braintree/braintree_ios/pull/1214/commits/c0a8e49f6d968aff9951ad18bcbb08771d3b66e4

jaxdesmarais:
Failing UI tests are unrelated

PR #1217, opened 2024-03-12
[main] Bump to Magnes v5.5.0

scannillo:
### Summary of changes

- Bump to using Magnes v5.5.0
- This version of Magnes contains the code signing & privacy manifest requirements imposed by Apple (see https://developer.apple.com/support/third-party-SDK-requirements/)

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo


Frameworks/XCFrameworks/PPRiskMagnes.xcframework/ios-arm64/PPRiskMagnes.framework/Headers/PPRiskMagnes-Swift.h
```diff
@@ -1,6 +1,6 @@
 #if 0
 #elif defined(__arm64__) && __arm64__
-// Generated by Apple Swift version 5.8 (swiftlang-5.8.0.124.2 clang-1403.0.22.11.100)
+// Generated by Apple Swift version 5.7.1 (swiftlang-5.7.1.135.3 clang-1400.0.29.51)
```
scannillo, line 3: I think this is fine since it's backward compatible with merchant apps that may use newer Swift versions. Please call out if you disagree!

scannillo: Blegh, I wonder with I don't _think_ [this](https://developer.apple.com/news/upcoming-requirements/?id=04292024a) will be an issue here, but coming up we should just as for a version from Magnes built with the latest Xcode and make that bump now to Xcode 15+

scannillo: I wonder if this will be an issue with this upcoming requirement - https://developer.apple.com/news/upcoming-requirements/?id=04292024a. I don't think so, but good to call out

jaxdesmarais: I don't _think_ it should be an issue as long as it works currently. It looks like [Xcode 15 supports down to Swift 4](https://developer.apple.com/support/xcode/)

jaxdesmarais (approved):

warmkesselj (approved):

PR #1218, opened 2024-03-13
[QL] Conform `BTPayPalClient` to `BTAppContextSwitchClient` Protocol

jaxdesmarais:
This work is into `paypal-app-switch-feature`

### Summary of changes

- Conform `BTPayPalClient` to `BTAppContextSwitchClient`
- Add `BTPayPalAppSwitchReturnURL` struct
- Add Unit Tests

### Checklist

### Authors

- @jaxdesmarais


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -431,3 +443,18 @@ import BraintreeDataCollector
...
+    @_documentation(visibility: private)
+    @objc public static func handleReturnURL(_ url: URL) {
+        payPalClient?.handleOpen(url)
+        BTPayPalClient.payPalClient = nil
+    }
+
+    /// :nodoc:
+    @_documentation(visibility: private)
+    @objc public static func canHandleReturnURL(_ url: URL) -> Bool {
+        url.scheme == "https" && (url.path.contains("cancel") || url.path.contains("success"))
```
jaxdesmarais, line 458: Based on the current contracts I don't believe we will ever get an error path back. In the future if we do we can consider adding it here. We can also consider adding a check here that the `host` contains the `universalLink` set on `BTAppContextSwitcher`. What do folks think?

Sources/BraintreeVenmo/BTVenmoAppSwitchReturnURL.swift
```diff
@@ -68,6 +68,8 @@ struct BTVenmoAppSwitchReturnURL {
     /// - Parameter url: an app switch return URL
     /// - Returns: `true` if the url represents a Venmo Touch app switch return
     static func isValid(url: URL) -> Bool {
-        url.host == "x-callback-url" && url.path.hasPrefix("/vzero/auth/venmo/")
+        url.scheme?.caseInsensitiveCompare(BTAppContextSwitcher.sharedInstance.returnURLScheme) == .orderedSame
```
jaxdesmarais, line 71: This is called from `BTVenmoClient.handleReturnURL` and enforces URL schemes for returning to the app

Sources/BraintreePayPal/BTPayPalClient.swift
scannillo: There won't be an error specific scenario in the URL, no.

Though we will want to surface the `cancel` as an error to the merchant, as well as continue with the flow to call the paypal_accounts API in the case of success. Are you planning to add those paths in this PR? If not, can we add TODOs?

Sources/BraintreeVenmo/BTVenmoAppSwitchReturnURL.swift
jaxdesmarais: It doesn't have to but the ticket listed "Consider updating VenmoClient’s handleReturnURL methods to enforce URL scheme & disallow https / universal link" 😄

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: > Though we will want to surface the cancel as an error to the merchant

🤔 how would we both surface the cancel as an error and continue the flow? If we surfaced it in our completion I would expect the execution of that path would also end? Or do we want to diverge from that pattern and latter in the flow return both a nonce and an error?

> Are you planning to add those paths in this PR? If not, can we add TODOs?

There is a TODO in `handleOpen` above where we will build out the logic in a future PR to handle the return URL. This is just the logic that determines if our SDK can process the URL we get back (though I did notice just now `BTAppContextSwitcher` wasn't registered properly and updated here along with deploying a success path so we can test that these methods are hit: https://github.com/braintree/braintree_ios/pull/1218/commits/9efbff214ecc7e4959617d8b54ea1ce4a0b43546)

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -4,7 +4,12 @@ import BraintreePayPal
 
 class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
 
-    lazy var payPalClient = BTPayPalClient(apiClient: apiClient)
+    var payPalClient: BTPayPalClient!
```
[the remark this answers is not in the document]
jaxdesmarais: With it being instantiated lazily and not used in the `universalLinkFlow` it's deallocated during that flow and none of the `BTAppContextSwitcher` methods are called. 

As an alternative we can call and just not use `payPalClient.tokenize(BTPayPalVaultRequest()) { _, _ in }` in `universalLinkFlow` for now. Eventually that method will be used in the `universalLinkFlow` method, so I am fine keeping this lazy and adding that call that will be used in the future. What do you think?

KunJeongPark (approved):

Demo/Application/Features/PayPalWebCheckoutViewController.swift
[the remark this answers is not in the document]
jaxdesmarais: Updated: https://github.com/braintree/braintree_ios/pull/1218/commits/429533e00359840f8a5f54e25375bab6197d44d7

Sources/BraintreePayPal/BTPayPalClient.swift
scannillo: > 🤔 how would we both surface the cancel as an error and continue the flow? If we surfaced it in our completion I would expect the execution of that path would also end? 

Yes! Like we do today when we get back `cancel` in the URL for the web-based flow, we end the flow for the merchant.

Sources/BraintreeVenmo/BTVenmoAppSwitchReturnURL.swift
scannillo: I think I overthought that when I originally wrote it, I don't think this logic has to change!

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -431,3 +445,18 @@ import BraintreeDataCollector
...
```
scannillo, line 460: Another thought - might be more readable to keep this similar to Venmo which has a separate class (`BTVenmoAppSwitchReturnURL`) for the `isValid()` logic & eventual param parsing? So adding something like a `BTVeniceAppSwitchReturnURL`

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -221,6 +229,12 @@ import BraintreeDataCollector
         performSwitchRequest(appSwitchURL: url, paymentType: paymentType, completion: completion)
     }
 
+    // MARK: - App Switch Methods
+
+    func handleOpen(_ url: URL) {
```
scannillo, line 234: I know this is the naming we use in Venmo, but it's a little confusing to me since without looking through the code callsites you don't know if this method is for opening the Venmo app, or opening the merchant app.

_Take it or leave it comment_ - but we could rename to `handleReturnURL()` or something similar

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: That makes since, sorry I read that wrong at first as we want to continue the flow on cancel, but that portion was for success. 🙈 

Do we think it'd be helpful to also validate that the `host` contains the value passed into `BTAppContextSwitcher.shared.universalLink` or do we think this is sufficient (checking for https and cancel or success)?

Sources/BraintreeVenmo/BTVenmoAppSwitchReturnURL.swift
jaxdesmarais: No worries, reverted here: https://github.com/braintree/braintree_ios/pull/1218/commits/a6aabdd86d68304e041a69481c000f1e7a9973e9

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: I'm down, updated here (I opted to call it `BTPayPalAppSwitchReturnURL` since I assume `Venice` is somewhat internal): https://github.com/braintree/braintree_ios/pull/1218/commits/4b3954130e9b9d91a2718efd1cd82ec6a9548aa5

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: That makes sense to me too: https://github.com/braintree/braintree_ios/pull/1218/commits/2d53f54c627d7e39fdc9c9835fc8bc43029577c8

Sources/BraintreePayPal/BTPayPalClient.swift
scannillo: Good question - maybe we can assume that if the switch successfully made it into the merchant app, that the host of the URL is fine. I vote for not doing that check, for now. But don't feel strongly if you have other thoughts

scannillo (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: I agree with that too, we can always add it in a future PR if we decide we need additional validation.

PR #1223, opened 2024-03-14
[V6] Accept email in `BTPayPalVaultRequest`

scannillo:
### Summary of changes

- This code change was already reviewed and approved in https://github.com/braintree/braintree_ios/pull/1188
- This PR instead moves the feature to the `main` branch to be released outside of the ShopperInsights beta feature release.

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* Add `BTPayPalVaultRequest.userAuthenticationEmail` optional property
+
 ## 6.14.0 (2024-03-13)
 * BraintreeDataCollector
   * Bump to PPRiskMagnes v5.5.0 with code signing & a privacy manifest file
--- a/Sources/BraintreePayPal/BTPayPalVaultRequest.swift
+++ b/Sources/BraintreePayPal/BTPayPalVaultRequest.swift
@@ -11,6 +11,9 @@ import BraintreeCore
 
     /// Optional: Offers PayPal Credit if the customer qualifies. Defaults to `false`.
     public var offerCredit: Bool
+    
+    /// Optional: User email to initiate a quicker authentication flow in cases where the user has a PayPal Account with the same email.
+    public var userAuthenticationEmail: String?
 
     // MARK: - Initializer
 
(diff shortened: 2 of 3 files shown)


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -75,6 +75,10 @@ import BraintreeDataCollector
         _ request: BTPayPalVaultRequest,
         completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void
     ) {
+        if request.userAuthenticationEmail != nil {
+            webAuthenticationSession.prefersEphemeralWebBrowserSession = true
+        }
```
jaxdesmarais, line 80: @scannillo Curious if you also run into weird behavior with MPE email without this. Even with clearing cookies on safari I see a "transition failed" message in the webview. It seems like we can't have any cookies in the session for this feature.

tdchow: were you seeing this in sandbox?

jaxdesmarais: Production - when prod is fixed I'll remove this and see if the behavior is still unexpected. This flag removes all cookies from the `ASWebAuthenticationSession` but may not be necessary when prod is fixed.

jaxdesmarais (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
scannillo: I honestly forgot this ephemeral session setting was changed for this feature. I'm not sure we want to do this anymore 🤔

scannillo:
```suggestion
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -75,6 +75,7 @@ import BraintreeDataCollector
         _ request: BTPayPalVaultRequest,
         completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void
     ) {
+
```
scannillo, line 78:
```suggestion
```

tdchow (approved):

PR #1228, opened 2024-03-19
[QL] Add additional POST params to `/setup_billing_agreement` API call

scannillo:
### Summary of changes

- Add 4 top-level POST params to `v1/paypal_hermes/setup_billing_agreement` if a merchant opts into the PP App Switch flow

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

PR #1232, opened 2024-03-22
[QL] Update `BTAppContextSwitcher.sharedInstance.universalLink` to URL Type

jaxdesmarais:
### Summary of changes

- Replace `BTAppContextSwitcher.sharedInstance.universalLink` with init:  `BTPayPalVaultRequest(userAuthenticationEmail:enablePayPalAppSwitch:universalLink:offerCredit:)`

- OG convo here: https://github.com/braintree/braintree_ios/pull/1228#discussion_r1530809502

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,9 @@
 
 ## unreleased
 * Require Xcode 15.0+ (per [App Store requirements](https://developer.apple.com/news/?id=khzvxn8a))
-* BraintreeCore
-  * Add property `BTAppContextSwitcher.sharedInstance.universalLink` for the PayPal app switch flow
 * BraintreePayPal
-  * Add `BTPayPalVault.enablePayPalAppSwitch`
-    * If set to `true` we will attempt to use the PayPal App Switch flow
+  * Add `BTPayPalVaultRequest.init(userAuthenticationEmail:enablePayPalAppSwitch:universalLink:offerCredit:)`
+    * This init should be used for the PayPal App Switch flow
 
 ## 6.16.0 (2024-03-19)
 * Add `BTPayPalVaultRequest.userAuthenticationEmail` optional property
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -87,7 +87,12 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
 
     @objc func universalLinkFlow(_ sender: UIButton) {
         // TODO: implement in a future PR - used here so we don't have to remove lazy instantiation
-        let request = BTPayPalVaultRequest(enablePayPalAppSwitch: true)
+        // TODO: replace URL with https://braintree-ios-demo.fly.dev/braintree-payments
+        let request = BTPayPalVaultRequest(
+            userAuthenticationEmail: "sally@gmail.com",
+            enablePayPalAppSwitch: true,
+            universalLink: URL(string: "https://paypal.com")!
+        )
         payPalClient.tokenize(request) { _, _ in }
         UIApplication.shared.open(URL(string: "https://braintree-ios-demo.fly.dev/braintree-payments/success")!)
     }
--- a/Sources/BraintreePayPal/BTPayPalVaultRequest.swift
+++ b/Sources/BraintreePayPal/BTPayPalVaultRequest.swift
@@ -17,7 +17,27 @@ import BraintreeCore
     /// Optional: User email to initiate a quicker authentication flow in cases where the user has a PayPal Account with the same email.
     public var userAuthenticationEmail: String?
 
+    /// The URL to use for the PayPal app switch flow. Must be a valid HTTPS URL dedicated to Braintree app switch returns.
+    var universalLink: URL?
+
     // MARK: - Initializer
+
+    /// Initializes a PayPal Vault request
+    /// - Parameters:
+    ///   - offerCredit: Optional: Offers PayPal Credit if the customer qualifies. Defaults to `false`.
+    ///   - userAuthenticationEmail: Optional: User email to initiate a quicker authentication flow in cases where the user has a PayPal Account with the same email.
+    ///   - enablePayPalAppSwitch: Optional: Used to determine if the customer will use the PayPal app switch flow. Defaults to `false`.
+    ///   This property is currently in beta and may change or be removed in future releases.
+    ///   - universalLink: The URL to use for the PayPal app switch flow. Must be a valid HTTPS URL dedicated to Braintree app switch returns.
+    public convenience init(
+        offerCredit: Bool = false,
+        userAuthenticationEmail: String,
+        enablePayPalAppSwitch: Bool,
+        universalLink: URL
+    ) {
+        self.init(offerCredit: offerCredit, userAuthenticationEmail: userAuthenticationEmail, enablePayPalAppSwitch: enablePayPalAppSwitch)
+        self.universalLink = universalLink
+    }
 
     /// Initializes a PayPal Vault request
     /// - Parameters:
@@ -38,12 +58,12 @@ import BraintreeCore
             baseParameters["payer_email"] = userAuthenticationEmail
         }
         
-        if enablePayPalAppSwitch, let universalLinkURL = BTAppContextSwitcher.sharedInstance.universalLink {
+        if enablePayPalAppSwitch, let universalLink {
             let appSwitchParameters: [String: Any] = [
                 "launch_paypal_app": enablePayPalAppSwitch,
                 "os_version": UIDevice.current.systemVersion,
                 "os_type": UIDevice.current.systemName,
-                "merchant_app_return_url": universalLinkURL.absoluteString
+                "merchant_app_return_url": universalLink.absoluteString
             ]
             return baseParameters.merging(appSwitchParameters) { $1 }
         }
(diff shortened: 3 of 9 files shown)


Sources/BraintreePayPal/BTPayPalVaultRequest.swift
```diff
@@ -38,12 +38,12 @@ import BraintreeCore
             baseParameters["payer_email"] = userAuthenticationEmail
         }
         
-        if enablePayPalAppSwitch {
+        if enablePayPalAppSwitch, let universalLinkURL = BTAppContextSwitcher.sharedInstance.universalLink {
```
scannillo, line 41: Do you think we should return an error if they try to opt into `enablePayPalAppSwitch` but don't have a universal link set?

jaxdesmarais: 🤔 that's a good question - I think it depends on the merchant experience we want. If this is not set we would get `nil` back as the `paypalAppApprovalUrl` so would just continue with the ASWeb based flow, but nothing would blow up. On the other side of things I can see a merchant ``_not_ wanting the ASWeb flow if they have set `enablePayPalAppSwitch` in which case we should add a check on the client that if true this value is also set. Which way are you leaning?

scannillo: I lean toward returning an error if they don't have the universal link set.

Actually brings me to another thought - we don't _need_ them to pass us the URL in their app delegate, do we? We shouldn't need it that early. We only need it at the time of making the GQ hermes POST request. We _could_ enforce this at the `BTPayPalRequest` init level. We can add an additional init specific to this flow that takes in everything we need: email, flag, link

jaxdesmarais: Yeah, that is a great question - I was just rolling with the design we had outlined. Technically I don't think the `returnURLScheme` for the Venmo flow needs to be set in `BTAppContextSwitcher` either. It's also not used until the `tokenize` method is called. Do you know historically why those were set in that manner instead of on the clients?

I am down to move this into a new init - I think that makes sense in this PR. We would also need to make a very fast follow docs PR since that was merged in today to be released Wednesday. I think if we make it required in the init we won't need to throw an error from the request or do a check in the client and throw the error there.

If that sounds good to you I can make the change here and get up a docs update as well?

jaxdesmarais: @scannillo this update is reflected in the PR - if we are good with this change, I will open a new docs PR and ping the docs team as well as update our HLD. In `BTPayPalVaultRequest` I added a new `init` with the required properties and made `universalLink` an internal var on the class so it can only be set on the `init` - this will enforce it as required without us needing to throw an error in the `init` for this flow.

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
```diff
@@ -9,24 +9,44 @@ import BraintreeCore
...
+    /// Optional: User email to initiate a quicker authentication flow in cases where the user has a PayPal Account with the same email.
+    public var userAuthenticationEmail: String?
+
+    // MARK: - Internal Properties
+
     /// Optional: Used to determine if the customer will use the PayPal app switch flow.
     /// Defaults to `false`.
     /// - Note: This property is currently in beta and may change or be removed in future releases.
-    public var enablePayPalAppSwitch: Bool
+    var enablePayPalAppSwitch: Bool = false
```
jaxdesmarais, line 20: Should we consider defaulting this to `true`? Do we even need this property at all or can we assume if the universal link is passed a merchant is opting into the flow?

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
```diff
@@ -9,24 +9,44 @@ import BraintreeCore
...
 
-    // MARK: - Initializer
+    // MARK: - Initializers
+
+    /// Initializes a PayPal Vault request
+    /// - Note: This initializer should be used for merchants using the PayPal App Switch flow
+    /// - Parameters:
+    ///   - userAuthenticationEmail: Required: User email to initiate a quicker authentication flow in cases where the user has a PayPal Account with the same email.
+    ///   - enablePayPalAppSwitch: Required: Used to determine if the customer will use the PayPal app switch flow.
+    ///   This property is currently in beta and may change or be removed in future releases.
```
agedd, line 32: nit(optional): can we move this line after all the parameter comments

UnitTests/BraintreePayPalTests/BTPayPalRequest_Tests.swift
```diff
@@ -96,8 +96,15 @@ class BTPayPalRequest_Tests: XCTestCase {
...
-        let request = BTPayPalVaultRequest()
+    func testEnablePayPalAppSwitch_whenInitialized_setsAllRequiredValues() {
+        let request = BTPayPalVaultRequest(
+            userAuthenticationEmail: "fake@gmail.com",
+            enablePayPalAppSwitch: true,
+            universalLink: URL(string: "my-website-is-cool.com")!
+        )
+
+        XCTAssertEqual(request.userAuthenticationEmail, "fake@gmail.com")
         XCTAssertFalse(request.enablePayPalAppSwitch)
```
agedd, line 107: qq here - should we `XCTAssertTrue` instead since we set `enablePayPalAppSwitch` on line 102

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
agedd: hmm yea that's a good point ~ imo i feel it's safe to remove this property since we can assume whether or not to launch the app switch flow based on if `universalLink` is passed in

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -87,7 +87,12 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
 
     @objc func universalLinkFlow(_ sender: UIButton) {
         // TODO: implement in a future PR - used here so we don't have to remove lazy instantiation
-        let request = BTPayPalVaultRequest(enablePayPalAppSwitch: true)
+        // TODO: replace URL with https://braintree-ios-demo.fly.dev/braintree-payments
```
warmkesselj, line 90: Curious why we can't use: "https://braintree-ios-demo.fly.dev/braintree-payments"

jaxdesmarais: The gateway has not yet allowlisted our URL for this flow, so we are using a URL that is included in their allowlist currently.

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
jaxdesmarais: I'm actually going to move it with the main note: https://github.com/braintree/braintree_ios/pull/1232/commits/631b3e30ff403465db9d203588194de35cfbdc71. Here is what a merchant will see:
Screenshot 2024-03-26 at 12 15 36 PM

UnitTests/BraintreePayPalTests/BTPayPalRequest_Tests.swift
jaxdesmarais: Yep, we actually weren't setting this on the init, which is why this test was passing. Updated here: https://github.com/braintree/braintree_ios/pull/1232/commits/89fef878b314959db2ccb3aefc83c5527b64eff3

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
scannillo: > Do you know historically why those were set in that manner instead of on the clients?

Unfortunately I don't know why it was that way

scannillo (approved):

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
jaxdesmarais: No worries, it's a great question in any case! I'm going to add a `NEXT_MAJOR_VERSION` note to `returnURLScheme` so we can move that to the clients in the future. 🚀

richherrera (approved):

CHANGELOG.md
```diff
@@ -2,11 +2,9 @@
 
 ## unreleased
 * Require Xcode 15.0+ (per [App Store requirements](https://developer.apple.com/news/?id=khzvxn8a))
-* BraintreeCore
-  * Add property `BTAppContextSwitcher.sharedInstance.universalLink` for the PayPal app switch flow
 * BraintreePayPal
-  * Add `BTPayPalVault.enablePayPalAppSwitch`
-    * If set to `true` we will attempt to use the PayPal App Switch flow
+  * Add `BTPayPalVaultRequest.init(userAuthenticationEmail:enablePayPalAppSwitch:universalLink:offerCredit:)`
```
```suggestion
  * Add `BTPayPalVaultRequest(userAuthenticationEmail:enablePayPalAppSwitch:universalLink:offerCredit:)`
```

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
jaxdesmarais: I think we are going to leave this as is for now to make it apparent to merchants that they are opting into this flow with both the `bool` and a `universalLink`. With just the `universalLink` it may make merchants think we are using that on the existing web flow (which we are not)

PR #1233, opened 2024-03-22
App installed check

stechiu:
### Summary of changes

We want to check if the PayPal app is installed. This branch will be merged into `paypal-app-switch-feature`. This PR doesn't cover the actual app switching to the PayPal app. To test, change the url scheme to `maps://`

- Add appInstalled check if enablePayPalAppSwitch=true, fallback to existing tokenize() flow if not

### Checklist

- [x] Added a changelog entry

- @stechiu @jaxdesmarais


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -83,7 +87,12 @@ import BraintreeDataCollector
         _ request: BTPayPalVaultRequest,
         completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void
     ) {
-        tokenize(request: request, completion: completion)
+        // Check if PayPal app is installed
+        if request.enablePayPalAppSwitch == true {
```
scannillo, line 91: We don't want to diverge in the flow _just yet_. We still need to get the response back from[ this API call](https://github.com/braintree/braintree_ios/blob/4d4f107700b1483addeae0b3b7c741014f398a5d/Sources/BraintreePayPal/BTPayPalClient.swift#L248) which will either contain the Venice URL, or it won't. Based on if the URL present, we will or won't proceed with app switch flow.

scannillo: This is where we instead want to do the `appInstalled` check. If that check fails, we don't want to bother encoding [these additional parameters](https://github.com/braintree/braintree_ios/pull/1228)

stechiu: Thanks! I've updated the PR with the changes

.github/workflows/build.yml
```diff
@@ -5,22 +5,22 @@ concurrency:
   cancel-in-progress: true
 jobs:
   cocoapods:
-    name: CocoaPods (Xcode 14.3)
+    name: CocoaPods (Xcode 15.1)
```
scannillo, line 8: Looks like the base branch needs to be updated for this PR

stechiu: That's strange... I'll update the `app-installed-check` branch with the main and see if that helps

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* BraintreeCore
```
jaxdesmarais, line 4: It seems like this branch has an outdated version of what's on `paypal-app-switch-feature` you may want to pull down the latest from the feature branch and then merge it in. These entries on the latest version will be replaces with the [diff from this PR in the CHANGELOG](https://github.com/braintree/braintree_ios/pull/1232/files).

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -728,6 +728,27 @@ class BTPayPalClient_Tests: XCTestCase {
     }
     
 
+    func testIsiOSAppSwitchAvailable_whenApplicationCanOpenPayPalInAppURL_returnsTrue() {
```
scannillo, line 731: Instead of asserting that the function `isPayPalAppInstalled` returns some boolean, we rather want our unit tests to assert on our public API. Implementation details (such as if the app installed check lives in a method or not) should be able to change without breaking our unit tests.

So this test should assert on the end result behavior, which is that the API POST request to `/v1/paypal_hermes/create_payment_resource` _does_ include the app switch params if the app is installed, and does _not_ include the app switch params if the app is not installed.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -154,9 +157,16 @@ import BraintreeDataCollector
             }
         }
     }
-    
+
+    @objc public func isPayPalAppInstalled() -> Bool {
```
scannillo, line 161: This can be private (and non @objc) once below test changes are addressed

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
stechiu: Thanks for the feedback! I've made changes to test the endpoints

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -857,6 +857,63 @@ class BTPayPalClient_Tests: XCTestCase {
...
+        XCTAssertEqual(lastPostParameters["merchant_app_return_url"] as? String, "https://paypal.com")
+    }
+
+    func testIsiOSAppSwitchAvailable_whenApplicationCantOpenPayPalInAppURL_returnsFalse() {
+        let fakeApplication = FakeApplication()
+        payPalClient.application = fakeApplication
+
+        let vaultRequest = BTPayPalVaultRequest(
+            userAuthenticationEmail: "fake@gmail.com",
+            enablePayPalAppSwitch: payPalClient.payPalAppInstalled,
```
jaxdesmarais, line 895:
```suggestion
        fakeApplication.cannedCanOpenURL = false
        payPalClient.application = fakeApplication

        let vaultRequest = BTPayPalVaultRequest(
            userAuthenticationEmail: "fake@gmail.com",
            enablePayPalAppSwitch: true,
```

jaxdesmarais (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -276,6 +281,12 @@ import BraintreeDataCollector
                 return
             }
 
+            self.payPalAppInstalled = self.isPayPalAppInstalled()
+
+            if !self.payPalAppInstalled {
+                (request as? BTPayPalVaultRequest)?.enablePayPalAppSwitch = false
```
KunJeongPark, line 287: Nice!

KunJeongPark (approved):

richherrera (approved):

PR #1246, opened 2024-04-01
[QL] Handle Success or Failure of `application.open`

jaxdesmarais:
### Summary of changes

- Handle Success or Failure of `application.open` via `invokedOpenURLSuccessfully`
    - Add TODOs for future analytics PR
- Add new `BTPayPalError.appSwitchFailed`

### Checklist

### Authors

- @jaxdesmarais


scannillo:
Changes look good! But I think this needs 2 unit tests based on the `bool` result of `application.open(redirectURL, options: [:])`

agedd (approved):
lgtm!

jaxdesmarais:
> Changes look good! But I think this needs 2 unit tests based on the `bool` result of `application.open(redirectURL, options: [:])`

It looks like there is a test `testTokenizeVaultAccount_whenPayPalAppApprovalURLPresent_attemptsAppSwitchWithParameters` that is already asserting on this result. As far as testing the inverse, it looks like we always hardcode the result of `application.open` to `success` here so we don't actually have any tests around the failure case: https://github.com/braintree/braintree_ios/blob/4d4f107700b1483addeae0b3b7c741014f398a5d/UnitTests/BraintreeTestShared/FakeApplication.swift#L15

scannillo:
> As far as testing the inverse, it looks like we always hardcode the result of application.open to success here so we don't actually have any tests around the failure case:

Added a test for the failure case + error in this commit 4123bcf33938ee48c99af2d4805648c4178ff441

scannillo (approved):

PR #1262, opened 2024-04-08
Update Workflows to Use Xcode 15.0.1 + Swift Version 5.9

jaxdesmarais:
### Summary of changes

- Update all workflow to use Xcode 15.0.1 - currently using Xcode 15.1 for the release for Carthage will result in merchants using that version to run into errors building with Xcode 15.0 since Xcode 15.1 is built with Swift 5.9.2 and our minimally supported version is Swift 5.9.0 ([resource](https://xcodereleases.com/))
- Bump `Package.swift` and `Braintree.podspec` to use Swift 5.9

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,23 +5,23 @@ concurrency:
   cancel-in-progress: true
 jobs:
   cocoapods:
-    name: CocoaPods (Xcode 15.1)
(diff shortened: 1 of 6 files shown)


agedd (approved):
🚀

PR #1265, opened 2024-04-09
[QL] Add Missing BA Token to Query Parameters

jaxdesmarais:
### Summary of changes

- We were not passing the `ba_token` query parameter during the App Switch flow which is required

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -351,6 +351,7 @@ import BraintreeDataCollector
 
         var urlComponents = URLComponents(url: payPalAppRedirectURL, resolvingAgainstBaseURL: true)
         urlComponents?.queryItems = [
+            URLQueryItem(name: "ba_token", value: payPalContextID),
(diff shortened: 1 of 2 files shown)


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -351,6 +351,7 @@ import BraintreeDataCollector
 
         var urlComponents = URLComponents(url: payPalAppRedirectURL, resolvingAgainstBaseURL: true)
         urlComponents?.queryItems = [
+            URLQueryItem(name: "ba_token", value: payPalContextID),
```
scannillo, line 354: This paypalContextID could be either a ba_token or ec_token, based on [this logic](https://github.com/braintree/braintree_ios/blob/761fdce14760455129c864a7970d48989a80226c/Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift#L22-L37).

We should ensure that we only proceed if a `ba_token` is present

scannillo: DTBTSDK-3781

PR #1270, opened 2024-04-10
[QL] Update plist and Handle Fallback Properly

jaxdesmarais:
### Summary of changes

- Update Braintree-Demo-Info.plist with proper scheme name - currently our check for `isPayPalAppInstalled` was always returning `false` because we need the scheme added to the `plist` for the demo app

The above lead to the next bug that I discovered:
- If we got a `paypalAppApprovalUrl` back we were always app switching, this would cause issues if the `isPayPalAppInstalled` check returned false and the app truly wasn't installed, I tested this by removing the app and ensuring we fell back to web with the new logic - it can also be tested by updating the `BTPayPalClient.payPalInAppScheme` to something like "badscheme"

### Checklist

### Authors

- @jaxdesmarais


richherrera (approved):

agedd (approved):
🚀

PR #1271, opened 2024-04-11
[QL] Consolidate Return URL Logic

jaxdesmarais:
### Summary of changes

- Rename `BTPayPalAppSwitchReturnURL` to `BTPayPalReturnURL` since this is now used for both app switch and web based flows
    - Move return URL related logic into this struct
- Rename `BTPayPalAppSwitchReturnURL_Tests` to `BTPayPalReturnURL_Tests`

### Checklist

### Authors

- @jaxdesmarais


Sources/BraintreePayPal/BTPayPalReturnURL.swift
```diff
@@ -0,0 +1,86 @@
...
+enum BTPayPalAppSwitchReturnURLState {
+    case unknownPath
+    case succeeded
+    case canceled
+}
+
+/// This class interprets URLs received from the PayPal app via app switch returns.
+///
+/// PayPal app switch authorization requests should result in success or user-initiated cancelation. These states are communicated in the url.
+struct BTPayPalReturnURL {
```
jaxdesmarais, line 16: This file was renamed but is showing additional diff. Adding logic for the web flow in the `init` as well as adding `isValidURLAction` plus `action` methods here are the actual changes.

Sources/BraintreePayPal/BTPayPalReturnURL.swift
```diff
@@ -0,0 +1,86 @@
...
+            .dropLast(1) // remove the action (`success`, `cancel`, etc)
+            .joined(separator: "/")
+
+        if hostAndPath.count > 0 {
+            hostAndPath.append("/")
+        }
+
+        /// If we are using the deeplink/ASWeb based PayPal flow we want to check that the host and path matches
+        /// the static callbackURLHostAndPath. For the universal link flow we do not care about this check.
+        if hostAndPath != BTPayPalRequest.callbackURLHostAndPath && linkType == "deeplink" {
```
jaxdesmarais, line 62: @agedd I know you mentioned this logic for this specific if block being confusing before. I added the docstrings as well as passing/checking the `linkType` here as deeplink instead of passing in the request and doing some weird stuff with that. Please let me know if this makes it clearer what this check is doing!

UnitTests/BraintreePayPalTests/BTPayPalReturnURL_Tests.swift
```diff
@@ -0,0 +1,40 @@
+import XCTest
+@testable import BraintreePayPal
+
+final class BTPayPalReturnURL_Tests: XCTestCase {
```
jaxdesmarais, line 4: Most of this test class is not new, was just renamed. The `testInitWithSchemeURL` are the only new tests since we are passing both scheme based URLs and Universal Link URLs into the `init` now.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -182,17 +182,30 @@ import BraintreeDataCollector
...
         )
-        guard let url, isValidURLAction(url: url) else {
+
+        guard let url, BTPayPalReturnURL.isValidURLAction(url: url, linkType: linkType) else {
             notifyFailure(with: BTPayPalError.invalidURLAction, completion: completion)
             return
         }
         
-        guard let response = responseDictionary(from: url) else {
+        guard !url.absoluteString.contains("cancel") else {
```
jaxdesmarais, line 191: The `responseDictionary` method was a bit odd because it was both building the dictionary to pass into the post parameters as well as doing a check that the URL did not contain `cancel`. The only part that used the return URL was the `cancel` check so I broke this out to build the dictionary here which made more sense.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -182,17 +182,30 @@ import BraintreeDataCollector
...
             payPalInstalled: payPalAppInstalled
         )
-        guard let url, isValidURLAction(url: url) else {
+
+        guard let url, BTPayPalReturnURL.isValidURLAction(url: url, linkType: linkType) else {
             notifyFailure(with: BTPayPalError.invalidURLAction, completion: completion)
             return
         }
         
-        guard let response = responseDictionary(from: url) else {
```
jaxdesmarais, line 190: The `responseDictionary` method was a bit odd because it was both building the dictionary to pass into the post parameters as well as doing a check that the URL did not contain `cancel`. The only part that used the return URL was the `cancel` check so I broke this out to build the dictionary here which made more sense.

stechiu:
I see that `BTPayPalAppSwitchReturnURLState` wasn't renamed to `BTPayPalReturnURLState`. Did you mean to leave it the same?

Sources/BraintreePayPal/BTPayPalReturnURL.swift
```diff
@@ -0,0 +1,84 @@
...
+
+/// This class interprets URLs received from the PayPal app via app switch returns.
+///
+/// PayPal app switch authorization requests should result in success or user-initiated cancelation. These states are communicated in the url.
+struct BTPayPalReturnURL {
+
+    /// The overall status of the app switch - success, cancelation, or an unknown path
+    var state: BTPayPalAppSwitchReturnURLState = .unknownPath
+
+    /// Initializes a new `BTPayPalAppSwitchReturnURL`
```
stechiu, line 21: Should this refer to `BTPayPalReturnURL`?

Sources/BraintreePayPal/BTPayPalReturnURL.swift
```diff
@@ -0,0 +1,84 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+enum BTPayPalAppSwitchReturnURLState {
```
stechiu, line 7: I see that `BTPayPalAppSwitchReturnURLState` wasn't renamed to `BTPayPalReturnURLState`. Did you mean to leave it the same?

jaxdesmarais: ✅ : https://github.com/braintree/braintree_ios/pull/1271/commits/58403e1c50e1cd7ad2cea5deae6d46a6a3c898fe

sarahkoop (approved):

stechiu (approved):

Sources/BraintreePayPal/BTPayPalReturnURL.swift
```diff
@@ -0,0 +1,84 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+enum BTPayPalReturnURLState {
+    case unknownPath
+    case succeeded
+    case canceled
```
agedd, line 10: small nit:
```suggestion
    case cancelled
```

Sources/BraintreePayPal/BTPayPalReturnURL.swift
```diff
@@ -0,0 +1,84 @@
...
+                state = .unknownPath
+            }
+        }
+    }
+
+    // MARK: - Static Methods
+
+    /// Evaluates whether the url represents a valid PayPal return URL.
+    /// - Parameter url: an app switch return URL
+    /// - Returns: `true` if the url represents a valid PayPal app switch return
```
agedd, line 40: qq: should we also note that the `url` can be a ASWebAuthenticationSession url in the docstrings here

Sources/BraintreePayPal/BTPayPalReturnURL.swift
```diff
@@ -0,0 +1,84 @@
...
+            } else {
+                state = .unknownPath
+            }
+        }
+    }
+
+    // MARK: - Static Methods
+
+    /// Evaluates whether the url represents a valid PayPal return URL.
+    /// - Parameter url: an app switch return URL
```
jaxdesmarais, line 39:
```suggestion
    /// - Parameter url: an app switch or ASWebAuthenticationSession return URL
```

Sources/BraintreePayPal/BTPayPalReturnURL.swift
agedd: this looks great! @jaxdesmarais imo definitely a lot more clearer to read :)

agedd (approved):
looking great!! 🚀

Sources/BraintreePayPal/BTPayPalReturnURL.swift
jaxdesmarais: Our SDKs always use the 1 L canceled/cancelation

KunJeongPark (approved):

Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
```diff
@@ -36,8 +36,8 @@ struct BTPayPalApprovalURLParser {
         }
     }
     
-    init?(body: BTJSON) {
-        if let payPalAppRedirectURL = body["agreementSetup"]["paypalAppApprovalUrl"].asURL() {
+    init?(body: BTJSON, linkType: String?) {
+        if linkType == "universal", let payPalAppRedirectURL = body["agreementSetup"]["paypalAppApprovalUrl"].asURL() {
```
scannillo, line 40: Nit for another time - maybe another indicator we should move this to an enum

scannillo: Actually, I don't think the `BTPayPalApprovalURLParser` class should care if the merchant has opted into universal links or not. It's only determining which URL type it got back from the GW response. 

The flow logic for falling back to the web flow already is taken care of in `BTPayPalClient`.

jaxdesmarais: As of Thursday, we were still getting the `paypalAppApprovalUrl` from the Gateway regardless of the `BTPayPalClient`. This may have been because we were working with a mock and an update may have been deployed to just allow our client side handling. Worth digging into but I was still getting this URL with the `PPBeta` app uninstalled.

PR #1272, opened 2024-04-11
[QL] Allow Nonce to be Copied for E2E Testing

jaxdesmarais:
### Summary of changes

For E2E testing our sample merchant server is not set up with the same credentials we are using for this flow. To avoid breaking all other flows, we want to be able to copy the nonce and add it to a postman request to make the transaction sale API call (this will be built out internally be another team).

- Add `nonceCompletionBlock`
    - When the status bar is tapped and this completion block is not nil we will copy this value just for this flow. All other flows will use their normal completion block and a transaction will be completed on tap of the status bar for those flows. Verified this in the demo app.

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Demo/Application/Base/ContainmentViewController.swift
+++ b/Demo/Application/Base/ContainmentViewController.swift
@@ -18,6 +18,7 @@ class ContainmentViewController: UIViewController {
             updateStatus("Presenting \(type(of: currentViewController))")
             currentViewController.progressBlock = progressBlock
             currentViewController.completionBlock = completionBlock
+            currentViewController.nonceCompletionBlock = nonceCompletionBlock
 
             appendViewController(currentViewController)
             title = currentViewController.title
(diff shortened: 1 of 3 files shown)


Demo/Application/Base/ContainmentViewController.swift
```diff
@@ -107,8 +119,15 @@ class ContainmentViewController: UIViewController {
...
 
+        if let copiedNonce {
+            UIPasteboard.general.string = copiedNonce.nonce
+            self.copiedNonce = nil
+            return
+        }
+
         if let currentPaymentMethodNonce {
             let nonce = currentPaymentMethodNonce.nonce
+
```
jaxdesmarais, line 130:
```suggestion
```

sarahkoop (approved):

agedd (approved):

PR #1274, opened 2024-04-15
[QL] Stricter Logic For BA Token for App Switch

jaxdesmarais:
### Summary of changes

- Previous PR comment: https://github.com/braintree/braintree_ios/pull/1265/files#r1562730080
- For the PayPal App Switch we only ever want a BA token currently, separate out logic to check only for BA token in this flow
- Check that BA Token is present before attempting to start the App Switch flow, if not throw an error
- Add new `BTPayPalError.missingBAToken`
- Add unit test for this updated logic

### Checklist

### Authors

- @jaxdesmarais


Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
```diff
@@ -32,6 +32,17 @@ struct BTPayPalApprovalURLParser {
                 return ecToken
             }
 
+            return nil
+
+        case .payPalApp(let url):
```
scannillo, line 37: Maybe `BTPayPalApprovalURLParser` instead have an `ecToken` & `baToken` property and the logic on which to use lives in `BTPayPalClient`?

jaxdesmarais: We can for sure, I pushed a commit here (https://github.com/braintree/braintree_ios/pull/1274/commits/d676ee2462d0ab064c19a077824c2689d78275e0) so we can decide if we like this approach. It made sense to also set the URL in the init (like we do with redirectType) since the redirectType doesn't matter for extracting these properties. Curious to hear what you think of this approach!

agedd (approved):
lgtm! 🚀

Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
```diff
@@ -17,31 +17,41 @@ enum PayPalRedirectType {
...
-            if let baToken = queryItems?.filter({ $0.name == "ba_token" }).first?.value, !baToken.isEmpty {
-                return baToken
-            } else if let ecToken = queryItems?.filter({ $0.name == "token" }).first?.value, !ecToken.isEmpty {
-                return ecToken
-            }
+    let redirectType: PayPalRedirectType
+    let url: URL
 
-            return nil
+    var ecToken: String? {
```
tdchow, line 23: Would we want to have some test coverage on the logic in this getter?

Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
```diff
@@ -17,31 +17,41 @@ enum PayPalRedirectType {
...
+
+        if let ecToken = queryItems?.filter({ $0.name == "token" }).first?.value, !ecToken.isEmpty {
+            return ecToken
         }
+
+        return nil
     }
-    
+
+    var baToken: String? {
```
tdchow, line 35: Same here, should we have test coverage for the logic here?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -332,17 +332,17 @@ import BraintreeDataCollector
                     return
                 }
                 
-                self.payPalContextID = approvalURL.pairingID
+                self.payPalContextID = approvalURL.baToken ?? approvalURL.ecToken
```
tdchow, line 335: If the `baToken` is null here should we fail fast and return? Or do we need to perform the risk call before failing?

Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
jaxdesmarais: We don't actually have any tests on this struct as a whole. It's implicitly tested through the client. Thoughts on making a follow up ticket for this work so we can add a test class for this file?

tdchow: Gotcha. I'm not sure how the iOS testing framework is set up but I think having tests cover smaller scopes makes things easier to understand and maintain.

`BTPayPalApprovalURLParser` looks like it could be a dependency that can be mocked for the client unit tests. This way it'd be easy to return expected `ecToken` values for client tests. And the logic within the parser itself can be completely separate tests.

Completely flexible on how we want to handle it! (creating a follow-up ticket or maybe getting team alignment on how we want to approach testing)

jaxdesmarais: Oh sorry if I was not clear, yes we should _absolutely_ have tests for `BTPayPalApprovalURLParser` - since we have 0 now, it expands the scope of this PR to add them all here. I created DTBTSDK-3804 to address tests as a whole for this file.

tdchow: Ah got it! Thanks!

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: We don't need to perform the risk call but only care about the `baToken` if in the app switch flow, which we don't know yet at this point. For the web based flow the `payPalContextID` here is only used for analytics.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -357,17 +357,22 @@ import BraintreeDataCollector
...
-    private func launchPayPalApp(with payPalAppRedirectURL: URL, completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void) {
+    private func launchPayPalApp(with payPalAppRedirectURL: URL, baToken: String?, completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void) {
         apiClient.sendAnalyticsEvent(
             BTPayPalAnalytics.appSwitchStarted,
             linkType: linkType,
             payPalContextID: payPalContextID,
             payPalInstalled: payPalAppInstalled
         )
 
+        guard let baToken else {
```
scannillo, line 368: Take it or leave it - making `baToken` non-optional in this method and throwing this error before `launchPayPalApp()` is called

Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
```diff
@@ -17,31 +17,41 @@ enum PayPalRedirectType {
...
-                .queryItems?
-                .compactMap { $0 }
-
-            if let baToken = queryItems?.filter({ $0.name == "ba_token" }).first?.value, !baToken.isEmpty {
-                return baToken
-            } else if let ecToken = queryItems?.filter({ $0.name == "token" }).first?.value, !ecToken.isEmpty {
-                return ecToken
-            }
+    let redirectType: PayPalRedirectType
+    let url: URL
```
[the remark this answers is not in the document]
scannillo: Actually, do we need this additional property? `PayPalRedirectType` houses the url which we previously accessed like:

```
switch redirectType {
        case .webBrowser(let url), .payPalApp(let url):
        ...
}
```

UnitTests/BraintreePayPalTests/BTPayPalApprovalURLParser_Tests.swift
```diff
@@ -0,0 +1,48 @@
+import XCTest
+@testable import BraintreeCore
+@testable import BraintreePayPal
+
+final class BTPayPalApprovalURLParser_Tests: XCTestCase {
```
scannillo, line 5: I think these things are already tested through `BTPayPalClient_Tests`. Will comment on each test below with an example of where they are covered

jaxdesmarais: These were merged in from PR #1279 based on this feedback from Tim: https://github.com/braintree/braintree_ios/pull/1274#discussion_r1568922780

UnitTests/BraintreePayPalTests/BTPayPalApprovalURLParser_Tests.swift
```diff
@@ -0,0 +1,48 @@
...
+
+    func testInit_withPayPalAppApprovalUrl_setsRedirectTypeAndBAToken() {
```
scannillo, line 7: * [Test for parsing baToken](https://github.com/braintree/braintree_ios/blob/f7ed71ea853aeee0cc77cf445068ab203aa864b9/UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift#L276-L294)
* [Test for "redirect type" / app switch was attempted](https://github.com/braintree/braintree_ios/blob/f7ed71ea853aeee0cc77cf445068ab203aa864b9/UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift#L729-L763)

Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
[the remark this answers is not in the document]
jaxdesmarais: It felt a bit odd to me to switch on the redirect type here when we only actually care about the URL. It added an additional layer of logic that felt unnecessary to me at least. If you feel strongly that we should go back to a Switch for this, I can revert to that.

UnitTests/BraintreePayPalTests/BTPayPalApprovalURLParser_Tests.swift
```diff
@@ -0,0 +1,48 @@
...
+            "agreementSetup": ["paypalAppApprovalUrl": "https://www.paypal.com?ba_token=A_FAKE_BA_TOKEN"]
+        ])
+        let approvalURLParser = BTPayPalApprovalURLParser(body: payPalAppApprovalUrlJSON, linkType: "universal")
+
+        XCTAssertEqual(approvalURLParser?.redirectType, .payPalApp(url: URL(string: "https://www.paypal.com?ba_token=A_FAKE_BA_TOKEN")!))
+        XCTAssertEqual(approvalURLParser?.baToken, "A_FAKE_BA_TOKEN")
+        XCTAssertNil(approvalURLParser?.ecToken)
+    }
+
+    func testInit_withRedirectUrl_setsRedirectTypeAndECToken() {
```
scannillo, line 18: * [Test for parsing ecToken](https://github.com/braintree/braintree_ios/blob/f7ed71ea853aeee0cc77cf445068ab203aa864b9/UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift#L214-L230) from redirectURL

UnitTests/BraintreePayPalTests/BTPayPalApprovalURLParser_Tests.swift
```diff
@@ -0,0 +1,48 @@
...
+            "agreementSetup": ["approvalUrl": "https://www.paypal.com/agreements/approve?ba_token=A_FAKE_BA_TOKEN"]
+        ])
+        let approvalURLParser = BTPayPalApprovalURLParser(body: approvalUrlJSON, linkType: "universal")
+
+        XCTAssertEqual(approvalURLParser?.redirectType, .webBrowser(url: URL(string: "https://www.paypal.com/agreements/approve?ba_token=A_FAKE_BA_TOKEN")!))
+        XCTAssertEqual(approvalURLParser?.baToken, "A_FAKE_BA_TOKEN")
+        XCTAssertNil(approvalURLParser?.ecToken)
+    }
+
+    func testInit_withBadJSON_returnsNil() {
```
scannillo, line 40: * [Test for invalid approvalURLs](https://github.com/braintree/braintree_ios/blob/f7ed71ea853aeee0cc77cf445068ab203aa864b9/UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift#L191-L212)

UnitTests/BraintreePayPalTests/BTPayPalApprovalURLParser_Tests.swift
```diff
@@ -0,0 +1,48 @@
...
+            "paymentResource": ["redirectUrl": "https://www.paypal.com/checkout?token=A_FAKE_EC_TOKEN"]
+        ])
+        let approvalURLParser = BTPayPalApprovalURLParser(body: redirectUrlJSON, linkType: nil)
+
+        XCTAssertEqual(approvalURLParser?.redirectType, .webBrowser(url: URL(string: "https://www.paypal.com/checkout?token=A_FAKE_EC_TOKEN")!))
+        XCTAssertNil(approvalURLParser?.baToken)
+        XCTAssertEqual(approvalURLParser?.ecToken, "A_FAKE_EC_TOKEN")
+    }
+
+    func testInit_withApprovalUrl_setsRedirectTypeAndECToken() {
```
scannillo, line 29: This one I don't see a 1:1 equivalent for

Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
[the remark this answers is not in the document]
scannillo: Nope, works for me! I think we can still make it private though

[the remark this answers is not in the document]
jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/1274/commits/2cd9b40c2b112c97329a9fc58907fe80ed3c1901

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: ✅ : https://github.com/braintree/braintree_ios/pull/1274/commits/225f6226f797b0eeb238e06c3d2c90ed71840d5d

UnitTests/BraintreePayPalTests/BTPayPalApprovalURLParser_Tests.swift
jaxdesmarais: Added `testTokenize_whenApprovalUrlContainsBAToken_sendsBATokenAsPayPalContextIDInAnalytics` here: https://github.com/braintree/braintree_ios/pull/1274/commits/fabe9984f68cc600f95d0eee3b46d3a3fbd1df7f.

UnitTests/BraintreePayPalTests/BTPayPalApprovalURLParser_Tests.swift
jaxdesmarais: That's fine by me, removed here: https://github.com/braintree/braintree_ios/pull/1274/commits/fabe9984f68cc600f95d0eee3b46d3a3fbd1df7f

scannillo (approved):

PR #1281, opened 2024-04-22
`Do not embed` Core in `Shopper Insights` framework

richherrera:
### Summary of changes

- Change the embedding method used in the `Shopper Insights` framework to resolve distribution issues with the Demo app. This will standardize the embedding approach of the `Braintree Core` framework with those used in the other frameworks within the Demo app

### Checklist

### Authors

- @richherrera


richherrera:
> > resolve distribution issues with the Demo app
> 
> Just curious what distribution issues we saw that this change resolves

Sure, this issue arose when trying to upload the Demo App to TestFlight. Upon review, we noticed that the frameworks that depend on Core are configured as `Do Not Embed`. `Shopper Insights` is the only one with the dependency set to `Embed & Sign`

Screenshot 2024-04-24 at 9 40 43 a m

jaxdesmarais (approved):
Appreciate you sharing the error you saw! 👏

PR #1234, opened 2024-03-23
Open PayPal App URL if present in BT GW response

scannillo:
DTBTSDK-3612

### Summary of changes

- Move `/v1/paypal_hermes/create_payment_resource` & `/v1/paypal_hermes/setup_billing_agreement` URL response parsing into new file `BTPayPalApprovalURLParser`
    - Remove manual query param parsing logic in favor of Apple's built-in `URLComponents`
-  Add logic path for opening PayPal app if `paypalAppApprovalUrl` response param present from gateway

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

PR #1285, opened 2024-04-26
Fix millisecond timestamp rounding bug in analytics

scannillo:
### Summary of changes

- Fix small bug where analytics timestamp was always rounding milliseconds up
- See same manual testing done in [this PR](https://github.com/braintree/braintree_ios/pull/1234#discussion_r1538054171)
    - Unit testing for this rounding IMO requires more trouble than it's worth (creating & injecting a mock of `Date`) 

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
+++ b/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
@@ -53,7 +53,7 @@ class BTAnalyticsService: Equatable {
         linkType: String? = nil,
         payPalContextID: String? = nil
     ) async {
-        let timestampInMilliseconds = UInt64(Date().timeIntervalSince1970 * 1000)
+        let timestampInMilliseconds = Int(round(Date().timeIntervalSince1970 * 1000))
         let event = FPTIBatchData.Event(
             correlationID: correlationID,
             errorDescription: errorDescription,


sarahkoop (approved):

sshropshire (approved):

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -53,7 +53,7 @@ class BTAnalyticsService: Equatable {
         linkType: String? = nil,
         payPalContextID: String? = nil
     ) async {
-        let timestampInMilliseconds = UInt64(Date().timeIntervalSince1970 * 1000)
+        let timestampInMilliseconds = Int(round(Date().timeIntervalSince1970 * 1000))
```
sshropshire, line 56: nit: It might be good to keep `UInt64` explicit here, but at the same time there aren't any [32-bit iOS devices](https://stackoverflow.com/a/27440158) out there anymore so it'll implicitly be a 64-bit number. 32-bit timestamps are the new [Y2K bug](https://www.linkedin.com/pulse/2038-next-y2k-bug-legacy-time-overflow-issues-john-o-shea-948pf) for some machines.

warmkesselj (approved):
🍏

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
scannillo: Interesting! Yeh I think Apple will cover us here, according to their [docs for `Int`](https://developer.apple.com/documentation/swift/int#overview)

>  On 32-bit platforms, Int is the same size as Int32, and on 64-bit platforms, Int is the same size as Int64.

PR #1287, opened 2024-04-30
Remove config caching throttle delay

scannillo:
_Edited, please re-read:_

### Summary

**GOAL**: 
The main goal of this PR is to remove the 100ms delay added in PR https://github.com/braintree/braintree_ios/pull/807 to our configuration fetching. At the same time, we want to make sure we don't re-introduce the increased "Network connection lost" errors that PR https://github.com/braintree/braintree_ios/pull/807 aimed to solve.

**PROBLEM**: 
The problem with this 100ms delay is that it occurs every time we `BTAPIClient.fetchOrReturnRemoteConfig()`, which is many times throughout the SDK. For example: in the PP Vault flow, this method is called a total of 6 times. This could be adding **up to 600ms of delay**.

**BACKGROUND**: 
We know that the URLCache logic implemented in PR https://github.com/braintree/braintree_ios/pull/789 caused an LE merchant to see an increase in "Network connection lost" errors. PR https://github.com/braintree/braintree_ios/pull/807 reduced the frequency of those errors by delaying all retrievals from the URLCache by 100ms.

This bug was hard to replicate consistently, as described by Jax below.

**THEORY**: 
After digging in about how URLCache works, I learned that it is a highly asynchronous queue. From [this SO post](https://stackoverflow.com/questions/52938033/urlresponse-is-not-retrieved-after-storing-in-cache-using-storecachedresponse), "You have to let the main run loop return before it will be available, and possibly even wait a little while". 

I was able to write failing unit tests to witness this lag in URLCache, see commit bb7e2f21251d70f83d9e2195fe0c4daa47210c2b. Putting things in the queue and asserting their existence synchronously always failed. _Until_ I added a DispatchQueue with a ~100+ms delay on the assertion statements, then the tests started passing. Uncanny! To me, this is a signal that our fix in #807 was addressing a _symptom_ of the cache being updated too slowly and not the root issue.

**PROPOSED CHANGES**:
In this PR, I move away from URLCache and to a basic NSCache setup via a singleton. NSCache is thread-safe, auto-evicts if app memory overloads, and is fast-access since it is in-memory storage only. I mirrored some of the logic in our Android SDK ([see code](https://github.com/braintree/braintree_android/blob/main/BraintreeCore/src/main/java/com/braintreepayments/api/ConfigurationCache.kt)). The failing unit tests from commit bb7e2f21251d70f83d9e2195fe0c4daa47210c2b are now able to pass, resolving the issue of URLCache working too slowly for our needs.

I feel like this knowledge & fix + the fact that [0.19% of Venmo iOS users](https://github.com/braintree/braintree_ios/pull/1287#issuecomment-2088657438) are on old app versions (pre-loading screen), we are not at a risk to re-introduce the "Network connection lost" issue.

### Changes

- Add `ConfigurationCache` with simplified, custom caching logic
- Remove caching logic from `BTHTTP`
    - Call caching functions directly from `BTAPIClient.fetchOrReturnRemoteConfig()`
    - (Mirrors Android SDK)
- Cleanup
    - Remove unused `URLSession` config injected at `APIClient` level (this was never caching locally since `v1/configuration` has no Cache Control headers)

### Testing

After extensive local testing with Xcode Instruments signposts, we are seeing a range of 15ms - 550ms reduction in elapsed time with this change. Across all trials, we see an average of **-247ms** reduction in elapsed time.

### Checklist

- [x] Added a changelog entry

### Authors
@scannillo


Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -18,17 +18,17 @@ class BTHTTP: NSObject, NSCopying, URLSessionDelegate {
...
 
     /// Session exposed for testing
     lazy var session: URLSession = {
         let configuration: URLSessionConfiguration = URLSessionConfiguration.ephemeral
         configuration.httpAdditionalHeaders = defaultHeaders
+        configuration.requestCachePolicy = .returnCacheDataElseLoad
         
         let delegateQueue: OperationQueue = OperationQueue()
         delegateQueue.name = "com.braintreepayments.BTHTTP"
-        delegateQueue.maxConcurrentOperationCount = OperationQueue.defaultMaxConcurrentOperationCount
```
scannillo, line 31: Removing, since this is the default value.

scannillo:
@sarahkoop or @jaxdesmarais - do you remember which combination of iOS versions & Venmo app versions that this delay was initially added for? With the current Venmo app I'm not able to replicate any issues. I'm hoping since [this PR](https://github.com/braintree/braintree_ios/pull/807) is 2 years old that the version of the Venmo app in question then has minimal to no active usage now.

sarahkoop:
> @sarahkoop or @jaxdesmarais - do you remember which combination of iOS versions & Venmo app versions that this delay was initially added for? With the current Venmo app I'm not able to replicate any issues. I'm hoping since [this PR](https://github.com/braintree/braintree_ios/pull/807) is 2 years old that the version of the Venmo app in question then has minimal to no active usage now.

It was definitely older versions, but I can't find a list of actual version numbers. I think it's likely very minimal, if any, traffic, but might be worth confirming with the Venmo team

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -137,63 +132,32 @@ class BTHTTP: NSObject, NSCopying, URLSessionDelegate {
...
-
-            if self.cacheDateValidator.isCacheInvalid(cachedResponse ?? nil) {
-                URLCache.shared.removeAllCachedResponses()
-                cachedResponse = nil
-            }
-
-            // The increase in speed of API calls with cached configuration caused an increase in "network connection lost" errors.
-            // Adding this delay allows us to throttle the network requests slightly to reduce load on the servers and decrease connection lost errors.
-            DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
-                if let cachedResponse = cachedResponse {
```
KunJeongPark, line 162: Is this the main difference in this PR? 
Are there still concerns about "network connection lost" errors?

scannillo:
For some reason the Amex & 3DS UI tests are broken from this PR, but the Demo features are working OK on their own. Looking into it 🤔

scannillo:
Ok, here's our breakdown of iOS version usage across the BT SDKs from Ian's dashboard:

iOS 17 = 81.95%
iOS 16 = 14.89%
iOS 15 = 1.50%
iOS 14 = 0.04%

scannillo:
> It was definitely older versions, but I can't find a list of actual version numbers. I think it's likely very minimal, if any, traffic, but might be worth confirming with the Venmo team

I asked them in Slack! They said they'll have to do some digging to see which version the loading screen was added in.

I also found this in an old thread from you and Jax. It helps give an idea of what versions you tested on:

> iPhone 12 mini + iOS 15.4.1 + Venmo app version 9.18.0: 0 failures in 100 attempts with fix
iPhone 6 + iOS 13.2.2 + Venmo app version 7.42.0: 1 in 100 with fix, 50 in 100 on 5.7.0, 1 in 100 on 5.5.0
Phone X, iOS 14.0.1, Venmo app version 8.8.0: 0 in 60 with fix (Venmo app force updated), 8 in 100 on 5.7.0, 1 in 100 on 5.5.0
iPhone X, iOS 14.0.1, Venmo app version 9.18.0: 1 in 50 with fix

Sources/BraintreeCore/BTHTTP.swift
jaxdesmarais: > TBH - I am still struggling to understand how this 100ms delay fixed anything. What it was doing was adding 100ms of delay to v1/configuration calls, as well as 100ms of delay if the config was cached. It was not adding 100ms delay to other API calls.

This is correct, the issue was not subsequent API calls, the issue was that when we started caching properly, we were returning from the venmo app without the loading screen (older version) _before_ we got the response back from the config API call. This resulted in the config network connection being lost during the switch/return.

scannillo: Hm, I'm still a little confused. So the config network call is what was seeing "network connection lost"? But it sounds like in this error case, the config was cached prior to launching Venmo, so why were we doing another config network call?

It sounds like a concurrency issue (Steven alluded to [here](https://github.com/braintree/braintree_ios/pull/807#pullrequestreview-942345074)), and the problem was that we weren't always properly waiting for a config value to be returned (whether from cache or network) before launching Venmo. My guess is multiple threads were updating the cache & removing items from it at the same time, and we "thought" it had a value in it, but by the time Venmo finished, another thread already wiped it.

scannillo: Curious - how did you run those 100 tests? I imagine you used some automated tool in Xcode to run them in parallel?

scannillo:
Another update - I checked with Venmo and v9.18.0 introduced the initial spinner/loading screen for the app switch checkout flow. I just dug into their app usage stats, and versions 9.17.0 and _earlier_ make up 0.254% of active Venmo iOS users.

> Should we consider letting DoorDash know we are making this update? This was a P0 at the time so I feel like we should consider giving them some sort of heads up to ramp this update on their end.

This is a good call. Ideally we don't have to loop them in, since if we deem the traffic high enough, we will do our best to make sure this issue doesn't resurface. I am thinking about implementing a serial queue on the URLCache implementation _and_ removing the timeout as a way to solve both problems.

Sources/BraintreeCore/BTHTTP.swift
jaxdesmarais: Since the Venmo app can only be installed on physical devices all of our testing was manual. We basically both ran as many tests as quick as possible, so as soon as we returned and got a nonce we kicked off the flow again. At the time the Venmo team was not able to help with testing and we got very little information from the merchant on any device or version specifics.

scannillo:
Just had a 💡 moment. `main` wasn't only adding 100ms of delay to each network request to `v1/configuration`, it was adding 100ms of delay [every time](https://github.com/braintree/braintree_ios/blob/83a5709a06bca8ceb3b48f842ebefd4016294047/Sources/BraintreeCore/BTHTTP.swift#L161) `fetchOrReturnRemoteConfig()` is called, wether it was a cache hit or miss. That method is called **6 total times** from instantiating a feature client --> ASWebSession.start() in the PP Vault flow. That's ... 600 ms added.

Safe to say, this PR should severely improve latency! 

Also shoutout to @KunJeongPark - who helped with timing analysis that made this issue more obvious. 👏 The more folks we can have A/B test the two branches and add their result to our Excel doc, the better.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -143,8 +121,15 @@ import Foundation
         if let clientToken {
             configPath = clientToken.configURL.absoluteString
         }
+        
+        let authCredential = clientToken?.authorizationFingerprint ?? tokenizationKey
+        if let authCredential,
```
scannillo, line 126: I acknowledge this is a little ugly, but would rather tackle it in another PR - to move away from _both_ the clientToken & tokenizationKey being optional and leverage the `BTAPIClientAuthorization` enum with an associated string value instead.

This PR has a lot of change, though so I am opting to do it later

Sources/BraintreeCore/ConfigurationCache.swift
```diff
@@ -0,0 +1,50 @@
...
+class ConfigurationCache {
+        
+    // MARK: - Private Properties
+    
+    private let timeToLiveMinutes = 5.0
+    
+    // MARK: - Internal Properties
+    
+    /// Exposed for testing
+    var cachedConfigStorage: [String: BTConfiguration] = [:]
```
scannillo, line 12: Note to self - see if NSCache functions synchronously. It offers auto eviction settings.

If not - implement max size on this dictionary.

scannillo: Moved from Dictionary to NSCache for the following reasons:
* Dictionary would require manual memory management. NSCache auto evicts if app memory consumption is high.
* NSCache is thread-safe. I was able to trigger a race condition using a very basic Dictionary setup. It would require more manual code to add locking.
* We can specify a max # of items, and max size in bytes of our NSCache, if we want

Sources/BraintreeCore/BTHTTP.swift
scannillo: Resolving this comment in favor of discussion in PR description. Victoria, if you have more open questions on the original issue please let me know and re-open the thread!

Sources/BraintreeCore/ConfigurationCache.swift
```diff
@@ -0,0 +1,50 @@
+import Foundation
+
+class ConfigurationCache {
+        
+    // MARK: - Private Properties
+    
+    private let timeToLiveMinutes = 5.0
```
agedd, line 7: curious why we set this to 5 min 👀

Sources/BraintreeCore/BTAPIClient.swift
jaxdesmarais: Do we think for now we could clean this up slightly by using a `guard` here:
```swift
guard let authorization = clientToken?.authorizationFingerprint ?? tokenizationKey else {
    completion(nil, BTAPIClientError.invalidAuthorization)
    return
}
```

Sources/BraintreeCore/ConfigurationCache.swift
jaxdesmarais: It's been set as 5 minutes for a while on Android, so we followed suit on iOS. We figured 5 minutes was a reasonable time from a merchant configuring an APIClient to actually using the APIClient in a feature client to check out. But this can be whatever we want as long as it's consistent across platforms.

Sources/BraintreeCore/ConfigurationCache.swift
```diff
@@ -0,0 +1,50 @@
...
+    
+    // MARK: - Internal Properties
+    
+    /// Exposed for testing
+    let cacheInstance = NSCache<NSString, BTConfiguration>()
+
+    // MARK: - Singleton Setup
+    
+    static let shared = ConfigurationCache()
+    private init() { }
```
jaxdesmarais, line 17: I think since we are providing defaults to the properties this can be removed, right?

Sources/BraintreeCore/ConfigurationCache.swift
```diff
@@ -0,0 +1,50 @@
...
+    
+    // MARK: - Internal Methods
+    
+    func putInCache(authorization: String, configuration: BTConfiguration) throws {
+        let cacheKey = try createCacheKey(authorization)
+        cacheInstance.setObject(configuration, forKey: cacheKey)
+    }
+    
+    func getFromCache(authorization: String) throws -> BTConfiguration? {
```
jaxdesmarais, line 26: Since this method throws can we throw and error here instead of returning nil then make BTConfiguration non optional?

jaxdesmarais: Ah I see we are returning nil because at the callsite if we don't have a cached config we want to fetch a new config. This feels a little odd to me since this method is called `getFromCache` I would expect us to always have the ability to return a config at this point. 🤔

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -365,15 +320,6 @@ class BTHTTP: NSObject, NSCopying, URLSessionDelegate {
             return
         }
 
-        // We should only cache the response if we do not have an error and status code is 2xx
-        let successStatusCode: Bool = httpResponse.statusCode >= 200 && httpResponse.statusCode < 300
```
jaxdesmarais, line 369: Do we need to have this logic live elsewhere or is there no chance we could cache a config with a bad status code?

scannillo: We don't anymore, since now we're caching at the BTAPIClient level, and not the BTHTTP level. We only add a config to the cache if we can [successfully construct a Config object](https://github.com/braintree/braintree_ios/blob/95eb1b379bc568ee8a7a86f5369351b9e63c5320/Sources/BraintreeCore/BTAPIClient.swift#L149-L152), after the HTTP call completes

Sources/BraintreeCore/ConfigurationCache.swift
scannillo: Good callout. The "throw" is bubbled up from if the base64 encoding fails. These methods don't _have_ to throw, the throw is always handled by APIClient and never bubbles up to the merchant, but I thought it was more clear to have "throw" to mean an error occurred & `nil` to mean there's actually no object in the cache. We _could_ lump both cases into a `nil` response, but that felt more Obj-C-ey.

Maybe there's a better name than `getFromCache`? `checkInCache`?

Open to ideas for making this public API more clear!

scannillo: (Also added some docstrings)

Sources/BraintreeCore/ConfigurationCache.swift
scannillo: The `private init()`? This prevents other parts of the codebase from being able to instantiate an instance of `ConfigurationCache`, making it a true singleton.

jaxdesmarais (approved):

richherrera (approved):
Tested on a physical device and works as expected!!! 🚀

PR #1289, opened 2024-04-30
Add `venmo_installed Analytics` Tag

KunJeongPark:
### Summary of changes

- Add `venmo_installed` analytics tag

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Sources/BraintreeCore/Analytics/FPTIBatchData.swift
+++ b/Sources/BraintreeCore/Analytics/FPTIBatchData.swift
@@ -136,7 +136,9 @@ struct FPTIBatchData: Codable {
     }
 
     private static func isVenmoAppInstalled() -> Bool {
-        let venmoURL = URL(string: "com.venmo.touch.v2://")!
+        guard let venmoURL = URL(string: "\(BTCoreConstants.venmoScheme)://") else {
+            return false
+        }
         return UIApplication.shared.canOpenURL(venmoURL)
     }
 }
--- a/Sources/BraintreeCore/Analytics/FPTIBatchData.swift
+++ b/Sources/BraintreeCore/Analytics/FPTIBatchData.swift
@@ -37,7 +37,7 @@ struct FPTIBatchData: Codable {
         let payPalContextID: String?
         let timestamp: String
         let tenantName: String = "Braintree"
-        let venmoInstalled: Bool
+        let venmoInstalled: Bool = isVenmoAppInstalled()
 
         enum CodingKeys: String, CodingKey {
             case correlationID = "correlation_id"
@@ -48,16 +48,6 @@ struct FPTIBatchData: Codable {
             case timestamp = "t"
             case tenantName = "tenant_name"
             case venmoInstalled = "venmo_installed"
-        }
-
-        init(correlationID: String?, errorDescription: String?, eventName: String, linkType: String?, payPalContextID: String?, timestamp: String) {
-            self.correlationID = correlationID
-            self.errorDescription = errorDescription
-            self.eventName = eventName
-            self.linkType = linkType
-            self.payPalContextID = payPalContextID
-            self.timestamp = timestamp
-            self.venmoInstalled = FPTIBatchData.isVenmoAppInstalled()
         }
     }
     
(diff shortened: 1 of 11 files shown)


Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -37,6 +37,7 @@ struct FPTIBatchData: Codable {
         let payPalContextID: String?
         let timestamp: String
         let tenantName: String = "Braintree"
+        let venmoInstalled: Bool?
```
scannillo, line 40: IMO we can remove all of the passing down of this boolean from the feature clients and just do the app installed check here. (it would send for every analytic event, but I think that's fine)

KunJeongPark: So just rewrite what we have in `isVenmoAppInstalled` function of `BTVenmoClient` in `FPTIBatchData`?

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -37,6 +37,7 @@ struct FPTIBatchData: Codable {
         let payPalContextID: String?
         let timestamp: String
         let tenantName: String = "Braintree"
+        let venmoInstalled: Bool
```
jaxdesmarais, line 40: Looking at the tag in the tag directory of FPTI it looks like the expected underlying type here is a string.

KunJeongPark: I think it gets converted to string in JSON.

jaxdesmarais: Adding a breakpoint in the `toDictionary` method shows that is is sent as a `Bool` to FPTI, but it seems like that is okay. We do the same for `is_simulator` which is also listed as a `String` type in the tag dictionary. As long as we are seeing this new tag as expected in FPTI we should be okay.

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
KunJeongPark: I was also wondering if there is a performance cost to Venmo install check each time analytic event is created.

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -132,4 +134,9 @@ struct FPTIBatchData: Codable {
             case tokenizationKey = "tokenization_key"
         }
     }
+
+    private static func isVenmoAppInstalled() -> Bool {
+        let venmoURL = URL(string: "com.venmo.touch.v2://")!
```
scannillo, line 139: Can we move this into the `BTCoreConstants` file? This way we can access it in here & both VenmoClient, and reduce risk of copy paste errors

scannillo, line 139: Also - avoid `!` force unwrap for safety. If the URL creation fails, we can just not set this boolean for analytics.

KunJeongPark: I thought you wanted to make this always available, true or false, not an optional.

Sources/BraintreeCore/BTCoreConstants.swift
```diff
@@ -10,6 +10,8 @@ import Foundation
     /// :nodoc: This property is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     public static let callbackURLScheme: String = "sdk.ios.braintree"
 
+    public static let venmoScheme: String = "com.venmo.touch.v2"
```
agedd, line 13: small nit (take it or leave it): 
```suggestion
    public static let venmoURLScheme: String = "com.venmo.touch.v2"
```

scannillo (approved):

Sources/BraintreeCore/BTCoreConstants.swift
jaxdesmarais: ➕ 1 agree that adding URL would make this more clear

jaxdesmarais (approved):

richherrera (approved):
Besides Alekhya's comment, LGTM!!

PR #1290, opened 2024-05-03
Add `scene(_:willConnectTo:options:)` to the MockVenmo App

richherrera:
### Summary of changes

- The UITs sometimes failed because the MockVenmo app was not running, causing `AppSwitcher. openVenmoURL` to be `nil`. This PR adds `scene(_:willConnectTo:options:)`. When the app is launched for the first time, it calls scene(_:willConnectTo:options:)

[Apple Docs:](https://developer.apple.com/documentation/xcode/defining-a-custom-url-scheme-for-your-app)
> If your app has opted into Scenes, and your app is not running, the system delivers the URL to the scene(_:willConnectTo:options:) delegate method after launch, and to scene(_:openURLContexts:) when your app opens a URL while running or suspended in memory.

### Checklist

- [ ] Added a changelog entry

### Authors

- @richherrera


jaxdesmarais:
Added this logic to #1345 - closing in favor of that PR that fully fixes UI Test flakiness with some additional changes

PR #1292, opened 2024-05-06
[QL] Measure API Request Latency for PayPal Flows

jaxdesmarais:
### Summary of changes

- Send `start_time`, `end_time`, and `endpoint` to FPTI for tracking API request latency in the PayPal flow
- Add new `BTCoreAnalytics` enum
- Add new `BTAPITimingDelegate` for sending start/end timing from `BTHTTP` to `BTAPIClient`
- Update `UnitTests`

### Checklist

- [x] Added a changelog entry

- @scannillo @jaxdesmarais


Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -1,8 +1,12 @@
 import Foundation
 import Security
 
+protocol BTAPITimingDelegate: AnyObject {
```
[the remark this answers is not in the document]
scannillo: Another nit, from [Swift style guide](https://www.swift.org/documentation/api-design-guidelines/):
> Protocols that describe what something is should read as nouns (e.g. Collection).

> Protocols that describe a capability should be named using the suffixes able, ible, or ing (e.g. Equatable, ProgressReporting).

Do we want to keep `Delegate` in the name? I know Apple still does this. I think `HTTPTiming` or `HTTPNetworkTiming` might be more clear?

Sources/BraintreeCore/Date+MillisecondTimestamp.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+extension Date {
+    
+    var utcTimpestampMilliseconds: Int {
```
scannillo, line 5: We use this in a few places so I moved it into an extension. If you hate it, feel free to revert lol

UnitTests/BraintreeTestShared/MockAPIClient.swift
```diff
@@ -92,6 +93,10 @@ public class MockAPIClient: BTAPIClient {
         postedPayPalContextID = payPalContextID
         postedLinkType = linkType
         postedAnalyticsEvents.append(name)
+
+        if postedShouldSendAPIRequestLatency {
+            postedAnalyticsEvents.append(BTCoreAnalytics.apiRequestLatency)
+        }
```
scannillo, line 99: Because of this, our assertions added in `BTPayPalClient_Tests` are only testing this Mock behavior, and not executing any of our new code added in BTAPIClient & BTHTTP.

We _could_ override `fetchAPITiming()` in this Mock class instead, but it'll be a similar situation of only passing based on added Mock code.

I would update BTAPIClient & BHTTP tests instead, and then the best we can do at the PayPalClient level is assert that `shouldSendAPIRequestLatency` is true.

Sources/BraintreeCore/BTHTTP.swift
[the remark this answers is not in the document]
jaxdesmarais: Good call, updated here: https://github.com/braintree/braintree_ios/pull/1292/commits/116d09590f8fda77152b0deab04cd35f554c77e5

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -457,4 +463,21 @@ class BTHTTP: NSObject, NSCopying, URLSessionDelegate {
...
         }
     }
+
+    func urlSession(_ session: URLSession, task: URLSessionTask, didFinishCollecting metrics: URLSessionTaskMetrics) {
+        metrics.transactionMetrics.forEach { transaction in
+            if let startDate = transaction.requestStartDate,
+               let endDate = transaction.responseEndDate,
+               let path = transaction.request.url?.path,
+               path != "/v1/tracking/batch/events" {
+                let cleanedPath = path.replacingOccurrences(of: "/merchants/([A-Za-z0-9]+)/client_api", with: "", options: .regularExpression)
```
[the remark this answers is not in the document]
jaxdesmarais: For sure, updated here: https://github.com/braintree/braintree_ios/pull/1292/commits/550c7226c10b42e62406f0d36f9953098ab4e372

Sources/BraintreeCore/Date+MillisecondTimestamp.swift
agedd, line 5: small nit: 
```suggestion
    var utcTimestampMilliseconds: Int {
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -45,6 +45,8 @@ import BraintreeDataCollector
     /// - Parameter apiClient: The API Client
     @objc(initWithAPIClient:)
     public init(apiClient: BTAPIClient) {
+        apiClient.shouldSendAPIRequestLatency = true
```
agedd, line 48: qq: do we always want this to be set to true?

agedd, line 48: qq: do we always want this to be set to true? 👀

Sources/BraintreeCore/Date+MillisecondTimestamp.swift
jaxdesmarais: I had noticed the same, updated here: https://github.com/braintree/braintree_ios/pull/1292/commits/e2aa0ea330939b6cb7ffec336c88fea90167854c

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: For PayPal currently yes we want to send this for all `BTPayPalClient` flows so we can track this across both current and future QL initiatives. This branch will get merged into our other feature branches to ensure we are tracking these events there as well.

UnitTests/BraintreeTestShared/MockAPIClient.swift
jaxdesmarais: I updated `BTPayPalClient_Tests` to just test setting `shouldSendAPIRequestLatency` here: https://github.com/braintree/braintree_ios/pull/1292/commits/8af3d0b4452f740f97b6feb73d4fa9ad549defba. For `BTAPIClient_Tests` we are going to run into the same issue, currently those tests when testing analytics ([example here](https://github.com/braintree/braintree_ios/blob/2120f0b705d4fd1f6173b431d71a7d9705a7d096/UnitTests/BraintreeCoreTests/BTAPIClient_Tests.swift#L406-L414)) use `FakeAnalyticsService` ([here](https://github.com/braintree/braintree_ios/blob/main/UnitTests/BraintreeCoreTests/Analytics/FakeAnalyticsService.swift)) which just sends whatever event you tell it to send. So we would end up just testing our mock there as well.

scannillo:
Based on[ this diagram ](https://developer.apple.com/documentation/foundation/urlsessiontasktransactionmetrics#3162615)from Apple’s docs, maybe requestStart isn’t what we want, but rather fetchStart? The DNS lookup, TCP & TLS setup aren’t included in requestStart

jaxdesmarais:
> Based on[ this diagram ](https://developer.apple.com/documentation/foundation/urlsessiontasktransactionmetrics#3162615) from Apple’s docs, maybe requestStart isn’t what we want, but rather fetchStart?
> 
> The DNS lookup, TCP & TLS setup aren’t included in requestStart.

Nice find, updated here: https://github.com/braintree/braintree_ios/pull/1292/commits/7ffd8ab6c1a473264a6e4ce9d090dc859b260c99

UnitTests/BraintreeTestShared/MockAPIClient.swift
scannillo: But we can assert that `FakeAnalyticsService.sendAnalyticsEvent()` was called with the expected values, to test the logic that lives within `BTAPIClient.fetchAPITiming()`.

Rough outline like:
* when networkTimingDelegate.fetchAPITiming() delegate method invoked, sends analytic event
* when delegate method invoked, and analytics path, does not send analytic event
* when delegate method invoked, and shouldSendAPILatency = false, does not send analytic event
* when delegate method invoked, and path contains merchantID, strips merchantID from posted JSON body tag param

The easiest way would be to call `sut.fetchAPITiming()` directly in each test. The probably "right" way would be to set a MockTimingDelegate on sut and trigger the delegate call that way via the mock. Not sure it's worth the setup for the later option at this point though.

jaxdesmarais: > But we can assert that FakeAnalyticsService.sendAnalyticsEvent() was called with the expected values, to test the logic that lives within BTAPIClient.fetchAPITiming().

This method does not behave at least how I would expect, because my initial thought was the same. If we take a look at the test `testSendAnalyticsEvent_whenCalled_callsAnalyticsService` ([here](https://github.com/braintree/braintree_ios/blob/2120f0b705d4fd1f6173b431d71a7d9705a7d096/UnitTests/BraintreeCoreTests/BTAPIClient_Tests.swift#L406-L414)) it is calling `FakeAnalyticsService.sendAnalyticsEvent()` with `blahblah` and asserting that the `mockAnalyticsService.lastEvent` equals `blahblah`. 

Which will always be true because that's the input we are giving it. While it is an override we are not actually hitting the override unless we call it directly so our call to `BTAPIClient.fetchAPITiming()` never actually hits `FakeAnalyticsService.sendAnalyticsEvent()`. We can take a look at this a bit later if that doesn't make sense.

jaxdesmarais: Paired on this because I was overthinking it, updated here: https://github.com/braintree/braintree_ios/pull/1292/commits/ff199133838acd50aa135b63a273fa2e31103b2c

UnitTests/BraintreeTestShared/MockAPIClient.swift
```diff
@@ -13,6 +13,7 @@ public class MockAPIClient: BTAPIClient {
     public var postedAnalyticsEvents : [String] = []
     public var postedPayPalContextID: String? = nil
     public var postedLinkType: String? = nil
+    public var postedShouldSendAPIRequestLatency: Bool = false
```
scannillo, line 16: We can remove this

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -18,6 +18,7 @@ class BTPayPalClient_Tests: XCTestCase {
         mockAPIClient.cannedResponseBody = BTJSON(value: [
             "paymentResource": ["redirectUrl": "http://fakeURL.com"]
         ])
+        mockAPIClient.postedShouldSendAPIRequestLatency = true
```
scannillo, line 21: We can remove this

scannillo (approved):
Changes look good! But would mark as "DO NOT MERGE" until #1295 is merged.

UnitTests/BraintreeTestShared/MockAPIClient.swift
jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/1292/commits/58768861720cbaa7e3c1b16aba7fb6102b82b0c7

scannillo:
One thing came to my mind - once the batching logic is in place, maybe we should just send this analytic event for all API requests (not just PP)? Since our batching is time based and not based on the # of events, I don't think it will have too much impact on consuming too much networking bandwidth.

Do we think those metrics would also be useful to have?

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -467,8 +473,10 @@ import Foundation
 
             if let clientToken, let baseURL {
                 http = BTHTTP(url: baseURL, authorizationFingerprint: clientToken.authorizationFingerprint)
+                http?.networkTimingDelegate = self
             } else if let tokenizationKey, let baseURL {
                 http = BTHTTP(url: baseURL, tokenizationKey: tokenizationKey)
+                http?.networkTimingDelegate = self
             }
```
richherrera, line 480:
```suggestion
            if let clientToken, let baseURL {
                http = BTHTTP(url: baseURL, authorizationFingerprint: clientToken.authorizationFingerprint)
            } else if let tokenizationKey, let baseURL {
                http = BTHTTP(url: baseURL, tokenizationKey: tokenizationKey)
            }
            
            http?.networkTimingDelegate = self
```

richherrera (approved):

agedd (approved):
lgtm! 🚀

scannillo:
@jaxdesmarais following up here - we talked IRL about merging this in anyways, even if batching isn't complete. Do we want to run a few comparison tests using OS Signpost to make sure this change isn't adding too much additional time?

Getting actual data for these analytics from real merchant usage will help us a ton

jaxdesmarais:
> @jaxdesmarais following up here - we talked IRL about merging this in anyways, even if batching isn't complete. Do we want to run a few comparison tests using OS Signpost to make sure this change isn't adding too much additional time?
> 
> Getting actual data for these analytics from real merchant usage will help us a ton

For sure, I am using branch `api-request-latency-signpost` for testing. Results are on cellular:
- Production: 1.14s, 628.27ms, 957.79ms, 495.53ms, 395.75ms, 864.03ms, 486.59ms, 482.74ms, 717.69ms, 742.30ms = 678.96 average
 
Results on `main`:
- Production: 1.02s, 962.32ms, 571.74ms, 547.22ms, 443.80ms, 610.82ms, 461.21ms, 493.29ms, 432.27ms, 409.00ms = 593.37 average

So it looks like on average this would add 85.59ms of latency without batching. Curious if your results are the same.

scannillo:
Here are my results. I ran them on WiFi for more consistency b/w trials:

Screenshot 2024-05-14 at 1 24 27 PM

jaxdesmarais:
Are we okay with that variance on stable vs slightly less stable connections to merge this in/release before batching? I agree the sooner we can get more data on these hops the better.

jaxdesmarais:
For reference here are my local tests with merging this branch into `batch-analytics` locally:

- Production wifi: 572.52ms, 319.12ms, 426.20ms, 399.95ms, 320.96ms, 404.85ms, 614.79ms, 439.32ms, 439.06ms, 522.78ms = 449.95 average
- Production cellular: 931.14ms, 540.38ms, 548.63ms, 712.02ms, 769.38ms, 545.93ms, 542.22ms, 654.36ms, 692.68ms, 600.24ms = 653.63 average

jaxdesmarais:
Discussed with @scannillo irl and removing the "do not merge" tag as even in the worst case our latency is minimally impacted. This will allow us to get additional merchant data on these hops and we will fast follow with batching.

PR #1295, opened 2024-05-08
[QL] Batch Analytics Events

jaxdesmarais:
### Summary of changes

- Add new `BTAnalyticsEventsStorage` actor to store and access our array of events in a thread-safe manner
- Add `DispatchSourceTimer` to send analytics events every 20 seconds

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais @scannillo


Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -6,16 +6,27 @@ class BTAnalyticsService: Equatable {
...
     private let apiClient: BTAPIClient
+    private let timerInterval: Double
+
+    private var timer: Timer?
+    private var events: [FPTIBatchData.Event] = []
 
     // MARK: - Initializer
 
-    init(apiClient: BTAPIClient) {
+    init(apiClient: BTAPIClient, timerInterval: Double = 30) {
```
jaxdesmarais, line 23: I ran through a bunch of different flows to find the "sweet spot" here. 30s seemed to be the most ideal for every flow except card, which tended to be faster. We could also consider bumping this down to 20s, but that seemed to be about as low as we would want to go while still getting a significant number of batched events. Anything below 20 (5, 10, 15) there were times we would just send a single event. Curious to hear what other folks think.

scannillo:
I ran through a few flows (paypal, then card, then venmo, then paypal) with Xcode's network inspector. I am seeing more requests to `v1/tracking` than expected, and a few of them are returning a 400. I notice these have an empty `event_params` array:

---
Screenshot 2024-05-08 at 10 30 51 AM
---

Maybe part of the issue is re-setting a new `Timer` instance every time `performEventRequest` is called.

jaxdesmarais:
In this commit: https://github.com/braintree/braintree_ios/pull/1295/commits/09da56cf9e101246df9b784e423bde902ad3b8c4 I moved to use a `DispatchQueue`. After doing some additional research on using `Timer` I found that it was deallocating and resetting unexpectedly. Also each time the method was called it would reset the timer in relation to the 30s interval. Using a `DispatchQueue` resulted in much more consistent firing of analytics.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -6,16 +6,22 @@ class BTAnalyticsService: Equatable {
...
+
     /// The FPTI URL to post all analytic events.
     static let url = URL(string: "https://api-m.paypal.com")!
 
+    // MARK: - Private Properties
+
     private let apiClient: BTAPIClient
+    private let timerInterval: Double
+
+    private var events: [FPTIBatchData.Event] = []
```
scannillo, line 18: This queue of events is wiped if a merchant re-instantiates a new BTAPIClient. We should support the case where they create and destroy BTAPIClients on button tap

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -84,15 +92,29 @@ class BTAnalyticsService: Equatable {
...
             }
 
-            let postParameters = self.createAnalyticsEvent(config: configuration, sessionID: self.apiClient.metadata.sessionID, event: event)
-            self.http?.post("v1/tracking/batch/events", parameters: postParameters) { _, _, _ in }
+            // Send analytics events immediately if no timer is needed for unit tests
+            if self.timerInterval == 0 {
+                self.sendQueuedAnalyticsEvents(configuration: configuration)
+            } else {
+                DispatchQueue.main.asyncAfter(deadline: .now() + self.timerInterval) { [weak self] in
+                    self?.sendQueuedAnalyticsEvents(configuration: configuration)
```
scannillo, line 100: I changed the timer interval to 2 seconds and threw a print statement in this block. I would expect it to print to console every 2 seconds, but I'm not seeing that 👁️

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
scannillo: I think this can be solved w/ a static property, it will be universally accessible to all BTAPIClient instances. However, arrays in Swift aren't thread-safe by default. So we'll want either a light array wrapper to ensure thread safety (like a DispatchQueue wrapper) _or_ explore another data store option for maintaining our events list that _is_ thread safe by default.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -84,15 +92,29 @@ class BTAnalyticsService: Equatable {
                 return
             }
 
-            let postParameters = self.createAnalyticsEvent(config: configuration, sessionID: self.apiClient.metadata.sessionID, event: event)
-            self.http?.post("v1/tracking/batch/events", parameters: postParameters) { _, _, _ in }
+            // Send analytics events immediately if no timer is needed for unit tests
+            if self.timerInterval == 0 {
+                self.sendQueuedAnalyticsEvents(configuration: configuration)
+            } else {
+                DispatchQueue.main.asyncAfter(deadline: .now() + self.timerInterval) { [weak self] in
```
scannillo, line 99: I changed the timer interval to 2 seconds and threw a print statement in this block. I would expect it to print to console every 2 seconds, but I'm not seeing that 👁️

scannillo: I think DispatchQueue.main.asyncAfter doesn't repeat a blocks execution numerous times, but just delays it by a certain period of time.

Since this is getting called every time we call `sendAnalyticsEvent()`, I think it's hiding the fact that this isn't working on a loop timer.

jaxdesmarais: That's a good call. I moved to `DispatchSourceTimer` here which _actually_ seems to repeat as expected when I add print statements: https://github.com/braintree/braintree_ios/pull/1295/commits/8a816a2b610e01fc12dcb5545493a50705da54d6.

Pros of this approach:
- Behaves as expected firing an event every however many seconds until it get's the the `BTAnalyticsService.events.isEmpty` else block and is canceled (can comment this line out to verify)
- Can be cancelled (vs running forever or until memory can be allocated)
- We don't need to manage run loops like we would with `Timer`

TODO: still need to address the thread-safety of the static array

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -96,15 +105,44 @@ class BTAnalyticsService: Equatable {
...
             }
 
-            let postParameters = self.createAnalyticsEvent(config: configuration, sessionID: self.apiClient.metadata.sessionID, event: event)
-            self.http?.post("v1/tracking/batch/events", parameters: postParameters) { _, _, _ in }
+            if self.timerInterval == 0 {
+                Task {
+                    await self.sendQueuedAnalyticsEvents(configuration: configuration)
+                }
+            } else {
+                self.timer = DispatchSource.makeTimerSource(queue: self.http?.dispatchQueue)
```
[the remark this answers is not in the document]
jaxdesmarais: Currently yes since it reuses the existing connection created via `http.dispatchQueue` - if we want to set up another DispatchQueue for analytics specifically to use a singular queue/timer we can do that. It will add additional latency vs reusing the existing queue/connection based on my local testing. What do you think?

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -6,16 +6,23 @@ class BTAnalyticsService: Equatable {
...
-    
+
     /// The FPTI URL to post all analytic events.
     static let url = URL(string: "https://api-m.paypal.com")!
 
+    // MARK: - Private Properties
+
     private let apiClient: BTAPIClient
+    private let timerInterval: Int
+    private let events = BTAnalyticsEventsStorage()
```
scannillo, line 17: If I toggle b/w features in the demo (Card, PP, Apple Pay), a new `BTAnalyticsEventsStorage()` is created each time (new empty array of events). Is this expected?

jaxdesmarais: You're right, not expected. Addressed in commit https://github.com/braintree/braintree_ios/pull/1295/commits/312a19d872ac31b93bef62256d8bd290659392ef by making events static. From there if you add the following under `await BTAnalyticsService.events.append(event)` and click through a number of flows before the 20s timer goes of each client will use same collection of events:
```swift
        print("START")
        await print("\(BTAnalyticsService.events.allValues) count: \(BTAnalyticsService.events.allValues.count)")
        print("END")
```

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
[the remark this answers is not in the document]
scannillo: I updated this to check if the timer is `nil` before recreating it

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -162,6 +162,18 @@ import Foundation
             }
         }
     }
+    
+    func fetchConfiguration() async throws -> BTConfiguration {
```
scannillo, line 166: Added because this makes life a lot easier within the `BTAnalyticsService.performEventRequest` method for testing

scannillo: Would love to move all `fetchOrReturnRemoteConfig` calls to this syntax at some point

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -95,16 +113,45 @@ class BTAnalyticsService: Equatable {
             if let http = self.http, http.baseURL.absoluteString == "test://do-not-send.url" {
                 return
             }
+            
+            if shouldBypassTimerQueue {
+                await self.sendQueuedAnalyticsEvents(configuration: configuration)
```
KunJeongPark, line 118: what is this option? To turn off this implementation?

KunJeongPark: Another question I had was right now in main, individual analytics event is sent to `v1/tracking/batch/events` which handles batching of the events on the server side?

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -95,16 +113,45 @@ class BTAnalyticsService: Equatable {
...
+                BTAnalyticsService.timer = DispatchSource.makeTimerSource(queue: self.http?.dispatchQueue)
+                BTAnalyticsService.timer?.schedule(
+                    deadline: .now() + .seconds(self.timerInterval),
+                    repeating: .seconds(self.timerInterval),
+                    leeway: .seconds(1)
+                )
 
-            let postParameters = self.createAnalyticsEvent(config: configuration, sessionID: self.apiClient.metadata.sessionID, event: event)
-            self.http?.post("v1/tracking/batch/events", parameters: postParameters) { _, _, _ in }
+                BTAnalyticsService.timer?.setEventHandler {
```
KunJeongPark, line 130: Another question I had was right now in main, individual analytics event is sent to v1/tracking/batch/events which handles batching of the events on the server side?

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
jaxdesmarais: This just bypasses the timer to allow us to unit test sending analytics without needing to override the timer implementation or add a super long sleep.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
jaxdesmarais: We were using `v1/tracking/batch/events` to send individual events not batched events. That API does not handle batching at all, just allows you to send events optionally in batches. This PR starts sending events in batches when available, though sometimes based on the timer we may still send a singular event to this endpoint in some cases.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
scannillo: We should look into mocking DispatchTimerSource (either via subclassing or a protocol) and injecting an actual mock of it for our test setup, versus toggling this boolean. That would be the best way to test this, though it might not be possible with DispatchTimerSource. It is worth trying though - Jax did you take a stab at it?

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
KunJeongPark: Another question I had was, I am not familiar with this timer, what happens if timer gets fired while events are in flight, being sent to back end.

jaxdesmarais: It would send whatever the current array of events is and any subsequent events would be appended to the now empty array of events.

KunJeongPark:
This is really interesting PR Jax. I've run through all the scenarios I could think of and don't see any issues but
I'm a bit nervous about mixing old and new. I've seen warnings on using checkedContinuation, for example, sometimes.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
jaxdesmarais: Yeah, I poked around at it a bit but mocking `DispatchTimerSource` from what I could find it was a bit odd because `DispatchTimerSource` is a protocol. It also was only mocking the timer methods (so schedule, setEventHandler, resume, cancel) vs also needing to send the event immediately. I think we'd have to have all the methods we need in our own protocol before going about the inheritance in `DispatchTimerSource`, then an extension for the implementation. We also would need to mock `DispatchQueue` which is used in the `makeTimerSource` method to send the event immediately.

But also maybe I over complicated it and missed something super obvious. 🙈 Happy to pair on it and see if we can get anything together if you want!

scannillo (approved):

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -4,20 +4,38 @@ class BTAnalyticsService: Equatable {
...
+    /// Exposed for testing only
+    var shouldBypassTimerQueue = false
+
+    // MARK: - Private Properties
+
+    private static let events = BTAnalyticsEventsStorage()
+
     private let apiClient: BTAPIClient
+    /// Amount of time, in seconds, between batch API requests sent to FPTI
+    private let timerInterval = 20
```
scannillo, line 22: Take it or leave it:

I wouldn't mind going further down - maybe to 15 sec. It is still a large win for analytics. But, I have a fear of losing some metrics since we aren't doing any local persistence to encounter for backgrounding, app kill etc. _Also_ this timer gets set only after the first analytic event is triggered, _not_ when the SDK is initialized

KunJeongPark (approved):

KunJeongPark:
Another question I had was we have unit test failures, I see this on my [PR](https://github.com/braintree/braintree_ios/pull/1308) as well.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
jaxdesmarais: > Also this timer gets set only after the first analytic event is triggered, not when the SDK is initialized

With us now sending config events it does get triggered fairly close to initialization. The first event is now the `v1/config` call. That said, down to bump it down to 15s!

jaxdesmarais, line 22:
```suggestion
    private let timerInterval = 15
```

jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/1295/commits/61aa5f16f8c8983f47b02cde6201309f01b330fd

jaxdesmarais:
> Another question I had was we have unit test failures, I see this on my [PR](https://github.com/braintree/braintree_ios/pull/1308) as well.

@KunJeongPark seems like it was a flake since they are passing now. The UI test failures have been failing for most of this week. Running the failing UI test locally is passing.

PR #1296, opened 2024-05-09
Added new `isVault` analytics tag - Venmo, PayPal

stechiu:
Added `is_vault` FPTI tag at the `event_params` level of our FPTI payload. This value should be a boolean. True if the merchant launched a `tokenize()` call with a `BTPayPalVaultRequest` and false if the merchant launched a `tokenize(`) call with a `BTPayPalCheckoutRequest`.

The same check was added to `BTVenmoRequest`

### Summary of changes

- Added new analytics tag `is_vault`
- Added tag to Venmo

### Checklist

- [x] Added a changelog entry

- @stechiu


Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -48,6 +50,7 @@ class BTAnalyticsService: Equatable {
     /// Exposed to be able to execute this function synchronously in unit tests
     func performEventRequest(
         _ eventName: String,
+        request: Bool? = nil,
```
scannillo, line 53: Nitpick - I would call all of these input param names `isVaultRequest`, since `request` is a little vague

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -29,6 +29,9 @@ import BraintreeDataCollector
     /// Exposed for testing, the ASWebAuthenticationSession instance used for the PayPal flow
     var webAuthenticationSession: BTWebAuthenticationSession
 
+    /// Checks whether Merchant launched tokenzie request with Vault
```
scannillo, line 32:
```suggestion
    /// True if `tokenize()` was called with a VaultRequest object type
```

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -30,6 +30,8 @@ struct FPTIBatchData: Codable {
         let correlationID: String?
         let errorDescription: String?
         let eventName: String
+        /// Checks whether Merchant launched tokenzie request with Vault
```
scannillo, line 33: Nitpick, take or leave
```suggestion
        /// True if `tokenize()` was called with a VaultRequest object type
```

UnitTests/BraintreeCoreTests/Analytics/FPTIBatchData_Tests.swift
```diff
@@ -71,6 +73,7 @@ final class FPTIBatchData_Tests: XCTestCase {
         XCTAssertTrue((batchParams["ios_package_manager"] as! String).matches("Carthage or Other|CocoaPods|Swift Package Manager"))
         XCTAssertEqual(batchParams["api_integration_type"] as? String, "fake-integration-type")
         XCTAssertEqual(batchParams["is_simulator"] as? Bool, true)
+        XCTAssertEqual(batchParams["is_vault"] as? Bool, true)
```
scannillo, line 76: I think this test assertion should be at the event_params level below

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -232,7 +237,7 @@ import BraintreeDataCollector
         request: BTPayPalRequest,
         completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void
     ) {
-        self.apiClient.sendAnalyticsEvent(BTPayPalAnalytics.tokenizeStarted)
+        self.apiClient.sendAnalyticsEvent(BTPayPalAnalytics.tokenizeStarted, request: isVaultRequest)
```
richherrera, line 240: nitpick:
```suggestion
        apiClient.sendAnalyticsEvent(BTPayPalAnalytics.tokenizeStarted, request: isVaultRequest)
```

Podfile.lock
```diff
@@ -37,4 +37,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 224cc2d6666b79d3d6adb45cccd6dfbc6fe74d18
 
-COCOAPODS: 1.15.2
+COCOAPODS: 1.14.3
```
jaxdesmarais, line 40: Lets revert this

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
jaxdesmarais: ➕ 1

richherrera (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -29,6 +29,9 @@ import BraintreeDataCollector
     /// Exposed for testing, the ASWebAuthenticationSession instance used for the PayPal flow
     var webAuthenticationSession: BTWebAuthenticationSession
 
+    /// True if `tokenize()` was called with a VaultRequest object type
+    var isVaultRequest: Bool = false
```
scannillo, line 33: I know this wasn't mentioned in the ticket, but can we also set this property for the VenmoRequests in VenmoClient? Also all good if you want to make a new ticket for that

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -23,12 +23,14 @@ class BTAnalyticsService: Equatable {
     /// Sends analytics event to https://api.paypal.com/v1/tracking/batch/events/ via a background task.
     /// - Parameters:
     ///   - eventName: Name of analytic event.
+    ///   - request: Optional. Type of checkout request associated with checkout session.
```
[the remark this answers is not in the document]
stechiu: Thanks, I forgot to update the description

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -320,13 +320,15 @@ import Foundation
     @_documentation(visibility: private)
     public func sendAnalyticsEvent(
         _ eventName: String,
+        request: Bool? = nil,
```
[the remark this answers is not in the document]
stechiu: I forgot to update that parameter name, thanks!

Sources/BraintreePayPal/BTPayPalClient.swift
stechiu: I can make the change in this PR! Do you mean updating the parameter name from `request` to `venmoRequest`? or create a new internal property `var isVenmoRequest: Bool` in `BTVenmoRequest`?

jaxdesmarais: I think what Sammy is saying is to add a property `isVaultRequest` to `BTVenmoClient` and pass it as `true/false` in that flow as well. I don't think it needs to be renamed or anything, just passed in the `BTVenmoClient` for vaulting/non-vaulting flows.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -234,7 +239,7 @@ import BraintreeDataCollector
         request: BTPayPalRequest,
         completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void
     ) {
-        self.apiClient.sendAnalyticsEvent(BTPayPalAnalytics.tokenizeStarted)
+        apiClient.sendAnalyticsEvent(BTPayPalAnalytics.tokenizeStarted, isVaultRequest: isVaultRequest)
```
jaxdesmarais, line 242: I think we also want to send this anywhere we are calling `sendAnalyticsEvent` in this flow?

UnitTests/BraintreeCoreTests/Analytics/FPTIBatchData_Tests.swift
```diff
@@ -99,6 +101,8 @@ final class FPTIBatchData_Tests: XCTestCase {
         XCTAssertNil(eventParams[1]["error_desc"])
         XCTAssertEqual(eventParams[0]["correlation_id"] as? String, "fake-correlation-id-1")
         XCTAssertNil(eventParams[1]["correlation_id"])
+        XCTAssertEqual(eventParams[0]["is_vault"] as? Bool, true)
+        XCTAssertNil(eventParams[1]["is_vault"])
```
jaxdesmarais, line 105: Based on the above, I would expect this test is failing currently. Looks like there is an error with `MockAPIClient` not overriding the `sendAnalyticsEvent` properly that is blocking the test execution currently.

Sources/BraintreePayPal/BTPayPalClient.swift
stechiu: When you say "this flow", do you mean the BT PayPal Client flow?

stechiu: I reread the ticket and assume yes, so I've made the changes in `BTPayPalClient` and have done the same in `BTVenmoClient`

UnitTests/BraintreeCoreTests/Analytics/FPTIBatchData_Tests.swift
stechiu: Thanks, looks like the issue was that the `isVaultRequest` parameter wasn't in the right order and `MockAPIClient` didn't have the updated `request` parameter name

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -31,6 +31,9 @@ import BraintreeCore
     /// This allows us to set and return a completion in our methods that otherwise cannot require a completion.
     var appSwitchCompletion: (BTVenmoAccountNonce?, Error?) -> Void = { _, _ in }
 
+    /// True if `tokenize()` was called with a Vault request object type
+    var isVaultRequest: Bool = false
```
scannillo, line 35: Looks like we already have a property for this, `shouldVault` on line 28 we can re-use and just pass into our analytics calls

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -78,6 +81,7 @@ import BraintreeDataCollector
         completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void
     ) {
         tokenize(request: request, completion: completion)
+        isVaultRequest = true
```
richherrera, line 84: We need to set `isVaultRequest` before calling the `tokenize` method.

```suggestion
        isVaultRequest = true
        tokenize(request: request, completion: completion)
```

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 6.18.2 (2024-05-15)
 * BraintreePayPal
   * Send `start_time`, `end_time`, and `endpoint` to FPTI for tracking API request latency
+  * Send `isVaultRequest` to FPTI for tracking Vault vs One Time Checkout and Venmo Checkout in "Pay with PayPal" flows
```
```suggestion
  * Send `isVaultRequest` to FPTI for tracking in Venmo and PayPal flows
```

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -23,12 +23,14 @@ class BTAnalyticsService: Equatable {
     /// Sends analytics event to https://api.paypal.com/v1/tracking/batch/events/ via a background task.
     /// - Parameters:
     ///   - eventName: Name of analytic event.
+    ///   - isVaultRequest: Optional. Type of checkout request associated with checkout session.
```
jaxdesmarais, line 26:
```suggestion
    ///   - isVaultRequest: Optional. If the Venmo or PayPal request is being vaulted.
```

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -32,6 +32,8 @@ struct FPTIBatchData: Codable {
         let endTime: Int?
         let errorDescription: String?
         let eventName: String
+        /// True if `tokenize()` was called with a VaultRequest object type
```
jaxdesmarais, line 35:
```suggestion
        /// True if the PayPal or Venmo request is to be vaulted
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -120,6 +124,7 @@ import BraintreeDataCollector
         completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void
     ) {
         tokenize(request: request, completion: completion)
+        isVaultRequest = false
```
jaxdesmarais, line 127: Should we default `isVaultRequest` to `false` so we don't have to set it to false in all non-vaulting cases?

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: ➕ 1

UnitTests/BraintreeCoreTests/Analytics/FPTIBatchData_Tests.swift
jaxdesmarais, line 105: It looks like this is failing currently, should be:
```suggestion
        XCTAssertEqual(eventParams[0]["is_vault"] as? Bool, false)
        XCTAssertEqual(eventParams[1]["is_vault"] as? Bool, true)
```

Sources/BraintreeVenmo/BTVenmoClient.swift
stechiu: oh wow completely missed that variable. thanks!

Sources/BraintreePayPal/BTPayPalClient.swift
stechiu: Maybe I don't understand the PayPal flow well enough. Could there be a situation where a user checks out with Vault but doesn't complete transaction and checks out with 1 time checkout in the same transaction?

UnitTests/BraintreeCoreTests/Analytics/FPTIBatchData_Tests.swift
```diff
@@ -99,6 +101,8 @@ final class FPTIBatchData_Tests: XCTestCase {
         XCTAssertNil(eventParams[1]["error_desc"])
         XCTAssertEqual(eventParams[0]["correlation_id"] as? String, "fake-correlation-id-1")
         XCTAssertNil(eventParams[1]["correlation_id"])
+        XCTAssertEqual(eventParams[0]["is_vault"] as? Bool, false)
+                XCTAssertEqual(eventParams[1]["is_vault"] as? Bool, true)
```
scannillo, line 105:
```suggestion
        XCTAssertEqual(eventParams[0]["is_vault"] as? Bool, false)
        XCTAssertEqual(eventParams[1]["is_vault"] as? Bool, true)
```

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -638,6 +638,19 @@ class BTPayPalClient_Tests: XCTestCase {
         }
     }
 
+    func testRequestType_Checkout() {
```
scannillo, line 641: Ideally we'd want to test the behavior, not the implementation of BTPayPalClient. So the unit tests shouldn't care _how_ BTPayPalClient passes the vault value to the APIClient.sendAnalyticsEvent, just that it does.

Psuedocode example would be something like:
```
func testTokenize_whenVaultRequest_setsVaultAnalyticsTag() {
    let vaultRequest = BTPayPalVaultRequest()
    let _ = try? await sut.tokenize(vaultRequest)

    XCTAssertTrue(mockApiClient.lastPostedVaultType)
}

func testTokenize_whenCheckoutRequest_setsVaultAnalyticsTag() {
    let checkoutRequest = BTPayPalCheckoutRequest()
    let _ = try? await sut.tokenize(checkoutRequest)

    XCTAssertFalse(mockApiClient.lastPostedVaultType)
}
```

This will require some modification to our [MockAPIClient here](https://github.com/braintree/braintree_ios/blob/c2a675c8757f7a8b949a5853617fb55a3edec4c7/UnitTests/BraintreeTestShared/MockAPIClient.swift#L85-L95).

Also, instead of adding 2 new functions - we _could_ also just tack on the assertion to an existing test (since there are already some that use a BTPayPalCheckoutRequest & BTPayPalVaultRequest.

scannillo: This is not a huge deal for this PR, but I thought it might be useful to explain. We want to test the behavior so that if we ever need to change the internal logic/implementation, our tests still give us value.

scannillo (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: 🤔 That would certainly be an edge case, but probably a good call to handle it incase it does happen!

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
stechiu: Thanks for the detailed explanation! I wrote a test but Xcode is stuck in a loop trying when I test, but here it is
```
    func testTokenize_whenVaultRequest_setsVaultAnalyticsTag() async {
        let vaultRequest = BTPayPalVaultRequest()
        let _ = try? await payPalClient.tokenize(vaultRequest)

        XCTAssertTrue(mockAPIClient.lastPostedVaultType)
    }
```
Will do the same for Checkout and Venmo request

jaxdesmarais (approved):

PR #1298, opened 2024-05-10
Added "Origami Checkout" to Web Checkout Flow

stechiu:
This ticket allows for QA team to test the Origami Checkout flow. We want to add the tokenization key “sandbox_rz48bqvw_jcyycfw6f9j4nj9c" to launch the PayPal checkout Origami experience. To launch this, PayPal intent type must be “Sale”

This tokenization key should only be used when specified in settings (not new default tokenization key).

### Summary of changes

- Branched off of `payment-insights-feature`
- Added to Origami Tokenization Key to Authorization Type in Settings
- Added toggle switch for Origami XO experience to Web Checkout

### To Test
1. Settings: Integration > PayPal - Web Checkout, Authorization Type > Origami Tokenization Key
2. Buyer email (ex: jsdk@bt.com)
3. Toggle Origami Checkout to `on`
4. Tap "PayPal Checkout"

### Known Error
When Authorization Type is set to `Origami Tokenization Key`, there's a known error where HTTP POST request fails. This appears when tapping on `PayPalCheckout` Button, regardless of whether Origami Checkout or PayLater is switched on or off. **You need to enter a Buyer Email (ex: jsdk@bt.com) to before using any of the PayPal Checkout experiences.**

This error doesn't occur when using the other Authorization Types.

simulator_screenshot_995BA9C5-B5B9-44EC-BBED-24DDFF82FA41

### Checklist

- [ ] Added a changelog entry

- @stechiu

### DIFF
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -28,6 +28,15 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
     
     let payLaterToggle = UISwitch()
 
+    lazy var origamiCheckoutToggleLabel: UILabel = {
+        let label = UILabel()
+        label.text = "Origami Checkout"
+        label.font = .preferredFont(forTextStyle: .footnote)
+        return label
+    }()
+    
+    let origamiCheckoutToggle = UISwitch()
+
     override func createPaymentButton() -> UIView {
         let payPalCheckoutButton = createButton(title: "PayPal Checkout", action: #selector(tappedPayPalCheckout))
         let payPalVaultButton = createButton(title: "PayPal Vault", action: #selector(tappedPayPalVault))
@@ -36,6 +45,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
             UIStackView(arrangedSubviews: [emailLabel, emailTextField]),
             buttonsStackView(label: "1-Time Checkout", views: [
                 UIStackView(arrangedSubviews: [payLaterToggleLabel, payLaterToggle]),
+                UIStackView(arrangedSubviews: [origamiCheckoutToggleLabel, origamiCheckoutToggle]),
                 payPalCheckoutButton
             ]),
             buttonsStackView(label: "Vault",views: [payPalVaultButton])
@@ -65,6 +75,10 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         request.lineItems = [lineItem]
         
         request.offerPayLater = payLaterToggle.isOn
+
+        if origamiCheckoutToggle.isOn {
+            request.intent = .sale
+        }
 
         payPalClient.tokenize(request) { nonce, error in
             sender.isEnabled = true
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -30,7 +30,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
 
     lazy var origamiCheckoutToggleLabel: UILabel = {
         let label = UILabel()
-        label.text = "Origami Checkout"
+        label.text = "New PayPal Checkout"
         label.font = .preferredFont(forTextStyle: .footnote)
         return label
     }()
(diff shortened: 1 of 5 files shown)


sarahkoop:
Update on the known error - this is happening when we pass `""` instead of an email address or `nil` for the `userAuthenticationEmail` param.

jaxdesmarais (commented):
It looks like the [Android PR](https://github.com/braintree/braintree_android/pull/996/files) used the term "New PayPal checkout experience" for this - should we consider aligning with the same naming since I believe "Oragami" is more of an internal name?

jaxdesmarais:
> Update on the known error - this is happening when we pass `""` instead of an email address or `nil` for the `userAuthenticationEmail` param.

🤔 it's a bit odd to me that this new flow behaves differently than passing an empty string for MPE ([I had asked about this on Android here](https://github.com/braintree/braintree_android/pull/944#discussion_r1526688611)). Do we know if this is a change in previous behavior or specific to this new flow?

sarahkoop:
> > Update on the known error - this is happening when we pass `""` instead of an email address or `nil` for the `userAuthenticationEmail` param.
> 
> 🤔 it's a bit odd to me that this new flow behaves differently than passing an empty string for MPE ([I had asked about this on Android here](https://github.com/braintree/braintree_android/pull/944#discussion_r1526688611)). Do we know if this is a change in previous behavior or specific to this new flow?

I'll follow up internally about this - agreed we should be able to pass empty string similar to the vault flow

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -66,6 +76,10 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         
         request.offerPayLater = payLaterToggle.isOn
 
+        if origamiCheckoutToggle.isOn {
```
scannillo, line 79: Take it or leave it - but instead of doing the whole processInfoArgs & Root.plist song and dance, we could just create a new `BTAPIClient` and `BTPayPalClient` w/ the specified tokenization key if this toggle is turned on, all within the ViewController.

I don't feel strongly either way!

scannillo (approved):

stechiu:
> It looks like the [Android PR](https://github.com/braintree/braintree_android/pull/996/files) used the term "New PayPal checkout experience" for this - should we consider aligning with the same naming since I believe "Oragami" is more of an internal name?

Good point. I changed the label text that users would see in the Demo app but kept Origami for when it's used internally. Thought process is that there will be new Checkout experiences in the future so we'd need a way to identify the different experiences. Unless we know for sure that Origami will be the new Checkout experiences and the current one will be fazed out in a couple months? @sarahkoop would love your input here

sarahkoop:
> > It looks like the [Android PR](https://github.com/braintree/braintree_android/pull/996/files) used the term "New PayPal checkout experience" for this - should we consider aligning with the same naming since I believe "Oragami" is more of an internal name?
> 
> Good point. I changed the label text that users would see in the Demo app but kept Origami for when it's used internally. Thought process is that there will be new Checkout experiences in the future so we'd need a way to identify the different experiences. Unless we know for sure that Origami will be the new Checkout experiences and the current one will be fazed out in a couple months? @sarahkoop would love your input here

I would agree we should probably avoid using internal naming for flows, but also agree "new PayPal checkout experience" is somewhat vague - is there another name we could use to indicate what this flow is?

jaxdesmarais:
> I would agree we should probably avoid using internal naming for flows, but also agree "new PayPal checkout experience" is somewhat vague - is there another name we could use to indicate what this flow is?

Do we know what we are calling this to merchants? Stephanie asked me this but I wasn't sure what the external facing name for the new UI was.

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -66,6 +76,12 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         
         request.offerPayLater = payLaterToggle.isOn
 
+        if newPayPalCheckoutToggle.isOn {
+            request.intent = .sale
+        } else {
+            request.intent = .authorize
+        }
```
jaxdesmarais, line 83: nit:
```swift
request.intent = newPayPalCheckoutToggle.isOn ? .sale : .authorize
```

jaxdesmarais (approved):
Just one very minor take it or leave it comment 👏

Demo/Application/Features/PayPalWebCheckoutViewController.swift
stechiu: thanks!!

PR #1301, opened 2024-05-14
[QL] Don't Allow Empty String For `payer_email` in One-Time Checkout

agedd:
### Summary of changes

- If the `userAuthenticalEmail`, do not set the `payer_email` field in the `BTPayPalCheckoutRequest` for one-time checkout flows. 
- This change only applies to the one-time checkout flows and not the vault flows. 
- Relevant Ticket: [DTMOBILES-553](https://paypal.atlassian.net/browse/DTMOBILES-553)

### Checklist

- [ ]~Added a changelog entry~

- @agedd

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
+++ b/Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
@@ -134,7 +134,7 @@ import BraintreeCore
             checkoutParameters["currency_iso_code"] = currencyCode
         }
         
-        if let userAuthenticationEmail {
+        if let userAuthenticationEmail, !userAuthenticationEmail.isEmpty {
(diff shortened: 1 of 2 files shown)


scannillo (approved):

stechiu:
Can you add an alert popup notification or something that let's someone know what to do instead? Otherwise the user will keep trying to send an empty string

jaxdesmarais:
Removed the ticket link since we don't want internal links in public jira

jaxdesmarais (approved):

agedd:
> Can you add an alert popup notification or something that let's someone know what to do instead? Otherwise the user will keep trying to send an empty string

this is a good callout, afaik the ask was to prevent including the `payer_email` field if the user email is empty, so that the backend `create_payment_resource` GW API does not fail - will follow-up with product on this.

PR #1308, opened 2024-05-17
[QL] Investigate Triple v1/config call

KunJeongPark:
### Description
The SDK is sometimes making 3 sequential network calls to `v1/configuration`. This only happens when you create an APIClient, PayPalClient, and then tokenize() all immediately in sequence of one another (on button click).
Technically, our docs snippet shows instantiating the BTAPIClient earlier than button click (ex: viewDidLoad). But ideally our SDK can be more performant if merchants aren’t reading our docs. Also, our docs can better call out this recommendation.

I  was able to reproduce this issue by instantiating `APIClient` immediately prior to checkout flow. 
If the configuration is not cached from earlier instantiation of` APIClient`, `APIClient`  instantiation, `tokenize` call and `sendAnalyticsEvent` try to call `v1/configuration` endpoint since network calls are made in quick succession asynchronously without configuration having been added to cache yet.

This is what I saw:
- BTAPIClient init, `fetchOrGetRemoteConfiguration` about to be called
- bypassed cache (from BTAPIClient instantiation)
- called sendAnalyticsEvent => PayPal:tokenize:started
- bypassed cache (from calls inside sendAnalyticsEvent to fetchOrReturn function)
- bypassed cache (from call to fetchOrReturn in tokenize function)

I considered using actor for `ConfigurationCache` but I don't know of a way of handling this in conjunction
with `BTHTTP` get call in the `fetchOrGetRemoteConfiguration` function which returns via completion handler and
is utilized throughout the code base. Also, the issue is that network calls get made in quick succession before result is written into the cache (this only happens if merchants don't adhere to our recommendation of instantiating `BTAPIClient` early on in their application), so having serialized access in the `ConfigurationCache` would not solve the issue.

I went the route of adding a loading state for `v1/configuration` and adding completion handlers if more calls were made
during loading state and on network call return, returning the same result to completion handlers in the array of completion handlers.

[Commit 2](e703302d889f4f35d740b27da5d016eca4f07da3), I considered just putting sync serial queue around just cache writes but realized that the mutable properties like array and isLoading state were not thread safe.

[Commit 3](3d735fbb9d476a340f5f9416550a9ca1c50970ed) I used concurrent queue for outer block of `fetchOrGetRemoteConfiguration` function (for concurrent reads since it should only write into cache one time during session) and barrier option for tasks that any write to mutable properties

TODO: 
- Examine failing unit tests
- Check out if commit 3 works with wrapping in async await checked continuation for [Jax's PR on batched analytics](https://github.com/braintree/braintree_ios/pull/1295) ✅
Commit 3 worked fine with Jax's PR which wraps fetchOrReturnRemoteConfiguration function in async await for Venmo, ApplePay, PayPal Web checkout features in demo app

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -39,7 +39,7 @@ import Foundation
 
     private static var _analyticsService: BTAnalyticsService?
     private var isLoadingConfig: Bool = false
-    private var fetchConfigQueue: DispatchQueue = DispatchQueue(label: "configQueue", attributes: .concurrent)
+    private var fetchConfigQueue: DispatchQueue = DispatchQueue(label: "configQueue")
     private var configCompletionHandlers: [(BTConfiguration?, Error?) -> Void] = []
 
     // MARK: - Initializers
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -39,7 +39,7 @@ import Foundation
 
     private static var _analyticsService: BTAnalyticsService?
     private var isLoadingConfig: Bool = false
-    private var fetchConfigQueue: DispatchQueue = DispatchQueue(label: "configQueue")
+    private var fetchConfigQueue: DispatchQueue = DispatchQueue(label: "configQueue", attributes: .concurrent)
     private var configCompletionHandlers: [(BTConfiguration?, Error?) -> Void] = []
 
     // MARK: - Initializers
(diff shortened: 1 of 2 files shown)


Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -122,44 +125,79 @@ import Foundation
...
-                let configuration = BTConfiguration(json: body)
+            }
 
-                setupHTTPCredentials(configuration)
-                try? ConfigurationCache.shared.putInCache(authorization: authorization, configuration: configuration)
-                
-                completion(configuration, nil)
+            if let cachedConfig = try? ConfigurationCache.shared.getFromCache(authorization: authorization) {
+                setupHTTPCredentials(cachedConfig)
+                DispatchQueue.main.async {
```
KunJeongPark, line 151: I added doing the completions on main because get call from BTHTTP returned results via dispatchQueue which was defaulted to main and I didn't see any calls overriding that. I noticed this because our demo setup for ApplePay had a crash because it was expecting results via main. I thought keeping this consistent with completion from get call completion block below would cause unexpected behavior in merchant app. https://github.com/braintree/braintree_ios/blob/da4c1056d4ae199917fb355f8c3ebe7118e3f03c/Sources/BraintreeCore/BTHTTP.swift#L311

scannillo:
The long-term goal is to make BTAPIClient less of a "god" object, since it currently does so many things. Android sets a good pattern for this, with a dedicated [ConfiguationLoader class](https://github.com/braintree/braintree_android/blob/main/BraintreeCore/src/main/java/com/braintreepayments/api/ConfigurationLoader.kt).

This PR could be a good excuse to move to that pattern, since now we're adding even more logic specific just to getting a config from either the cache or the network.

The long-term goal also is to mirror Android's separation of BTHTTP - into a HTTPClient specific & and then BT specific layer. Currently `BTHTTP` couples both business & raw networking logic.

scannillo:
Another comment re: tests for this change. Can we add a unit test that fails against `main`, so we know that this change is properly addressing the issue (the test would then pass with the new changes)?

scannillo:
I think there's several ways we can address this issue:

* At the HTTP / URLSession level. If we have an incoming `URLSessionDataTask` request incoming and detect that a matching one is already pending progress, cancel the incoming request.
    * This one will prevent this issue from happening with other HTTP requests (beyond just the config). Is that something we should consider?
* At the BTAPIClient / ConfigurationLoader level like you're doing here.
* Our SDK avoids calling `fetchOrReturnRemoteConfiguration` so many times in sequence, and move to a dependency injection model for the `BTConfiguration` where possible (for ex: to initialize a `BTAnalyticsService`).

❓ - does Android have this same issue if a merchant initializes there client & calls `tokenize()` all upon button click? If so, does the same issue arise? Asking b/c the more we can align the better.

KunJeongPark:
> I think there's several ways we can address this issue:
> 
> 1. At the HTTP / URLSession level. If we have an incoming `URLSessionDataTask` request incoming and detect that a matching one is already pending progress, cancel the incoming request.
>    
>    * This one will prevent this issue from happening with other HTTP requests (beyond just the config). Is that something we should consider?
> 2. At the BTAPIClient / ConfigurationLoader level like you're doing here.
> 3. Our SDK avoids calling `fetchOrReturnRemoteConfiguration` so many times in sequence, and move to a dependency injection model for the `BTConfiguration` where possible (for ex: to initialize a `BTAnalyticsService`).
> 
> ❓ - does Android have this same issue if a merchant initializes there client & calls `tokenize()` all upon button click? If so, does the same issue arise? Asking b/c the more we can align the better.

I will look into Android code to see if they have same issue.

KunJeongPark:
> Another comment re: tests for this change. Can we add a unit test that fails against `main`? This way we know that this change is properly addressing the issue (the test would then pass with the new changes).

I just did a quick test on local branch where I instantiated the BTAPIClient on button click along with payment client and call to tokenize method but I will write a proper unit test.

sshropshire:
@KunJeongPark @scannillo I could see this being an issue on Android too. We don't have logic to guard against multiple calls for configuration fetching before it's cached, so it's possible both iOS and Android have this issue.

scannillo:
> @KunJeongPark @scannillo I could see this being an issue on Android too. We don't have logic to guard against multiple calls for configuration fetching before it's cached, so it's possible both iOS and Android have this issue.

Definitely agree, I think digging into Android will be useful to do now! This way we can pick a solution that we can align as much as possible. Rich is adding the same network timing analytics to Android in DTMOBILES-625, which will be helpful for us to compare against iOS.

For some background - here's where the iOS numbers are currently at. Note, these analytics were added in a recent version so we only have a few merchants sending them (~11). But that # of events for `v1/configuration` is _way_ too high, given the # of calls to the `paypal_hermes` endpoints. This looks like a bug in our SDK, and not just the merchant integration edge case Victoria's PR summary mentions.

Screenshot 2024-06-06 at 12 32 49 PM

sshropshire:
@scannillo oh wow yeah that's a lot. When we fix this bug we'll probably reduce a ton of gateway traffic. That'll be a big win.

KunJeongPark:
> > @KunJeongPark @scannillo I could see this being an issue on Android too. We don't have logic to guard against multiple calls for configuration fetching before it's cached, so it's possible both iOS and Android have this issue.
> 
> Definitely agree, I think digging into Android will be useful to do now! This way we can pick a solution that we can align as much as possible. Rich is adding the same network timing analytics to Android in DTMOBILES-625, which will be helpful for us to compare against iOS.
> 
> For some background - here's where the iOS numbers are currently at. Note, these analytics were added in a recent version so we only have a few merchants sending them (~11). But that # of events for `v1/configuration` is _way_ too high, given the # of calls to the `paypal_hermes` endpoints. This looks like a bug in our SDK, and not just the merchant integration edge case Victoria's PR summary mentions.
> 
> Screenshot 2024-06-06 at 12 32 49 PM
> 
> (Percentiles are the the p95 value for total time in milliseconds the SDK spends waiting for these API calls to complete).

We don't know if this is an edge case integration though.
And were the above stats before or after adding fetchAPITiming? I think I saw recursion paths from there.

I am thinking of three different integration patterns for BTAPIClient:
Scenarios #1 BTAPIClient instantiation used for different payment clients, instantiated earlier
Scenario #2 BTAPIClient instantiated with before paymentClient instantiation earlier
Scenario #3 BTAPIClient, paymentClient tokenize called all at once

KunJeongPark:
I'm going to continue on the spike on this branch because there are issues on main with the demo app
after shopper insights. After seeing some results, I can move over stuff to another branch.

KunJeongPark:
> Another comment re: tests for this change. Can we add a unit test that fails against `main`? This way we know that this change is properly addressing the issue (the test would then pass with the new changes).

Addressed here: c743823731d4ee902ae3f9240589265d30632017

jaxdesmarais:
👋 Sorry it's taken some time to review this. I have a couple of potential concerns with this approach:

1. `configCompletionHandlers` - while I think we've proven this technically fixes the issue, the solution at least for me to have an array of completions feels a bit odd. It seems like there could be an instance where we have a mix of configs and errors in the array that could lead to some unexpected behavior.
2. Mixing of queues - this PR introduces a `fetchConfigQueue`, calls to the main queue, and calls to a `barrier` flagged queue which to me is a bit of a code smell. We ideally don't need to wrap queues in other varying queues to have things behave as expected or can just have 1 blocking queue in 1 place to stop the execution of the configuration fetch until the first one is returned.

I think if we take the approach of using a `ConfigurationLoader` similar to Android we can resolve the two issues above. I do think the test `testFetchOrReturnRemoteConfiguration_with_multiple_calls` here will be great for confirming we solve this same problem when we implement the `ConfigurationLoader`.

scannillo:
> => Yes, I agree with implementing ConfigurationLoader. My thinking was that we should still address possible concurrency issues in ConfigurationLoader as well. And difficulty of implementing prefetch of config with ConfigurationLoader.

I think we should solve this issue in the ConfigurationLoader and not in the BTAPIClient. BTAPIClient already is a "god" object and does too many things. We want to achieve higher separation of concerns. 

Here's what I'm thinking for steps forward:

1. Step 1 
    - Refactor existing logic as-is within `fetchOrReturnRemoteConfiguration` into a `ConfigurationLoader` class. 
    - Model Android (constructor takes BTHTTP & `load()` method takes Authorization param). 
    - Keep public function signature for `BTAPIClient.fetchOrReturnRemoteConfiguration()` the same so feature clients don’t have to change.
    - `BTAPIClient.fetchOrReturnRemoteConfiguration()` will call out to the `ConfigurationLoader.getConfig(authorization)`
2. Step 2
    - Within `ConfigurationLoader`, implement blocking mechanism to prevent duplicate outbound `v1/config` GET requests from happening at the same time.

PR #1312, opened 2024-05-20
Remove unused support for "data" URI schemes

scannillo:
### Summary of changes

- Remove support for "data" type URI schemes. This was unused by any feature clients in the SDK.
- From looking in braintree_android, this URL type is not explicitly expected/supported there.
    - (Someone can double check me on this!)

### Checklist

- ~Added a changelog entry~

### DIFF
--- a/Sources/BraintreeCore/BTHTTP.swift
+++ b/Sources/BraintreeCore/BTHTTP.swift
@@ -165,14 +165,7 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
             throw BTHTTPError.missingBaseURL(errorUserInfo)
         }
         
-        let fullPathURL: URL?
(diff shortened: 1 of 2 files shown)


jaxdesmarais (approved):
I also can't find any use of this scheme type on Android. Looking more into this scheme type [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs):
> allow content creators to embed small files inline in documents.

I can't really think of a compelling reason why the SDK would get/receive that scheme type unless as some point we were receiving base64 encoded text. Seems like since Android doesn't handle data schemes we should be fine, but we could also triple check with the gateway just to cover our bases.

richherrera (approved):

PR #1313, opened 2024-05-20
[QL] Update PayPal App Installed Check

jaxdesmarais:
### Summary of changes

- Align PayPal app installed check with the Venmo app installed PR #1289 

### Checklist

### Authors

- @jaxdesmarais


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -357,18 +353,17 @@ import BraintreeDataCollector
     }
 
     private func isPayPalAppInstalled() -> Bool {
```
scannillo, line 355: Take it or leave it - maybe we could make these actual installed checks re-used? Maybe an extension on `UIApplication`? `application.isVenmoAppInstalled()`?

jaxdesmarais: I dig it, updated here: https://github.com/braintree/braintree_ios/pull/1313/commits/a578a7db52b4212fe809bc42ae51215d85048d7d

richherrera (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: So we have a lot of unit tests that depend on our implementation of `URLOpener` in `BTPayPalClient`. Adding the install checks into the protocol to not duplicate the app install checks gets really messy really fast. Not sure if this is reason to not go with an extension and just rely on the constants. Curious to hear your thoughts.

The current state of this PR has duplicated logic but does have all tests passing again.

richherrera: I believe what we could do is add `isVenmoAppInstalled()` and `isPayPalAppInstalled()` to the` URLOpener` protocol. In an extension of UIApplication, we can put the default method, making them instance methods instead of static.

```
public protocol URLOpener {
    
    func canOpenURL(_ url: URL) -> Bool
    func open(_ url: URL, options: [UIApplication.OpenExternalURLOptionsKey: Any], completionHandler completion: ((Bool) -> Void)?)
    func isPayPalAppInstalled() -> Bool
    func isVenmoAppInstalled() -> Bool
}

extension UIApplication: URLOpener {
    
        guard let venmoURL = URL(string: "\(BTCoreConstants.venmoURLScheme)://") else {
            return false
        }
        return canOpenURL(venmoURL)
    }

    /// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
    /// Indicates whether the PayPal App is installed.
    public func isPayPalAppInstalled() -> Bool {
        guard let payPalURL = URL(string: "\(BTCoreConstants.payPalURLScheme)://") else {
            return false
        }
        return canOpenURL(payPalURL)
    }
}
```

 Here on line 21, you have the UIApplication singleton instance, so it would only be necessary to call `application.isPayPalAppInstalled()` as Sammy mentioned. I think it would work. Let me know what do you think

jaxdesmarais: This does work and I had done the same but requires `FPTIBatchData` to have a `var application: URLOpener = UIApplication.shared`. As long as we are okay with that I am fine with this. Updated here: https://github.com/braintree/braintree_ios/pull/1313/commits/000c6ff07646ec21a1984d94ca74779a10faff12

richherrera: 💯 LGTM!!!

PR #1317, opened 2024-05-21
Using PayPalMessageConfig we set the integration name and version to support Messaging Analytics

warmkesselj:
### Summary of changes

Using PayPalMessageConfig we set the integration name and version to support Messaging Analytics

### Checklist

- [ ] Added a changelog entry

### Authors
@saperi22 
@jwarmkessel 
@KunJeongPark


Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
```diff
@@ -52,6 +58,8 @@ public class BTPayPalMessagingView: UIView {
                 self.notifyFailure(with: BTPayPalMessagingError.payPalClientIDNotFound)
                 return
             }
+            
+
```
jaxdesmarais, line 62:
```suggestion
```

Sources/BraintreeCore/BTCoreConstants.swift
```diff
@@ -27,4 +27,9 @@ import Foundation
     /// :nodoc: This property is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     /// Key for userInfo dictionary that contains the BTJSON body of the HTTP error response
     public static let jsonResponseBodyKey: String = "com.braintreepayments.BTHTTPJSONResponseBodyKey"
+    
+    /// :nodoc: This property is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    /// Named used for identifiying the integrating SDK e.g. BTPayPalMessageView
+    public static let braintreeSDKIdentifier: String = "BT_SDK"
```
jaxdesmarais, line 33: I see in the Android PR we just hardcode this value and here we have it extracted out. I don't think we will use this for anything but messaging so probably fine to just hardcode where we call `setGlobalAnalytics`

Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
```diff
@@ -36,6 +36,12 @@ public class BTPayPalMessagingView: UIView {
     /// - Parameter request: an optional `BTPayPalMessagingRequest`
     /// - Warning: use `BTPayPalMessagingDelegate` protocol to receive notifications for events
     public func start(_ request: BTPayPalMessagingRequest = BTPayPalMessagingRequest()) {
+
```
jaxdesmarais, line 39:
```suggestion
```

Sources/BraintreeCore/BTCoreConstants.swift
KunJeongPark: Yes we had a discussion about that, we didn't want to make a typo in any subsequent changes, but I'm ok with hardcoding.

scannillo (approved):

jaxdesmarais (approved):

PR #1320, opened 2024-05-24
[QL] Remove Stage Logic from `PayPalWebCheckoutViewController`

jaxdesmarais:
### Summary of changes

- Remove Stage Logic from `PayPalWebCheckoutViewController`

### Checklist

### Authors

- @jaxdesmarais


stechiu:
Out of curiosity, why are we removing staging?

jaxdesmarais:
> Out of curiosity, why are we removing staging?

We only had it in our demo app in the first place because sandbox and production for this feature was not released. Now that we can test in sandbox and production, we don't need to keep this around. Eventually the logic will be removed from the GW for this and our deployed backing services will expire so better for us to clean this up before merging to main.

tdchow:
Would we ever have a long-term need to support running the demo app in stage?

PR #1321, opened 2024-05-28
[QL] Measure API Request Latency for all modules

richherrera:
### Summary of changes

- Send `start_time`, `end_time`, and `endpoint` to FPTI for tracking API request latency for all modules
- Update `UnitTests`

### Checklist

- [x] Added a changelog entry

### Authors

- @richherrera

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
-* Send `start_time`, `end_time`, and `endpoint` to FPTI for tracking API request latency
 * BraintreeCore
   * Batch analytics events to FPTI
+  * Send `start_time`, `end_time`, and `endpoint` to FPTI for tracking API request latency
 
 ## 6.18.2 (2024-05-15)
 * BraintreePayPal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* Send `start_time`, `end_time`, and `endpoint` to FPTI for tracking API request latency
 * BraintreeCore
   * Batch analytics events to FPTI
 
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -19,10 +19,6 @@ import Foundation
     /// Client metadata that is used for tracking the client session
     public private(set) var metadata: BTClientMetadata
 
-    /// :nodoc: This property is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
-    @_documentation(visibility: private)
-    public var shouldSendAPIRequestLatency: Bool = false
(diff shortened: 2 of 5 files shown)


scannillo (approved):

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -510,7 +506,7 @@ import Foundation
     func fetchAPITiming(path: String, startTime: Int, endTime: Int) {
         let cleanedPath = path.replacingOccurrences(of: "/merchants/([A-Za-z0-9]+)/client_api", with: "", options: .regularExpression)
 
-        if (shouldSendAPIRequestLatency || path.contains("v1/configuration")) && cleanedPath != "/v1/tracking/batch/events" {
+        if cleanedPath != "/v1/tracking/batch/events" {
```
jaxdesmarais, line 509: It looks like we aren't currently sending anything for requests going through GraphQL (for example `Card` flows). What do we want to send in these instances? cc: @scannillo

scannillo: Ah, that's a great catch! (Yet another reason to move to 1 URLSession implementation 🐔)

We should have the same metrics for GraphQL calls too, but since "endpoint" is always the same, we'd probably need to log the mutation name instead. I'm OK if we want to make a separate ticket to tackle that.

agedd (approved):

Sources/BraintreeCore/BTAPIClient.swift
richherrera: with this change [5783145](https://github.com/braintree/braintree_ios/pull/1321/commits/5783145b461ebb16ab8703587856ac903f4e2fe3) we are able to send this event for GraphQL request, yeap as Sammy said, the path will always the same

richherrera: I'd prefer to address GraphQL mutation in a separate ticket/PR, but I'm open to whatever you prefer

jaxdesmarais: I'm find tackling this in a separate ticket. Created DTMOBILES-683 for this.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* Send `start_time`, `end_time`, and `endpoint` to FPTI for tracking API request latency
```
jaxdesmarais, line 4: We can move this under `BraintreeCore`

jaxdesmarais (approved):

CHANGELOG.md
richherrera: yes, we can [4607716](https://github.com/braintree/braintree_ios/pull/1321/commits/460771682793594fef03f858902cdf4286d0b6c1)

Sources/BraintreeCore/BTAPIClient.swift
scannillo: Awesome, thanks for creating that ticket

PR #1324, opened 2024-05-28
Fix PayPal Unit Tests

jaxdesmarais:
### Summary of changes

- Unit tests for this PR have been running since last week: https://github.com/braintree/braintree_ios/actions/runs/9215293641/job/25353361230, this is causing failures across multiple PRs where Unit Tests are running continuously
- I observed the same locally and tracked it down to an issue where new tests were not properly setting `webAuthenticationSession` this was causing the tests to never stop running both locally and on CI
- Renamed `lastPostedVaultType` to `postedIsVaultRequest` to properly represent the posted tag name for that event

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
+++ b/UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
@@ -19,6 +19,7 @@ class BTPayPalClient_Tests: XCTestCase {
             "paymentResource": ["redirectUrl": "http://fakeURL.com"]
         ])
         payPalClient = BTPayPalClient(apiClient: mockAPIClient)
+        payPalClient.webAuthenticationSession = MockWebAuthenticationSession()
     }
 
     func testTokenizePayPalAccount_whenRemoteConfigurationFetchFails_callsBackWithConfigurationError() {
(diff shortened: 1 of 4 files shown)


UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -685,15 +685,19 @@ class BTPayPalClient_Tests: XCTestCase {
 
     func testTokenize_whenVaultRequest_setsVaultAnalyticsTag() async {
         let vaultRequest = BTPayPalVaultRequest()
+        payPalClient.webAuthenticationSession = MockWebAuthenticationSession()
```
scannillo, line 688: Can we do this in the setup instead?

scannillo: Also IMO I would rather the `postedIsVaultRequest` assertions just be tacked onto another test vs added a new test entirely for - but can go either way

jaxdesmarais: https://github.com/braintree/braintree_ios/pull/1324/commits/c522adfa0e78f3865008b9dc7ca3b04948bdf572

> Also IMO I would rather the postedIsVaultRequest assertions just be tacked onto another test vs added a new test entirely for - but can go either way

I don't have a strong preference, looks like new tests were created based on this comment: https://github.com/braintree/braintree_ios/pull/1296#discussion_r1606882074. If you'd rather, I can throw these assertions into existing tests.

scannillo (approved):

agedd (approved):
thanks for fixing this @jaxdesmarais! 🚀

PR #1333, opened 2024-06-07
[QL] Refactor Spike for Multiple v1/config calls

KunJeongPark:
### Summary of changes

- Recreate merchant integration that instantiates BTAPIClient on button press along with Payment Method Client instantiation
- Inject config object into BTAnalyticsService to avoid additional v1/config calls

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark @richherrera


KunJeongPark:
I will write a test that fails against https://github.com/braintree/braintree_ios/pull/1333/commits/2a2bf1e5c17d265f45f427471540e728b9ba8532 before I made changes on for config injection to BTAnalyticsService.

 It should still fail since there is still duplication. 

Plan is to create ConfigurationLoader per Sammy's suggestion and
incorporate elements of previous PR on Investigate triple v1/config call ticket in the ConfigurationLoader. https://github.com/braintree/braintree_ios/pull/1308

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -22,11 +22,14 @@ class BTAnalyticsService: Equatable {
...
 
     private static var timer: DispatchSourceTimer?
-    
+
+    private let configuration: BTConfiguration
+
     // MARK: - Initializer
 
-    init(apiClient: BTAPIClient) {
+    init(apiClient: BTAPIClient, configuration: BTConfiguration) {
```
scannillo, line 30: 👏 I like this a lot!

Demo/Application/Features/VenmoViewController.swift
```diff
@@ -67,6 +66,8 @@ class VenmoViewController: PaymentButtonBaseViewController {
     func checkout(request: BTVenmoRequest)  {
         Task {
             do {
+                let apiClient = BTAPIClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")!
+                let venmoClient = BTVenmoClient(apiClient: apiClient)
```
scannillo, line 70: While this helps us test this merchant integration style, it now renders our Settings screen obsolete since the env & auth type settings will be ignored for every feature VC.

We could include these changes in a git patch and apply them locally for our own testing? This PR can be just the BTAnalyticsService changes - what do you think?

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -102,44 +105,38 @@ class BTAnalyticsService: Equatable {
         )
 
         await BTAnalyticsService.events.append(event)
-        
-        do {
-            let configuration = try await apiClient.fetchConfiguration()
```
scannillo, line 107: This `apiClient.fetchConfiguration()` isn't used anymore, if we want to delete it entirely

Demo/Application/Features/VenmoViewController.swift
KunJeongPark: Oh yeah, these were not meant to be permanent changes. It was just to reproduce the conditions

KunJeongPark: Do you want me to write a test to fail before the BTAnalyticsService change on this PR?
BTAnalyticsService's omission of that config call is just one of 3 that I usually see with this merchant integration scenario.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
richherrera: Nice, SOLID principle 💯 .

I was thinking about what would happen when the configuration expires, although the injection of the BTConfiguration solves a problem, we need to verify what would happen in this scenario where Configuration cache expires

KunJeongPark: That's a good point. I do think we want to avoid calling fetchOrGetRemoteConfig in sendAnalyticsEvent.

Thinking out loud, could we check if configuration passed in here against latest cached value when we call sendQueuedAnalyticsEvents? 5 minutes for cache expiration is a random time we set. 
fetchOrGetRemoteConfig gets called in every client classes in functions like tokenize. 

This is all assuming that these values need to be expired during a session which I am not 100% sure about.

KunJeongPark: I was thinking about notification pattern in each payment client. That seems like path that doesn't require a lot of code churn.

Demo/Application/Features/VenmoViewController.swift
KunJeongPark: Let me look into git patch 😅
And as to writing unit test failing on main, I don't see a way of simulating quick fetchOrRemoteConfig calls involving ConfigCache right now. Maybe with ConfigutationLoader, we can test this scenario.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
KunJeongPark: That's effectively what happens with barrier or any kind of locking mechanism. We just need to move it to ConfigurationLoader class. And this needs to be done before `sendAnalyticsEvent` as I suggested, like in Android if we are sending values from the config into the analytics.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -65,12 +65,15 @@ import Foundation
...
         super.init()
-        BTAPIClient._analyticsService = BTAnalyticsService(apiClient: self)
+
         http?.networkTimingDelegate = self
 
         // Kickoff the background request to fetch the config
         fetchOrReturnRemoteConfiguration { configuration, error in
             // No-op
+            if let configuration {
+                BTAPIClient._analyticsService = BTAnalyticsService(apiClient: self, configuration: configuration)
```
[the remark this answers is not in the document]
KunJeongPark: I incorporated my implementation of fetchOrReturnRemoteConfiguration with barrier block from previous investigation PR here. I don't think there is still a guarantee that the analyticsService will be set before first analytic event from tokenize function that I moved into completion block of fetch function in tokenize.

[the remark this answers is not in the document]
KunJeongPark: I can be explicit about returning completion handlers on main in the completion block but handleCompletionAsync in BTHTTP  is set to main by default and not override anywhere else.

Demo/Application/Features/VenmoViewController.swift
scannillo: > And as to writing unit test failing on main, I don't see a way of simulating quick succession of fetchOrRemoteConfig calls involving ConfigCache right now. 

That's OK. This has been a tricky problem to replicate. I think it's OK if we can't write the perfect failing test and we can still make small changes that we think will help this problem.

KunJeongPark: I wrote the test. 45eface812323ba501935cf61dac303722ce87f3

KunJeongPark:
I am thinking about whether injection of config in BTAnalytics is a good idea.

 I think our concerns for using cache still holds for values in sendAnalyticsEvents. 
The barrier I am using in fetchOrGetRemoteConfiguration should stop multiple network calls in one shot.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -65,15 +69,20 @@ import Foundation
         }
         
         super.init()
+
         BTAPIClient._analyticsService = BTAnalyticsService(apiClient: self)
         http?.networkTimingDelegate = self
 
         // Kickoff the background request to fetch the config
         fetchOrReturnRemoteConfiguration { configuration, error in
```
KunJeongPark, line 77: I think we are deciding to disregard prefetch - this would simplify the implementation.

KunJeongPark: Would this save time though for some merchant integrations? Do we want to make sure that BTAnalyticsService is instantiated here? If so, how should we handle the injection. I wrote updateConfiguration function here but also thinking about how merchant would integrate DataCollector which also uses fetchOrReturnRemoteConfiguration.

richherrera: I wonder how much time we are saving the merchant. If Android `v5` doesn't have a prefetch, I think we could follow the same approach. If our concern is to always have `BTAnalyticsService` updated with the configuration values, we could implement an observable pattern

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
scannillo: > However, upon review, it appears that we only need the BTConfiguration object to send the merchantID and the FPTI environment

Some merchants do use multiple merchantIDs (for ex: Gap / Old Navy are 1 company but use different auth credentials for different merchantIDs for each sub-business)

Sources/BraintreeCore/BTAPIClient.swift
KunJeongPark: Yes, I think it's hard to measure except to assume that there won't be a delay for some merchants' integrations where they'd already have config from v1/configurations kicked off in the background and cached.

So for the merchant integration that instantiates BTAPIClient and payment clients earlier before tokenize functions, they'd reduce latency by typical duration of the v1/configuration call when they call functions like tokenize.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
richherrera: I understand that some merchants use different merchant IDs. My question is whether this can change at runtime. For example, is it possible for a user to be making a purchase using our SDK and the merchant ID changes from GAP to Old Navy in a subsequent request after our cache expires?

Demo/Application/Features/VenmoViewController.swift
KunJeongPark: I'm concerned about concurrent cache writes that can happen with current implementation for some merchant integrations that might trigger quick successive calls to fetchOrReturnRemoteConfiguration.

PR #1336, opened 2024-06-14
[Do not review] Create Config loader

richherrera:

### Summary of changes

- 

### Checklist

- [ ] Added a changelog entry

-

### DIFF
--- a/Sources/BraintreeCore/Authorization/ConfigurationLoader.swift
+++ b/Sources/BraintreeCore/Authorization/ConfigurationLoader.swift
@@ -1,0 +1,45 @@
+
+protocol ConfigurationLoadable {
+    func getConfiguration() async throws -> BTConfiguration
+}
+
+final class ConfigurationLoader: ConfigurationLoadable {
+    
+    public var authorization: ClientAuthorization
+    private var cache = ConfigurationCache.shared
+    private lazy var http: BTHTTP? = BTHTTP(authorization: authorization)
+    
+    init(authorization: ClientAuthorization) {
+        self.authorization = authorization
+    }
+    
+    func getConfiguration() async throws -> BTConfiguration {
+        try await withCheckedThrowingContinuation { continuation in
+            let configPath = "v1/configuration"
+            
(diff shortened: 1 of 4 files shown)


Sources/BraintreeCore/Authorization/ConfigurationLoader.swift
```diff
@@ -0,0 +1,45 @@
...
+    public var authorization: ClientAuthorization
+    private var cache = ConfigurationCache.shared
+    private lazy var http: BTHTTP? = BTHTTP(authorization: authorization)
+    
+    init(authorization: ClientAuthorization) {
+        self.authorization = authorization
+    }
+    
+    func getConfiguration() async throws -> BTConfiguration {
+        try await withCheckedThrowingContinuation { continuation in
```
KunJeongPark, line 17: Nice! Even disregarding prefetch,  I don't think this addresses the multiple network calls if somehow this function is called concurrently.

KunJeongPark: If we assume most merchants have this integration - BTAPIClient, BTPaymentMethodClient, tokenize in one click, I'm happy to disregard prefetching, makes things a lot easier.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -278,27 +261,16 @@ import Foundation
     ///   HTTP response and `error` will be `nil`; on failure, `body` and `response` will be
     ///   `nil` and `error` will contain the error that occurred.
     @_documentation(visibility: private)
-    @objc(POST:parameters:httpType:completion:)
+    @objc(POST:parameters:httpType:configuration:completion:)
     public func post(
         _ path: String,
         parameters: [String: Any]? = nil,
         httpType: BTAPIClientHTTPService = .gateway,
+        configuration: BTConfiguration? = nil,
```
KunJeongPark, line 269: this is very clean! We are assuming same config per user action like tokenize, right?

PR #1339, opened 2024-06-18
Add customFields for 3DS verifyCard

warmkesselj:
### Summary of changes

Accept a raw JSON format for customFields in BTThreeDSecureRequest for VerifyCard.

### Checklist

- [X] Added a changelog entry

### Authors
@agedd 
@jwarmkessel 
@warmkesselj

### DIFF
--- a/Demo/Application/Features/ThreeDSecureViewController.swift
+++ b/Demo/Application/Features/ThreeDSecureViewController.swift
@@ -92,7 +92,12 @@ class ThreeDSecureViewController: PaymentButtonBaseViewController {
     private func createThreeDSecureRequest(with nonce: String) -> BTThreeDSecureRequest {
         let request = BTThreeDSecureRequest()
         
-        request.customFields = ["poop": "poop"]
+        let customFields = [
+            "test" : "test",
+            "test1": "test1",
+        ]
+        
+        request.customFields = customFields
         request.threeDSecureRequestDelegate = self
         request.amount = 10.32
         request.nonce = nonce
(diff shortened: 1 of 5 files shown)


Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -11,6 +11,9 @@ import BraintreeCore
 
     /// A nonce to be verified by ThreeDSecure
     public var nonce: String?
+    
+    /// Object where each key is the name of a custom field which has been configured in the Control Panel. In the Control Panel you can configure 3D Secure Rules which trigger on certain values.
+    public var customFields: [String: Any]? = [:]
```
scannillo, line 16: Similar comment [to here](https://github.com/braintree/braintree_android/pull/1044/files#r1650990591), with making this a more standard type that merchants use to model JSON - `Encodable`. A dictionary will also work though.

Which do you think is better merchant dev-ex?

Demo/Application/Features/ThreeDSecureViewController.swift
```diff
@@ -91,6 +91,8 @@ class ThreeDSecureViewController: PaymentButtonBaseViewController {
 
     private func createThreeDSecureRequest(with nonce: String) -> BTThreeDSecureRequest {
         let request = BTThreeDSecureRequest()
+        
+        request.customFields = ["poop": "poop"]
```
sshropshire, line 95:
```suggestion
        request.customFields = ["💩": "💩"]
```

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
warmkesselj: I just got more information that we'll be supporting no nested JSON values and n number of custom values. I'm waiting on clarification to determine if we're expecting only String values.

warmkesselj: Confirmed we can enforce string key values. I think [String: String] is fine. The merchant will be setting these values in their control panel. Btw, the Cardinal team is taking anything but stringifying everything coming through the gateway.

sshropshire (approved):

KunJeongPark (approved):
Great job Justin!

PR #1338, opened 2024-06-17
Fix Button Overlap on `ShopperInsightsViewController`

jaxdesmarais:
### Summary of changes

- There was some button overlap on the `ShopperInsightsViewController`

### Visuals
| Before | After |
|-----|-----|
|Simulator Screenshot - iPhone SE (3rd generation) - 2024-06-17 at 14 29 32|Simulator Screenshot - iPhone SE (3rd generation) - 2024-06-17 at 15 46 33|
|Simulator Screenshot - iPhone 15 Pro Max - 2024-06-17 at 11 19 31|Simulator Screenshot - iPhone 15 - 2024-06-17 at 15 47 59

### Checklist

- ~[ ] Added a changelog entry~

### Authors

- @jaxdesmarais

PR #1340, opened 2024-06-20
Fix Squished `PayPalWebCheckoutViewController`

jaxdesmarais:
### Summary of changes

- Fix Squished `PayPalWebCheckoutViewController` (introduced in #1338)
- Add a `hightConstraint` property to `PaymentButtonBaseViewController` so we can set heights dynamically vs all views using the same height

| Before | After |
|-----|-----|
|Simulator Screenshot - iPhone 15 - 2024-06-20 at 12 50 21|Simulator Screenshot - iPhone 15 - 2024-06-20 at 12 55 09|

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Demo/Application/Base/PaymentButtonBaseViewController.swift
+++ b/Demo/Application/Base/PaymentButtonBaseViewController.swift
@@ -4,7 +4,9 @@ import BraintreeCore
 class PaymentButtonBaseViewController: BaseViewController {
 
     let apiClient: BTAPIClient
-    
+
+    var heightConstraint: CGFloat?
+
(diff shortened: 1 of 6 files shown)


KunJeongPark:
Nice! I think I only saw this issue with some simulators, is that right?

KunJeongPark (approved):

jaxdesmarais:
> Nice! I think I only saw this issue with some simulators, is that right?

No this was all simulators, the shopper insights bug was only on some that didn't have enough space for both views it was displaying.

scannillo (approved):

PR #1342, opened 2024-06-21
Fix Constraints Error in `PayPalWebCheckoutViewController`

jaxdesmarais:
### Summary of changes

- Fix constraint error caused by not having a width constraint on our nested `buttonsStackView`s. This was caused because we have `layoutMargins` on these nested `buttonsStackView`s but those views didn't have a width set to constrain the `layoutMargins` to. Setting leading and trailing anchors to the parent stack view allows the layout to know where to inset the `layoutMargins` from.
- The following errors in the console are now gone:
    Screenshot 2024-06-18 at 1 11 05 PM

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Demo/Application/Base/AppDelegate.swift
+++ b/Demo/Application/Base/AppDelegate.swift
@@ -64,7 +64,6 @@ import BraintreeCore
         if let preferences = settings.object(forKey: "PreferenceSpecifiers") as? Array<[String: Any]> {
             var defaultsToRegister: [String: Any] = [:]
             preferences.forEach { prefSpecification in
-                print(prefSpecification)
                 if let key = prefSpecification["Key"] as? String, prefSpecification.keys.contains("DefaultValue") {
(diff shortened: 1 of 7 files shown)


Demo/Application/Base/AppDelegate.swift
```diff
@@ -64,7 +64,6 @@ import BraintreeCore
         if let preferences = settings.object(forKey: "PreferenceSpecifiers") as? Array<[String: Any]> {
             var defaultsToRegister: [String: Any] = [:]
             preferences.forEach { prefSpecification in
-                print(prefSpecification)
```
jaxdesmarais, line 67: This was printing out a ton of not very helpful things to the console. Removing this should allow us to more easily debug things printed to the console.

richherrera (approved):

agedd (approved):
🚀

PR #1344, opened 2024-06-24
Fix dispatchSource

richherrera:
### Summary of changes

- Create a class called `RepeatingTimer` that uses `DispatchSource`, moving all the logic from the `BTAnalyticsService` class. This way, we can reuse the same logic wherever needed. This class ensures it can be properly deallocated from memory or the implementation of pause/resume functionality. Also, `RepeatingTimer` aims to fix an intermittent crash.

Steps to reproduce the crash:
`Xcode 15.4`
`iOS 17.2`
`Simulator iPhone 14`

This [branch](https://github.com/braintree/braintree_ios/tree/instantiation-clients) was created where the `APIClient` and `PaymentClient` are instantiated almost at the same time, based on a potential implementation by the merchants. The steps to reproduce the crash are to `tokenize` and then `cancel` (crash is very intermittent, so you'll need to run the application multiple times and follow the tokenization process repeatedly).

Screenshot 2024-06-25 at 9 09 32 a m

Update: This [branch](https://github.com/braintree/braintree_ios/tree/test_multiple_instances) includes the changes from the branch that reproduces the crash, as well as the changes from this branch.

### Checklist

- [x] Added a changelog entry

### Authors

- @richherrera


Sources/BraintreeCore/RepeatingTimer.swift
```diff
@@ -0,0 +1,58 @@
...
+    }
+    
+    private var state: State = .suspended
+    
+    deinit {
+        timer.setEventHandler {}
+        timer.cancel()
+        /*
+         If the timer is suspended, calling cancel without resuming
+         triggers a crash. This is documented here https://forums.developer.apple.com/thread/15902
```
jaxdesmarais, line 39: Based on this comment this seems like it could be the main source of the crash. Have we tried just adding a `resume` call to our `deinit`?

jaxdesmarais (commented):
Are the steps we can follow to replicate this error? Will be helpful for folks testing/reviewing this PR.

Sources/BraintreeCore/RepeatingTimer.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+
+final class RepeatingTimer {
```
scannillo, line 3: I like this idea of moving the timer wrapping & management logic into it's own class

Sources/BraintreeCore/RepeatingTimer.swift
scannillo, line 39: Nitpick

```suggestion
         If the timer is suspended, calling cancel without resuming afterwards
         triggers a crash. This is documented here https://forums.developer.apple.com/thread/15902
```

Sources/BraintreeCore/RepeatingTimer.swift
```diff
@@ -0,0 +1,58 @@
...
+    }()
+    
+    var eventHandler: (() -> Void)?
+    
+    private enum State {
+        case suspended
+        case resumed
+    }
+    
+    private var state: State = .suspended
```
scannillo, line 32: Curious why we need to manage this state ourselves?

richherrera: GCD timers can be somewhat sensitive. If you try and `resume`/`suspend` an already `resumed`/`suspended` timer, you will get a crash with a reason like:

> BUG IN CLIENT OF LIBDISPATCH: Over-resume of an object

This tells us we have tried to resume an already resumed timer. We just need to balance the calls to suspend and resume, like the [documentation states](https://developer.apple.com/library/content/documentation/General/Conceptual/ConcurrencyProgrammingGuide/GCDWorkQueues/GCDWorkQueues.html#//apple_ref/doc/uid/TP40008091-CH103-SW8):

> As a result, you must balance each call to dispatch_suspend with a matching call to dispatch_resume before event delivery resumes.

Sources/BraintreeCore/RepeatingTimer.swift
richherrera: Added: [ae55b24](https://github.com/braintree/braintree_ios/pull/1344/commits/ae55b247310f288e119a4709d18050ce4ac52750)

richherrera: Yes, however, I think moving the logic to a separate class would help us segregate the responsibilities of each class. Another benefit is that we can better manage the instantiation of `DispatchSource` (RepeatingTimer). Previously, it was possible to instantiate it more than once, even when we [validated](https://github.com/braintree/braintree_ios/blob/f03147676b81f938c04c22525ffc60befd07441d/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift#L121) that it was equal to nil.

Braintree.xcodeproj/project.pbxproj
```diff
@@ -53,6 +53,8 @@
 		428F976626727333001042E1 /* BTMockOpenURLContext.m in Sources */ = {isa = PBXBuildFile; fileRef = 428F976526727333001042E1 /* BTMockOpenURLContext.m */; };
 		42FC218B25CDE0290047C49A /* BTPayPalRequest_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42FC218A25CDE0290047C49A /* BTPayPalRequest_Tests.swift */; };
 		42FC237125CE0E110047C49A /* BTPayPalCheckoutRequest_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 42FC237025CE0E110047C49A /* BTPayPalCheckoutRequest_Tests.swift */; };
+		457D7FC82C29CEC300EF6523 /* RepeatingTimer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 457D7FC72C29CEC300EF6523 /* RepeatingTimer.swift */; };
```
scannillo, line 56: Nit, this file can go in the `Analytics` directory within `BraintreeCore`

Sources/BraintreeCore/RepeatingTimer.swift
scannillo: Gotcha, thanks for the background here. This makes sense

UnitTests/BraintreeCoreTests/RepeatingTimer_Tests.swift
```diff
@@ -0,0 +1,56 @@
+import XCTest
+@testable import BraintreeCore
+
+class RepeatingTimerTests: XCTestCase {
+
+    func testTimer_resumesAndSuspends() {
```
scannillo, line 6: Nitpick - we try to name tests like this format: `test<MethodName>_{optional when<Condition>}_<ExpectedResult>`

This test might be more readable as two separate tests. For ex: `testResume_callsEventHandler` or `testSuspend_eventHandlerNotCalled`

Sources/BraintreeCore/RepeatingTimer.swift
scannillo: Can we include that link & those details in a comment?

Sources/BraintreeCore/RepeatingTimer.swift
jaxdesmarais, line 3:
```suggestion
final class BTRepeatingTimer {
```

Sources/BraintreeCore/RepeatingTimer.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+
+final class RepeatingTimer {
+    
+    /// Amount of time, in seconds.
+    private let timeInterval: Int
+    
+    init(timeInterval: Int) {
```
jaxdesmarais, line 8: Since we've separate out this logic can we remove the `timeInterval` property from `BTAnalyticsService` and set the default in this class? That way the timer logic is fully contained here and we are only instantiating it in `BTAnalyticsService`.

Sources/BraintreeCore/RepeatingTimer.swift
```diff
@@ -0,0 +1,58 @@
...
+    private lazy var timer: DispatchSourceTimer = {
+        let timerSource = DispatchSource.makeTimerSource(queue: DispatchQueue.main)
+        timerSource.schedule(
+            deadline: .now() + .seconds(timeInterval),
+            repeating: .seconds(timeInterval),
+            leeway: .seconds(1)
+        )
+        timerSource.setEventHandler(handler: { [weak self] in
+            self?.eventHandler?()
+        })
```
jaxdesmarais, line 21:
```suggestion
        timerSource.setEventHandler { [weak self] in
            self?.eventHandler?()
        }
```

Sources/BraintreeCore/RepeatingTimer.swift
```diff
@@ -0,0 +1,58 @@
...
+    
+    private enum State {
+        case suspended
+        case resumed
+    }
+    
+    private var state: State = .suspended
+    
+    deinit {
+        timer.setEventHandler {}
```
jaxdesmarais, line 35:
```suggestion
        timer.setEventHandler { }
```

Sources/BraintreeCore/RepeatingTimer.swift
```diff
@@ -0,0 +1,58 @@
...
+        timer.cancel()
+        /*
+         If the timer is suspended, calling cancel without resuming afterwards
+         triggers a crash. This is documented here https://forums.developer.apple.com/thread/15902
+         */
```
jaxdesmarais, line 40:
```suggestion
         // If the timer is suspended, calling cancel without resuming afterwards
         // triggers a crash. This is documented here https://forums.developer.apple.com/thread/15902
```

UnitTests/BraintreeCoreTests/RepeatingTimer_Tests.swift
```diff
@@ -0,0 +1,65 @@
...
+        
+        DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
+            expectation.fulfill()
+        }
+        
+        waitForExpectations(timeout: 4, handler: nil)
+        XCTAssertFalse(handlerCalled, "Event handler should not be called after timer is suspended")
+    }
+
+    func testEventHandler_isCalled() {
```
scannillo, line 41: Does this test the same this as the `testResume_callsEventHandler` test above?

Sources/BraintreeCore/RepeatingTimer.swift
```diff
@@ -0,0 +1,70 @@
...
+        )
+        timerSource.setEventHandler { [weak self] in
+            self?.eventHandler?()
+        }
+        return timerSource
+    }()
+    
+    var eventHandler: (() -> Void)?
+    
+    private enum State {
```
scannillo, line 27: Another nitpick/styling - to try to use pragma marks according to our internal style guide. I'll grab a link and send this to you internally

Screenshot 2024-06-26 at 8 17 02 AM

jaxdesmarais (commented):
Lets make sure to add a CHANGELOG entry for this

Demo/Application/Base/AppDelegate.swift
```diff
@@ -64,7 +64,6 @@ import BraintreeCore
         if let preferences = settings.object(forKey: "PreferenceSpecifiers") as? Array<[String: Any]> {
             var defaultsToRegister: [String: Any] = [:]
             preferences.forEach { prefSpecification in
-                print(prefSpecification)
```
jaxdesmarais, line 67: Looks like some changes that were made on `main` are showing as diffs here and below, think we need to resolve those since they shouldn't show as a diff here.

Sources/BraintreeCore/RepeatingTimer.swift
scannillo: Yeh, we can just hard code it to 15 in the `RepeatingTimer` class. BTAnalyticsService doesn't need to change the interval

Sources/BraintreeCore/RepeatingTimer.swift
scannillo: Yeh, we can just hard code it to 15 in the RepeatingTimer class. BTAnalyticsService doesn't need to change the interval

jaxdesmarais: @richherrera wanted to make sure you saw this

Braintree.xcodeproj/project.pbxproj
scannillo: @richherrera just wanted to follow up on the few remaining open comments

Sources/BraintreeCore/RepeatingTimer.swift
```diff
@@ -0,0 +1,71 @@
...
+
+       BUG IN CLIENT OF LIBDISPATCH: Over-resume of an object
+
+     Such errors indicate an attempt to resume an already resumed timer. According to the
+     documentation, each call to `dispatch_suspend` must be matched with a corresponding
+     call to `dispatch_resume` to ensure proper event delivery and avoid crashes.
+     For more information: https://developer.apple.com/library/archive/documentation/General/Conceptual/ConcurrencyProgrammingGuide/GCDWorkQueues/GCDWorkQueues.html#//apple_ref/doc/uid/TP40008091-CH103-SW8
+    */
+    
+    func resume() {
```
KunJeongPark, line 58: This PR is great work Rich. I am wondering if we want to synchronize resume and suspends as well since analytics could be sent from different threads. We could use actor but it would require waits on function calls and we could also use serial dispatch queue. wdyt?

richherrera: Excellent suggestion. For now, I think `DispatchSource` is a task that runs periodically, so we could proceed without needing _actor_'s. Also, based on Jax's implementation, the `DispatchSource` is created to run on the main thread. From what I see, the task of sending analytics can be on different threads, but both only enqueue events. The task of sending them is handled by the single instance of `DispatchSource`.

KunJeongPark: Interesting. I didn't know that this line `let timerSource = DispatchSource.makeTimerSource(queue: DispatchQueue.main)` ensures that the handler is run on main as well.

KunJeongPark (approved):

Braintree.xcodeproj/project.pbxproj
scannillo: I took care of this!

scannillo (approved):
👏 We want to try to get this merged & released by EOW as an LE merchant is integrating w/ App Switch next week

Sources/BraintreeCore/RepeatingTimer.swift
jaxdesmarais: I am not sure that I agree with this. `BTAnalyticsService` only responsibility should be sending events, it should not care about how often we send events at all. On the other hand the `RepeatingTimer` should control all time related logic including the interval of the timer. I don't think `BTAnalyticsService` needs to care how often events are sent only that it's sending events.

richherrera: Sure, I completely agree with you. I was thinking more about making this class usable in multiple places if needed, considering that in some other context, 15 seconds might be too short. But I'm happy to make the change. Do you think it would be good to put it in the initializer?

PR #1347, opened 2024-06-26
Fix Xcode 16-beta2 Sendable warning

scannillo:
### Summary 

Running our Demo app on Xcode 16-beta2 results in the following warning:

```
Stored property 'pinnedCertificates' of 'Sendable'-conforming class 'BTHTTP' has non-sendable type '[NSData]'; this is an error in the Swift 6 language mode.
```

### Changes

 - Update BTHTTP.pinnedCertificates array to Swift `Data` type, over `NSData`

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


jaxdesmarais (approved):
Have we also updated the `braintree-ios-certificate-generation` repo with this change or will that be a follow up? Just want to make sure next time we need to generate the certs we don't revert this change.

agedd (approved):
🚀

scannillo:
> Have we also updated the braintree-ios-certificate-generation repo with this change or will that be a follow up? Just want to make sure next time we need to generate the certs we don't revert this change.

Good catch. I will update this there as well!

PR #1353, opened 2024-07-02
Add Configuration Loader

richherrera:
### Summary of changes

- Refactor the existing logic by creating a `ConfigurationLoader` object, moving all the logic from the `fetchOrReturnRemoteConfiguration` method to this new object. The public method (`fetchOrReturnRemoteConfiguration`) remain unchanged to avoid requiring modifications from the `clients`. This will help us maintain similarities with `Android` and isolate the logic, allowing us to better implement a mechanism to avoid multiple calls to the `v1/configuration` endpoint.

### Checklist

- [X] Added a changelog entry

### Authors

- @richherrera


Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -50,21 +52,22 @@ import Foundation
...
                 self.authorization = clientToken
-
-                http = BTHTTP(authorization: self.authorization)
             } catch {
                 return nil
             }
         }
         
+        let btHttp = BTHTTP(authorization: self.authorization)
+        http = btHttp
```
scannillo, line 68:
```suggestion
        http = BTHTTP(authorization: self.authorization)
```

Sources/BraintreeCore/ConfigurationLoader.swift
```diff
@@ -0,0 +1,66 @@
...
+    
+    init(http: BTHTTP) {
+        self.http = http
+    }
+    
+    deinit {
+        http.session.finishTasksAndInvalidate()
+    }
+    
+    func getConfig(_ authorization: ClientAuthorization, completion: @escaping (BTConfiguration?, Error?) -> Void) {
```
scannillo, line 19: Just a few more MARKs we usually use, but open to if folks don't find these useful

```suggestion
    // MARK: - Intitializer

    init(http: BTHTTP) {
        self.http = http
    }
    
    deinit {
        http.session.finishTasksAndInvalidate()
    }
    
    // MARK: - Internal Methods
    
    func getConfig(_ authorization: ClientAuthorization, completion: @escaping (BTConfiguration?, Error?) -> Void) {
```

Sources/BraintreeCore/ConfigurationLoader.swift
```diff
@@ -0,0 +1,66 @@
...
+        // Fetches or returns the configuration and caches the response in the GET BTHTTP call if successful
```
scannillo, line 20: Nit - this should probably be a docstring instead of a comment

Sources/BraintreeCore/BTAPIClient.swift
richherrera: yeah, i dont like it neeither, I've created this:

`let btHttp = BTHTTP(authorization: self.authorization)
`
because `BTAPIClient.http` is optional and` ConfigurationLoader.init` needs a `non-optional` instance of BTHTTP.

> configurationLoader = ConfigurationLoader(http: btHttp)

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -22,6 +22,8 @@ import Foundation
     var graphQLHTTP: BTGraphQLHTTP?
     var payPalHTTP: BTHTTP?
```
scannillo, line 24:
```suggestion
```

scannillo (approved):
👏 Some minor comments, but this looks great!

Sources/BraintreeCore/BTAPIClient.swift
richherrera: This ensures that I have the same instance for both `BTAPIClient` and `ConfigurationLoader`, and it avoids having to use "!" when instantiating the `ConfigurationLoader`.

```
let btHttp = BTHTTP(authorization: self.authorization)
http = btHttp
configurationLoader = ConfigurationLoader(http: btHttp)
```

scannillo: Ah, that makes sense!

Sources/BraintreeCore/ConfigurationLoader.swift
richherrera: Thats a good idea, updated: [418b39d](https://github.com/braintree/braintree_ios/pull/1353/commits/418b39d4f9c799335b6393f2fd9195fa034a674f)

Sources/BraintreeCore/ConfigurationLoader.swift
richherrera: I like them, added: [418b39d](https://github.com/braintree/braintree_ios/pull/1353/commits/418b39d4f9c799335b6393f2fd9195fa034a674f)

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreeCore
+  * Add `ConfigurationLoader` class
+
```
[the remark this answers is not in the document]
richherrera: Agree, updated: [646a1ef](https://github.com/braintree/braintree_ios/pull/1353/commits/646a1ef158fc8bb8831ad98ae0ec2dde3b9966d8)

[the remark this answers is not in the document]
richherrera: Agree, updated: [646a1ef](https://github.com/braintree/braintree_ios/pull/1353/commits/646a1ef158fc8bb8831ad98ae0ec2dde3b9966d8)

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -50,21 +51,22 @@ import Foundation
...
-
-                http = BTHTTP(authorization: self.authorization)
             } catch {
                 return nil
             }
         }
         
+        let btHttp = BTHTTP(authorization: self.authorization)
+        http = btHttp
+        configurationLoader = ConfigurationLoader(http: btHttp)
```
KunJeongPark, line 68: I am thinking about the fact that ConfigurationCache is a singleton and thought of ConfigurationLoader as being one as well. In a weird scenario that there are multiple instances of BTAPIClient. wdyt?

sshropshire: This is something we want to do on Android too. Making `ConfigurationLoader` a singleton makes the most sense for our architecture as well.

Sources/BraintreeCore/ConfigurationLoader.swift
```diff
@@ -0,0 +1,78 @@
...
+    private let http: BTHTTP
+    
+    // MARK: - Intitializer
+    
+    init(http: BTHTTP) {
+        self.http = http
+    }
+    
+    deinit {
+        http.session.finishTasksAndInvalidate()
```
KunJeongPark, line 18: Just curious, is this necessary?

UnitTests/BraintreeCoreTests/ConfigurationLoader_Tests.swift
```diff
@@ -0,0 +1,141 @@
...
+        }
+        
+        do {
+            _ = try await asyncTask.value
+            XCTFail("Expected error to be thrown")
+        } catch {
+            XCTAssertEqual(error as NSError, mockError)
+        }
+    }
+}
```
KunJeongPark, line 141: Great job on these tests! Very thorough!

Sources/BraintreeCore/BTAPIClient.swift
richherrera: Something that happened in MXO iOS was that almost everything was a singleton. At least in iOS, we need to define the architecture. I think most of the code was a migration from Objective-C to Swift (IMO there are several areas of opportunity that we could refactor.). For now, I think we can proceed this way and discuss whether it would be necessary to convert it to a singleton in the future. Like any design pattern, it has its pros and cons. I'm not entirely a fan because it violates the SRP (Single Responsibility Principle), and we have to think of a creative way to mock

sshropshire: @richherrera I'm 100% onboard with that concern. The one thing I noticed with MXO though is that the interface was static e.g. `ConfigurationLoader.load()` . If we go for `ConfigurationLoader.shared.load()` we'll preserve the flexibility needed for mocking if we inject the singleton reference in the constructor (we do this quite a bit in Android with [getInstance()](https://github.com/braintree/braintree_android/blob/586cea65db35fd15013e1317c014e97566325aee/BraintreeCore/src/main/java/com/braintreepayments/api/AnalyticsClient.kt#L27) accessors).

KunJeongPark: Yes, passing it in with default initializer makes testing possible.

KunJeongPark: Necessary, not sure, but I think there might be some weirdness in possibly having multiple instances of BTAPIClient and multiple instances of Loader.

richherrera: @sestevens yeah DIP rules, i like that, something I’d like to know is how we could use the same instance of BTHTTP in both BTAPIClient and ConfigurationLoader. Are there two different instances in Android (for BraintreeClient and ConfigurationLoader)?

KunJeongPark: I think for as long as there can be multiple instances of BTAPIClient, we should address those issues. 

I guess with singleton, you'd need to keep track of BTHTTP instance creation in ConfigurationLoader by authorization or something and check with the getConfiguration function.

I think maybe considering multiple instances of BTAPIClient and configurationLoader, we might need to synchronize ConfigurationCache functions as well some time. 

Great work on this Rich.

KunJeongPark (approved):

Sources/BraintreeCore/BTAPIClient.swift
sshropshire: > Are there two different instances in Android (for BraintreeClient and ConfigurationLoader)?

@richherrera On Android we have a single `BraintreeHttpClient` instance that's shared between `BraintreeClient` and `ConfigurationLoader`. That would be equivalent to `BTAPIClient` and `ConfigurationLoader` sharing a `BTHTTP` instance on iOS.

KunJeongPark: I think that's how we want to go eventually.

PR #1362, opened 2024-07-11
Update Custom Fields Implementation

jaxdesmarais:
### Summary of changes

- I noticed on recent CI runs that all of the UI tests for 3D Secure were failing, this is due to the additional of custom fields that are not set up in the gateway - updated to remove these from the ViewController since they were not set on the Android PR: https://github.com/braintree/braintree_android/pull/1044
- Additionally I noticed that with this removed 3DS was still not launching, this is because we were not checking that `customFields` was not nil before adding it to the request dictionary - this is implemented correctly in the Android PR

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## unreleased
 * BraintreeShopperInsights (BETA)
   * Add error when using an invalid authorization type
+* BraintreeThreeDSecure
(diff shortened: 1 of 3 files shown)


stechiu (approved):

agedd (approved):
lgtm! thanks for fixing this 🚀

richherrera (approved):
Thanks for addressing this. I was thinking that maybe we could turn off the `Venmo` UI tests (while we fix them) :/

PR #1363, opened 2024-07-11
Prevent duplicate `v1/configuration` GET requests

scannillo:
### Summary

See Victoria's original PR https://github.com/braintree/braintree_ios/pull/1308 for full description and background.

**Short summary**: our SDK would trigger multiple of the same outbound API request to `GET v1/configuration` if a merchant instantiated their `BTAPIClient` and also called `tokenize()` all upon button click. This is a common pattern for merchants, since nowhere do our docs recommend otherwise.

**Goal**: Prevent duplicate network requests to `GET v1/configuration` if one is already pending over the network.

### Changes

- Keep track of pending requests via an array of completionHandler references in `ConfigurationLoader`

### Manual Testing

- I created this git patch ([trigger_duplicate_config_calls.patch](https://github.com/user-attachments/files/16181529/trigger_duplicate_config_calls.patch)) which you can download and apply to both the `main` branch and then the branch of this PR `config-loader-dedupe` to compare behavior.
   - Apply a patch like: `git apply /path/to/trigger_duplicate_config_calls.patch`
   - This patch includes a log statement for the path of each HTTP request executed, as well as a Demo app update to more easily replicate the duplicate `v1/configuration` calls in quick sequence.

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeCore/ConfigurationCallbackStorage.swift
+++ b/Sources/BraintreeCore/ConfigurationCallbackStorage.swift
@@ -1,23 +1,26 @@
 import Foundation
 
 /// Used to store, access, and manage an array of to-be-invoked `BTConfiguration` GET result callbacks in a thread-safe manner
-actor ConfigurationCallbackStorage {
+class ConfigurationCallbackStorage {
     
+    private let queue = DispatchQueue(label: "com.braintreepayments.ConfigurationCallbackStorage")
     private var pendingCompletions: [(BTConfiguration?, Error?) -> Void] = []
-
+    
     /// The number of completions that are waiting to be invoked
     var count: Int {
-        return pendingCompletions.count
+        queue.sync { pendingCompletions.count }
     }
     
     /// Adds a pending, yet to-be-invoked completion handler
     func add(_ completion: @escaping (BTConfiguration?, Error?) -> Void) {
-        pendingCompletions.append(completion)
+        queue.sync { pendingCompletions.append(completion) }
     }
     
     /// Executes and clears all pending completion handlers
     func invoke(_ configuration: BTConfiguration?, _ error: Error?) {
-        pendingCompletions.forEach { $0(configuration, error) }
-        pendingCompletions.removeAll()
+        queue.sync {
+            pendingCompletions.forEach { $0(configuration, error) }
+            pendingCompletions.removeAll()
+        }
     }
 }
--- a/Sources/BraintreeCore/ConfigurationCallbackStorage.swift
+++ b/Sources/BraintreeCore/ConfigurationCallbackStorage.swift
@@ -1,0 +1,23 @@
+import Foundation
+
+/// Used to store, access, and manage an array of to-be-invoked `BTConfiguration` GET result callbacks in a thread-safe manner
+actor ConfigurationCallbackStorage {
+    
+    private var pendingCompletions: [(BTConfiguration?, Error?) -> Void] = []
+
+    /// The number of completions that are waiting to be invoked
+    var count: Int {
+        return pendingCompletions.count
+    }
+    
+    /// Adds a pending, yet to-be-invoked completion handler
+    func add(_ completion: @escaping (BTConfiguration?, Error?) -> Void) {
+        pendingCompletions.append(completion)
+    }
+    
+    /// Executes and clears all pending completion handlers
+    func invoke(_ configuration: BTConfiguration?, _ error: Error?) {
+        pendingCompletions.forEach { $0(configuration, error) }
+        pendingCompletions.removeAll()
+    }
+}
--- a/Sources/BraintreeCore/ConfigurationLoader.swift
+++ b/Sources/BraintreeCore/ConfigurationLoader.swift
@@ -28,15 +28,14 @@ class ConfigurationLoader {
     /// 3. If fetching the configuration fails, it returns an error.
     ///
     /// - Parameters:
-    ///   - authorization: An `ClientAuthorization` object required to access the configuration.
     ///   - completion: A completion handler that is called with the fetched or cached `BTConfiguration` object or an `Error`.
     ///
     /// - Completion:
     ///   - `BTConfiguration?`: The configuration object if it is successfully fetched or retrieved from the cache.
     ///   - `Error?`: An error object if fetching the configuration fails or if the instance is deallocated.
     @_documentation(visibility: private)
-    func getConfig(_ authorization: ClientAuthorization, completion: @escaping (BTConfiguration?, Error?) -> Void) {
-        if let cachedConfig = try? configurationCache.getFromCache(authorization: authorization.bearer) {
+    func getConfig(completion: @escaping (BTConfiguration?, Error?) -> Void) {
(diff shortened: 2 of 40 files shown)


Sources/BraintreeCore/ConfigurationLoader.swift
```diff
@@ -39,26 +41,33 @@ class ConfigurationLoader {
...
-
-            if let error {
-                completion(nil, error)
-                return
-            } else if response?.statusCode != 200 || body == nil {
-                completion(nil, BTAPIClientError.configurationUnavailable)
-                return
-            } else {
-                let configuration = BTConfiguration(json: body)
+                if let error {
```
KunJeongPark, line 56: We want to handle the pendingCompletions for errors as well.

Sources/BraintreeCore/ConfigurationLoader.swift
```diff
@@ -39,26 +41,33 @@ class ConfigurationLoader {
...
+                } else {
+                    let configuration = BTConfiguration(json: body)
 
-                try? configurationCache.putInCache(authorization: http.authorization.bearer, configuration: configuration)
-                
-                completion(configuration, nil)
-                return
+                    try? configurationCache.putInCache(authorization: http.authorization.bearer, configuration: configuration)
+                    
+                    _ = pendingCompletions.map { $0(configuration, nil) }
```
KunJeongPark, line 67: For me, maps are a bit harder to reason about at quick glance than forEach.

KunJeongPark: Great job! Very clean code.

Sources/BraintreeCore/ConfigurationLoader.swift
scannillo: Done!

Sources/BraintreeCore/ConfigurationLoader.swift
```diff
@@ -39,26 +41,34 @@ class ConfigurationLoader {
...
-            if let error {
-                completion(nil, error)
-                return
-            } else if response?.statusCode != 200 || body == nil {
-                completion(nil, BTAPIClientError.configurationUnavailable)
-                return
-            } else {
-                let configuration = BTConfiguration(json: body)
+                if let error {
+                    completion(nil, error)
```
KunJeongPark, line 58: I meant don't we need to handle the pendingCompletions here?

Sources/BraintreeCore/ConfigurationLoader.swift
```diff
@@ -39,26 +41,34 @@ class ConfigurationLoader {
...
+                    return
+                } else if response?.statusCode != 200 || body == nil {
+                    completion(nil, BTAPIClientError.configurationUnavailable)
```
KunJeongPark, line 61: and here?

Sources/BraintreeCore/ConfigurationLoader.swift
scannillo: We definitely should! Updated

Sources/BraintreeCore/ConfigurationLoader.swift
scannillo: Updated!

Sources/BraintreeCore/ConfigurationLoader.swift
```diff
@@ -74,4 +82,11 @@ class ConfigurationLoader {
             }
         }
     }
+    
+    // MARK: - Private Methods
+    
+    func notifyCompletions(_ configuration: BTConfiguration?, _ error: Error?) {
+        pendingCompletions.forEach { $0(configuration, error) }
```
KunJeongPark, line 89: This is a nit pick, you can disregard, but I think we are assuming this being called from main
as we do our async completion in the get call. Do we want to make those consistent?

KunJeongPark (approved):

KunJeongPark:
Great job Sammy!

Sources/BraintreeCore/ConfigurationLoader.swift
```diff
@@ -39,26 +41,32 @@ class ConfigurationLoader {
             completion(cachedConfig, nil)
```
KunJeongPark, line 41: nit pick : this is where I meant to put the comment about putting this on main since get closure is called from main by default in our code base, I think.

scannillo: I'm a little confused what this means - can you clarify?

KunJeongPark:
One last thing for future improvement is to make array handling thread safe.

scannillo:
> One last thing for future improvement is to make array handling thread safe by using locks or DispatchQueues.

Yeh, this is another good call. I'm thinking of using an Actor like we did[ here](https://github.com/braintree/braintree_ios/pull/1295/files) for our analytics events. This will be easier if everything is in async-await syntax. I might make that change in another PR, and Draft this one for now.

Sources/BraintreeCore/ConfigurationLoader.swift
KunJeongPark: https://github.com/braintree/braintree_ios/blob/b4ff6ce555b099b57c2ef74e225998fd262164b1/Sources/BraintreeCore/BTHTTP.swift#L279
This is defaulted to main and I haven't seen anywhere where we override it. So I was thinking if merchant's app expects this and other completions to be returned on main.

scannillo: Hm, I don't think there's anything additional we need to be doing at the Config level, since this level is essentially just propagating up whatever result we get back from the BTHTTP layer. If the BTHTTP layer is switching over to the main thread to deliver it's result, then the ConfigLoader is also delivering on the main thread once the result is received.

I don't think we should change any thread management logic in this PR.

scannillo:
Ok, I trialed several ways of making the array of completions thread-safe. 

1st I tried actors, similar to [BTAnalyticsEventsStorage](https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreeCore/Analytics/BTAnalyticsEventsStorage.swift). The problem with this is needing to call all actor methods with `await`. This means all callsites must either be in a `Task` block, or in an async function. Wrapping all callsites in `Task` blocks borks our test suite since there is no way for the test to know when a Task block completed. There are ways around this, but it got messy and it didn't feel worth it. 

Converting our ConfigurationLoader.fetchConfig method to only use async-await style syntax was another thing I tried, but we can't store references to the async Task to call at a later point, like we can with completions. There _might_ be a way to do this I'm not aware of.

Lastly, I ended up going with the tried and true `DispatchQueue` for ensuring all array operations are serial.

Sources/BraintreeCore/ConfigurationCallbackStorage.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+/// Used to store, access, and manage an array of to-be-invoked `BTConfiguration` GET result callbacks in a thread-safe manner
+class ConfigurationCallbackStorage {
```
jaxdesmarais, line 4: 🔥

jaxdesmarais (approved):

PR #1370, opened 2024-07-18
Add `config_cached` FPTI tag

scannillo:
### Summary of changes

- Folks from other teams have been confused by our E2E (tokenize() --> web view presentation) timing metrics since the overall metrics may or may not include the `v1/configuration` call, depending on if the config was retrieved from an in-memory cache. This largely depends on how the merchant integrates our SDK.
- This PR adds a boolean tag to our FPTI payload to make it easy to decipher sessions where the config was cached vs not in our analytics queries. The goal is to update our latency dashboard to separate our config cached vs config not-cached sessions into 2 different charts.
- Note:
    - This PR _only_ sends the `config_cached` tag with `paypal:tokenize:browser-presentation:succeeded` events. Would like folks feedback if you agree on only adding it here for now, since it is the only place we currently need it. 

### Checklist

### Authors
@scannillo


scannillo:
@jaxdesmarais Would love your input on this - so I honestly don't _want_ to keep adding analytics specific to 1 flow for this latency workstream, but it keeps coming up. 

The only other way I can think of to be able to decipher sessions based on if the config was cached or not (using existing analytics), is to compare `core:api-request-latency` events for `v1/configuration` with `paypal:tokenize:started` events.

We would compare the API `start_time` of the  `core:api-request-latency` event to the `t` timestamp of the `paypal:tokenize:started` event. If the v1/config start time is _after_ the tokenize:started time, we can assume the config was not cached.

---
If you think that idea ^ is sound, then we can scrap this PR. What do you think?

jaxdesmarais:
🤔 I think what you mentioned sounds like a sound way to rely on it, but I feel like if we have even 1 request that deviates from that we are back to needing to add a new property to FPTI. The only other thing I could potentially see is adding a new event here: https://github.com/braintree/braintree_ios/blob/79a0864fd827d5d433e418efc8af6e1fdb7d4e75/Sources/BraintreeCore/Configuration/ConfigurationLoader.swift#L40

Obviously that increases some complexity where the ConfigurationLoader then needs to know about APIClient. Unless there is a way we can set a flag in ConfigurationLoader that we can check elsewhere and then send a new event if the config was cached? Agree any way about it isn't particularly clean.

scannillo:
> Agree any way about it isn't particularly clean.

Yeh, so we have 3 options:

1) Rely on existing FPTI events. If `core:api-request-latency`.`start_time` after `paypal:tokenize:started`.`t`, assume a network call was made / no cache hit.
2) Add a new tag (like this PR).
3) Send a new event entirely. I think the logic would be similar to this PR where one of the Cache layers sends an indicator if the config was pulled from the cache or not, but fire the actual event [here](https://github.com/braintree/braintree_ios/blob/79a0864fd827d5d433e418efc8af6e1fdb7d4e75/Sources/BraintreePayPal/BTPayPalClient.swift#L310).

A potential wrinkle in 1 is that the `t` timestamp we send along with an analytic event gets calculated in the BTAnalyticsService layer, not at the exact callsite of `tokenize()` start.

jaxdesmarais:
@scannillo my thoughts on each:

> Rely on existing FPTI events. If core:api-request-latency.start_time after paypal:tokenize:started.t, assume a network call was made / no cache hit.

This one is obviously the lowest effort. I think if we are 100% confident that this can always be relied on we should go with this. Our timestamps are in milliseconds which seems like it should be reliable to do this. Always feels risky to say we are 100% confident with limited data though. 😅 Do we feel like saying we are 99% confident this will always be accurate that will get us the data we need?

> Add a new tag (like this PR).
> 
> Send a new event entirely. I think the logic would be similar to this PR where one of the Cache layers sends an indicator if the config was pulled from the cache or not, but fire the actual event [here](https://github.com/braintree/braintree_ios/blob/79a0864fd827d5d433e418efc8af6e1fdb7d4e75/Sources/BraintreePayPal/BTPayPalClient.swift#L310).

These are the same just slightly different approaches so grouping them together. This seems like the easiest way to be 100% confident we are always able to rely on having data for this. But it is a huge trade off where now we are sending a ton more events then we were even a few months ago. At what point would this potentially have a hit to customers especially with poor connections? Obviously it's just 1 event but we've been adding 1 additional event or flag across several PRs.

Are you leaning one direction or the other?

scannillo:
@jaxdesmarais These are all great points, especially about sending a ton of events now & adding more load on user's network usage. After looking at individual data points, I have low confidence in option 1 right now since we had that[ triple config call bug](https://github.com/braintree/braintree_ios/pull/1363). But once the fix goes out, it might be more doable. Maybe we ask Sruthy to poke around see if they can deduce any meaningful results and then go from there?

scannillo:
Moving this out of draft. This gap of not being able to distinguish config-cache vs config-not-cached cases keeps coming up in latency conversations and we need this data sooner than later to help show that the SDK is not taking up time outside of the network requests we make.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -447,6 +452,7 @@ import BraintreeDataCollector
             if didAppear {
                 apiClient.sendAnalyticsEvent(
                     BTPayPalAnalytics.browserPresentationSucceeded,
+                    isConfigFromCache: isConfigFromCache,
```
jaxdesmarais, line 455: Do we only care about this for `browserPresentationSucceeded` or are there any other events we may want to send this for?

jaxdesmarais (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
scannillo: I wanted to get folks feedback on that. Currently this is the only event we are using it for in our dashboard, but do you think there are others it may be useful?

Was operating under not adding it to the others unless there was a clear need - but open to your thoughts

jaxdesmarais: In that case, we should just keep it to this event. If we need to add it elsewhere we can at a later date.

agedd (approved):

PR #1372, opened 2024-07-19
Set `URLSessionConfiguration` timeouts to 30 seconds

scannillo:
### Summary of changes

- Now that we are collecting data ([original PR](https://github.com/braintree/braintree_ios/pull/1292)) on how much time the SDK spends waiting for each network request to complete, we observed some incredibly long outlier datapoints. See the below image for examples:
Screenshot 2024-07-19 at 10 25 09 AM
- The goal of this PR is to reduce the max timeout for each API request. Currently, we're using Apple's defaults which are:
     - The default [timeoutIntervalForRequest](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1408259-timeoutintervalforrequest) is 60 seconds.
    - The default [timeoutIntervalForResource](https://developer.apple.com/documentation/foundation/nsurlsessionconfiguration/1408153-timeoutintervalforresource) is 7 days.
- This PR matches our [Android SDK](https://github.com/braintree/braintree_android/blob/556ec64baf80e6811438e3eef1313bc42ce199dc/SharedUtils/src/main/java/com/braintreepayments/api/sharedutils/HttpRequest.java#L45) and reduces the timeout to 30 seconds.
    - This will allow our merchants to get notified of network errors sooner and allow them to retry or proceed however they please.
    
### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 ## unreleased
 * BraintreeThreeDSecure
   * Add error code and error message for `exceededTimeoutLimit`  
-* Prevent duplicate outbound `v1/configuration` requests
+* BraintreeCore
+  * Prevent duplicate outbound `v1/configuration` requests
+  * Add network timeout of 30 seconds
 
 ## 6.23.0 (2024-07-15)
 * BraintreeShopperInsights (BETA)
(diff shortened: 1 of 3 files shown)


jaxdesmarais (commented):
General question but do we have anywhere where we can surface an error in these cases or will the experience be a no-op for customers/merchants?

scannillo:
> General question but do we have anywhere where we can surface an error in these cases or will the experience be a no-op for customers/merchants?

The error from URLSession when the timeout is reached will bubble up to the merchant via their completionHandler. It should be the same timeout error that they would have already gotten with the default 60 sec / 7 day values, but now that error will come earlier.

sshropshire (approved):

jaxdesmarais (approved):

PR #1373, opened 2024-07-22
Recurring Billing Agreement Add total amount.

warmkesselj:
### Summary of changes

Additional field needed to be added to the iOS feature branch, total_amount. 

### Authors
@jwarmkessel 
@KunJeongPark

### DIFF
--- a/Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
+++ b/Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
@@ -29,7 +29,7 @@ public struct BTPayPalRecurringBillingDetails {
     ///   - shippingAmount: The shipping amount for the billing cycle at the time of checkout.
     ///   - productAmount: The item price for the product associated with the billing cycle at the time of checkout.
     ///   - taxAmount: The taxes for the billing cycle at the time of checkout.
-    ///   - totalAmount: The total amount to pay at the time of checkout.
+    ///   - totalAmount: The total amount associated with the billing cycle at the time of checkout.
     public init(
         billingCycles: [BTPayPalBillingCycle],
         currencyISOCode: String,
@@ -40,7 +40,7 @@ public struct BTPayPalRecurringBillingDetails {
         shippingAmount: String?,
         productAmount: String?,
         taxAmount: String?,
-        totalAmount: String?
+        totalAmount: String
     ) {
         self.billingCycles = billingCycles
         self.currencyISOCode = currencyISOCode
--- a/Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
+++ b/Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
@@ -12,7 +12,7 @@ public struct BTPayPalRecurringBillingDetails {
     private let productQuantity: Int?
     private let oneTimeFeeAmount: String?
     private let shippingAmount: String?
-    private let productAmount: String?
+    private let productPrice: String?
     private let taxAmount: String?
     private let totalAmount: String?
     
@@ -27,7 +27,7 @@ public struct BTPayPalRecurringBillingDetails {
     ///   - productQuantity: Quantity associated with the product.
     ///   - oneTimeFeeAmount: Price and currency for any one-time charges due at plan signup.
     ///   - shippingAmount: The shipping amount for the billing cycle at the time of checkout.
-    ///   - productAmount: The item price for the product associated with the billing cycle at the time of checkout.
+    ///   - productPrice: The item price for the product associated with the billing cycle at the time of checkout.
     ///   - taxAmount: The taxes for the billing cycle at the time of checkout.
     ///   - totalAmount: The total amount associated with the billing cycle at the time of checkout.
     public init(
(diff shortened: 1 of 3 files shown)


sshropshire (approved):

Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
```diff
@@ -37,7 +39,8 @@ public struct BTPayPalRecurringBillingDetails {
         oneTimeFeeAmount: String?,
         shippingAmount: String?,
         productAmount: String?,
-        taxAmount: String?
+        taxAmount: String?,
+        totalAmount: String?
```
sarahkoop: Is this based on the final GW PR for this? I think the `productAmount` field also changed and there is a `productPrice` filed now

scannillo (approved):

Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
warmkesselj: Good catch. The param being sent is `"product_price"` while the interface is `productAmount`. I'll update this to use `productPrice`.
```
if let productAmount {
            parameters["product_price"] = productAmount
        }
```

scannillo: The SDK is sending `product_price` in the JSON blob to the GW.

I intentionally opted to name the public SDK property `amount` instead since it is more consistent with how we name dollar amounts throughout the SDK. We don't use `price` anywhere else.

Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
```diff
@@ -12,8 +12,9 @@ public struct BTPayPalRecurringBillingDetails {
     private let productQuantity: Int?
     private let oneTimeFeeAmount: String?
     private let shippingAmount: String?
-    private let productAmount: String?
+    private let productPrice: String?
```
jaxdesmarais, line 15: Sounds like we want to revert this change and keep `productAmount` based on this comment: https://github.com/braintree/braintree_ios/pull/1373#discussion_r1688004735

Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
```diff
@@ -36,8 +38,9 @@ public struct BTPayPalRecurringBillingDetails {
         productQuantity: Int?,
         oneTimeFeeAmount: String?,
         shippingAmount: String?,
-        productAmount: String?,
-        taxAmount: String?
+        productPrice: String?,
+        taxAmount: String?,
+        totalAmount: String
```
jaxdesmarais, line 43: Total amount is required here and optional above. Should these be the same?

Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
sarahkoop: Ahh sorry my mistake!

Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
warmkesselj: Yes, it should be a required property.

Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
```diff
@@ -14,6 +14,7 @@ public struct BTPayPalRecurringBillingDetails {
     private let shippingAmount: String?
     private let productAmount: String?
     private let taxAmount: String?
+    private let totalAmount: String?
```
jaxdesmarais, line 17:
```suggestion
    private let totalAmount: String
```

jaxdesmarais: Lets also move this to the top with the other required params

Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
jaxdesmarais: Cool, can we move it to the top of the init with the other required params? Along with moving up the docstring.

warmkesselj: ✅

Sources/BraintreePayPal/RecurringBillingMetadata/BTPayPalRecurringBillingDetails.swift
```diff
@@ -86,6 +90,8 @@ public struct BTPayPalRecurringBillingDetails {
             parameters["tax_amount"] = taxAmount
         }
         
+        parameters["total_amount"] = totalAmount
```
jaxdesmarais, line 93: nit: we should be able to add this directly to the base parameters in line 62

agedd (approved):

jaxdesmarais (approved):

PR #1374, opened 2024-07-23
Cleanup - remove BTAnalyticsService singleton pattern

scannillo:
### Summary of changes

- Remove [this weird singleton instance logic](https://github.com/braintree/braintree_ios/blob/94e22368d4aea7f1dd79f2c74177eff17afe5e21/Sources/BraintreeCore/BTAPIClient.swift#L27-L35) in BTAPIClient
    - This was a remnant from our Objective-C days. It's original intent was to make sure we persisted the pending queue/list of events, even if a new BTAPIClient instance was instantiated. This is now taken care of by using a static array of events (see `BTAnalyticsEventsStorage` in this PR https://github.com/braintree/braintree_ios/pull/1295)
- Remove cyclic dependency b/w BTAPIClient <--> BTAnalyticsService
- Move to dependency injection of `BTConfiguration` into `BTAnalyticsService` vs having `BTAnalyticsService` make another fetch (or reference to ConfigurationLoader)
    - Rich, Jax, and I discussed this internally. This is OK because each APIClient instance is associated with 1 auth credential. The auth credential is the unique key to our ConfigurationCache.

Note: I was trying to fix the bug in DTMOBILES-882. This PR doesn't fix that bug, but cleans things up. We will need an additional fix to address that issue.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
+++ b/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
@@ -13,6 +13,9 @@ class BTAnalyticsService {
     var http: BTHTTP
 
     /// Exposed for testing only
+    var configurationLoader: ConfigurationLoader
+    
+    /// Exposed for testing only
     var shouldBypassTimerQueue = false
 
     // MARK: - Private Properties
@@ -25,7 +28,6 @@ class BTAnalyticsService {
     private static let timer = RepeatingTimer(timeInterval: timeInterval)
     
     private let authorization: ClientAuthorization
-    private let configurationLoader: ConfigurationLoader
     private let metadata: BTClientMetadata
     
     // MARK: - Initializer
--- a/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
+++ b/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
@@ -11,9 +11,6 @@ class BTAnalyticsService {
 
     /// The HTTP client for communication with the analytics service endpoint. Exposed for testing.
     var http: BTHTTP
-
-    /// Exposed for testing only
-    var configurationLoader: ConfigurationLoader
     
     /// Exposed for testing only
     var shouldBypassTimerQueue = false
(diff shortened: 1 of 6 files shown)


Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -131,38 +135,23 @@ class BTAnalyticsService: Equatable {
...
-                    events: Self.events.allValues
-                )
-                http?.post("v1/tracking/batch/events", parameters: postParameters) { _, _, _ in }
-                await Self.events.removeAll()
-            } catch {
-                return
-            }
+            let postParameters = await createAnalyticsEvent(sessionID: metadata.sessionID, events: Self.events.allValues)
+            http.post("v1/tracking/batch/events", parameters: postParameters) { _, _, _ in }
+            await Self.events.removeAll()
```
agedd, line 140: love how clean and readable this is! 👏

agedd (approved):
lgtm!! thanks for cleaning this up 🚀

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -63,14 +56,19 @@ import Foundation
...
         super.init()
-        BTAPIClient._analyticsService = BTAnalyticsService(apiClient: self)
         http?.networkTimingDelegate = self
 
         // Kickoff the background request to fetch the config
-        fetchOrReturnRemoteConfiguration { _, _ in
-            // No-op
+        fetchOrReturnRemoteConfiguration { configuration, _ in
+            if let configuration {
+                self.analyticsService = BTAnalyticsService(
```
KunJeongPark, line 66: Even with this, if we are moving toward injection of configuration into analytics, we have no guarantee that configuration is fetched before first call to sendAnalytics in tokenize function, for example. (batch sending mitigates it but I think there might still be intermittent issues)

KunJeongPark: We could do what Android does, persist the analytic events in local storage, don't know the detail implementation but I remember seeing that.

scannillo: It's a great callout - I'm probably going to close this PR in favor of something else

scannillo:
We decided to stick with the singleton pattern for now since it does resolve several of our bugs. Replacing with https://github.com/braintree/braintree_ios/pull/1379

PR #1375, opened 2024-07-23
Update 3DS and GraphQL Endpoint Sent to FPTI

jaxdesmarais:
### Summary of changes

- 3D Secure: remove token/unique ID from path
- GraphQL: remove reliance on `operationName` since this is only present for card, causing us not to send the correct path for the Venmo flow
- Both endpoints sending as expected have been verified in FPTI and via breakpoints, additionally all endpoint have been checked to ensure we are sending the expected paths now

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -403,7 +403,21 @@ import Foundation
     func fetchAPITiming(path: String, connectionStartTime: Int?, requestStartTime: Int?, startTime: Int, endTime: Int) {
         let cleanedPath = path.replacingOccurrences(of: "/merchants/([A-Za-z0-9]+)/client_api", with: "", options: .regularExpression)
 
-        if cleanedPath != "/v1/tracking/batch/events" {
+        if cleanedPath.contains("three_d_secure") {
+            let threeDSecurePath = cleanedPath.replacingOccurrences(
+                of: "payment_methods/.*/three_d_secure",
+                with: "payment_methods/three_d_secure",
+                options: .regularExpression
+            )
+            analyticsService?.sendAnalyticsEvent(
+                BTCoreAnalytics.apiRequestLatency,
+                connectionStartTime: connectionStartTime,
+                endpoint: threeDSecurePath,
+                endTime: endTime,
+                requestStartTime: requestStartTime,
+                startTime: startTime
+            )
+        } else if cleanedPath != "/v1/tracking/batch/events" {
             analyticsService?.sendAnalyticsEvent(
                 BTCoreAnalytics.apiRequestLatency,
                 connectionStartTime: connectionStartTime,
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -401,23 +401,14 @@ import Foundation
     // MARK: BTAPITimingDelegate conformance
 
     func fetchAPITiming(path: String, connectionStartTime: Int?, requestStartTime: Int?, startTime: Int, endTime: Int) {
-        let cleanedPath = path.replacingOccurrences(of: "/merchants/([A-Za-z0-9]+)/client_api", with: "", options: .regularExpression)
+        var cleanedPath = path.replacingOccurrences(of: "/merchants/([A-Za-z0-9]+)/client_api", with: "", options: .regularExpression)
+        cleanedPath = cleanedPath.replacingOccurrences(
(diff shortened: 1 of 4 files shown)


Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -446,8 +446,12 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
         let json = try? JSONSerialization.jsonObject(with: data)
         let body = BTJSON(value: json)
         
-        guard let mutationName = body["operationName"].asString() else { return nil }
```
jaxdesmarais, line 449: The card GraphQL request is the only flow that uses `operationName` so for other flows such as Venmo we were sending the endpoint as `/graphql`. This update allows us to now send the expected query/mutation name.

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -446,8 +446,12 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
-        
-        return "mutation \(mutationName)"
+        guard let query = body["query"].asString() else {
+            return nil
+        }
+
+        let queryDiscardHolder = query.replacingOccurrences(of: #"^[^\(]*"#, with: "", options: .regularExpression)
```
jaxdesmarais, line 453: Open to alternate names for this, essentially this is to hold the parts of the query we want to remove. For example:
```swift
mutation CreateVenmoPaymentContext($input: CreateVenmoPaymentContextInput!) { createVenmoPaymentContext(input: $input) { venmoPaymentContext { id } } }
// now passed as
mutation CreateVenmoPaymentContext
```
Took some messing around with regex in a playground with multiple queries to get the correct syntax

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -403,7 +403,21 @@ import Foundation
     func fetchAPITiming(path: String, connectionStartTime: Int?, requestStartTime: Int?, startTime: Int, endTime: Int) {
         let cleanedPath = path.replacingOccurrences(of: "/merchants/([A-Za-z0-9]+)/client_api", with: "", options: .regularExpression)
 
-        if cleanedPath != "/v1/tracking/batch/events" {
+        if cleanedPath.contains("three_d_secure") {
```
jaxdesmarais, line 406: This logic feels a bit clunky but gets us the expected results, open to alternates but seemed like the cleanest way to achieve what we needed

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -403,7 +403,21 @@ import Foundation
...
+            let threeDSecurePath = cleanedPath.replacingOccurrences(
+                of: "payment_methods/.*/three_d_secure",
+                with: "payment_methods/three_d_secure",
+                options: .regularExpression
+            )
```
scannillo, line 411: Do we need the `if cleanedPath.contains("three_d_secure")` check at all? I would assume that the regex on lines 407-411 would be fine to run against _all_ paths, and it would only modify the ones that matched. This way we only need to call `analyticsService?.sendAnalyticsEvent` in 1 place

jaxdesmarais: 🤦‍♂️ you're totally right, updated here: https://github.com/braintree/braintree_ios/pull/1375/commits/00fade197d79f4004aafd7f845a6a5c9bd2a3354

Sources/BraintreeCore/BTHTTP.swift
agedd: just curious - would this change affect us being able to fetch the mutation name for the card graphql flow 👀

Sources/BraintreeCore/BTHTTP.swift
agedd: take it or leave it - do we wanna also update the function name to `getGraphQLQueryName` to reflect this change

Sources/BraintreeCore/BTHTTP.swift
jaxdesmarais: We still pass in a mutation as expected for the card flow, we just extract it from here: https://github.com/braintree/braintree_ios/blob/7b81129478fb3302b2ffacdbb4ade80e561faef7/Sources/BraintreeCard/BTCard.swift#L260 vs concatenating the mutation string to the `operationName`. So card is passed as `mutation TokenizeCreditCard` same as prior to this change.

richherrera (approved):

PR #1381, opened 2024-08-05
Cleanup App Switch

jaxdesmarais:
### Summary of changes

- Move `LinkType` to enum
- Remove logic to pass `linkType` into Approval URL parser
- Move app installed check into Vault Request

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeCore/Analytics/LinkType.swift
+++ b/Sources/BraintreeCore/Analytics/LinkType.swift
@@ -1,5 +1,6 @@
 /// Used to describe the link type for analytics
 /// :nodoc: This class is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+@_documentation(visibility: private)
 public enum LinkType: String {
     case universal
     case deeplink
--- a/Sources/BraintreeCore/Analytics/LinkType.swift
+++ b/Sources/BraintreeCore/Analytics/LinkType.swift
@@ -1,0 +1,6 @@
+/// Used to describe the link type for analytics
+/// :nodoc: This class is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+public enum LinkType: String {
+    case universal = "universal"
+    case deeplink = "deeplink"
+}
--- a/Sources/BraintreeCore/Analytics/LinkType.swift
+++ b/Sources/BraintreeCore/Analytics/LinkType.swift
@@ -1,6 +1,6 @@
 /// Used to describe the link type for analytics
 /// :nodoc: This class is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
 public enum LinkType: String {
-    case universal = "universal"
-    case deeplink = "deeplink"
+    case universal
+    case deeplink
 }
--- a/Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
+++ b/Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
@@ -45,9 +45,8 @@ struct BTPayPalApprovalURLParser {
         return nil
     }
 
-    // TODO: do we need this
-    init?(body: BTJSON, linkType: LinkType?) {
-        if linkType == .universal, let payPalAppRedirectURL = body["agreementSetup"]["paypalAppApprovalUrl"].asURL() {
+    init?(body: BTJSON) {
+        if let payPalAppRedirectURL = body["agreementSetup"]["paypalAppApprovalUrl"].asURL() {
             redirectType = .payPalApp(url: payPalAppRedirectURL)
             url = payPalAppRedirectURL
         } else if let approvalURL = body["paymentResource"]["redirectUrl"].asURL() ??
--- a/Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
+++ b/Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
@@ -45,8 +45,9 @@ struct BTPayPalApprovalURLParser {
         return nil
     }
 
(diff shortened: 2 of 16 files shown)


Sources/BraintreeCore/Analytics/LinkType.swift
```diff
@@ -0,0 +1,6 @@
+/// Used to describe the link type for analytics
+/// :nodoc: This class is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+public enum LinkType: String {
```
scannillo, line 3:
```suggestion
@_documentation(visibility: private)
public enum LinkType: String {
```

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
```diff
@@ -19,6 +19,9 @@ import BraintreeCore
     /// - Warning: This property is currently in beta and may change or be removed in future releases.
     var enablePayPalAppSwitch: Bool = false
 
+    /// exposed for mocking in tests
+    var application: URLOpener = UIApplication.shared
```
scannillo, line 23: IMO it's a little strange for VaultRequest to do the app installed inspection. It adds logic into our "model" type

jaxdesmarais: Agree, but it also felt odd that we were overriding a property in the client to ensure we don't pass the app switch params if the app is not installed here: https://github.com/braintree/braintree_ios/blob/6a375f2b7733e685e31cb0f1cd4d8909208ba4fb/Sources/BraintreePayPal/BTPayPalClient.swift#L336-L338

Do you know where else this logic could live that would make more sense? We could potentially pass it as a property when we call `parameters()` similar to `universalLink`. Curious to hear your thoughts.

scannillo: IMO the way it exists now is the highest separation of concerns (business logic in PayPalClient, model only in VaultRequest)

Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
```diff
@@ -45,8 +45,8 @@ struct BTPayPalApprovalURLParser {
         return nil
     }
 
-    init?(body: BTJSON, linkType: String?) {
```
richherrera, line 48: 👍

richherrera (approved):
LGTM!!

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
scannillo: I understand the concern now - currently we're changing the merchant-facing `VaultRequest. enablePayPalAppSwitch` setting without them knowing, which could be confusing. Jax I like the idea of injecting into the `parameters()` method more than having a reference to `UIApplication` in the `VaultRequest`!

jaxdesmarais: Yeah exactly - sorry I realize I should have done a better job of explaining the intent of the change initially. Pushed the change here: https://github.com/braintree/braintree_ios/pull/1381/commits/893d7363221beaa8711f453415fd6249d22d43bf

Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift
```diff
@@ -45,8 +45,8 @@ struct BTPayPalApprovalURLParser {
...
-        if linkType == "universal", let payPalAppRedirectURL = body["agreementSetup"]["paypalAppApprovalUrl"].asURL() {
+    init?(body: BTJSON) {
+        if let payPalAppRedirectURL = body["agreementSetup"]["paypalAppApprovalUrl"].asURL() {
```
KunJeongPark, line 49: For EditFI work, we are not using payPalAppApprovalUrl because app switching is not in scope yet for that work stream. But if the gateway returns this value as well as approvalURL, using this Parser we would get payPalAppRedirectURL instead of the webURL. For right now, I just forced the web Url fetching  by passing in linkType.

KunJeongPark: The specs on Confluence show: Response - agreementSetup {tokenId, paypalAppApprovalUrl, approvalUrl}
as response for edit FI url endpoint. Maybe the doc is not updated.

jaxdesmarais: Okay, then removing the link type will not impact that flow and we should not need to pass in a link type, it will always fail this first check for a `payPalAppApprovalUrl` and move on to the next block which includes the web URL.

KunJeongPark: I've posted the question on the collaboration channel so we can clarify with them.

jaxdesmarais: Fwiw we also have to opt in to the app switch flow for the Gateway to return that property, so I would expect it only exists to future proof this feature. We need to pass them everything here to say the merchant opted into app switch which we won't be doing for Edit FI currently: https://github.com/braintree/braintree_ios/blob/38297e4a697a1b89a1addab9c42c7af8553b8a57/Sources/BraintreePayPal/BTPayPalVaultRequest.swift#L56-L64

The gateway only returns a `payPalAppApprovalUrl` if those properties are passed AND the merchant is eligible on the gateway side.

KunJeongPark: Great! thank you for the clarification. I do want to make sure that this is true for editFI endpoint as well since there are no details given in the specs. I want to clarify that this will not be returned for generate_edit_fi_url endpoint if these parameters are not passed.

KunJeongPark (approved):

PR #1383, opened 2024-08-05
Add Optional merchantAccountID to PayPalVaultEditRequest

KunJeongPark:
### Summary of changes

- Add optional merchantAccountID and docstrings to PayPalVaultEditRequest

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
+++ b/Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
@@ -2,7 +2,7 @@ import Foundation
 
 /// Options for the PayPal edit funding instrument flow
 /// - Warning: This feature is currently in beta and may change or be removed in future releases.
-public class BTPayPalVaultEditRequest {
+public struct BTPayPalVaultEditRequest {
 
     private let editPayPalVaultID: String
     private let merchantAccountID: String?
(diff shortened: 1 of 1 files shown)


tdchow (approved):

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
```diff
@@ -5,13 +5,16 @@ import Foundation
 public class BTPayPalVaultEditRequest {
 
     private let editPayPalVaultID: String
+    public var merchantAccountID: String?
```
jaxdesmarais, line 8: We can make this private

sshropshire, line 8:
```suggestion
    private let merchantAccountID: String?
```
I agree with Jax. Also any reason we can't make this a `data class`?

sshropshire: Oh wow yeah I'm getting confused I thought this was Kotlin lol. Sorry. I forgot that Objective-C and Swift structs don't mix well. Plz disregard.

KunJeongPark: No worries. You do bring up a good point about use of classes for requests.

scannillo: I think this could be a struct. For new features, we're trying to experiment w/ seeing if we get merchant feedback by not offering Objective-C support (ex: ShopperInsights & CreditMessaging don't support Obj-C). If we need to switch it to class later for ObjC bridging we can, but I like the idea of using a struct!

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
```diff
@@ -5,13 +5,16 @@ import Foundation
 public class BTPayPalVaultEditRequest {
 
     private let editPayPalVaultID: String
+    private let merchantAccountID: String?
 
     //   TODO: specify endpoint for merchant to retrieve the token
     /// Initializes a PayPal Edit Request for the edit funding instrument flow
     /// - Parameters:
     ///   - editPayPalVaultID: Required: The `edit_paypal_vault_id` returned from the server side request
+    ///   merchantAccountIdD: optional ID of the merchant account; if one is not provided the default will be used
```
jaxdesmarais, line 14:
```suggestion
    ///   merchantAccountID: optional ID of the merchant account; if one is not provided the default will be used
```

Issue #1382, opened 2024-08-05
Crash in ConfigurationCallbackStorage After Upgrading to 6.23.2


sadaf-behbahani:
### Braintree SDK Version

6.23.2

### Environment

Both

### Xcode Version

15.4

### OS Version & Device

iOS 17.*

### Integration type

Swfit Package Manager

### Development Processor

Apple Silicon (M-series chips)

### Describe the bug

We encountered a critical issue after upgrading Braintree from version `6.18.2` to `6.23.2`. The crash occurs in `ConfigurationCallbackStorage` at `line 16` when attempting to add a completion to the queue. Below is a detailed description of the problem:

1. Create a `BraintreeClient`.
2. During the initialization of `BraintreeClient`, we instantiate a `BTAPIClient`.
3. Inside the `BTAPIClient` initialization, the method `fetchOrReturnRemoteConfiguration` is called without a completion handler, kicking off the background refresh and adding a completion to the pending completions. In our case, the internet is slow, and we don't get the configuration right away.
4. We proceed to create a `BTCardClient` from the `BTAPIClient` instance.
5. We then create a `BTApplePayClient` and start the payment process, which will call `tokenize`. This again calls `fetchOrReturnRemoteConfiguration`, adding a second completion to the pending completions.
6. After this point, the configuration network call succeeds.
7. `ConfigurationCallbackStorage` will sync/lock the queue and start invoking the completion blocks:
   - The first block works because it's a No-op.
   - The second block calls `BTApplePayClient.tokenize`, where the crash occurs. `Tokenize` calls `self.apiClient.post`, and inside the `post` function, it calls `fetchOrReturnRemoteConfiguration`.

_Just a small reminder to ensure that a strong self is not captured inside of a block to avoid potential retain cycles and other issues._

### To reproduce

Check the explanation above

### Expected behavior

No crash should occur, and the completion handlers should be managed correctly.

dane-thomas-vs:
I have a project that is also experiencing this crash. We upgraded from `6.21.0` to `6.23.1` to resolve a different crash and now this one has started to occur. For us, it occurs when using the `BTApplePayClient.tokenize` method. Is there a safe version to downgrade to while this gets investigated? Please advise.

scannillo:
[...]

PR #1385, opened 2024-08-06
Fix `ConfigurationCallbackStorage` Crash

scannillo:
### Summary of changes

- Fixes https://github.com/braintree/braintree_ios/issues/1382
- Remove `ConfigurationCallbackStorage`
- Add `async/throws` `get` method
- Update `getConfig()` to use `async/throws` - use task block
- Update tests to use `async/await` methods

### To Test
- The configuration is being cached and we use the cache if available - verify via FPTI
- When triggering a memory warning a crash is not present
- There are no crashes when instantiating multiple instances at the same time

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo @jaxdesmarais @KunJeongPark


Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
```diff
@@ -35,56 +36,33 @@ class ConfigurationLoader {
...
-                
-                if let error {
-                    notifyCompletions(nil, error)
-                    return
-                } else if response?.statusCode != 200 || body == nil {
-                    notifyCompletions(nil, BTAPIClientError.configurationUnavailable)
-                    return
+     
+        let task = Task {
+            do {
```
KunJeongPark, line 49: do we need this do catch block? It seems to be catching errors that are already thrown.

KunJeongPark, line 49: Do we need do-catch block?

Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
```diff
@@ -37,16 +37,16 @@ class ConfigurationLoader {
...
-        if let existingTask {
-            return try await existingTask.value
-        }
-
         if let cachedConfig = try? configurationCache.getFromCache(authorization: http.authorization.bearer) {
             return cachedConfig
         }
+        
+        if let existingTask {
+            return try await existingTask.value
```
KunJeongPark, line 45: I had put existingTask check first because I wanted to avoid cache simultaneous read/write scenario, however rare.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -93,19 +93,16 @@ import Foundation
...
     @_documentation(visibility: private)
     public func fetchOrReturnRemoteConfiguration(_ completion: @escaping (BTConfiguration?, Error?) -> Void) {
-        configurationLoader.getConfig { [weak self] configuration, error in
-            guard let self else {
-                completion(nil, BTAPIClientError.deallocated)
-                return
-            }
-            
-            if let error {
+        // TODO: - Consider updating all feature clients to use async version of this method?
```
jaxdesmarais, line 96: Created DTMOBILES-980 for this TODO. Want to focus on tackling the crash first, then we can follow up with this change

UnitTests/BraintreeCoreTests/Configuration/ConfigurationLoader_Tests.swift
```diff
@@ -132,11 +123,16 @@ class ConfigurationLoader_Tests: XCTestCase {
...
-        sut.getConfig() { _, _ in }
-        sut.getConfig() { _, _ in }
-        sut.getConfig() { _, _ in }
+    func testGetConfig_whenCalledInQuickSequence_onlySendsOneNetworkRequest() async {
+        do {
+            async let functionOne = sut.getConfig()
+            async let two = sut.getConfig()
+            async let three = sut.getConfig()
+            async let four = sut.getConfig()
+            let _ = try await (functionOne, two, three, four)
```
jaxdesmarais, line 132: @KunJeongPark are you able to walk me through how this would be triggered by a merchant? The `getConfig()` method is internal to the SDK so I am trying to grock how we would end up with this situation.

KunJeongPark: Sure, they wouldn't call this on its own but we are testing this function but itself on this test.
In our SDK we are calling getConfig from Task annotated @MainActor so it ensures that getConfig calls are serialized. But we have a test here that tests that there are no duplicate calls to this function.

So making the test this way instead of awaiting each function call that we had in earlier commit, is really testing if concurrent calls or quick successive calls would not be problematic and makes just one network call.

jaxdesmarais: Gotcha - so really just an update to protect ourself from accidentally doing something weird in the future and causing issues. Thanks for the explanation.

richherrera (approved):
Amazing!!!! 🚀 👏

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -93,23 +93,20 @@ import Foundation
...
+
+        Task { @MainActor in
```
KunJeongPark, line 98: I wonder if we remove this and just run the setupHTTPCredentials and completions on MainActor, we wouldn't need @MainActor annotation on fetchConfiguration. I am concerned about potential latency issues of sendingAnalytics on main.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -93,23 +93,20 @@ import Foundation
...
-                return
             }
-            
-            setupHTTPCredentials(configuration)
-            completion(configuration, nil)
         }
     }
     
-    func fetchConfiguration() async throws -> BTConfiguration {
+    @MainActor func fetchConfiguration() async throws -> BTConfiguration {
```
KunJeongPark, line 109: im a still thinking about this. that it might block main thread. 
Can we try to move @MainActor from the top Task block and just run the completion and setupHTTPCredentials on main?

Sources/BraintreeCore/BTAPIClient.swift
KunJeongPark, line 98: Just making sure that these don't cause deadlocks with nested calls to fetchOrReturnConfig.

jaxdesmarais: `@MainActor` is equivalent to running on the main dispatch queue: https://developer.apple.com/documentation/swift/mainactor. So should behave the exact same.

KunJeongPark: I think main difference for our purposes is that @MainActor can have suspension points so we wouldn't have deadlocks with nested completion handlers with fetchOrReturnRemoteConfiguration

KunJeongPark: I'm just puzzled why there are differences in analytics output with this[ commit](20d7a46d9381de1d6e01fd601c52f14fa668dbf5).

KunJeongPark: I guess we are just doing same on two call sites, so coordinating the two functions calls.

KunJeongPark (approved):

Sources/BraintreeCore/BTAPIClient.swift
jaxdesmarais: Yep exactly, we just want them to both behave the same, will also make it easier to move to the async method when we tackle that ticket.

jaxdesmarais:
Ran through another set of manual tests on sand and prod while triggering memory warnings and all looks good from my side. Let me know once you complete going through the flows as well @KunJeongPark and I'll merge this in.

Sources/BraintreeCore/BTAPIClient.swift
KunJeongPark: the ConfigurationLoader's getConfig method is annotated with globalActor, so I think they should be serialized.

KunJeongPark:
Roger! running tests and will let you know.

KunJeongPark:
I do want to make a note here for the future, I want to keep an eye on whether @MainActor annotation on `fetchConfiguration` async function and  entire Task block in `fetchOrReturnRemoteConfiguration` function are necessary or potentially cause slight latency issues. 

`Configuration.getConfiguration()` is already marked with @globalActor and it should be enough to serialize these calls.
In the Task block of` fetchOrReturnRemoteConfiguration`, it would be sufficient to use `MainActor.run` for blocks that include completion handlers and `setupHTTPCredentials` function. This allows UI operations and class property assignments to be thread safe while moving network calls to background thread.

Sources/BraintreeCore/BTAPIClient.swift
scannillo: Will this be a blocking action on the main thread? If so, that would interrupt a merchant doing UI updates

scannillo: cc: @jaxdesmarais @KunJeongPark

Sources/BraintreeCore/BTAPIClient.swift
scannillo, line 109: Similar ❓ - why did we need these MainActor attributes?

Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
```diff
@@ -35,56 +40,38 @@ class ConfigurationLoader {
     ///   - `BTConfiguration?`: The configuration object if it is successfully fetched or retrieved from the cache.
     ///   - `Error?`: An error object if fetching the configuration fails or if the instance is deallocated.
     @_documentation(visibility: private)
-    func getConfig(completion: @escaping (BTConfiguration?, Error?) -> Void) {
+    @ConfigurationActor
+    func getConfig() async throws -> BTConfiguration {
+        if let existingTask {
```
scannillo, line 45: Shouldn't we check the cache before bothering checking if there are any pending tasks?

@jaxdesmarais @KunJeongPark

KunJeongPark: Jax had the same question. I was thinking of scenario where there may be read/write cache race condition if there is a pending task.

Sources/BraintreeCore/BTAPIClient.swift
KunJeongPark: I don't think it's blocking, because MainActor allows suspension points and Configuration.getConfiguration runs on its own actor, actual network call is not performed on main thread.

I think there might be unnecessary thread hopping.

scannillo: MainActor should be reserved for UI tasks only ([doc](https://www.avanderlee.com/swift/mainactor-dispatch-main-thread/)). Why was this needed at all?

Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
```diff
@@ -35,56 +40,38 @@ class ConfigurationLoader {
...
-                }
-                
-                if let error {
-                    notifyCompletions(nil, error)
-                    return
-                } else if response?.statusCode != 200 || body == nil {
-                    notifyCompletions(nil, BTAPIClientError.configurationUnavailable)
-                    return
+     
+        let task = Task { [weak self] in
```
scannillo, line 53: Can we just do `existingTask = Task { [weak self] in ...` instead? 

Line 75 is a little confusing why we're doing the await on the `task` constant and not the re-assigned `existingTask` variable.

Sources/BraintreeCore/BTAPIClient.swift
jaxdesmarais: @scannillo from that doc: `Dispatching to the main actor only takes place if needed to ensure the best performance possible.`. It is worth noting this has always taken place on the main thread, see the following test: https://github.com/braintree/braintree_ios/blob/c17642dad8f8a45f1b2d51c0c762873d35e17f6e/UnitTests/BraintreeCoreTests/BTAPIClient_Tests.swift#L264-L268

We may be able to clean this up slightly, but this is not new behavior to have `fetchOrReturnRemoteConfiguration` run on main.

jaxdesmarais: Digging further into this and the `@MainActor` for the async `fetchConfiguration` we were not properly caching the config and pulling from the cached config if not using `@MainActor`, additionally removing this causes the test `testGetConfig_whenCalledInQuickSequence_onlySendsOneNetworkRequest` to fail as we are making 4 network request. We expect to be doing these calls off of the main thread (again the main thread is an expectation in the test above). I am going to leave these as is to ensure our current behavior based on our tests and my observations from FPTI.

KunJeongPark: That test` testGetConfig_whenCalledInQuickSequence_onlySendsOneNetworkRequest` just runs on the function on ConfigurationLoader `getConfig`. I'm not sure how removing MainActor annotation on the function that calls it, fetchConfiguration affects that test. I've shaven't seen that test fail on its own.

KunJeongPark: That test you quoted above,` testCallbacks_useMainDispatchQueue()`, it expects the completion handler to return on main.
So entire Task block of  fetchOrReturnRemoteConfiguration doesn't need to be annoytated @MainActor,
just the setupHTTPCredentials and completion invocation need to be wrapped in await MainActor.run to make this
test pass.

KunJeongPark: I haven't personally seen weirdness or difference in the FPTI between what was on main before, before this crash test was implemented, after adding it, with or without MainActor annotations for fetchConfig functions.

But I trust that you saw something, there is a lot going on with sending analytics. 
Let's keep an eye on latency metrics and then we can revisit. What are your thoughts on that?

jaxdesmarais: > That test you quoted above, testCallbacks_useMainDispatchQueue(), it expects the completion handler to return on main.
So entire Task block of fetchOrReturnRemoteConfiguration doesn't need to be annotated @mainactor,
just the setupHTTPCredentials and completion invocation need to be wrapped in await MainActor.run to make this test pass, as I mentioned at top of this thread.

imo, adding `await MainActor.run` for just the completions/setupHTTPCredentials complicates the callsite vs running the task on `MainActor`. There is no observed latency since `getConfig` is a global isolated actor already. Having a cleaner callsite is more important as there is no difference in wrapping the completions in a `MainActor.run` vs what exists today, but this method will also entirely be removed in a future PR.

> I haven't personally seen weirdness or difference in the FPTI between what was on main before, before this crash fix was implemented, after adding it, with or without MainActor annotations for fetchConfig functions.

If  you look at ticket DTMOBILES-998 you will see this same behavior where we make multiple network called and the config isn't cached properly. There are screenshots in that ticket with additional details from FPTI.

> My main concerns are potential for more latency for fetching configuration for sendQueuedAnalyticsEvents with @mainactor annotation. But it's hard to tell. Let's keep an eye on latency metrics and then we can revisit.

There is clearly a need to annotate as is, removing it causes the issues we can clearly see in FPTI noted above. Removing the duplicate config call and just relying on the async method should resolve any remaining confusion. There is not additional latency added that did not exist previously.

KunJeongPark: Thank you for the screenshots. Do these extra calls happen in 6.23.2? I can check.
I believe you that you observed these, I just didn't see it myself. I am trying to figure out why.
I want to make sure I understand what @MainActor annotation is fixing.

KunJeongPark: I made comments on your jira ticket with screenshots. I tested 6.23.3 with @mainactor annotation as it is currently on main wit no changes, 6.23.3 without @mainactor annotation but with await MainActor.run on setHTTPCredentials and completion calls and 6.23.2 with no changes.

I didn't incorporate your changes on this branch, I am going to try that as well.

I found that there are no duplicate calls to v1/configuration in any of them, they are all fetching from cache but 6.23.3 without @mainactor annotation results in analytic events being listed out of order, this was also true for 6.23.2, the analytics were out of order.
I just ran through each once, so I will try several times for verification.

PR #1386, opened 2024-08-07
Update BTPayPalVaultEditResult Object

KunJeongPark:
### Summary of changes

- Add properties to BTPayPalVaultEdit Result
- Change from class to struct

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalVaultEditResult.swift
+++ b/Sources/BraintreePayPal/BTPayPalVaultEditResult.swift
@@ -1,8 +1,33 @@
 import Foundation
+import BraintreeCore
 
 /// A result of the Edit FI flow used to display a customers updated payment details in your UI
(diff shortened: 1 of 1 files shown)


sshropshire (approved):

Sources/BraintreePayPal/BTPayPalVaultEditResult.swift
```diff
@@ -1,8 +1,33 @@
 import Foundation
+import BraintreeCore
```
jaxdesmarais, line 2: We need to use the `canImport` syntax here to resolve the cocoapods error

KunJeongPark: Oh, nice catch.

jaxdesmarais (approved):

PR #1388, opened 2024-08-12
Move payPalInstalled and venmoInstalled to batch_params

KunJeongPark:
### Summary of changes

- Move venmo_installed and paypal_installed FPTI metadata params to `batch_prams` from `event_params`

on device (less events here since I didn't go through actual Apple Pay on device, but both sets matched up number of events = these fields):
Screenshot 2024-08-13 at 12 13 40 PM

on simulator 
Screenshot 2024-08-13 at 12 06 40 PM

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark @jwarmkessel

### DIFF
--- a/Sources/BraintreeCore/Analytics/FPTIBatchData.swift
+++ b/Sources/BraintreeCore/Analytics/FPTIBatchData.swift
@@ -29,8 +29,6 @@ struct FPTIBatchData: Codable {
     /// Encapsulates a single event by it's name and timestamp.
     struct Event: Codable {
 
-        static var application: URLOpener = UIApplication.shared
-
         /// UTC millisecond timestamp when a networking task started establishing a TCP connection. See [Apple's docs](https://developer.apple.com/documentation/foundation/urlsessiontasktransactionmetrics#3162615).
         ///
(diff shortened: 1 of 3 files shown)


jaxdesmarais (commented):
Did we verify in FPTI that we are still seeing these events as expected?

scannillo:
I'm surprised no unit tests failed. Can we add these params to [this unit test](https://github.com/braintree/braintree_ios/blob/5cd32cab424b91a28aa99eb1999a5f049baa95da/UnitTests/BraintreeCoreTests/Analytics/FPTIBatchData_Tests.swift#L55)?

scannillo (approved):
Changes LGTM. Left comment re: adding unit test
Vic reply:  I thought about that but these values would always return false in simulator test, I think.
(sorry I thought I was replying to this instead of appending to your comment)

agedd (approved):
🚀

KunJeongPark:
> Did we verify in FPTI that we are still seeing these events as expected?

Yes, I did see them all returning false on simulator but I will test on device as well. (I have both apps installed on my device)

PR #1394, opened 2024-08-20
Configuration Cleanup

jaxdesmarais:
### Summary of changes

- Cleanup comments left post merge from Sammy in PR: https://github.com/braintree/braintree_ios/pull/1385

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
+++ b/Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
@@ -13,6 +13,9 @@ class ConfigurationLoader {
     private let configurationCache = ConfigurationCache.shared
     private let http: BTHTTP
 
+    /// Used to hold an in-flight task to fetch a configuration or return an error
+    private var existingTask: Task<BTConfiguration, Error>?
+
     // MARK: - Initializer
     
     init(http: BTHTTP) {
@@ -32,19 +35,18 @@ class ConfigurationLoader {
     /// 2. If no cached configuration is found, it fetches the configuration from the server and caches the successful response.
     /// 3. If fetching the configuration fails, it returns an error.
     ///
-    /// - Parameters:
-    ///   - completion: A completion handler that is called with the fetched or cached `BTConfiguration` object or an `Error`.
-    ///
-    /// - Completion:
-    ///   - `BTConfiguration?`: The configuration object if it is successfully fetched or retrieved from the cache.
-    ///   - `Error?`: An error object if fetching the configuration fails or if the instance is deallocated.
+    /// - Returns: A `BTConfiguration` if it is successfully fetched or retrieved from the cache.
+    /// - Throws: An `Error` describing the failure; if fetching the configuration fails or if the instance is deallocated.
     @_documentation(visibility: private)
     @ConfigurationActor
     func getConfig() async throws -> BTConfiguration {
-        var existingTask: Task<BTConfiguration, Error>
-
         if let cachedConfig = try? configurationCache.getFromCache(authorization: http.authorization.bearer) {
             return cachedConfig
+        }
+
+        /// if we are writing to the cache at this time, we can return the existing task
+        if let existingTask {
+            return try await existingTask.value
         }
 
         existingTask = Task { [weak self] in
@@ -65,6 +67,10 @@ class ConfigurationLoader {
             } catch {
                 throw error
             }
+        }
+
+        guard let existingTask else {
+            throw BTAPIClientError.configurationUnavailable
         }
 
         return try await existingTask.value
--- a/Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
+++ b/Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
@@ -13,8 +13,6 @@ class ConfigurationLoader {
     private let configurationCache = ConfigurationCache.shared
     private let http: BTHTTP
 
-    private var existingTask: Task<BTConfiguration, Error>?
-
     // MARK: - Initializer
     
     init(http: BTHTTP) {
@@ -43,21 +41,17 @@ class ConfigurationLoader {
     @_documentation(visibility: private)
     @ConfigurationActor
     func getConfig() async throws -> BTConfiguration {
+        var existingTask: Task<BTConfiguration, Error>
+
         if let cachedConfig = try? configurationCache.getFromCache(authorization: http.authorization.bearer) {
             return cachedConfig
         }
 
-        /// if we are writing to the cache at this time, we can return the existing task
-        if let existingTask {
-            return try await existingTask.value
-        }
-
-        let task = Task { [weak self] in
+        existingTask = Task { [weak self] in
             guard let self else {
                 throw BTAPIClientError.deallocated
             }
 
-            defer { self.existingTask = nil }
             do {
                 let (body, response) = try await http.get(configPath, parameters: BTConfigurationRequest())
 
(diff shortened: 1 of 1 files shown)


jaxdesmarais:
@KunJeongPark would love your thoughts on these changes to make sure I didn't miss anything! I went through several tests with triggering a memory warning, but would love a second set of eyes if you're able to pull down this branch and run through some tests. :teamwork:

Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
```diff
@@ -33,29 +35,28 @@ class ConfigurationLoader {
...
-    ///
-    /// - Completion:
-    ///   - `BTConfiguration?`: The configuration object if it is successfully fetched or retrieved from the cache.
-    ///   - `Error?`: An error object if fetching the configuration fails or if the instance is deallocated.
+    /// - Returns: A `BTConfiguration` if it is successfully fetched or retrieved from the cache.
+    /// - Throws: An `Error` describing the failure; if fetching the configuration fails or if the instance is deallocated.
     @_documentation(visibility: private)
     @ConfigurationActor
     func getConfig() async throws -> BTConfiguration {
+        if let cachedConfig = try? configurationCache.getFromCache(authorization: http.authorization.bearer) {
```
jaxdesmarais, line 43: I think we can move forward with checking the cache first, if a cache save is in flight we would use the existing task value below. Added a docstring for clarity for this.

Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
```diff
@@ -33,29 +35,28 @@ class ConfigurationLoader {
...
         if let existingTask {
             return try await existingTask.value
         }
 
-        if let cachedConfig = try? configurationCache.getFromCache(authorization: http.authorization.bearer) {
-            return cachedConfig
-        }
-     
-        let task = Task { [weak self] in
+        existingTask = Task { [weak self] in
```
jaxdesmarais, line 52: I think like Sammy suggested we do not need a new property here that we then set to the existing task. I went through several flows to ensure we weren't reintroducing any bugs but think we should be covered here. It will also reduce confusion from the previous PR where we were setting `existingTask = task` but then using `task.value`

KunJeongPark: I think it's same logic, I just thought it was easier to read that way that a task is being created there, maybe for debugging also, but I am not attached to it.

Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
KunJeongPark: I was concerned about chance of data race between reading from cache and writing to cache from the Task block.
If we are doing getFromCache while there is an existingTask, we run a small risk of simultaneous read/write to cache.

KunJeongPark:
I think Sammy also had concerns whether @MainActor annotations were necessary in two places.

jaxdesmarais:
> I think Sammy also had concerns whether @mainactor annotations were necessary in two places.

I addressed those in the original PR: https://github.com/braintree/braintree_ios/pull/1385#discussion_r1723330766

KunJeongPark:
I saw your comments in PR and commented on them: https://github.com/braintree/braintree_ios/pull/1385#discussion_r1726940870

Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
jaxdesmarais: There should not be a data race in that case, in that case the sequence of events would be:
1. We check the cache and nothing is there because we are actively writing to the cache
2. There _is_ and `existingTask` so we would return that on like 49

There wouldn't be an instance were we are reading and writing at the same time because `getFromCache` would be empty in this case while `existingTask` would not

KunJeongPark: I was thinking about the rare case where we are reading from cache and there is an existing task that is writing into cache while it's reading cache. Simultaneous read/write cache would result.

jaxdesmarais: `NSCache` is multi-threading friendly so that shouldn't be an issue. But also, if we were writing to the cache at the exact same moment as attempting to read from the cache the object would not yet exist in the `NSCache` so we would instead return the `existingTask.value`. The `NSCache` also only stores a singular object per key, which is the `authorizationFingerprint` or `tokenizationKey`.

KunJeongPark: I didn't know that about NSCache! Thanks.

KunJeongPark:
I made comments on your jira ticket. I tested 6.23.3 with @MainActor annotation, 6.23.3 without @MainActor annotation but with await MainActor.run on setHTTPCredentials and completion calls and 6.23.2. 
I didn't incorporate your changes on this branch, I am going to try that as well.

I found that there are no duplicate calls to v1/configuration in any of them, they are all fetching from cache but 6.23.3 without @MainActor annotation results in analytic events being listed out of order.

jaxdesmarais:
>  I tested 6.23.3 with @mainactor annotation as it is currently on main wit no changes, 6.23.3 without @mainactor annotation but with await MainActor.run on setHTTPCredentials and completion calls and 6.23.2 with no changes.

Both `Task { @MainActor in` and `MainActor.run` around the completion/`setupHTTPCredentials` do the exact same thing so these will return the same results. I don't think it's worth overcomplicating the callsite as this method will also soon be removed with DTMOBILES-980. Additionally since both behave the same we can keep the callsite simple for now until we move to only using the async config call.

> I found that there are no duplicate calls to v1/configuration in any of them, they are all fetching from cache

You'll need to ensure you're letting the cache clear, it's valid for 5 minutes, but first fetch will be `false` and subsequent will be `true`. In any case moving to a singular fetch config method will remove this as a grey area if you're unable to replicate due to setup issues.

KunJeongPark:
> > I tested 6.23.3 with @mainactor annotation as it is currently on main wit no changes, 6.23.3 without @mainactor annotation but with await MainActor.run on setHTTPCredentials and completion calls and 6.23.2 with no changes.
> 
> Both `Task { @MainActor in` and `MainActor.run` around the completion/`setupHTTPCredentials` do the exact same thing so these will return the same results. I don't think it's worth overcomplicating the callsite as this method will also soon be removed with DTMOBILES-980. Additionally since both behave the same we can keep the callsite simple for now until we move to only using the async config call.
 = I'm fine with it, but I don't think they're same and we should keep an eye on latency, although I don't think it will be noticeable. 
> > I found that there are no duplicate calls to v1/configuration in any of them, they are all fetching from cache
> 
> You'll need to ensure you're letting the cache clear, it's valid for 5 minutes, but first fetch will be `false` and subsequent will be `true`. In any case moving to a singular fetch config method will remove this as a grey area if you're unable to replicate due to setup issues.
  => I did try it, clearing cache and I got same result. I made note on ticket about low memory setting and NSCache. 
I'm good to ✅ this ticket after running one test. Thank you for the clean up!

As I mentioned in the ticket, I just wanted to make sure that there were no inherent problems elsewhere without the @MainActor annotation, if that annotation is masking something, if that makes sense. Not holding this one back.

KunJeongPark:
I saw a `v1/config` duplicate call from one test out of 3 on your branch, so I will do more testing and also test what's on main.

KunJeongPark:
Thank you for the work! 🎉

KunJeongPark (approved):

agedd (approved):
🚀

PR #1395, opened 2024-08-20
Edit FI Gateway Endpoints

KunJeongPark:
### Summary of changes

- implement Edit-FI client API's 
- UI for vault ID and riskCorrelationID text entry in demo app

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark @jwarmkessel

### DIFF
--- a/Demo/Application/Base/BaseViewController.swift
+++ b/Demo/Application/Base/BaseViewController.swift
@@ -5,7 +5,6 @@ class BaseViewController: UIViewController {
 
     var progressBlock: ((String?) -> Void) = { _ in }
     var completionBlock: ((BTPaymentMethodNonce?) -> Void) = { _ in }
-    var paypalEditCompletionBlock: ((String) -> Void) = { _ in  }
 
     init(authorization: String) {
         super.init(nibName: nil, bundle: nil)
--- a/Demo/Application/Base/BaseViewController.swift
+++ b/Demo/Application/Base/BaseViewController.swift
@@ -5,6 +5,7 @@ class BaseViewController: UIViewController {
 
     var progressBlock: ((String?) -> Void) = { _ in }
     var completionBlock: ((BTPaymentMethodNonce?) -> Void) = { _ in }
+    var paypalEditCompletionBlock: ((String) -> Void) = { _ in  }
 
     init(authorization: String) {
         super.init(nibName: nil, bundle: nil)
--- a/Demo/Application/Base/ContainmentViewController.swift
+++ b/Demo/Application/Base/ContainmentViewController.swift
@@ -18,7 +18,7 @@ class ContainmentViewController: UIViewController {
             updateStatus("Presenting \(type(of: currentViewController))")
             currentViewController.progressBlock = progressBlock
             currentViewController.completionBlock = completionBlock
-            
+
             appendViewController(currentViewController)
             title = currentViewController.title
         }
--- a/Demo/Application/Base/ContainmentViewController.swift
+++ b/Demo/Application/Base/ContainmentViewController.swift
@@ -42,10 +42,6 @@ class ContainmentViewController: UIViewController {
         updateStatus("Got a nonce. Tap to make a transaction.")
     }
 
-    func paypalEditCompletionBlock(_ correlationID: String) {
-        updateStatus("Edit FI completed.\n riskCorrelationID: \(correlationID)")
-    }
-
     override func viewDidLoad() {
         super.viewDidLoad()
 
--- a/Demo/Application/Base/ContainmentViewController.swift
+++ b/Demo/Application/Base/ContainmentViewController.swift
@@ -18,8 +18,7 @@ class ContainmentViewController: UIViewController {
             updateStatus("Presenting \(type(of: currentViewController))")
             currentViewController.progressBlock = progressBlock
             currentViewController.completionBlock = completionBlock
-            currentViewController.paypalEditCompletionBlock = paypalEditCompletionBlock
-
+            
             appendViewController(currentViewController)
             title = currentViewController.title
         }
--- a/Demo/Application/Base/ContainmentViewController.swift
+++ b/Demo/Application/Base/ContainmentViewController.swift
@@ -18,6 +18,7 @@ class ContainmentViewController: UIViewController {
             updateStatus("Presenting \(type(of: currentViewController))")
             currentViewController.progressBlock = progressBlock
             currentViewController.completionBlock = completionBlock
+            currentViewController.paypalEditCompletionBlock = paypalEditCompletionBlock
 
             appendViewController(currentViewController)
             title = currentViewController.title
@@ -40,6 +41,10 @@ class ContainmentViewController: UIViewController {
     func completionBlock(_ nonce: BTPaymentMethodNonce?) {
         currentPaymentMethodNonce = nonce
         updateStatus("Got a nonce. Tap to make a transaction.")
+    }
+
+    func paypalEditCompletionBlock(_ correlationID: String) {
+        updateStatus("Edit FI completed.\n riskCorrelationID: \(correlationID)")
     }
 
     override func viewDidLoad() {
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -49,13 +49,14 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
     override func createPaymentButton() -> UIView {
         let payPalCheckoutButton = createButton(title: "PayPal Checkout", action: #selector(tappedPayPalCheckout))
         let payPalVaultButton = createButton(title: "PayPal Vault", action: #selector(tappedPayPalVault))
+        let payPalEditVaultButton = createButton(title: "Edit PayPal Vault", action: #selector(tappedPayPalEditVault))
         let payPalAppSwitchButton = createButton(title: "PayPal App Switch", action: #selector(tappedPayPalAppSwitch))
         let oneTimeCheckoutStackView = buttonsStackView(label: "1-Time Checkout", views: [
             UIStackView(arrangedSubviews: [payLaterToggleLabel, payLaterToggle]),
             UIStackView(arrangedSubviews: [newPayPalCheckoutToggleLabel, newPayPalCheckoutToggle]),
             payPalCheckoutButton
         ])
-        let vaultStackView = buttonsStackView(label: "Vault",views: [payPalVaultButton, payPalAppSwitchButton])
+        let vaultStackView = buttonsStackView(label: "Vault",views: [payPalVaultButton, payPalAppSwitchButton, payPalEditVaultButton])
 
 
         let stackView = UIStackView(arrangedSubviews: [
@@ -159,7 +160,27 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
             self.completionBlock(nonce)
         }
     }
-    
+
+    // MARK: Edit FI flow
+
+    @objc func tappedPayPalEditVault(_ sender: UIButton) {
+        progressBlock("Tapped PayPal - Vault using BTPayPalClient")
+        sender.setTitle("Processing...", for: .disabled)
+        sender.isEnabled = false
+
+        let request = BTPayPalVaultEditRequest(editPayPalVaultID: "YJbRTegvI/dIDEyFZRa52Twflbn0q2pSktu1llbZmMg=")
+        payPalClient.edit(request) { nonce, error in
+            sender.isEnabled = true
+
+            guard let nonce else {
+                self.progressBlock(error?.localizedDescription)
+                return
+            }
+
+         //   self.completionBlock(nonce)
+        }
+    }
+
     // MARK: - Helpers
     
     private func buttonsStackView(label: String, views: [UIView]) -> UIStackView {
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -49,7 +49,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
     override func createPaymentButton() -> UIView {
         let payPalCheckoutButton = createButton(title: "PayPal Checkout", action: #selector(tappedPayPalCheckout))
         let payPalVaultButton = createButton(title: "PayPal Vault", action: #selector(tappedPayPalVault))
-        let payPalEditVaultButton = createButton(title: "Edit PayPal Vault", action: #selector(tappedPayPalEditVault))
+        let payPalEditVaultButton = createButton(title: "Edit FI", action: #selector(tappedPayPalEditVault))
         let payPalAppSwitchButton = createButton(title: "PayPal App Switch", action: #selector(tappedPayPalAppSwitch))
         let oneTimeCheckoutStackView = buttonsStackView(label: "1-Time Checkout", views: [
             UIStackView(arrangedSubviews: [payLaterToggleLabel, payLaterToggle]),
@@ -168,7 +168,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         sender.setTitle("Processing...", for: .disabled)
         sender.isEnabled = false
 
-        let request = BTPayPalVaultEditRequest(editPayPalVaultID: "YJbRTegvI/dIDEyFZRa52Twflbn0q2pSktu1llbZmMg=")
+        let request = BTPayPalVaultEditRequest(editPayPalVaultID: "+fZXfUn6nzR+M9661WGnCBfyPlIExIMPY2rS9AC2vmA=")
         payPalClient.edit(request) { editResult, error in
             sender.isEnabled = true
 
@@ -177,7 +177,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
                 return
             }
 
-         //   self.completionBlock(editResult)
+            self.paypalEditCompletionBlock(editResult.riskCorrelationID)
         }
     }
 
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -44,12 +44,14 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
     lazy var payPalVaultIDLabel: UILabel = {
         let label = UILabel()
         label.text = "PayPal Vault ID:"
+        label.font = .preferredFont(forTextStyle: .footnote)
         return label
     }()
 
     lazy var payPalVaultIDTextField: UITextField = {
         let textField = UITextField()
         textField.text = "+fZXfUn6nzR+M9661WGnCBfyPlIExIMPY2rS9AC2vmA="
+        textField.font = .preferredFont(forTextStyle: .footnote)
         textField.backgroundColor = .systemBackground
         return textField
     }()
@@ -57,6 +59,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
     lazy var riskCorrelationIDLabel: UILabel = {
         let label = UILabel()
         label.text = "Risk Correlation ID:"
+        label.font = .preferredFont(forTextStyle: .footnote)
         label.isHidden = true
         return label
     }()
@@ -64,6 +67,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
     lazy var riskCorrelationIDTextField: UITextField = {
         let textField = UITextField()
         textField.text = "test"
+        textField.font = .preferredFont(forTextStyle: .footnote)
         textField.backgroundColor = .systemBackground
         textField.isHidden = true
         return textField
@@ -72,6 +76,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
     lazy var errorHandlingToggleLabel: UILabel = {
         let label = UILabel()
         label.text = "Error Handling Flow"
+        label.font = .preferredFont(forTextStyle: .footnote)
         return label
     }()
 
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -227,7 +227,6 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         sender.setTitle("Processing...", for: .disabled)
         sender.isEnabled = false
 
-
         let vaultID = payPalVaultIDTextField.text ?? "+fZXfUn6nzR+M9661WGnCBfyPlIExIMPY2rS9AC2vmA="
         let request = BTPayPalVaultEditRequest(editPayPalVaultID: vaultID)
 
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -239,7 +239,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
                 return
             }
 
-            self.paypalEditCompletionBlock(editResult.riskCorrelationID)
+            self.progressBlock(("Edit FI completed.\n riskCorrelationID: \(editResult.riskCorrelationID)"))
         }
     }
 
@@ -260,7 +260,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
                 return
             }
 
-            self.paypalEditCompletionBlock(editResult.riskCorrelationID)
+            self.progressBlock(("Edit FI completed.\n riskCorrelationID: \(editResult.riskCorrelationID)"))
         }
     }
 
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -41,29 +41,65 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
     
     let newPayPalCheckoutToggle = UISwitch()
 
+    lazy var payPalVaultIDLabel: UILabel = {
+        let label = UILabel()
+        label.text = "PayPal Vault ID:"
+        return label
+    }()
+
+    lazy var payPalVaultIDTextField: UITextField = {
+        let textField = UITextField()
+        textField.text = "+fZXfUn6nzR+M9661WGnCBfyPlIExIMPY2rS9AC2vmA="
+        textField.backgroundColor = .systemBackground
+        return textField
+    }()
+
+    lazy var riskCorrelationIDLabel: UILabel = {
+        let label = UILabel()
+        label.text = "Risk Correlation ID:"
+        label.isHidden = true
+        return label
+    }()
+
+    lazy var riskCorrelationIDTextField: UITextField = {
+        let textField = UITextField()
+        textField.text = "test"
+        textField.backgroundColor = .systemBackground
+        textField.isHidden = true
+        return textField
+    }()
+
+    lazy var errorHandlingToggleLabel: UILabel = {
+        let label = UILabel()
+        label.text = "Error Handling Flow"
+        return label
+    }()
+
+    let errorHandlingToggle = UISwitch()
+
     override func viewDidLoad() {
-        super.heightConstraint = 300
+        super.heightConstraint = 600
         super.viewDidLoad()
+        errorHandlingToggle.addTarget(self, action: #selector(toggleErrorHandling(_:)), for: .valueChanged)
     }
 
     override func createPaymentButton() -> UIView {
         let payPalCheckoutButton = createButton(title: "PayPal Checkout", action: #selector(tappedPayPalCheckout))
         let payPalVaultButton = createButton(title: "PayPal Vault", action: #selector(tappedPayPalVault))
         let payPalEditVaultButton = createButton(title: "Edit FI", action: #selector(tappedPayPalEditVault))
-        let payPalEditVaultErrorHandlingButton = createButton(title: "Edit FI (Error Handling)", action: #selector(tappedPayPalEditVaultErrorHandling))
         let payPalAppSwitchButton = createButton(title: "PayPal App Switch", action: #selector(tappedPayPalAppSwitch))
         let oneTimeCheckoutStackView = buttonsStackView(label: "1-Time Checkout", views: [
             UIStackView(arrangedSubviews: [payLaterToggleLabel, payLaterToggle]),
             UIStackView(arrangedSubviews: [newPayPalCheckoutToggleLabel, newPayPalCheckoutToggle]),
             payPalCheckoutButton
         ])
-        let vaultStackView = buttonsStackView(label: "Vault",views: [payPalVaultButton, payPalAppSwitchButton, payPalEditVaultButton, payPalEditVaultErrorHandlingButton])
-
-
+        let vaultStackView = buttonsStackView(label: "Vault",views: [payPalVaultButton, payPalAppSwitchButton])
+        let editFIStackView = buttonsStackView(label: "Edit FI Flow", views: [payPalVaultIDLabel, payPalVaultIDTextField, UIStackView(arrangedSubviews: [errorHandlingToggleLabel, errorHandlingToggle]), riskCorrelationIDLabel, riskCorrelationIDTextField, payPalEditVaultButton])
         let stackView = UIStackView(arrangedSubviews: [
             UIStackView(arrangedSubviews: [emailLabel, emailTextField]),
             oneTimeCheckoutStackView,
-            vaultStackView
+            vaultStackView,
+            editFIStackView
         ])
 
         NSLayoutConstraint.activate([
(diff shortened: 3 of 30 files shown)


Sources/BraintreePayPal/BTPayPalError.swift
```diff
@@ -112,6 +117,8 @@ public enum BTPayPalError: Error, CustomNSError, LocalizedError, Equatable {
             return "Missing BA Token for PayPal App Switch."
         case .missingPayPalRequest:
             return "The PayPal Request was missing or invalid."
+        case .missingClientMetadataID:
+            return "Missing ClientMetadataID"
```
jaxdesmarais, line 121: Can we make this a bit more explicit of a message? Maybe something like:
```suggestion
            return "Unable to collect riskCorrelationID during the Edit FI or Error Handling flow."
```

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
```diff
@@ -1,20 +1,40 @@
...
+import BraintreeCore
+#endif
+
 /// Options for the PayPal edit funding instrument flow
 /// - Warning: This feature is currently in beta and may change or be removed in future releases.
-public struct BTPayPalVaultEditRequest {
+public class BTPayPalVaultEditRequest {
 
+    var riskCorrelationID: String?
+    let hermesPath: String = "v1/paypal_hermes/generate_edit_fi_url"
```
jaxdesmarais, line 12: Lets just have this live at the callsite since it's not necessary to live here for the edit flow

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
```diff
@@ -1,20 +1,40 @@
...
     private let editPayPalVaultID: String
-    private let merchantAccountID: String?
```
jaxdesmarais, line 8: Why is this being removed? How will merchants use their non default merchantAccountID?

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
```diff
@@ -1,20 +1,40 @@
...
+    public func parameters() -> [String: Any] {
+        var parameters: [String: Any] = [:]
+
+        parameters["edit_paypal_vault_id"] = editPayPalVaultID
+
+        parameters["return_url"] = BTCoreConstants.callbackURLScheme + "://\(BTPayPalRequest.callbackURLHostAndPath)success"
+        parameters["cancel_url"] = BTCoreConstants.callbackURLScheme + "://\(BTPayPalRequest.callbackURLHostAndPath)cancel"
+
+        parameters["risk_correlation_id"] = riskCorrelationID
+        return parameters
```
jaxdesmarais, line 38: Since we already know the values we can construct the dictionary like so:
```suggestion
        [
            "edit_paypal_vault_id": editPayPalVaultID,
            "return_url": BTCoreConstants.callbackURLScheme + "://\(BTPayPalRequest.callbackURLHostAndPath)success",
            "cancel_url": BTCoreConstants.callbackURLScheme + "://\(BTPayPalRequest.callbackURLHostAndPath)cancel",
            "risk_correlation_id": riskCorrelationID
        ]
```

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
```diff
@@ -1,20 +1,40 @@
...
     ///   `gateway.payment_method.find("payment_method_token")` or `gateway.customer.find("customer_id")`
     ///   - merchantAccountID: Optional: ID of the merchant account; if one is not provided the default will be used
     /// - Warning: This feature is currently in beta and may change or be removed in future releases.
-    public init(editPayPalVaultID: String, merchantAccountID: String? = nil) {
+    public init(editPayPalVaultID: String) {
         self.editPayPalVaultID = editPayPalVaultID
-        self.merchantAccountID = merchantAccountID
+    }
+
+    public func parameters() -> [String: Any] {
```
jaxdesmarais, line 29: Instead of having `riskCorrelationID` be an object we set, we can instead pass it into this method:
```suggestion
    public func parameters(riskCorrelationID: String) -> [String: Any] {
```

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
jaxdesmarais: It looks like we are defining `callbackURLHostAndPath` but just pulling it here from `BTPayPalRequest`. Which do we want to use?

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -159,7 +161,49 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
...
-    
+
+    // MARK: Edit FI flow
+
+    @objc func tappedPayPalEditVault(_ sender: UIButton) {
+        progressBlock("Tapped PayPal - Edit FI")
+        sender.setTitle("Processing...", for: .disabled)
+        sender.isEnabled = false
+
+        let request = BTPayPalVaultEditRequest(editPayPalVaultID: "+fZXfUn6nzR+M9661WGnCBfyPlIExIMPY2rS9AC2vmA=")
```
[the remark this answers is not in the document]
KunJeongPark: Good call. Right now, we just have one that's working, so I can make the default value.

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
KunJeongPark: This was to be used in second endpoint that is now being moved to server side, so we no longer need it. Since we are passing in encrypted billing agreement ID, I think we don't need it.
This endpoint never took that as a param to Gateway function.

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
```diff
@@ -1,20 +1,40 @@
...
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
 /// Options for the PayPal edit funding instrument flow
 /// - Warning: This feature is currently in beta and may change or be removed in future releases.
-public struct BTPayPalVaultEditRequest {
+public class BTPayPalVaultEditRequest {
 
+    var riskCorrelationID: String?
```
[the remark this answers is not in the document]
KunJeongPark: I thought it is a convenient way to just have a parameter function in there and add this value in the request object in our edit function.

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
KunJeongPark: oh yeah, what I did doesn't make much sense. should use the local property.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -196,8 +196,68 @@ import BraintreeDataCollector
...
+                self.notifyEditFIFailure(with: BTPayPalError.fetchConfigurationFailed, completion: completion)
+                return
+            }
+
+            self.isConfigFromCache = configuration.isFromCache
+
+            guard json["paypalEnabled"].isTrue else {
+                self.notifyEditFIFailure(with: BTPayPalError.disabled, completion: completion)
+                return
+            }
```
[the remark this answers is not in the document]
KunJeongPark: I ran into issues earlier extracting this into common function with subclassing or with protocols because of the difference in riskCorrelationID property. I got errors regarding ambiguous context.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -221,6 +281,102 @@ import BraintreeDataCollector
...
+    ///   - request: A `BTPayPalVaultErrorHandlingEditRequest`
+    ///   - completion: This completion will be invoked exactly once when the edit flow is complete or an error occurs.
+    /// - Warning: This feature is currently in beta and may change or be removed in future releases.
+    public func edit(
+        _ request: BTPayPalVaultErrorHandlingEditRequest,
+        completion: @escaping (BTPayPalVaultEditResult?, Error?) -> Void
+    ) {
+        // TODO: Analytics event
+
+        apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
```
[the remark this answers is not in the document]
KunJeongPark: I had trouble with correlationID being present one and not the other when creating super class or protocol as a parameter to a common function.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -196,8 +196,68 @@ import BraintreeDataCollector
...
+                    self.notifyEditFIFailure(with: BTPayPalError.httpPostRequestError(dictionary), completion: completion)
+                    return
+                }
+
+                guard let body, let approvalURL = BTPayPalApprovalURLParser(body: body) else {
+                    self.notifyEditFIFailure(with: BTPayPalError.invalidURL("Missing approval URL in gateway response."), completion: completion)
+                    return
+                }
+
+                switch approvalURL.redirectType {
```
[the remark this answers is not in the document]
KunJeongPark: The BTPayPalApprovalURLParser checks paypalAppApprovalUrl first. We are assuming this is null without providing params needed for it, is that ok?

Sources/BraintreePayPal/BTPayPalClient.swift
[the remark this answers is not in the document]
KunJeongPark: Let me try this.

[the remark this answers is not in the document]
KunJeongPark: I'm finding that having one method that takes a superclass of two options that are public facing a bit confusing, and might be confusing for merchant. 

Subclassing itself is easy, making a single function is fine but we have one signature that takes BTPayPalVaultEditRequest and it's subclassed request which is error handling, I think it maybe confusing for the merchant.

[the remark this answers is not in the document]
jaxdesmarais: The public interface shouldn't change, just the private method we call under the hood. This should be the exact same as what we do for tokenize today expect it instead will use the edit request.

Sources/BraintreePayPal/BTPayPalError.swift
```diff
@@ -112,6 +117,8 @@ public enum BTPayPalError: Error, CustomNSError, LocalizedError, Equatable {
             return "Missing BA Token for PayPal App Switch."
         case .missingPayPalRequest:
             return "The PayPal Request was missing or invalid."
+        case .missingRiskCorrelationID:
```
jaxdesmarais, line 120: Lets confirm with product that we should fail if we cannot fetch a `riskCorrelationID`. If so, we should throw this error before making the call to `generate_edit_fi_url`

jaxdesmarais: If we are guaranteeing the value of `riskCorrelationID` on line 489 we shouldn't need to use the optional `clientMetadataID` at all. Instead we can just pass the `riskCorrelationID` into the `handlePayPalEditFIRequest`, `performEditSwitchRequest`, and `performEditSwitchRequest` methods. While we don't expect this to ever happen if for some reason it does it'd be odd to throw this error so late in the flow. If we instead do the above we can remove the error entirely.

KunJeongPark: Do we need clientMetadataID for analytics for edit flow?

KunJeongPark: You are right. It is not right to have public error for merchants for non-existent scenario.

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -41,40 +41,86 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
...
+        return label
+    }()
+
+    let errorHandlingToggle = UISwitch()
+
     override func viewDidLoad() {
-        super.heightConstraint = 300
+        super.heightConstraint = 600
         super.viewDidLoad()
+        errorHandlingToggle.addTarget(self, action: #selector(toggleErrorHandling(_:)), for: .valueChanged)
```
jaxdesmarais, line 88:
```suggestion
        errorHandlingToggle.addTarget(self, action: #selector(toggleErrorHandling), for: .valueChanged)
```

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -159,7 +205,65 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
...
+
+    private func handlePayPalEditVault(_ sender: UIButton) {
+        progressBlock("Tapped PayPal - Edit FI")
+        sender.setTitle("Processing...", for: .disabled)
+        sender.isEnabled = false
+
+
+        let vaultID = payPalVaultIDTextField.text ?? "+fZXfUn6nzR+M9661WGnCBfyPlIExIMPY2rS9AC2vmA="
+        let request = BTPayPalVaultEditRequest(editPayPalVaultID: vaultID)
+        payPalClient.edit(request) { editResult, error in
```
jaxdesmarais, line 233: super nit just to add a space here:
```suggestion
        let request = BTPayPalVaultEditRequest(editPayPalVaultID: vaultID)
        
        payPalClient.edit(request) { editResult, error in
```

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -159,7 +205,65 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
...
+    }
+
+    private func handlePayPalErrorHandlingEditVault(_ sender: UIButton) {
+        progressBlock("Tapped PayPal - Edit FI, ErrorHandling")
+        sender.setTitle("Processing...", for: .disabled)
+        sender.isEnabled = false
+
+        let vaultID = payPalVaultIDTextField.text ?? "+fZXfUn6nzR+M9661WGnCBfyPlIExIMPY2rS9AC2vmA="
+        let riskCorrelationID = riskCorrelationIDTextField.text ?? "test"
+
```
jaxdesmarais, line 252:
```suggestion
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -221,6 +221,43 @@ import BraintreeDataCollector
...
+    ///   - completion: This completion will be invoked exactly once when the edit flow is complete or an error occurs.
+    /// - Warning: This feature is currently in beta and may change or be removed in future releases.
+    public func edit(
+        _ request: BTPayPalVaultErrorHandlingEditRequest,
+        completion: @escaping (BTPayPalVaultEditResult?, Error?) -> Void
+    ) {
+        // TODO: Analytics event
+        edit(request: request, completion: completion)
+    }
+    /// Allows a customer to edit their PayPal payment method.
```
jaxdesmarais, line 240:
```suggestion
    }

    /// Allows a customer to edit their PayPal payment method.
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -319,6 +356,19 @@ import BraintreeDataCollector
...
+        with url: URL,
+        riskCorrelationID: String,
+        completion: @escaping (BTPayPalVaultEditResult?, Error?) -> Void
+    ) {
+        // Defensive programming in case PayPal returns a non-HTTP URL so that ASWebAuthenticationSession doesn't crash
+        if let scheme = url.scheme, !scheme.lowercased().hasPrefix("http") {
+            notifyEditFIFailure(with: BTPayPalError.asWebAuthenticationSessionURLInvalid(scheme), completion: completion)
+            return
+        }
+        performEditSwitchRequest(editURL: url, riskCorrelationID: riskCorrelationID, completion: completion)
```
jaxdesmarais, line 369:
```suggestion
        }

        performEditSwitchRequest(editURL: url, riskCorrelationID: riskCorrelationID, completion: completion)
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -414,6 +464,69 @@ import BraintreeDataCollector
...
+            }
+
+            let dataCollector = BTDataCollector(apiClient: self.apiClient)
+
+            var riskCorrelationID: String
+            var parameters: [String: Any]
+
+            if let editRequest = request as? BTPayPalVaultErrorHandlingEditRequest {
+                riskCorrelationID = dataCollector.clientMetadataID(editRequest.riskCorrelationID)
+                parameters = editRequest.parameters()
```
jaxdesmarais, line 494: Should we be passing the `riskCorrelationID` here?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -414,6 +464,69 @@ import BraintreeDataCollector
...
+                guard let body, let approvalURL = body["paymentResource"]["redirectUrl"].asURL() ??  body["agreementSetup"]["approvalUrl"].asURL() else {
+                    self.notifyEditFIFailure(with: BTPayPalError.invalidURL("Missing approval URL in gateway response."), completion: completion)
+                    return
+                }
+
+                self.handlePayPalEditFIRequest(with: approvalURL, riskCorrelationID: riskCorrelationID,  completion: completion)
+            }
+        }
+
+
```
jaxdesmarais, line 527:
```suggestion
```

Sources/BraintreePayPal/BTPayPalVaultErrorHandlingEditRequest.swift
```diff
@@ -16,9 +18,17 @@ public struct BTPayPalVaultErrorHandlingEditRequest {
...
     /// - Warning: This feature is currently in beta and may change or be removed in future releases.
-    public init(editPayPalVaultID: String, riskCorrelationID: String, merchantAccountID: String? = nil) {
-        self.editPayPalVaultID = editPayPalVaultID
-        self.merchantAccountID = merchantAccountID
+    public init(editPayPalVaultID: String, riskCorrelationID: String) {
         self.riskCorrelationID = riskCorrelationID
+        super.init(editPayPalVaultID: editPayPalVaultID)
+    }
+
+    public func parameters() -> [String: Any] {
```
jaxdesmarais, line 26: I don't think we need this here since it exists in the super class unless we aren't supposed to pass `risk_correlation_id` for error handling

Sources/BraintreePayPal/BTPayPalClient.swift
KunJeongPark: That was what I understood from @sarahkoop

Sources/BraintreePayPal/BTPayPalError.swift
KunJeongPark: Yes, I think we still need the different parameters functions because one uses its request property (sending in the raw riskCorrelationID sent in by merchant in error handling flow) and one generated internally by us in regular edit flow (requiring argument in parameters function)

Sources/BraintreePayPal/BTPayPalClient.swift
KunJeongPark: I don't think so, I think I caused confusion with unnecessary call to DataCollector.clientMetadataID function there.
We are just using raw riskCorrelationID passed into the request there.

KunJeongPark: I was wrong about what Sarah told us - she told us to use CMID function with the merchant passed in riskCorrelationID. 
I've made corrections on this commit e5f98e80d94363d2b6e0307b06ea721695e17004 
to use the data collector CMID function for error handling flow with merchant passed in value.

I've posted question on collaboration channel just to double check on
1. whether we should be passing in riskCorrelationID for both edit flows
2. whether both flows should be using the data collector CMID function to generate this value passed into gateway

Sources/BraintreePayPal/BTPayPalVaultErrorHandlingEditRequest.swift
KunJeongPark: Made corrections to omit this.

Demo/Application/Base/BaseViewController.swift
```diff
@@ -5,6 +5,7 @@ class BaseViewController: UIViewController {
 
     var progressBlock: ((String?) -> Void) = { _ in }
     var completionBlock: ((BTPaymentMethodNonce?) -> Void) = { _ in }
+    var paypalEditCompletionBlock: ((String) -> Void) = { _ in  }
```
jaxdesmarais, line 8: I don't think we need this, we can just use the `progressBlock`

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -159,7 +205,65 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
...
+        payPalClient.edit(request) { editResult, error in
+            sender.isEnabled = true
+
+            guard let editResult else {
+                self.progressBlock(error?.localizedDescription)
+                return
+            }
+
+            self.paypalEditCompletionBlock(editResult.riskCorrelationID)
```
jaxdesmarais, line 242:
```suggestion
            self.progressBlock("Edit FI completed.\n riskCorrelationID: \(editResult.riskCorrelationID)")
```

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -159,7 +205,65 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
...
```
jaxdesmarais, line 263:
```suggestion
            self.progressBlock("Edit FI completed.\n riskCorrelationID: \(editResult.riskCorrelationID)")
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -221,6 +221,44 @@ import BraintreeDataCollector
...
+        // TODO: Analytics event
+        edit(request: request, completion: completion)
+    }
+
+    /// Allows a customer to edit their PayPal payment method.
+    ///
+    /// On success, you will receive an instance of `BTPayPalVaultEditResult`; on failure or user cancelation you will receive an error.
+    /// If the user cancels out of the flow, the error code will be `.canceled`.
+    ///
+    /// - Parameter request: A `BTPayPalVaultEditRequest`
```
jaxdesmarais, line 246:
```suggestion
    /// - Parameter request: A `BTPayPalVaultErrorHandlingEditRequest`
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -414,6 +466,66 @@ import BraintreeDataCollector
...
+                self.notifyEditFIFailure(with: BTPayPalError.disabled, completion: completion)
+                return
+            }
+
+            let dataCollector = BTDataCollector(apiClient: self.apiClient)
+
+            var riskCorrelationID: String
+
+            if let editRequest = request as? BTPayPalVaultErrorHandlingEditRequest {
+                riskCorrelationID = dataCollector.clientMetadataID( editRequest.riskCorrelationID)
```
jaxdesmarais, line 494:
```suggestion
                riskCorrelationID = dataCollector.clientMetadataID(editRequest.riskCorrelationID)
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -530,6 +642,58 @@ import BraintreeDataCollector
         }
     }
 
+    private func performEditSwitchRequest(
+        editURL: URL,
```
jaxdesmarais, line 646: nit just to clean up the naming a bit:
```suggestion
        with url: URL,
```

Sources/BraintreePayPal/BTPayPalVaultErrorHandlingEditRequest.swift
```diff
@@ -1,13 +1,15 @@
...
 /// Options for the PayPal edit funding instrument flow with retry on failed attempts
 ///
 /// - Warning: This feature is currently in beta and may change or be removed in future releases.
-public struct BTPayPalVaultErrorHandlingEditRequest {
+public class BTPayPalVaultErrorHandlingEditRequest: BTPayPalVaultEditRequest {
 
-    private let editPayPalVaultID: String
-    private let riskCorrelationID: String
-    private let merchantAccountID: String?
+    public let riskCorrelationID: String
```
jaxdesmarais, line 12: I think we want to keep this private and have it only be set via the init

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
```diff
@@ -1,20 +1,35 @@
...
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
 /// Options for the PayPal edit funding instrument flow
 /// - Warning: This feature is currently in beta and may change or be removed in future releases.
-public struct BTPayPalVaultEditRequest {
+public class BTPayPalVaultEditRequest {
+
+    let editPayPalVaultID: String
```
jaxdesmarais, line 11: Does this need to be internal for access in the subclass or can it remain private?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -24,13 +24,13 @@ import BraintreeDataCollector
     var approvalURL: URL? = nil
 
     /// Exposed for testing the clientMetadataID associated with this request.
-    /// Used in POST body for FPTI analytics & `/paypal_account` fetch.
+    /// Used in POST body for FPTI analytics & `/paypal_account` fetch & editFI POST body.
```
jaxdesmarais, line 27: It doesn't look like we use this at all for the edit FI post body anymore?

Sources/BraintreePayPal/BTPayPalVaultErrorHandlingEditRequest.swift
KunJeongPark: With private, I run into issue when I access it in the dataCollector CMID function. 
I can make it internal I think

Sources/BraintreePayPal/BTPayPalVaultEditRequest.swift
KunJeongPark: I see no errors setting it private.

Sources/BraintreePayPal/BTPayPalClient.swift
KunJeongPark: I just followed convention we had with performSwitchRequest.
I wanted to clarify what the url is for also.

Demo/Application/Base/BaseViewController.swift
KunJeongPark: But it's really a completion Block, not progressBlock, which is used to show progress. 
I thought it might be confusing.

jaxdesmarais: The progress block is also used for errors, which are part of the completion. It looks like `completionBlock` here represents anything that's a `BTPaymentMethodNonce` and `progressBlock` is used for anything that's a string, so I think reusing that is fine.

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: I think it's pretty clear since the method is called `performEditSwitchRequest`. Also the other one isn't particularly accurate since it's called `appSwitchURL` and is opening a web URL. What do you think?

KunJeongPark: I don't have strong preference.

Demo/Application/Base/ContainmentViewController.swift
```diff
@@ -42,6 +42,10 @@ class ContainmentViewController: UIViewController {
         updateStatus("Got a nonce. Tap to make a transaction.")
     }
 
+    func paypalEditCompletionBlock(_ correlationID: String) {
+        updateStatus("Edit FI completed.\n riskCorrelationID: \(correlationID)")
+    }
+
```
jaxdesmarais, line 48:
```suggestion
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -26,11 +26,11 @@ import BraintreeDataCollector
     /// Exposed for testing the clientMetadataID associated with this request.
     /// Used in POST body for FPTI analytics & `/paypal_account` fetch.
     var clientMetadataID: String? = nil
-    
+
     /// Exposed for testing the intent associated with this request
     var payPalRequest: BTPayPalRequest? = nil
 
-    /// Exposed for testing, the ASWebAuthenticationSession instance used for the PayPal flow
```
jaxdesmarais, line 33: Looks like this space got removed accidentally

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -383,6 +536,13 @@ class BTPayPalClient_Tests: XCTestCase {
         XCTAssertNotNil(payPalClient.clientMetadataID)
     }
 
+    func testEditFI_whenGenerateFIURLSuccessful_performsSwitchRequest() {
+        let request = BTPayPalVaultEditRequest(editPayPalVaultID: "test-ID")
+        payPalClient.edit(request) { _, _ in }
+
+        XCTAssertNotNil(payPalClient.webAuthenticationSession)
```
jaxdesmarais, line 543: Are we able to assert that the risk correlation ID is not nil here?

KunJeongPark: We are not persisting it as PayPalClient property anymore but using a local variable.

jaxdesmarais: Shouldn't it exist in here?
```swift
payPalClient.edit(request) { result, error in 
    XCTAssertNotNil(result.riskCorrelationID)
    XCTAssertNil(error)
}
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -221,6 +221,44 @@ import BraintreeDataCollector
         }
     }
 
+    /// Allows a customer to edit their PayPal payment method.
```
jaxdesmarais, line 224:
```suggestion
    /// Allows a customer to retry alternate funding instruments on failed attempts.
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -221,6 +221,44 @@ import BraintreeDataCollector
...
+    /// - Warning: This feature is currently in beta and may change or be removed in future releases.
+    public func edit(
+        _ request: BTPayPalVaultErrorHandlingEditRequest,
+        completion: @escaping (BTPayPalVaultEditResult?, Error?) -> Void
+    ) {
+        // TODO: Analytics event
+        edit(request: request, completion: completion)
+    }
+
+    /// Allows a customer to edit their PayPal payment method.
```
jaxdesmarais, line 241:
```suggestion
    /// Allows a customer to retry alternate funding instruments on failed attempts.
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -568,6 +732,24 @@ import BraintreeDataCollector
         )
         completion(nil, BTPayPalError.canceled)
     }
+
+    private func notifyEditFISuccess(
+        with result: BTPayPalVaultEditResult,
+        completion: @escaping (BTPayPalVaultEditResult?, Error?) -> Void
+    ) {
+       // TODO: add Analytics Event
```
jaxdesmarais, line 740: When we do the analytics piece, lets make sure we are also passing the `correlationID` as the result of the data collector for success, failure, and cancel.

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
jaxdesmarais: Wanted to make sure this doesn't get missed

KunJeongPark: Did you see a test I pushed?

KunJeongPark:
@jaxdesmarais Did you see the tests that I added for checking correlationID in success result and cancellation? (10dcbe263584fb35988a98f45bc2bf1b2661ceff and  b5505a9ee2d6ad6d711615a7804c6093f12ad5ad.
There are already tests for error  cases:

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -414,6 +466,66 @@ import BraintreeDataCollector
         }
     }
 
+    private func edit(request: BTPayPalVaultEditRequest, completion: @escaping (BTPayPalVaultEditResult?, Error?) -> Void
+    ) {
```
jaxdesmarais, line 470:
```suggestion
    private func edit(request: BTPayPalVaultEditRequest, completion: @escaping (BTPayPalVaultEditResult?, Error?) -> Void) {
```

jaxdesmarais (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -319,6 +357,20 @@ import BraintreeDataCollector
         performSwitchRequest(appSwitchURL: url, paymentType: paymentType, completion: completion)
     }
 
+    func handlePayPalEditFIRequest(
+        with url: URL,
+        riskCorrelationID: String,
+        completion: @escaping (BTPayPalVaultEditResult?, Error?) -> Void
+    ) {
+        // Defensive programming in case PayPal returns a non-HTTP URL so that ASWebAuthenticationSession doesn't crash
+        if let scheme = url.scheme, !scheme.lowercased().hasPrefix("http") {
```
agedd, line 366: nit(optional): 
```suggestion
        if let urlScheme = url.scheme, !urlScheme.lowercased().hasPrefix("http") {
```

Sources/BraintreePayPal/BTPayPalVaultErrorHandlingEditRequest.swift
```diff
@@ -1,24 +1,24 @@
...
 /// Options for the PayPal edit funding instrument flow with retry on failed attempts
 ///
 /// - Warning: This feature is currently in beta and may change or be removed in future releases.
-public struct BTPayPalVaultErrorHandlingEditRequest {
+public class BTPayPalVaultErrorHandlingEditRequest: BTPayPalVaultEditRequest {
 
-    private let editPayPalVaultID: String
-    private let riskCorrelationID: String
-    private let merchantAccountID: String?
+    let riskCorrelationID: String
```
agedd, line 12: qq: should this be private as well?

jaxdesmarais: Great question, this is used internally in `BTPayPalClient` on line 493: `dataCollector.clientMetadataID(editRequest.riskCorrelationID)` so needs to be internal since it wouldn't be accessible there if it was private.

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: Updated: https://github.com/braintree/braintree_ios/pull/1395/commits/2bccefcec847524713fe700b255b5cbbdba74029

agedd (approved):
lgtm!!

PR #1399, opened 2024-08-27
Async fetch configuration

KunJeongPark:
### Summary of changes

- Implement async await fetchConfiguration and replace all usage of completion handler fetchOrReturnRemoteConfiguration with fetchConfiguration. 

Replacing of completion handler version of `fetchOrReturnRemoteConfiguration` required wrapping the functions in Task blocks within functions that return results via completion handler to the merchants.

There were many changes to unit tests required as tests with mocks returned results synchronously and asynchronous nature of the Task blocks required slight delays for the assertions.

I think there are a lot of considerations - 

Consider refactoring functions wrapped with withCheckedContinuation to true asynchronous/await syntax where possible

How far down do we refactor? Within the task blocks, there are many subsequent function calls that pass down the completion handlers. I am unsure about top down approach to the refactor where we have task block for top level tokenize function for instance, as opposed to bottom up, like the apiClient’s post function or the functions that tokenize calls.

We should do careful and extensive testing to make sure that there are no unintended consequences especially in app switching context. Major difference with creating a Task block is that it does not inherit the context of its caller by default. We should test thoroughly to ensure that there are no subtle bugs introduced like deadlocks and priority inversions.

I will open another ticket to continue to test and improve on the refactor started here.

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark @jaxdesmarais


Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
```diff
@@ -43,21 +41,17 @@ class ConfigurationLoader {
     @_documentation(visibility: private)
     @ConfigurationActor
     func getConfig() async throws -> BTConfiguration {
+        var existingTask: Task<BTConfiguration, Error>
```
KunJeongPark: I just noticed this commit with moving existingTask as property in this function. Would this eliminate benefits of deduplication of the task in quick successive calls?

jaxdesmarais:
Closing this PR until this work is reprioritized. Will reopen or open a new PR when we pick this back up.

PR #1411, opened 2024-09-10
App switch paypal installed checks

nmehtapp:

### Summary of changes
- Modify the isPayPalInstalled func to check for universal link along with deeplink for app switch flows 
- log and event if app can open deeplink but not the universal link

### Checklist

- [ ] Added a changelog entry

- @nmehtapp

### DIFF
--- a/Sources/BraintreeCore/BTCoreConstants.swift
+++ b/Sources/BraintreeCore/BTCoreConstants.swift
@@ -15,6 +15,9 @@ import Foundation
 
     /// URL Scheme for PayPal App
     public static let payPalURLScheme: String = "paypal-app-switch-checkout"
+    
+    /// Universal link for PayPal app (app switch flow)
+    public static let payPalUniversalLink: String = "https://paypal.com/app-switch-checkout"
 
     static let apiVersion: String = "2016-10-07"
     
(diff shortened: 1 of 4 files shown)


Sources/BraintreeCore/BTCoreConstants.swift
```diff
@@ -15,6 +15,9 @@ import Foundation
 
     /// URL Scheme for PayPal App
     public static let payPalURLScheme: String = "paypal-app-switch-checkout"
+    
+    /// Universal link for PayPal app (app switch flow)
+    public static let payPalUniversalLink: String = "https://paypal.com/app-switch-checkout"
```
jaxdesmarais, line 20: This will always return true regardless of if the app is installed or not. You can test this by doing the following in simulator without the PayPal app installed:
```swift
print(UIApplication.shared.canOpenURL(URL(string: "https://paypal.com/app-switch-checkout")!))
```
This is a valid URL string so it's always possible to open it.

jaxdesmarais:
Closing based on above comment: https://github.com/braintree/braintree_ios/pull/1411#discussion_r1761974587. Will look to rework this logic to check the app is installed as expected vs checking a static URL.

PR #1414, opened 2024-09-16
[QL] Edit FI Analytics

jaxdesmarais:
### Summary of changes

- Send analytics events for the edit FI feature

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeCore/BTURLUtils.swift
+++ b/Sources/BraintreeCore/BTURLUtils.swift
@@ -32,7 +32,7 @@ import Foundation
                         encode(subValue.description)
                     )
                 }
-            } else if value as? NSNull != nil {
+            } else if value is NSNull {
                 queryString = queryString.appendingFormat("%@=&", encodedKey)
             } else {
                 queryString = queryString.appendingFormat("%@=%@&", encodedKey, encode(String(describing: value).description))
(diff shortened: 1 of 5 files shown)


stechiu (approved):
LTGM!

Sources/BraintreeCore/BTURLUtils.swift
```diff
@@ -32,7 +32,7 @@ import Foundation
                         encode(subValue.description)
                     )
                 }
-            } else if value as? NSNull != nil {
+            } else if value is NSNull {
```
sarahkoop, line 35: Potentially me misunderstanding swift but this looks like we are changing the check from not equal to nil to equal to nil - is that right?

jaxdesmarais: This logic is super weird for sure, but the original code was checking that the value could be cast to `NSNull` successfully. In this commit https://github.com/braintree/braintree_ios/commit/006ec11c4388c1defbc183edd5307bcad330e8f0#diff-35c217704427fba4f0917b3d88910e3e99533a538b2a0a404ab575881a4c2400 it was originally `else if let _ = value as? NSNull`. SwiftLint doesn't like setting a let to an unnamed property so it was changed to `if value as? NSNull != nil`, instead we can just check if the value is `NSNull` which ultimately makes this code easier to read and more in line with the logic we had initially.

sarahkoop (approved):

PR #1417, opened 2024-09-18
[QL] Update `interval` and `intervalCount`

jaxdesmarais:
### Summary of changes

- End to end testing revealed that there are cases where `billing_frequency` and `billing_frequency_unit` are not required for `unscheduled` cases for RBA metadata - updated and tested that these fields can be omitted without error when the `recurringBillingPlanType` is `.unscheduled` and the `pricingModel` is `.autoReload` per the test case
- Updated docstrings

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Demo/Application/Base/ContainmentViewController.swift
+++ b/Demo/Application/Base/ContainmentViewController.swift
@@ -182,7 +182,7 @@ class ContainmentViewController: UIViewController {
                 }
             }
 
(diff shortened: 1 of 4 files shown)


richherrera (approved):

agedd (approved):
lgtm!

PR #1421, opened 2024-09-26
Remove Completion Edit FI Methods

jaxdesmarais:
### Summary of changes

- Remove completion Edit FI methods
- Update unit tests to use async/await Edit FI methods

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## unreleased
 * BraintreePayPal
   * Add PayPal vault edit flow (BETA)
-    * Add `BTPayPalClient.edit(_:completion:)` method that takes in a `BTPayPalVaultEditRequest` or a `BTPayPalVaultErrorHandlingEditRequest`
+    * Add `BTPayPalClient.edit(_:)` method that takes in a `BTPayPalVaultEditRequest` or a `BTPayPalVaultErrorHandlingEditRequest`
     * **Note:** This feature is currently in beta and may change or be removed in future releases
   * Send `isVaultRequest` for App Switch events to PayPal's analytics service (FPTI)
 
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -189,29 +189,13 @@ import BraintreeDataCollector
     /// On success, you will receive an instance of `BTPayPalVaultEditResult`; on failure or user cancelation you will receive an error.
     /// If the user cancels out of the flow, the error code will be `.canceled`.
     ///
-    /// - Parameters:
-    ///   - request: A `BTPayPalVaultEditRequest`
-    ///   - completion: This completion will be invoked exactly once when the edit flow is complete or an error occurs.
-    /// - Warning: This feature is currently in beta and may change or be removed in future releases. Additionally, this feature is currently only supported with client token authorization.
-    public func edit(
-        _ request: BTPayPalVaultEditRequest,
-        completion: @escaping (BTPayPalVaultEditResult?, Error?) -> Void
-    ) {
-        edit(request: request, completion: completion)
-    }
-
-    /// Allows a customer to edit their PayPal payment method.
-    ///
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## unreleased
 * BraintreePayPal
   * Add PayPal vault edit flow (BETA)
-    * Add `BTPayPalClient.edit(_:completion:)` method that takes in a `BTPayPalVaultEditRequest` or a `BTPayPalVaultErrorHandlingEditRequest`
+    * Add `BTPayPalClient.edit(_:)` method that takes in a `BTPayPalVaultEditRequest` or a `BTPayPalVaultErrorHandlingEditRequest`
```
jaxdesmarais, line 6: Looks like this is how apple documents async/await methods. Used this as a reference: https://developer.apple.com/documentation/foundation/urlsession/3767353-data

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -211,7 +195,7 @@ import BraintreeDataCollector
     /// - Warning: This feature is currently in beta and may change or be removed in future releases. Additionally, this feature is currently only supported with client token authorization.
     public func edit(_ request: BTPayPalVaultEditRequest) async throws -> BTPayPalVaultEditResult {
         try await withCheckedThrowingContinuation { continuation in
-            edit(request) { result, error in
+            edit(request: request) { result, error in
```
jaxdesmarais, line 198: I opted to just change this for now vs going down the route of making everything downstream use async/await since that'd also require changes to how we use ASWeb. Though this is something to consider with how we want to handle async/await across the board as it'll require some larger changes.

warmkesselj (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
scannillo: Good callout - would be nice to create a plan for slowly migrating everything internally to async-await at some point. Maybe after we drop completionHandler syntax in the next major version 😈 👀

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -152,21 +152,20 @@ class BTPayPalClient_Tests: XCTestCase {
...
-        clientTokenPayPalClient.edit(editRequest) { result, error in
-            XCTAssertNil(error)
-            XCTAssertNotNil(result?.riskCorrelationID)
-            expectation.fulfill()
+
+        do {
+            let _ = try await clientTokenPayPalClient.edit(request)
+        } catch {
+            // no-op
         }
```
scannillo, line 160: I think you can simplify this by doing:

```suggestion
        let _ = try? await clientTokenPayPalClient.edit(request)
```

scannillo: > You use try? to handle an error by converting it to an optional value. If an error is thrown while evaluating the try? expression, the value of the expression is nil. 

[Apple docs](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/errorhandling/#Converting-Errors-to-Optional-Values)

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: Agree! We can also slowly migrate things even in a minor version, but will be much easier to do it from the bottom up and migrate the internal/private methods first and work our way up to the public methods. Maybe I will take a look at some of these next open dev...

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
jaxdesmarais: ❤️ nice, updated them all here: https://github.com/braintree/braintree_ios/pull/1421/commits/2181f9de4fd1fd4ee57a236e64ddb4ea05fca2d9

PR #1426, opened 2024-10-04
[OPEN DEV] Add Custom Authorization Setting

jaxdesmarais:
### Summary of changes

- Allow us to pass custom tokenization keys in sandbox if needed. This would allow us to do quick clarity checks on flows that may not be set up for our sample merchant server. It could also allow us to test flows on stage (that are routed through sandbox on the gateway side) for things like app switch before it is rolled out. We weren't using the previous custom setup at all, so this gives us a bit more use.
- You can test this by using your own tokenization key or using `sandbox_jy4fvpfg_v7x2rb226dx4pr7b` and the PayPal Vault flow, this is routed to hit stage and you should see the stage URL in the demo app.

### Checklist

### Authors

- @jaxdesmarais


Demo/Application/Base/Settings/BraintreeDemoSettings.swift
```diff
@@ -35,28 +35,19 @@ class BraintreeDemoSettings: NSObject {
 
     static var currentEnvironmentName: String {
         switch currentEnvironment {
-        case .sandbox:
+        case .sandbox, .custom:
```
jaxdesmarais, line 38: Right now I have this set up to hit sandbox since that seemed like what we would hit the most. Could also consider building logic to parse the first part of the environment from the tokenization key if folks think we may also want to use this for prod testing. Thoughts?

jaxdesmarais: Worth noting on the Braintree Gateway side stage routes through sandbox as well.

agedd (approved):
lgtm! 🙌🏽

richherrera (approved):

PR #1428, opened 2024-10-07
Fix ShopperInsights - sessionID loss analytics

scannillo:
### Issue [LI-64570]
- There was a bug found with an LE merchant's ShopperInsights integration where the PaymentReady API team logs were tracking _more_ unique sessionIDs used to call their API from our SDK, than our SDK logs were tracking alone
    - The SDK sends our `session_id` to the PaymentReady API via[ this header](https://github.com/braintree/braintree_ios/blob/971196de71e88b1ccf2c35bf79ca7fc4c4c60b77/Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift#L58)
    - Our analytics batch upload logic was including different events from varying sessionIDs in a single batch payload, and since the `session_id` was at the `batch_params` level, the unique sessionIDs per API call were squashed/lost only logging the latest sessionID to be created/used.

### Summary of changes

- Move `session_id` from `batch_params` to `event_params` level

### Note

- Let's do the same on Android, for parity (DTMOBILES-1110)

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreeCore
+  * Fix bug where `session_id` in analytics payload would not always be accurate; move to `event_params` dict
+
 ## 6.23.4 (2024-09-24)
 * BraintreePayPal
   * Send `isVaultRequest` for App Switch events to PayPal's analytics service (FPTI)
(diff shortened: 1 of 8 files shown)


scannillo:
Replaced by https://github.com/braintree/braintree_ios/pull/1430

PR #1435, opened 2024-10-16
Add App Switch Analytics events

richherrera:
### Summary of changes

- Add the same events for the Venmo flow that were added for the PayPal app switch flow:
`venmo:tokenize:app-switch:started` and `venmo:tokenize:handle-return:started`
- Add `appSwitchUrl` in the `appSwitchSucceeded` and `appSwitchFailed` events.

<p>
  
  
</p>

### Checklist

- [x] Added a changelog entry
- [x] Relevant test coverage

### Authors

- @richherrera

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreeVenmo
+  * Send `appSwitchUrl` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
+
 ## 6.24.0 (2024-10-15)
 * BraintreePayPal
   * Add `BTPayPalRecurringBillingDetails` and `BTPayPalRecurringBillingPlanType` opt-in request objects. Including these details will provide transparency to users on their billing schedule, dates, and amounts, as well as launch a modernized checkout UI.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 * BraintreeVenmo
-  * Send `appSwitchUrl` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
+  * Send `url` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
 
 ## 6.24.0 (2024-10-15)
 * BraintreePayPal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
-
 * BraintreePayPal
   * Add `BTPayPalRequest.userPhoneNumber` optional property
 * BraintreeVenmo
--- a/Sources/BraintreeCore/Analytics/FPTIBatchData.swift
+++ b/Sources/BraintreeCore/Analytics/FPTIBatchData.swift
@@ -29,6 +29,7 @@ struct FPTIBatchData: Codable {
     /// Encapsulates a single event by it's name and timestamp.
     struct Event: Codable {
 
+        let appSwitchURL: URL?
         /// UTC millisecond timestamp when a networking task started establishing a TCP connection. See [Apple's docs](https://developer.apple.com/documentation/foundation/urlsessiontasktransactionmetrics#3162615).
         /// `nil` if a persistent connection is used.
         let connectionStartTime: Int?
@@ -61,6 +62,7 @@ struct FPTIBatchData: Codable {
         let tenantName: String = "Braintree"
         
         init(
+            appSwitchURL: URL? = nil,
             connectionStartTime: Int? = nil,
             correlationID: String? = nil,
             endpoint: String? = nil,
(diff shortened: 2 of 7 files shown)


Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -93,6 +96,7 @@ struct FPTIBatchData: Codable {
         }
 
         enum CodingKeys: String, CodingKey {
+            case appSwitchURL = "appSwitchUrl"
```
jaxdesmarais, line 99: `appSwitchUrl` is not an approved tag in FPTI. We will want to use `url` here instead.

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -29,6 +29,7 @@ struct FPTIBatchData: Codable {
     /// Encapsulates a single event by it's name and timestamp.
     struct Event: Codable {
 
+        let appSwitchURL: URL?
```
jaxdesmarais, line 32: I think here and in the init this would be a string. And when we send the event should instead be using `appSwitchURL.absoluteString`.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreeVenmo
+  * Send `appSwitchUrl` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
```
```suggestion
  * Send `url` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
```

UnitTests/BraintreeTestShared/MockAPIClient.swift
```diff
@@ -17,6 +17,7 @@ public class MockAPIClient: BTAPIClient {
     public var postedIsVaultRequest = false
     public var postedMerchantExperiment: String? = nil
     public var postedPaymentMethodsDisplayed: String? = nil
+    public var postedAppSwitchURL: [String: URL?] = [:]
```
jaxdesmarais, line 20: I think this would just be a string of the URL, right?

richherrera: You're right, updated: [4712ec0](https://github.com/braintree/braintree_ios/pull/1435/commits/4712ec0f96259bd101dc9c8a8713c3860d96012c)

CHANGELOG.md
richherrera: Updated here: [ec399f0](https://github.com/braintree/braintree_ios/pull/1435/commits/ec399f0bfe6a6dfdff6ac01a0e5db67845cbe6d2)

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
richherrera: Got it, thanks for the feedback, updated: [714dd27](https://github.com/braintree/braintree_ios/pull/1435/commits/714dd27c44918283e04160d130bec7687a3d49bd)

UnitTests/BraintreeTestShared/MockAPIClient.swift
```diff
@@ -17,6 +17,7 @@ public class MockAPIClient: BTAPIClient {
     public var postedIsVaultRequest = false
     public var postedMerchantExperiment: String? = nil
     public var postedPaymentMethodsDisplayed: String? = nil
+    public var postedAppSwitchURL: [String: String?] = [:]
```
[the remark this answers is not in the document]
richherrera: This is a dictionary because we currently don't have a way to verify the parameters sent in each event. The way the other parameters are handled (for example, `postedIsVaultRequest`) only allows us to evaluate the parameters sent in the last event (in cases where more than one event is sent, like in the `invokedOpenURLSuccessfully` method in my unit test)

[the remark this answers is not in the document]
richherrera: In my unit test, the events appSwitchFailed and tokenizeFailed are sent, in that order. The second event doesn’t have the appSwitchUrl parameter, which returns nil and prevents me from validating that the events are being sent correctly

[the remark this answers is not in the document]
jaxdesmarais: Ah, I see you are saying the test is to verify that we are sending it for that specific event and not just that it is being set, which makes sense. Though I don't think it would be set as nil, just tested locally and it would still be set if it's being passed for any of the events in `mockAPIClient`. If we are instead trying to verify the URL is sent for this specific event I think what you have makes sense.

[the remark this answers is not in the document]
richherrera: `appSwitchURL` isn’t sent in all events, and yes, we’re testing that the parameter is sent in the events specified in this ticket

jaxdesmarais (approved):

warmkesselj (approved):
LGTM

PR #1440, opened 2024-10-21
Venmo Universal Link Return

jaxdesmarais:
### Summary of changes

- Add `BTVenmoClient(apiClient:universalLink:)` to use Universal Links when redirecting back from the Venmo flow
- Update Demo App
- Update Unit Tests
- Android PR: https://github.com/braintree/braintree_android/pull/1190

### Checklist

- [x] Added a changelog entry

### Authors

- @scannillo @jaxdesmarais


Sources/BraintreeVenmo/BTVenmoAppSwitchRedirectURL.swift
```diff
@@ -57,6 +55,16 @@ struct BTVenmoAppSwitchRedirectURL {
             "braintree_sdk_data": base64EncodedBraintreeData ?? "",
             "customerClient": "MOBILE_APP"
         ]
+
+        if let universalLink {
+            queryParameters["x-success"] = universalLink.absoluteString + "/success"
+            queryParameters["x-error"] = universalLink.absoluteString + "/error"
+            queryParameters["x-cancel"] = universalLink.absoluteString + "/cancel"
```
scannillo, line 62: Can we use URLComponents, or some other Apple built-in function, to do a path append? This way, we don't care if a merchant passes in a URL with an end ing slash `/` or not?

scannillo:
I think URL schemes on the way back from Venmo was the only remaining need for a merchant to set `BTAppContextSwitcher.sharedInstance.returnURLScheme` 🥹. Let's deprecate it! 🔥

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -46,16 +46,28 @@ import BraintreeCore
     /// Used for sending the type of flow, universal vs deeplink to FPTI
     private var linkType: LinkType?
 
+    private var universalLink: URL?
+
     // MARK: - Initializer
 
-    /// Creates an Apple Pay client
+    /// Creates an Venmo client
```
agedd, line 53: small nit (take it or leave it): 
```suggestion
    /// Creates a Venmo client
```

agedd (approved):
lgtm!! 🚀

richherrera (approved):

Sources/BraintreeCore/BTAppContextSwitcher.swift
```diff
@@ -14,6 +14,11 @@ import UIKit
     /// The URL scheme to return to this app after switching to another app or opening a SFSafariViewController.
     /// This URL scheme must be registered as a URL Type in the app's info.plist, and it must start with the app's bundle ID.
     /// - Note: This property should only be used for the Venmo flow.
+    @available(
+        *,
+        deprecated,
+        message: "returnURLScheme is deprecated and will be removed in a future version. Use BTVenmoClient(apiClient:universalLink:)."
```
scannillo, line 20: Should this deprecation go in the CHANGELOG?

scannillo (approved):

Sources/BraintreeCore/BTAppContextSwitcher.swift
jaxdesmarais: So I forgot about this swift limitation until now, but we cannot add this message since it causes cocoapods to fail ([previously discussed here](https://github.com/braintree/braintree_ios/pull/1130)). We cannot use the same solution in that PR because `BTAppContextSwitcher` lives in `Core` so the property would need to be public and the solution is to keep it internal. There may be some wonky workarounds but still no out of the box solutions. Any ideas?

scannillo: We could do the same thing as that PR ^, but no-doc the "internal" (actually public) interior property that it references?

jaxdesmarais: Ah, you're right, I had also added a deprecation message to the public property which resulted in the same error. 🤦

jaxdesmarais: Updated: https://github.com/braintree/braintree_ios/pull/1440/commits/2705af71d15b7b0ee218ad97d9d27db7ff362c70

jaxdesmarais:
Adding do not merge while we wait on RCIP approval for merge/release

scannillo:
RCIP case RITM3021997 has been closed. This PR is ready for merge / release.

jaxdesmarais:
Merging in and will be released once moratorium is lifted

PR #1445, opened 2024-10-23
[V7] Make `BTClientToken` internal + drop Obj-C exposure

scannillo:
DTMOBILES-1137

### Summary of changes

- Drop Objective-C support for `BTClientToken`
    - It was only exposed to Objective-C for our DropIn SDK to use, not for merchant use. 
    - The DropIn has not been compatible with the BT iOS SDK since it was re-written in Swift (v6). The decision was made to remove the Obj-C exposure that still exists for Drop-In since we have no current plan of updating it, and if we do it will likely require a re-write. 
- Make `BTClientToken` internal
- Remove unused protocol conformances & related unit tests

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


jaxdesmarais (approved):
🔥

agedd (approved):

warmkesselj (approved):

PR #1449, opened 2024-10-30
Update heroku server endpoints

warmkesselj:
### Summary of changes

Updating the url endpoints for the new Heroku apps ruby server.


jaxdesmarais (commented):
Are we able to add a description to this PR with why we are making this change? I also assume we've tested all the flows to ensure things are still working as expected in the demo app?

warmkesselj:
> Are we able to add a description to this PR with why we are making this change? I also assume we've tested all the flows to ensure things are still working as expected in the demo app?

I've run through the demo app but would be nice to have a second pair of eyes run through as well.

scannillo (approved):

jaxdesmarais (approved):
Ran through the flows as well in production and sandbox and both appear to be working as expected.

PR #1452, opened 2024-11-01
[v7] Update `BTThreeDSecureRequest` Parameters

agedd:
### Summary of changes

- Update `BTThreeDSecureRequest` properties to have everything in the init vs using dot syntax
- Update related unit tests 
- Confirmed with the gateway on questions around required vs optional parameters

### Checklist

- [x] Added a changelog entry

- @agedd


Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -7,97 +7,89 @@ import BraintreeCore
...
+        email: String? = nil,
+        exemptionRequested: Bool = false,
+        mobilePhoneNumber: String? = nil,
+        nonce: String? = nil,
+        renderTypes: [BTThreeDSecureRenderType]? = nil,
+        requestedExemptionType: BTThreeDSecureRequestedExemptionType = .unspecified,
+        shippingMethod: BTThreeDSecureShippingMethod = .unspecified,
+        uiType: BTThreeDSecureUIType = .both,
+        v2UICustomization: BTThreeDSecureV2UICustomization? = nil
+    ) {
```
agedd, line 75: note: alphabetized the order of the parameters in the initializer

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -424,9 +424,9 @@ import BraintreeCore
             requestParameters["customFields"] = customFields
         }
 
-        if request._cardAddChallenge == .requested || request.cardAddChallengeRequested == true {
+        if request.cardAddChallengeRequested == true {
             requestParameters["cardAdd"] = true
-        } else if request._cardAddChallenge == .notRequested {
+        } else if request.cardAddChallengeRequested == false {
```
agedd, line 429: question for the team: does this logic sound okay? (updated this after removing the `cardAddChallenge` property based on its doc strings)

jaxdesmarais: There is a separate ticket for this: DTMOBILES-1140. So we can only do the init portion in this PR.

agedd: ahh gotcha! thanks, addressed in b2bec4a7a7638818e90bb3b13d658912b0b9a9a4

jaxdesmarais: I think we can revert the changes here and uncomment `_cardAddChallenge` for now, then we can tackle all of it in a single PR.

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
warmkesselj: What do you think about grouping the required properties first and then the non-optional properties?

warmkesselj (approved):
One comment but it's not a blocker. Approving ahead of time.

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -66,38 +42,74 @@ import BraintreeCore
...
-    // MARK: - Internal Properties
+    // MARK: - Initializer
     
-    /// The dfReferenceID for the session. Exposed for testing.
-    var dfReferenceID: String?
+    /// Creates a `BTThreeDSecureRequest`
+    /// - Parameters:
+    ///    - accountType: Optional. The account type selected by the cardholder. Some cards can be processed using either a credit or debit account and cardholders have the option to choose which account to use.
+    ///    - additionalInformation: Optional. The additional information used for verification.
+    ///    - amount: The amount for the transaction.
```
jaxdesmarais, line 60: If this is optional do we want to note that in the docstring?

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -66,38 +42,74 @@ import BraintreeCore
...
+    ///    - billingAddress: Optional. The billing address used for verification
+    ///    - cardAddChallengeRequested: Optional.  An authentication created using this flag should only be used for vaulting operations (creation of customers' credit cards or payment methods) and not for creating transactions. If set to `true`, a card-add challenge will be requested from the issuer. If set to `false`, a card-add challenge will not be requested. If the parameter is missing, a card-add challenge will only be requested for $0 amount.
+    ///    - challengeRequested: Optional. If set to true, an authentication challenge will be forced if possible.
+    ///    - customFields: Object where each key is the name of a custom field which has been configured in the Control Panel. In the Control Panel you can configure 3D Secure Rules which trigger on certain values.
```
jaxdesmarais, line 64: Same here, do we want to note this as optional in the docstring?

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -66,38 +42,74 @@ import BraintreeCore
...
+    ///    - dataOnlyRequested:  Optional. Indicates whether to use the data only flow. In this flow, frictionless 3DS is ensured for Mastercard cardholders as the card scheme provides a risk score for the issuer to determine whether to approve. If data only is not supported by the processor, a validation error will be raised. Non-Mastercard cardholders will fallback to a normal 3DS flow.
+    ///    - dfReferenceID: The dfReferenceID for the session. Exposed for testing.
```
jaxdesmarais, line 66: Should this be exposed for merchants? It looks like it was internal previously.

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -66,38 +42,74 @@ import BraintreeCore
...
+    ///    - email: Optional. The email used for verification.
+    ///    - exemptionRequested: Optional. If set to true, an exemption to the authentication challenge will be requested.
+    ///    - mobilePhoneNumber: Optional. The mobile phone number used for verification. Only numbers. Remove dashes, parentheses and other characters.
+    ///    - nonce: A nonce to be verified by ThreeDSecure
```
jaxdesmarais, line 70: Do we want to add optional to the docstring here if this is optional?

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -66,38 +42,74 @@ import BraintreeCore
...
+    ///    - renderTypes: Optional: List of all the render types that the device supports for displaying specific challenge user interfaces within the 3D Secure challenge. When using `BTThreeDSecureUIType.both` or `BTThreeDSecureUIType.html`, all `BTThreeDSecureRenderType` options must be set. When using `BTThreeDSecureUIType.native`, all `BTThreeDSecureRenderType` options except `.html` must be set.
+    ///    - requestedExemptionType: Optional. The exemption type to be requested. If an exemption is requested and the exemption's conditions are satisfied, then it will be applied.
+    ///    - shippingMethod: Optional. The shipping method chosen for the transaction
+    ///    - uiType: Optional: Sets all UI types that the device supports for displaying specific challenge user interfaces in the 3D Secure challenge. Defaults to `.both`
+    ///    - v2UICustomization: Optional. UI Customization for 3DS2 challenge views.
+    public init(
+        accountType: BTThreeDSecureAccountType = .unspecified,
+        additionalInformation: BTThreeDSecureAdditionalInformation? = nil,
+        amount: NSDecimalNumber? = 0,
```
jaxdesmarais, line 79: Based on this logic it looks like both amount and the delegate may be required: https://github.com/braintree/braintree_ios/blob/5ecb18dc594019e5bf9292fd67871b8f15812fd1/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift#L62-L74

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
jaxdesmarais: Ah, I see it looks like this is exposed for merchants using `prepareLookup`: https://github.com/braintree/braintree_ios/blob/5ecb18dc594019e5bf9292fd67871b8f15812fd1/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift#L126-L128. We should update this docstring or have 2 inits for this. What do you think?

agedd: ahhh that's a good point! i've updated the docstrings to say, that the `dfReferenceID` is helpful for merchants using the 3DS `prepareLookup` in b6f5382aefb7dc7a3467a891735988bba0a80a0d

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
agedd: ahh yep, good catch! the amount does seem to be required (i've also confirmed this by raising a BT 3D Secure General Inquiry workflow request) - as for the delegate, it seems that it was also previously optional so i wonder if it's look to leave that as is 👀

CHANGELOG.md
```diff
@@ -6,12 +6,14 @@
...
-  * BraintreeLocalPayment
-    * Update `BTLocalPaymentRequest` to make all properties accessible on the initializer only vs via the dot syntax.
+* BraintreeVenmo
+  * Update `BTVenmoRequest` to make all properties accessible on the initializer only vs via the dot syntax.
+* BraintreeSEPADirectDebit
+  * Update `BTSEPADirectDebitRequest` to make all properties accessible on the initializer only vs via the dot syntax.
+* BraintreeLocalPayment
+  * Update `BTLocalPaymentRequest` to make all properties accessible on the initializer only vs via the dot syntax.
+* BraintreeThreeDSecure
+  * Update `BraintreeThreeDSecure` to make all properties accessible on the initializer only vs via the dot syntax.
```
jaxdesmarais, line 16: Lets keep the above spacing. These are breaking changes so we want them to be nested under the breaking changes section. See here for similar formatting with v6: https://github.com/braintree/braintree_ios/blob/main/CHANGELOG.md#600-beta4-2023-06-01

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -66,38 +42,76 @@ import BraintreeCore
...
+    ///    - challengeRequested: Optional. If set to true, an authentication challenge will be forced if possible.
+    ///    - customFields: Optional. Object where each key is the name of a custom field which has been configured in the Control Panel. In the Control Panel you can configure 3D Secure Rules which trigger on certain values.
+    ///    - dataOnlyRequested:  Optional. Indicates whether to use the data only flow. In this flow, frictionless 3DS is ensured for Mastercard cardholders as the card scheme provides a risk score for the issuer to determine whether to approve. If data only is not supported by the processor, a validation error will be raised. Non-Mastercard cardholders will fallback to a normal 3DS flow.
+    ///    - dfReferenceID: Optional. The dfReferenceID for the session, particularly useful for merchants peforming 3DS lookup. Exposed for testing.
```
jaxdesmarais, line 66: Let's remove the testing comment from the public docs.
```suggestion
    ///    - dfReferenceID: Optional. The dfReferenceID for the session, particularly useful for merchants peforming 3DS lookup.
```

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -66,38 +42,76 @@ import BraintreeCore
...
-    // MARK: - Internal Properties
+    // MARK: - Initializer
     
-    /// The dfReferenceID for the session. Exposed for testing.
-    var dfReferenceID: String?
+    /// Creates a `BTThreeDSecureRequest`
+    /// - Parameters:
+    ///    - accountType: Optional. The account type selected by the cardholder. Some cards can be processed using either a credit or debit account and cardholders have the option to choose which account to use.
+    ///    - additionalInformation: Optional. The additional information used for verification.
+    ///    - amount: Required. The amount for the transaction.
```
jaxdesmarais, line 60: If this is required lets put it first and not provide a default. It's also worth considering if we want to continue to use `NSDecimalNumber` or if we can switch to `Double`?

agedd: ahh yep, so i just verified with the public docs and it looks like `amount` is a `string` and required (same with the `nonce` property) - updated the initializer and other places in the code accordingly

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -66,38 +42,76 @@ import BraintreeCore
...
+    ///    - shippingMethod: Optional. The shipping method chosen for the transaction
+    ///    - uiType: Optional: Sets all UI types that the device supports for displaying specific challenge user interfaces in the 3D Secure challenge. Defaults to `.both`
+    ///    - v2UICustomization: Optional. UI Customization for 3DS2 challenge views.
+    public init(
+        amount: String,
+        nonce: String,
+        accountType: BTThreeDSecureAccountType = .unspecified,
+        additionalInformation: BTThreeDSecureAdditionalInformation? = nil,
+        billingAddress: BTThreeDSecurePostalAddress? = nil,
+        _cardAddChallenge: BTThreeDSecureCardAddChallenge = .unspecified,
```
jaxdesmarais, line 82: nit: we can probably just leave this out of the init for now then remove it in the ticket to remove the property completely. If that's more work than it's worth though, feel free to leave and we can hopefully pick up that cleanup ticket soon!

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -66,38 +42,76 @@ import BraintreeCore
...
+    ///    - mobilePhoneNumber: Optional. The mobile phone number used for verification. Only numbers. Remove dashes, parentheses and other characters.
+    ///    - renderTypes: Optional: List of all the render types that the device supports for displaying specific challenge user interfaces within the 3D Secure challenge. When using `BTThreeDSecureUIType.both` or `BTThreeDSecureUIType.html`, all `BTThreeDSecureRenderType` options must be set. When using `BTThreeDSecureUIType.native`, all `BTThreeDSecureRenderType` options except `.html` must be set.
+    ///    - requestedExemptionType: Optional. The exemption type to be requested. If an exemption is requested and the exemption's conditions are satisfied, then it will be applied.
+    ///    - shippingMethod: Optional. The shipping method chosen for the transaction
+    ///    - uiType: Optional: Sets all UI types that the device supports for displaying specific challenge user interfaces in the 3D Secure challenge. Defaults to `.both`
+    ///    - v2UICustomization: Optional. UI Customization for 3DS2 challenge views.
+    public init(
+        amount: String,
```
jaxdesmarais, line 77: 🤔  I am getting an error with the amount as a string in the demo app. Are we sure this is valid? It looks like a `Double` works as expected.

agedd: hmm yea i'm a bit confused on this - i also feel amount as `Double` makes more sense (and this aligns with what is mentioned in our iOS BT Dev Docs) but after asking here  #[bt-service-3dsecure](https://paypal.slack.com/archives/C02743S2L/p1731618309941569) and checking the android [equivalent](https://github.com/braintree/braintree_android/blob/c9d4246eccb6926a0985b5e6d409147d6fb38470/ThreeDSecure/src/main/java/com/braintreepayments/api/threedsecure/ThreeDSecureRequest.kt#L56) it looks like they define amount as a string

jaxdesmarais: Ah, you're right, I must have been doing something weird the first time I tested it as a string on main that resulted in an error. If we want to follow what android/the service channel recommends that's okay with me.

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -59,7 +59,7 @@ import BraintreeCore
                 return
             }
 
-            if request.amount?.decimalValue.isNaN == true || request.amount == nil {
+            if request.amount.isEmpty {
                 NSLog("%@ BTThreeDSecureRequest amount can not be nil or NaN.", BTLogLevelDescription.string(for: .critical))
                 let error = BTThreeDSecureError.configuration("BTThreeDSecureRequest amount can not be nil or NaN.")
```
jaxdesmarais, line 64: Lets update these strings to accurately reflect this block now that amount cannot be nil or NaN. We can instead update to something like "BTThreeDSecureRequest amount cannot be an empty string"

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -103,7 +103,7 @@ import BraintreeCore
             return
         }
 
-        guard request.nonce != nil else {
+        if request.nonce.isEmpty {
             notifyFailure(
                 with: BTThreeDSecureError.configuration("BTThreeDSecureRequest nonce can not be nil."),
```
jaxdesmarais, line 108: Same here, can also update to something like "BTThreeDSecureRequest nonce can not be an empty string"

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@
     * Update `BTLocalPaymentRequest` to make all properties accessible on the initializer only vs via the dot syntax.
   * BraintreeCard 
     * Remove `BTCardRequest`, use `BTCard` directly instead
+  * BraintreeThreeDSecure
+    * Update `BTThreeDSecureRequest` to make all properties accessible on the initializer only vs via the dot syntax.
```
jaxdesmarais, line 19: Let's add an entry as well for any change in types like `BTThreeDSecureRequest.amount` now being a string

PR #1464, opened 2024-11-12
Remove `fetchPaymentMethodNonces`

stechiu:
### Summary of changes

- Removed both the async and completion fetchPaymentMethodNonces methods
- Removed any associated classes that are no longer needed
- We do not need to move this method to the Drop-in on iOS as it is not currently being maintained

Remove all tests that reference fetchPaymentMethodNonces

### Checklist

- [ ] Added a changelog entry

- @stechiu


jaxdesmarais (commented):
I believe with this we should also be able to remove `BTPaymentMethodNonceParser`

jaxdesmarais (approved):

jaxdesmarais (commented):
Let's add a changelog for removing this method and the parser since they are both public. Other than that looks great!

PR #1465, opened 2024-11-14
Update `BTPayPalRequest` with App Switch Properties

agedd:
### Summary of changes

- Update `BTPayPalRequest` with app-switch properties, `enablePayPalAppSwitch` and `userAuthenticationEmail`
- Remove these properties from `BTPayPalVaultRequest`
- Add init to `BTPayPalCheckoutRequest` with these properties
- Tested and verified app switch flow for vaulting works as expected (the checkout flow is not yet expected to work)

### Checklist

- [x] Added a changelog entry

- @agedd

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,9 @@
 ## unreleased
 * BraintreePayPal
   * Add `BTPayPalRequest.userPhoneNumber` optional property
-  * Update `BTPayPalRequest` with App-Switch properties, `userAuthenticationEmail` and `enablePayPalAppSwitch`
-  * Add an initializer that requires both App-Switch properties to `BTPayPalCheckoutRequest` 
-  * Remove these properties and related logic from the `BTPayPalVaultRequest`
-  
+  * Add PayPal App Switch checkout Flow (BETA)
+    * Add `BTPayPalCheckoutRequest(userAuthenticationEmail:enablePayPalAppSwitch:amount:intent:userAction: offerPayLater:currencyCode:requestBillingAgreement:)`
+    * **Note:** This feature is currently in beta and may change or be removed in future releases.
 * BraintreeVenmo
   * Send `url` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## unreleased
 * BraintreePayPal
   * Add `BTPayPalRequest.userPhoneNumber` optional property
+  * Update `BTPayPalRequest` with App-Switch properties, `userAuthenticationEmail` and `enablePayPalAppSwitch`
+  * Add an initializer that requires both App-Switch properties to `BTPayPalCheckoutRequest` 
+  * Remove these properties and related logic from the `BTPayPalVaultRequest`
+  
 * BraintreeVenmo
   * Send `url` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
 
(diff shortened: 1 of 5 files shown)


jaxdesmarais (commented):
This should have the base branch as the feature branch. Lets also try not to have the internal ticket number in the branch name in the future.

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 ## unreleased
 * BraintreePayPal
   * Add `BTPayPalRequest.userPhoneNumber` optional property
+  * Update `BTPayPalRequest` with App-Switch properties, `userAuthenticationEmail` and `enablePayPalAppSwitch`
+  * Add an initializer that requires both App-Switch properties to `BTPayPalCheckoutRequest` 
+  * Remove these properties and related logic from the `BTPayPalVaultRequest`
```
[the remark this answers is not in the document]
agedd: ahhh yep that reads much clearer! addressed in 91287a2b9feb1800ef00f84b4d29c0110cf8a53e

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -98,22 +132,31 @@ import BraintreeCore
     ///   See https://developer.paypal.com/docs/api/reference/currency-codes/ for a list of supported currency codes.
     ///   - requestBillingAgreement: Optional: If set to `true`, this enables the Checkout with Vault flow, where the customer will be prompted to consent to a billing agreement
     ///   during checkout. Defaults to `false`.
+    ///   - userAuthenticationEmail: Optional: User email to initiate a quicker authentication flow in cases where the user has a PayPal Account with the same email.
+    ///   - enablePayPalAppSwitch: Optional: Used to determine if the customer will use the PayPal app switch flow. Defaults to `false`.
```
jaxdesmarais, line 136: I think we only want the `enablePayPalAppSwitch` property on the app switch init, this will ensure merchants pass us the correct properties if opting in for this flow. I believe we do the same in `BTPayPalVaultRequest`. What do you think?

agedd: ahhh gotcha, i see that in the `BTPayPalVaultRequest` we pass in both app-switch properties: https://github.com/braintree/braintree_ios/blob/2f3dd67f32218888f9738d8629a1a89897956eaf/Sources/BraintreePayPal/BTPayPalVaultRequest.swift#L37-L41

just wanted to confirm, did you mean we can exclude the `userAuthenticationEmail` from the app-switch init on `BTPayPalCheckoutRequest` 👀

jaxdesmarais: Sorry I should have been more specific. We _do_ want it on the app switch convenience init. In the init linked above you will see we set `self.enablePayPalAppSwitch = enablePayPalAppSwitch`. But on the other init below that here: https://github.com/braintree/braintree_ios/blob/2f3dd67f32218888f9738d8629a1a89897956eaf/Sources/BraintreePayPal/BTPayPalVaultRequest.swift#L52-L57 we do not have the `enablePayPalAppSwitch` property included. This allows us to ensure that if `enablePayPalAppSwitch = true` the merchant is also passing a currently required `userAuthenticationEmail`.

Let me know if that doesn't make sense though and we can always grab some time to pair on it!

Sources/BraintreePayPal/BTPayPalRequest.swift
```diff
@@ -51,62 +51,71 @@ import BraintreeCore
...
     @_documentation(visibility: private)
     public var paymentType: BTPayPalPaymentType
     
     /// Optional: A user's phone number to initiate a quicker authentication flow in the scenario where the user has a PayPal account
     /// identified with the same phone number.
     public var userPhoneNumber: BTPayPalPhoneNumber?
-
+    
+    /// Optional: User email to initiate a quicker authentication flow in cases where the user has a PayPal Account with the same email.
+    public var userAuthenticationEmail: String?
```
richherrera, line 105: can we consider using `let` instead of `var`?

richherrera (approved):
LGTM!!!

scannillo (approved):

PR #1468, opened 2024-11-20
[V7] Add `Encodable` protocol for Local Payments.

richherrera:
### Summary of changes

- Add encodable types for `post` bodies used in the Local Payment flow.
- Local Payment flow has been tested in the Demo app to ensure the dictionaries are still constructed as expected.

### Checklist

### Authors

- @richherrera

### DIFF
--- a/Sources/BraintreeLocalPayment/BTLocalPaymentRequest.swift
+++ b/Sources/BraintreeLocalPayment/BTLocalPaymentRequest.swift
@@ -9,7 +9,12 @@ import BraintreeDataCollector
 #endif
 
 /// Used to initialize a local payment flow
-@objcMembers public class BTLocalPaymentRequest: NSObject {
+/// The POST body for v1/local_payments/create
+@objcMembers public class BTLocalPaymentRequest: NSObject, Encodable {
+    
+    // MARK: - Public Properties
+    
+    public weak var localPaymentFlowDelegate: BTLocalPaymentRequestDelegate?
     
     // MARK: - Internal Properties
     
@@ -26,11 +31,19 @@ import BraintreeDataCollector
     let phone: String?
     let isShippingAddressRequired: Bool
     let bic: String?
-    
-    public weak var localPaymentFlowDelegate: BTLocalPaymentRequestDelegate?
 
     var paymentID: String?
     var correlationID: String?
+    
+    // MARK: - Private Properties
+    
+    private let intent: String
+    private let returnURL: String
+    private let cancelURL: String
+    
+    private lazy var experienceProfile: ExperienceProfile = {
+        .init(noShipping: !isShippingAddressRequired, brandName: displayName)
+    }()
     
     // MARK: - Initializer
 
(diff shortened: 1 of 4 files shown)


Sources/BraintreeLocalPayment/BTLocalPaymentRequest.swift
```diff
@@ -80,5 +93,75 @@ import BraintreeDataCollector
...
+        self.cancelURL = BTCoreConstants.callbackURLScheme + "://x-callback-url/braintree/local-payment/cancel"
+    }
+    
+    enum CodingKeys: String, CodingKey {
+        case paymentType = "funding_source"
+        case amount
+        case currencyCode = "currency_iso_code"
+        case paymentTypeCountryCode = "payment_type_country_code"
+        case merchantAccountID = "merchant_account_id"
+        case email = "payer_email"
```
richherrera, line 107: After reviewing Android, I noticed we’re not sending the same parameters, such as `email`, `phone`, `givenName`, `surname`, `bic`, and `merchantAccountId`. I verified with the LPMS team the keys needed to send these parameters. It seems the gateway supports both `camel case` and `snake case`, web and Android use `camel case`.

agedd (approved):
lgtm! 🚀

Sources/BraintreeLocalPayment/BTLocalPaymentRequest.swift
```diff
@@ -9,7 +9,12 @@ import BraintreeDataCollector
 #endif
 
 /// Used to initialize a local payment flow
-@objcMembers public class BTLocalPaymentRequest: NSObject {
+/// The POST body for v1/local_payments/create
+@objcMembers public class BTLocalPaymentRequest: NSObject, Encodable {
```
jaxdesmarais, line 13: Take it or leave it but we could also consider separating the encodable portion out into a new file, we've done that in most places where we've made this change, like here for example: https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreeSEPADirectDebit/SEPADebitAccountsRequest.swift

Sources/BraintreeLocalPayment/Models/LocalPaymentPOSTBody.swift
```diff
@@ -0,0 +1,119 @@
...
+        try container.encodeIfPresent(experienceProfile, forKey: .experienceProfile)
+          
+        if let address {
+            try container.encodeIfPresent(address.streetAddress, forKey: .streetAddress)
+            try container.encodeIfPresent(address.extendedAddress, forKey: .extendedAddress)
+            try container.encodeIfPresent(address.locality, forKey: .locality)
+            try container.encodeIfPresent(address.countryCodeAlpha2, forKey: .countryCodeAlpha2)
+            try container.encodeIfPresent(address.postalCode, forKey: .postalCode)
+            try container.encodeIfPresent(address.region, forKey: .region)
+        }
```
scannillo, line 102: Take it or leave it - instead of needing a custom Encoder implementation, we could have the `init` take in a `BTPostalAddress` param, but then the private properties on this struct are the actual encodable types.

Roughly something like:
```
private let amount: String
private var streetAddress: String?
private var extendedAddress: String?

init(amount: String, ... the other stuff, address: BTPostalAddress?) {
    self.amount = amount 
    self.streetAddress = address.streetAddress
    self.extendedAddress = address.extendedAddress
}
```

Then we wouldn't need to override `encode()`.

Which is more readable? I go back and forth sometimes with the custom encoder being more or less explicit.

richherrera: Updated: [3e00b0e](https://github.com/braintree/braintree_ios/pull/1468/commits/3e00b0e96d9cd3d3ca2e5601a1bc8cf79acb3d31)

Sources/BraintreeLocalPayment/BTLocalPaymentRequest.swift
richherrera: Sure, updated: [3e00b0e](https://github.com/braintree/braintree_ios/pull/1468/commits/3e00b0e96d9cd3d3ca2e5601a1bc8cf79acb3d31) what do you think??

jaxdesmarais (approved):
:shipit:

PR #1479, opened 2024-12-04
Add Shopper Session ID to `BTShopperInsightsClient` [RP1]

stechiu:
### Summary of changes

- Added `shopperSessionID` property to `BTShopperInsightsClient`
- Passed `session_id` to FPTI

### Checklist

- [x] Added a changelog entry

- @stechiu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * BraintreeVenmo
   * Send `url` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
 * BraintreeShopperInsights (BETA)
-  * Send `sessionID` to PayPal's analytics service (FPTI)
+  * Add `shopperSessionID` to `BTShopperInsightsClient` initializer
   * Add `isPayPalAppInstalled()` and/or `isVenmoAppInstalled()`
 
 ## 6.24.0 (2024-10-15)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
   * Add `BTPayPalRequest.userPhoneNumber` optional property
 * BraintreeVenmo
   * Send `url` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
+* BraintreeShopperInsights (BETA)
+  * Send `sessionID` to PayPal's analytics service (FPTI)
 
 ## 6.24.0 (2024-10-15)
 * BraintreePayPal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   * Send `url` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
 * BraintreeShopperInsights (BETA)
   * Send `sessionID` to PayPal's analytics service (FPTI)
+  * Add `isPayPalAppInstalled()` and/or `isVenmoAppInstalled()`
 
 ## 6.24.0 (2024-10-15)
 * BraintreePayPal
--- a/Demo/Application/Features/ShopperInsightsViewController.swift
+++ b/Demo/Application/Features/ShopperInsightsViewController.swift
@@ -6,7 +6,7 @@ import BraintreeShopperInsights
 
 class ShopperInsightsViewController: PaymentButtonBaseViewController {
     
-    lazy var shopperInsightsClient = BTShopperInsightsClient(apiClient: apiClient)
+    lazy var shopperInsightsClient = BTShopperInsightsClient(apiClient: apiClient, sessionID: "123456")
     lazy var payPalClient = BTPayPalClient(apiClient: apiClient)
     lazy var venmoClient = BTVenmoClient(apiClient: apiClient)
     
@@ -36,19 +36,11 @@ class ShopperInsightsViewController: PaymentButtonBaseViewController {
         view.textField.text = "4082321001"
         return view
     }()
-
-    lazy var sessionIDView: TextFieldWithLabel = {
-        let view = TextFieldWithLabel()
-        view.label.text = "Shopper Session ID"
-        view.textField.placeholder = "Session ID"
-        view.textField.text = "123456"
-        return view
-    }()
     
     lazy var shopperInsightsButton = createButton(title: "Fetch Shopper Insights", action: #selector(shopperInsightsButtonTapped))
     
     lazy var shopperInsightsInputView: UIStackView = {
-        let stackView = UIStackView(arrangedSubviews: [emailView, countryCodeView, nationalNumberView, sessionIDView])
+        let stackView = UIStackView(arrangedSubviews: [emailView, countryCodeView, nationalNumberView])
         stackView.axis = .vertical
         stackView.spacing = 10
         stackView.distribution = .fillEqually
--- a/Demo/Application/Features/ShopperInsightsViewController.swift
+++ b/Demo/Application/Features/ShopperInsightsViewController.swift
@@ -6,7 +6,7 @@ import BraintreeShopperInsights
 
 class ShopperInsightsViewController: PaymentButtonBaseViewController {
     
-    lazy var shopperInsightsClient = BTShopperInsightsClient(apiClient: apiClient, sessionID: "123456")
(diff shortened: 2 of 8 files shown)


stechiu:
I'm unsure if `sessionID` is being passed in the right methods and if the unit tests I wrote covers the newly added code

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -110,6 +113,7 @@ struct FPTIBatchData: Codable {
             case requestStartTime = "request_start_time"
             case timestamp = "t"
             case tenantName = "tenant_name"
+            case sessionID = "session_id"
```
sarahkoop, line 116: I don't think we should need a new property for this since a "session_id" tag already exists in the `Metadata` - we should be able to pass the session ID from `ShopperInsightsClient` to that existing field instead. @jaxdesmarais correct me if I'm misunderstanding the requirements

Demo/Application/Features/ShopperInsightsViewController.swift
```diff
@@ -36,11 +36,19 @@ class ShopperInsightsViewController: PaymentButtonBaseViewController {
         view.textField.text = "4082321001"
         return view
     }()
+
+    lazy var sessionIDView: TextFieldWithLabel = {
```
sarahkoop, line 40: In my opinion we don't need a text field input for this value because a user would never be inputting this data, it would come from the merchants server. For now we can probably hard code a value and then update the demo app client to include the value from the server when it's available

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
jaxdesmarais: You're correct, we should be setting the existing session ID in Metadata as this value vs creating a new property.

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -17,6 +17,9 @@ public class BTShopperInsightsClient {
     // MARK: - Private Properties
     
     private let apiClient: BTAPIClient
+
+    /// This value should be the shopper session ID returned from your server SDK request
+    private let shopperSessionID: String? = ""
```
jaxdesmarais, line 22: We instead want to add this to the initializer for this class as a required property so merchants will set it when initializing the shopper insights client. Then in our analytics we can set this property when making the analytics request.

Demo/Application/Features/ShopperInsightsViewController.swift
stechiu: Do you think a toggle would be better? Or don't add anything to the UI and hard code into the demo app instead?

sarahkoop: Hmmm we could add a toggle, or also might not be necessary because the future iteration of this will remove the existing non-sessionID pattern so we will always be passing the session ID from the server (so no toggle will be needed). I think I lean just hard code it to always pass a session ID now (no toggle), but open to other's thoughts!

stechiu: ok! I thought it would be the Merchant entering in a session ID but if it comes from the server, then hard coding makes sense until server SDK updates on their end

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
stechiu: thanks, don't know how I didn't see `session_id` in the metadata at all, otherwise I wouldn't have added it to events

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * BraintreeVenmo
   * Send `url` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
 * BraintreeShopperInsights (BETA)
+  * Send `sessionID` to PayPal's analytics service (FPTI)
```
jaxdesmarais, line 9:
```suggestion
  *  Add `shopperSessionID` to `BTShopperInsightsClient` initializer
```

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -178,6 +177,7 @@ struct FPTIBatchData: Codable {
 
         let platform = "iOS"
 
+        /// This value should be the shopper session ID returned from your server SDK request
```
jaxdesmarais, line 180: Since we also use this for other modules that aren't shopper insights maybe something like:
```suggestion
        /// Either a randomly generated session ID or the shopper session ID passed in by a merchant
```

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -53,7 +53,6 @@ struct FPTIBatchData: Codable {
         /// Used for linking events from the client to server side request
         /// This value will be PayPal Order ID, Payment Token, EC token, Billing Agreement, or Venmo Context ID depending on the flow
         let payPalContextID: String?
-
```
jaxdesmarais, line 56: Lets keep this linebreak

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -17,12 +17,20 @@ public class BTShopperInsightsClient {
     // MARK: - Private Properties
     
     private let apiClient: BTAPIClient
+    private var shopperSessionID: String?
     
     /// Creates a `BTShopperInsightsClient`
-    /// - Parameter apiClient: A `BTAPIClient` instance.
+    /// - Parameters:
+    ///     - `apiClient`: A `BTAPIClient` instance.
+    ///     - `shopperSessionID`: This value should be the shopper session ID returned from your server SDK request
```
jaxdesmarais, line 25:
```suggestion
    ///     - apiClient: A `BTAPIClient` instance.
    ///     - shopperSessionID: This value should be the shopper session ID returned from your server SDK request
```

UnitTests/BraintreeShopperInsightsTests/BTShopperInsightsClient_Tests.swift
```diff
@@ -223,11 +223,16 @@ class BTShopperInsightsClient_Tests: XCTestCase {
         XCTAssertEqual(mockAPIClient.postedAnalyticsEvents.first, "shopper-insights:venmo-presented")
     }
     
-    func testSendVenmoSelectedEvent_sendsAnalytic() {
+    func testSendVenmoSelectedEvent() {
```
jaxdesmarais, line 226: I think we can revert this test name change

stechiu: Oops didn't mean to change that

UnitTests/BraintreeShopperInsightsTests/BTShopperInsightsClient_Tests.swift
```diff
@@ -228,6 +228,11 @@ class BTShopperInsightsClient_Tests: XCTestCase {
         XCTAssertEqual(mockAPIClient.postedAnalyticsEvents.first, "shopper-insights:venmo-selected")
     }
 
+    func testShopperInsightsClient_withSessionID_setSessionIdInMetadata() {
```
jaxdesmarais, line 231:
```suggestion
    func testShopperInsightsClient_withSessionID_setSessionIDInMetadata() {
```

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -17,12 +17,20 @@ public class BTShopperInsightsClient {
     // MARK: - Private Properties
     
     private let apiClient: BTAPIClient
+    private var shopperSessionID: String?
```
jaxdesmarais, line 20:
```suggestion
    private let shopperSessionID: String?
```

jaxdesmarais (approved):
Couple of very minor cleanup comments but overall great work!

sarahkoop (approved):

PR #1480, opened 2024-12-05
Add `BTThreeDSecureRequest.requestorAppURL`

jaxdesmarais:
### Summary of changes

- Add `BTThreeDSecureRequest.requestorAppURL` - this is a new property required by Mastercard for 3DS for out of band transactions. There was an internal request from a large merchant for exposing this field.
- There is currently not a good way to test values set on `cardinalConfiguration` so those are not included. I did test setting this value locally to ensure this value is being set on `cardinalConfiguration` when passed.

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
@@ -93,6 +93,10 @@ import BraintreeCore
     /// When using `BTThreeDSecureUIType.native`, all `BTThreeDSecureRenderType` options except `.html` must be set.
     public var renderTypes: [BTThreeDSecureRenderType]?
 
+    /// Optional. Three DS Requester APP URL Merchant app declaring their URL within the CReq message
+    /// so that the Authentication app can call the Merchant app after OOB authentication has occurred. 
+    public var threeDSRequestorAppURL: String?
+
     /// A delegate for receiving information about the ThreeDSecure payment flow.
     public weak var threeDSecureRequestDelegate: BTThreeDSecureRequestDelegate?
     
--- a/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
+++ b/Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
@@ -94,7 +94,7 @@ import BraintreeCore
     public var renderTypes: [BTThreeDSecureRenderType]?
 
     /// Optional. Three DS Requester APP URL Merchant app declaring their URL within the CReq message
-    /// so that the Authentication app can call the Merchant app after OOB authentication has occurred. 
+    /// so that the Authentication app can call the Merchant app after out of band authentication has occurred. 
     public var requestorAppURL: String?
 
     /// A delegate for receiving information about the ThreeDSecure payment flow.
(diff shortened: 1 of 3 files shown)


Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -93,6 +93,10 @@ import BraintreeCore
     /// When using `BTThreeDSecureUIType.native`, all `BTThreeDSecureRenderType` options except `.html` must be set.
     public var renderTypes: [BTThreeDSecureRenderType]?
 
+    /// Optional. Three DS Requester APP URL Merchant app declaring their URL within the CReq message
+    /// so that the Authentication app can call the Merchant app after OOB authentication has occurred.
```
jaxdesmarais, line 97: I pulled these docstrings from the cardinal property and docs but it's not super detailed. Open to suggestions but my understanding is that this property is for out of band cardinal 3DS attempts for some users.

warmkesselj: I think the acronyms are confusing to me. OOB is "out of band"? Do you happen to know what a sample value for requestorAppURL would look like? e.g. Lululemon.com?

warmkesselj: ```/// Optional. A value to enable Out-Of-Band authentication used by the Authentication app to message the Merchant app once the secondary authentication has competed.```? Not sure how to fit CReq in there. Basing my understanding of Out-of-band on this [link](https://www.techtarget.com/searchsecurity/definition/out-of-band-authentication)

warmkesselj: Sure~

jaxdesmarais, line 97:
```suggestion
    /// so that the Authentication app can call the Merchant app after out of band authentication has occurred. 
```

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
```diff
@@ -93,6 +93,10 @@ import BraintreeCore
     /// When using `BTThreeDSecureUIType.native`, all `BTThreeDSecureRenderType` options except `.html` must be set.
     public var renderTypes: [BTThreeDSecureRenderType]?
 
+    /// Optional. Three DS Requester APP URL Merchant app declaring their URL within the CReq message
+    /// so that the Authentication app can call the Merchant app after out of band authentication has occurred. 
+    public var requestorAppURL: String?
```
scannillo, line 98: Does this function like an app switch universal or deep link return? Or is just a web-hook/callback to that URL?

richherrera (approved):

Sources/BraintreeThreeDSecure/BTThreeDSecureRequest.swift
jaxdesmarais: It's a bit hard to tell, this docstring is pulled directly from the Cardinal docs and they don't make it super clear. It seems like maybe? Apparently it's part of a new Mastercard requirement for merchants that use `prepareLookup`/`initializeChallenge`.

PR #1481, opened 2024-12-06
Shopper insights rp1 feature include session

warmkesselj:
### Summary of changes

Add public property `shopperSessionId` to BTPaypalRequest. Value should be passed to both the create_payment_resource and setup_billing_agreement endpoints. Demo app should be updated enable/disable passing this value. 

### Checklist

- [x] Added a changelog entry

### Authors
@stechiu 
@warmkesselj 
@jwarmkessel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## unreleased
 * BraintreePayPal
   * Add `BTPayPalRequest.userPhoneNumber` optional property
+  * Add `shopperSessionInsight`
 * BraintreeVenmo
   * Send `url` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
 * BraintreeShopperInsights (BETA)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## unreleased
 * BraintreePayPal
   * Add `BTPayPalRequest.userPhoneNumber` optional property
-  * Add `shopperSessionInsight`
+  * Add `shopperSessionID` to `BTPayPalCheckoutRequest` and `BTPayPalVaultRequest`
 * BraintreeVenmo
   * Send `url` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
 * BraintreeShopperInsights (BETA)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
   * Add `BTPayPalRequest.userPhoneNumber` optional property
 * BraintreeVenmo
   * Send `url` in `event_params` for App Switch events to PayPal's analytics service (FPTI)
+* BraintreeShopperInsights (BETA)
+  * Add `isPayPalAppInstalled()` and/or `isVenmoAppInstalled()`
 
 ## 6.24.0 (2024-10-15)
 * BraintreePayPal
--- a/Demo/Application/Features/PayPalNativeCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalNativeCheckoutViewController.swift
@@ -47,7 +47,7 @@ class PayPalNativeCheckoutViewController: PaymentButtonBaseViewController {
 		sender.setTitle("Processing...", for: .disabled)
 		sender.isEnabled = false
 
-		let request = BTPayPalNativeVaultRequest()   
+		let request = BTPayPalNativeVaultRequest()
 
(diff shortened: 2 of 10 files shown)


Demo/Application/Features/PayPalNativeCheckoutViewController.swift
```diff
@@ -47,7 +47,7 @@ class PayPalNativeCheckoutViewController: PaymentButtonBaseViewController {
 		sender.setTitle("Processing...", for: .disabled)
 		sender.isEnabled = false
 
-		let request = BTPayPalNativeVaultRequest()
+		let request = BTPayPalNativeVaultRequest()
```
jaxdesmarais, line 50:
```suggestion
		let request = BTPayPalNativeVaultRequest()
```

Demo/Application/Features/ShopperInsightsViewController.swift
```diff
@@ -13,6 +13,8 @@ class ShopperInsightsViewController: PaymentButtonBaseViewController {
     lazy var payPalVaultButton = createButton(title: "PayPal Vault", action: #selector(payPalVaultButtonTapped))
     lazy var venmoButton = createButton(title: "Venmo", action: #selector(venmoButtonTapped))
     
+    private var shopperSessionId = "test-shopper-session-id"
```
jaxdesmarais, line 16:
```suggestion
    private var shopperSessionID = "test-shopper-session-id"
```

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## unreleased
 * BraintreePayPal
   * Add `BTPayPalRequest.userPhoneNumber` optional property
+  * Add `shopperSessionInsight`
```
```suggestion
  * Add `shopperSessionID` to `BTPayPalCheckoutRequest` and `BTPayPalVaultRequest`
```

Sources/BraintreePayPal/BTPayPalRequest.swift
```diff
@@ -179,6 +184,10 @@ import BraintreeCore
         if let userPhoneNumberDict = try? userPhoneNumber?.toDictionary() {
             parameters["phone_number"] = userPhoneNumberDict
         }
+        
+        if let shopperSessionID {
+            parameters["shopper_session_id"] = shopperSessionID
```
jaxdesmarais, line 189: @scannillo @sarahkoop do we also need to be setting the `apiClient.metadata.sessionID` as this value as well in the `BTPayPalClient`? Was just thinking if a merchant uses 2 different `BTAPIClient`s for those flows the PayPal flow I believe would randomly generate a new session ID vs reusing the session ID from the `BTShopperInsightsClient`?

sarahkoop: Ah yes I think we will want to update to set the sessionID for PayPal too

sarahkoop: I think it would be easier to tackle that in separate tickets though (will be needed for Android too)

warmkesselj: Do we want to update in this same PR or create another ticket to update PayPal?

sarahkoop: I would vote a separate ticket and separate PRs

richherrera (approved):

Demo/Application/Features/ShopperInsightsViewController.swift
```diff
@@ -13,6 +13,8 @@ class ShopperInsightsViewController: PaymentButtonBaseViewController {
     lazy var payPalVaultButton = createButton(title: "PayPal Vault", action: #selector(payPalVaultButtonTapped))
     lazy var venmoButton = createButton(title: "Venmo", action: #selector(venmoButtonTapped))
     
+    private var shopperSessionID = "test-shopper-session-id"
```
jaxdesmarais, line 16: Can we also pass this in the shopper insights client in this class?
```suggestion
    private let shopperSessionID = "test-shopper-session-id"
```

jaxdesmarais (approved):
Just one minor cleanup comment. 👏

Sources/BraintreePayPal/BTPayPalRequest.swift
scannillo: What do y'all think about just using a new tag for `shopper_session_id`? The existing `session_id` is very SDK specific and IMO will be confusing to start mixing with shopper session ID for some flows, but not all

PR #1485, opened 2024-12-13
ShopperInsights - send `shopper_session_id` to FPTI

scannillo:
### Summary 

- This PR addresses some previous confusion around requirements for how to log the `shopperSessionID`

### Changes
- Send a new FPTI param with the tag `shopper_session_id`
    - Send this anywhere a shopperSessionID is accessible (in the PayPalClient & ShopperInsightsClient)
- Revert change to of overriding the SDK sessionID with the shopperSessionID
    - Keeps `/v2/payments/find-eligible-methods` API call as it was prior
    - Note: This API call will be swapped in the future to one that actually requires a shopperSessionID. Will get tackled in RP3. This is when we can require the shopperSessionID at the init level.

This feature has been a bit confusing, so please let me know if there are questions!

### Checklist
- ~Added a changelog entry~
- ~Tested and confirmed payment flows affected by this change are functioning as expected~

### Authors
@scannillo


jaxdesmarais (approved):

scannillo:
Verified this new tag is making it to FPTI:

Screenshot 2024-12-16 at 10 25 10 AM

warmkesselj (approved):

PR #1488, opened 2024-12-16
Create encodable for credit cards post

warmkesselj:
### Summary of changes

Move `/credit_cards` POST to Encodable.

### Checklist

- [X] Added a changelog entry

### Authors
@warmkesselj 
@jwarmkessel


Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,276 @@
...
+            
+                var merchantAccountId: String?
+                
+                init() {
+                    self.merchantAccountId = nil
+                }
+                
+                init(merchantAccountId: String) {
+                    self.merchantAccountId = merchantAccountId
+                }
```
jaxdesmarais, line 208: I wonder if providing a default here would clean this up. It looks like previously we did `variables["authenticationInsightInput"] = [:]` if the merchant account ID was nil.

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,133 @@
+// swiftlint:disable all
+import Foundation
+
+struct CreditCardPOSTBody: Encodable {
+    var authenticationInsight: Bool?
+    var merchantAccountId: String?
```
jaxdesmarais, line 6:
```suggestion
    var merchantAccountID: String?
```

UnitTests/BraintreeCardTests/BTCardClient_Tests.swift
```diff
@@ -530,7 +530,6 @@ class BTCardClient_Tests: XCTestCase {
                 return
             }
             lastPostParameters.removeValue(forKey: "clientSdkMetadata")
-            XCTAssertEqual(lastPostParameters as NSObject, card.graphQLParameters() as NSObject)
```
jaxdesmarais, line 533: Did we mean to remove this or just replace with the new method name?

UnitTests/BraintreeCardTests/BTCard_Tests.swift
```diff
@@ -4,12 +4,12 @@ import XCTest
 class BTCard_Tests: XCTestCase {
 
     func testInitialization_withoutParameters() {
-        let card = BTCard(
-            number: "4111111111111111",
-            expirationMonth: "12",
-            expirationYear: "2038",
-            cvv: "123"
-        )
```
jaxdesmarais, line 12: We will want to revert this vs changing back to the dot syntax.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
warmkesselj: Yup, I was aware of an empty customer being sent. I'll have to check this again to be sure but I think this gives us the desired outcome as when `variables["authenticationInsightInput"] = [:]` was being set. 
```
{
  "requestedExemptionType" : "low_value",
  "requestedThreeDSecureVersion" : "2",
  "accountType" : "credit",
  "additionalInfo" : {
    "billingState" : "CA",
    "mobilePhoneNumber" : null,
    "billingLine2" : "#5",
    "billingCity" : "Oakland",
    "billingGivenName" : "Jill",
    "email" : "test@example.com",
    "billingPostalCode" : "12345",
    "billingSurname" : "Doe",
    "billingCountryCode" : "US",
    "shippingMethod" : "01",
    "billingLine1" : "555 Smith St.",
    "billingPhoneNumber" : "8101234567"
  },
  "dfReferenceId" : "0_e6783c0d-2f58-42c2-9d63-3945502e2201",
  "dataOnlyRequested" : false,
  "challengeRequested" : false,
  "amount" : "10.32",
  "customer" : {

  },
  "exemptionRequested" : false
}
```

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,276 @@
...
+    var variables: Variables
+    var query: String
+    var operationName: String
+    
+    init(card: BTCard,
+         shouldValidate: Bool,
+         authenticationInsightRequested: Bool,
+         merchantAccountID: String?
+    ) {
+        var cardBody = CreditCardGraphQLBody.Variables.Input.CreditCard(
```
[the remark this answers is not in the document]
warmkesselj: I'm okay with changing this. I figured it might be easier to see the nested structure of the body.

Sources/BraintreeCard/BTCard.swift
```diff
@@ -116,181 +120,29 @@ import Foundation
...
-        }
-        
-        if !expirationMonth.isEmpty {
-            cardDictionary[isGraphQL ? "expirationMonth" : "expiration_month"] = expirationMonth
-        }
+        let meta = CreditCardPOSTBody.Meta(
+            integration: apiClient.metadata.integration.stringValue,
+            source: apiClient.metadata.source.stringValue,
+            sessionId: apiClient.metadata.sessionID
+        )
```
richherrera, line 130: i think we can remove this, since we are not using it

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,276 @@
+// swiftlint:disable all
+import Foundation
+
+struct CreditCardGraphQLBody: Encodable {
+
+    var variables: Variables
```
richherrera, line 6: can we add a mark ? `// MARK: - Internal Properties`
I like to start with private properties and then change them to internal if needed

Sources/BraintreeCard/BTCard.swift
```diff
@@ -116,181 +120,29 @@ import Foundation
...
         
-        if !expirationYear.isEmpty {
-            cardDictionary[isGraphQL ? "expirationYear" : "expiration_year"] = expirationYear
+        if authenticationInsightRequested {
+            creditCardBody.authenticationInsight = true
+            creditCardBody.merchantAccountId = merchantAccountID
```
richherrera, line 134: I'd like to know what you think. Maybe we could pass `authenticationInsightRequested` and `merchantAccountId` in the init of `CreditCardPOSTBody`, and handle this logic within the structure itself.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,276 @@
...
+    var query: String
+    var operationName: String
+    
+    init(card: BTCard,
```
richherrera, line 10: can we ad a mark? `// MARK: - Initializer`

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,133 @@
...
+        case creditCard = "credit_card"
+    }
+
+    init(
+        card: BTCard,
+        authenticationInsight: Bool? = nil,
+        merchantAccountId: String?  = nil,
+        meta: Meta? = nil,
+        usesGraphQL: Bool = false
+    ) {
```
richherrera, line 25:
```suggestion
    init(
        card: BTCard,
        authenticationInsight: Bool,
        merchantAccountId: String
    ) {
```

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,133 @@
+// swiftlint:disable all
+import Foundation
+
+struct CreditCardPOSTBody: Encodable {
+    var authenticationInsight: Bool?
+    var merchantAccountId: String?
+    var meta: Meta?
+    let creditCard: CreditCard?
+    
+    private var usesGraphQL: Bool
```
richherrera, line 10: Since this is the entity we'll use for POST APIs, I think we can omit this boolean

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,133 @@
...
+        self.usesGraphQL = usesGraphQL
+        
+    }
+
+    struct Meta: Encodable {
+        var integration: String
+        var source: String
+        var sessionId: String
+
+        init(integration: String, source: String, sessionId: String) {
```
richherrera, line 39: i think we can just pass `BTClientMetadata` as a parameter

```suggestion
        init(metadata: BTMetadata) {
```

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,133 @@
...
+        card: BTCard,
+        authenticationInsight: Bool? = nil,
+        merchantAccountId: String?  = nil,
+        meta: Meta? = nil,
+        usesGraphQL: Bool = false
+    ) {
+        self.creditCard = CreditCard(card: card)
+        self.authenticationInsight = authenticationInsight
+        self.merchantAccountId = merchantAccountId
+        self.meta = meta
```
richherrera, line 29: what if we instantiate this in the same way we are doing with `creditCard` on line 26?

Sources/BraintreeCard/BTCard.swift
```diff
@@ -1,296 +1,114 @@
...
+        return creditCardBody
+    }
 
-        mutation.append(
-            """
-              }
-            }
-            """
+    func parameters() -> CreditCardGraphQLBody {
+        return CreditCardGraphQLBody(
```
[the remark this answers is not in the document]
warmkesselj: I kind of like this function name. Do you think this must be updated?

UnitTests/BraintreeCardTests/BTCardClient_Tests.swift
warmkesselj: Yeah, intentionally removed.

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,134 @@
...
+    enum CodingKeys: String, CodingKey {
+        case authenticationInsight
+        case meta = "_meta"
+        case merchantAccountID = "merchantAccountId"
+        case creditCard = "credit_card"
+    }
+
+    init(
+        card: BTCard,
+        metaData: BTClientMetadata
```
jaxdesmarais, line 23: nit since this is the casing we use elsewhere:
```suggestion
        metadata: BTClientMetadata
```

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,134 @@
...
+        
+        self.meta = Meta(metaData: metaData)
+    }
+
+    struct Meta: Encodable {
+        var integration: String
+        var source: String
+        var sessionId: String
+
+        init(metaData: BTClientMetadata) {
```
jaxdesmarais, line 40:
```suggestion
        init(metadata: BTClientMetadata) {
```

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,134 @@
...
+            self.merchantAccountID = card.merchantAccountID
+        }
+        
+        self.meta = Meta(metaData: metaData)
+    }
+
+    struct Meta: Encodable {
+        var integration: String
+        var source: String
+        var sessionId: String
```
jaxdesmarais, line 38: nit to match the casing we use in the rest of the SDK:
```suggestion
        var sessionID: String
```

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,134 @@
...
+        let number: String?
+        let expirationMonth: String?
+        let cvv: String?
+        let options: Options?
+        let expirationYear: String?
+        let cardHolderName: String?
+
+        init(
+            card: BTCard
+        ) {
```
jaxdesmarais, line 59:
```suggestion
        init(card: BTCard) {
```

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,134 @@
...
+            let locality: String?
+            let region: String?
+            let countryName: String?
+            let countryCodeAlpha2: String?
+            let countryCodeAlpha3: String?
+            let countryCodeNumeric: String?
+            
+            init(
+                card: BTCard
+            ) {
```
jaxdesmarais, line 95:
```suggestion
            init(card: BTCard) {
```

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
```diff
@@ -40,6 +40,7 @@ struct PayPalCheckoutPOSTBody: Encodable {
     
     // MARK: - Initializer
     
+    // swiftlint:disable cyclomatic_complexity
```
jaxdesmarais, line 43: Did we have a warning here? Was curious since we didn't update this in this PR but we may have missed it elsewhere.

UnitTests/BraintreeCardTests/BTCardClient_Tests.swift
jaxdesmarais: 🤔 It seems like we would now be missing the part where we are testing the codable object though, right? But maybe I am misunderstanding this part of the test.

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
warmkesselj: Yeah, that's the warning I saw.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,223 @@
+// swiftlint:disable all
+import Foundation
+
+struct CreditCardGraphQLBody: Encodable {
+
+    // MARK: - Internal Properties
+    
+    var variables: Variables
+    var query: String
+    var operationName: String
```
scannillo, line 10: Slightly outside the scope of this PR, but do we want to make the things required for GraphQL calls part of a protocol?

This is the 1st GraphQL call in the SDK that moved to Encodable, so we can set the pattern for the rest of them.

warmkesselj: I’m open to other patterns! Though my understand is that the operationName is sent as part of the body for the network call so I'm not sure how we would separate this. If you're referencing the inclusion of:
```
static func cardTokenizationGraphQLMutation(authenticationInsightRequested: Bool) -> String {
        var mutation = "mutation TokenizeCreditCard($input: TokenizeCreditCardInput!"

        if authenticationInsightRequested {
            mutation.append(", $authenticationInsightInput: AuthenticationInsightInput!")
        }

        // swiftlint:disable indentation_width
        mutation.append(
            """
            ) {
              tokenizeCreditCard(input: $input) {
                token
                creditCard {
                  brand
                  expirationMonth
                  expirationYear
                  cardholderName
                  last4
                  bin
```

...then in the spirit of encapsulating the logic for setting values on the Encodable object, it made sense to me to move that logic out of BTCard and into the Encodable logic. However, if we wanted more flexibility and maintainability we could move this function to BTCard.

jaxdesmarais: I think we can punt on this for now. I made internal ticket DTMOBILES-1410 to establish a protocol pattern for Graph QL requests.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,223 @@
+// swiftlint:disable all
```
jaxdesmarais, line 1: We should remove this, we don't want to disable all of our linter rules for any files

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,136 @@
+// swiftlint:disable all
```
jaxdesmarais, line 1: We should remove this, we don't want to disable all of our linter rules for any files

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,223 @@
...
+                var expirationMonth: String?
+                var cvv: String?
+                var options: Options?
+                var expirationYear: String?
+                var cardholderName: String?
+
+                init(card: BTCard) {                    
+                    if card.firstName != nil {
+                        self.billingAddress = BillingAddress(card: card)
+                    }
```
richherrera, line 55: Why do we evaluate whether the firstName is not null to create the BillingAddress? I believe that BillingAddress is optional and does not depend on the firstName. What do you think?
```suggestion
                    self.billingAddress = BillingAddress(card: card)
```

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,223 @@
...
+                }
+            }
+            
+            struct AuthenticationInsightInput: Encodable {
+            
+                var merchantAccountID: String?
+                
+                init(card: BTCard) {
+                    
+                    guard card.authenticationInsightRequested else {
```
richherrera, line 144: Can we check if this validation is necessary? It's already being done above in line 36.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
warmkesselj: This was the original logic. If we don't need it that would be fantastic. @jaxdesmarais can you weigh in on whether this is a business rule?

Sources/BraintreeCard/CreditCardGraphQLBody.swift
warmkesselj: @jaxdesmarais can you weigh in on whether this is a business rule? I think it is and is part of the original logic.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,223 @@
...
+                    if let cardName = card.cardholderName, !cardName.isEmpty{
+                        self.cardholderName = card.cardholderName
+                    }
+                }
+                
+                func encode(to encoder: any Encoder) throws {
+                    var container = encoder.container(keyedBy: CodingKeys.self)
+                    if let billingAddress {
+                        try container.encodeIfPresent(billingAddress, forKey: .billingAddress)
+                    }
```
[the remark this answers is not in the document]
warmkesselj: I'll need to retest this but I think generally trying to prevent the key from existing on the params if BillingAddress doesn't exist.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
jaxdesmarais: I don't see where the logic to check the first name existed for billing address unless I am missing it: https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreeCard/BTCard.swift. Looks like that was changed in this PR.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
jaxdesmarais: I think this is the logic that we want to replicate here: https://github.com/braintree/braintree_ios/blob/50ca7f8197237b3283408c2273aae34bdccff030/Sources/BraintreeCard/BTCard.swift#L108-L114. So I think here instead of checking `authenticationInsightRequested` again @richherrera is saying we can just check `merchantAccountID` but Rich please correct me if I misunderstood your comment.

richherrera: yep, exactly

Sources/BraintreeCard/CreditCardGraphQLBody.swift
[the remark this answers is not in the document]
warmkesselj: Btw, here's original code which means we purposefully omit the billingAddress key.
```
        if !billingAddressDictionary.isEmpty {
            cardDictionary["billingAddress"] = billingAddressDictionary
        }
```

Sources/BraintreeCard/CreditCardGraphQLBody.swift
warmkesselj: Let me check this out. I think it's necessary. The pain for these things is the original logic supported passing empty json bodies with Keys and I don't know if the Gateway is doing any conditional logic on these params.

warmkesselj: The original case
```
        if authenticationInsightRequested {
            if let merchantAccountID {
                variables["authenticationInsightInput"] = ["merchantAccountId": merchantAccountID]
            } else {
                variables["authenticationInsightInput"] = [:]
            }
        }
```

If false then no existing key on the body of the params authenticationInsightInput.
If true then key will exist with an empty body without a merchantAccountId or in the case merchantAccountID exists then with obviously with the accountID.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
warmkesselj: I think you're right. I'm not sure why I remembered some logic that encased the BillingInfo...

warmkesselj: Okay, I think we can remove this condition here in `func encode`
```
                 if let billingAddress {
                        try container.encodeIfPresent(billingAddress, forKey: .billingAddress)
                    }
```
But we still need to keep a conditional check here 
```
init(card: BTCard) {
    if [someCondition]                 
     self.billingAddress = BillingAddress(card: card)
```

Doing so will keep this original logic intact for not sending an empty BillingAddress json object 
```
          if !billingAddressDictionary.isEmpty {
            cardDictionary["billingAddress"] = billingAddressDictionary
        }
```

jaxdesmarais: Gotcha, that makes sense to me. Thanks for tracking that down. Maybe here we can do `if !card.billingAddress. isEmpty` to match the rest of the object? What do you think?

Sources/BraintreeCard/CreditCardGraphQLBody.swift
jaxdesmarais: 🤔 I think I am still not following. In that case, since we are only creating the `authenticationInsightInput` if `authenticationInsightRequested` above is true here we should just be able to do `merchantAccountID = card.merchantAccountID != nil ? card.merchantAccountID : nil`

Sources/BraintreeCard/CreditCardGraphQLBody.swift
jaxdesmarais: Ah you may be right, I don't think it exists yet. Maybe then doing this makes the most sense:
```swift
if let billingAddress {
    try container.encodeIfPresent(billingAddress, forKey: .billingAddress)
}
```
I think there are cases when merchants pass a billing address but don't pass a first name, and with the logic as is we just wouldn't pass the other details along even if they existed if no firstName exists.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
warmkesselj: Got rid of the check completely and setting merchantAccountID. This maintains the original logic and removes some code.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,178 @@
+// swiftlint:disable nesting
+
+import Foundation
+
+struct CreditCardGraphQLBody: Encodable {
```
jaxdesmarais, line 5: Can we move this swiftlint disable here instead:
```suggestion
// swiftlint:disable nesting
struct CreditCardGraphQLBody: Encodable {
```

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,139 @@
+// swiftlint:disable nesting
```
jaxdesmarais, line 1: I think we can remove this since we have it below as well

Sources/BraintreeCard/CreditCardGraphQLBody.swift
jaxdesmarais, line 5: Can we add a docstring for this? Can use `/// The POST body for <endpoint_here>` like we have in other files.

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,139 @@
...
+
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+// swiftlint:disable nesting
+struct CreditCardPOSTBody: Encodable {
```
jaxdesmarais, line 10: Can we add a docstring for this? Can use `/// The POST body for <endpoint_here>` like we have in other files.

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        enum CodingKeys: String, CodingKey {
+            case integration
+            case source
+            case sessionID = "sessionId"
+        }
+    }
+
+    /// POST Body Model
```
jaxdesmarais, line 57: Take it or leave it but not sure this comment is helpful. Can likely remove once we add the one to the top of this struct.

UnitTests/BraintreeCardTests/BTCard_Tests.swift
```diff
@@ -151,158 +146,121 @@ class BTCard_Tests: XCTestCase {
...
-                            "postalCode": "94107"
-                        ],
-                    ] as [String: Any],
-                    "options": ["validate": false]
-                ]
-            ]
-        ] as [String: Any] as NSObject)
+        let params = card.graphQLParameters()
+        
+        XCTAssertEqual(params.variables.input.options.validate,  false)
```
jaxdesmarais, line 151:
```suggestion
        XCTAssertEqual(params.variables.input.options.validate, false)
```

UnitTests/BraintreeCardTests/BTCard_Tests.swift
```diff
@@ -151,158 +146,121 @@ class BTCard_Tests: XCTestCase {
...
             authenticationInsightRequested: false,
             merchantAccountID: "some id"
         )
+         
+        let params = card.graphQLParameters()
+        
+        XCTAssertEqual(params.variables.input.creditCard.number, "6111111111111111")
+        XCTAssertEqual(params.operationName, "TokenizeCreditCard")
+        XCTAssertNotNil(params.query)
+        XCTAssertEqual(params.variables.input.options.validate,  false)
```
jaxdesmarais, line 207:
```suggestion
        XCTAssertEqual(params.variables.input.options.validate, false)
```

UnitTests/BraintreeCardTests/BTCard_Tests.swift
```diff
@@ -151,158 +146,121 @@ class BTCard_Tests: XCTestCase {
...
-                        "expirationYear": "2038",
-                        "number": "4111111111111111",
-                    ],
-                    "options": [ "validate": false ],
-                ],
-                "authenticationInsightInput": NSDictionary()
-            ]
-        ] as [String: Any] as NSObject)
+        let params = card.graphQLParameters()
+        printEncodableObject(params)
```
jaxdesmarais, line 223: Do we need this? It doesn't look like we are using the output.

UnitTests/BraintreeCardTests/BTCard_Tests.swift
```diff
@@ -151,158 +146,121 @@ class BTCard_Tests: XCTestCase {
...
-                    "options": [ "validate": false ],
-                ] as [String: Any]
-            ]
-        ] as [String: Any] as NSObject)
+        let params = card.graphQLParameters()
+        
+        XCTAssertEqual(params.variables.input.creditCard.number, "8111111111111111")
+        XCTAssertEqual(params.operationName, "TokenizeCreditCard")
+        XCTAssertNotNil(params.query)
+        XCTAssertEqual(params.variables.input.options.validate,  false)
```
jaxdesmarais, line 261:
```suggestion
        XCTAssertEqual(params.variables.input.options.validate, false)
```

UnitTests/BraintreeTestShared/MockAPIClient.swift
```diff
@@ -55,6 +56,7 @@ public class MockAPIClient: BTAPIClient {
     public override func post(_ path: String, parameters: Encodable, headers: [String: String]? = nil, httpType: BTAPIClientHTTPService = .gateway, completion completionBlock: ((BTJSON?, HTTPURLResponse?, Error?) -> Void)? = nil) {
         lastPOSTPath = path
         lastPOSTParameters = try? parameters.toDictionary()
+        lastPOSTEncodableParameters = parameters
```
jaxdesmarais, line 59: General question. Do we need this? I think the other encodable PRs were using `lastPOSTParameters` which converts the `Encodable` class to a dictionary so we wouldn't need to change tests. Just trying to make sure I understand why we needed to make large changes to the tests here but not in the other PRs.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,178 @@
...
+
+    // MARK: - Internal Properties
```
jaxdesmarais, line 7: imo we can remove this

UnitTests/BraintreeTestShared/MockAPIClient.swift
warmkesselj: I wanted to introduce and test the Encodable types and values from the unit tests. I can remove though.

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -48,7 +48,10 @@ struct CreditCardGraphQLBody: Encodable {
                 var cardholderName: String?
 
                 init(card: BTCard) {
-                    self.billingAddress = BillingAddress(card: card)
+                    if let firstName = card.firstName, !firstName.isEmpty {
```
warmkesselj, line 51: @jaxdesmarais @richherrera I wonder what alternative check we want to add here. Perhaps will have to iterate through all of card to check for billingAddress information and omit if all are nil and include if there is any info.

warmkesselj: If we include BillingAddress with an empty JSON we get a failure with `integrationTests` `testTokenizeCard_whenCardHasValidationDisabledAndCardIsValid_tokenizesSuccessfully` which is related to using `sandboxTokenVersion3`

warmkesselj: Could just do this:
```
                    if
                        let firstName = card.firstName, !firstName.isEmpty,
                        let lastName = card.lastName, !lastName.isEmpty,
                        let company = card.company, !company.isEmpty,
                        let postalCode = card.postalCode, !postalCode.isEmpty,
                        let streetAddress = card.streetAddress, !streetAddress.isEmpty,
                        let extendedAddress = card.extendedAddress, !extendedAddress.isEmpty,
                        let locality = card.locality, !locality.isEmpty,
                        let region = card.region, !region.isEmpty,
                        let countryName = card.countryName, !countryName.isEmpty,
                        let countryCodeAlpha2 = card.countryCodeAlpha2, !countryCodeAlpha2.isEmpty,
                        let countryCodeAlpha3 = card.countryCodeAlpha3, !countryCodeAlpha3.isEmpty,
                        let countryCodeNumeric = card.countryCodeNumeric, !countryCodeNumeric.isEmpty
                    {
                        self.billingAddress = BillingAddress(card: card)
                    }
```

warmkesselj: I've added this function for now to check the billing fields.

```
                func hasNonEmptyBillingField(in card: BTCard) -> Bool {
                    let fields: [String?] = [
                        card.firstName,
                        card.lastName,
                        card.company,
                        card.postalCode,
                        card.streetAddress,
                        card.extendedAddress,
                        card.locality,
                        card.region,
                        card.countryName,
                        card.countryCodeAlpha2,
                        card.countryCodeAlpha3,
                        card.countryCodeNumeric
                    ]
                    
                    for field in fields {
                        if let value = field, !value.isEmpty {
                            return true
                        }
                    }
                    return false
                }
```

richherrera: That function seems like a good alternative. However, I think it should be part of the BillingAddress object. 

However, another option we could try is: have the init of the `BillingAddress` object be -> `init?(card: BTCard)` and validate the necessary properties within the init. I'm thinking of something like the following:

``` swift

    init?(card: BTCard) {
        let properties = [
             card.firstName, card.lastName, card.company, card.postalCode, 
             card.streetAddress, card.extendedAddress, card.locality, card.region, 
             card.countryName, card.countryCodeAlpha2, card.countryCodeAlpha3, card.countryCodeNumeric
       ]

        if properties.allSatisfy({ $0 == nil || $0?.trimmingCharacters(in: .whitespacesAndNewlines) == "" }) {
            return nil
        }

        self.firstName = card.firstName
        self.lastName = card.lastName
        self.company = card.company
        self.postalCode = card.postalCode
        self.streetAddress = card.streetAddress
        self.extendedAddress = card.extendedAddress
        self.locality = card.locality
        self.region = card.region
        self.countryName = card.countryName
        self.countryCodeAlpha2 = card.countryCodeAlpha2
        self.countryCodeAlpha3 = card.countryCodeAlpha3
        self.countryCodeNumeric = card.countryCodeNumeric
    }
```

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,186 @@
...
+                            card.extendedAddress,
+                            card.locality,
+                            card.region,
+                            card.countryName,
+                            card.countryCodeAlpha2,
+                            card.countryCodeAlpha3,
+                            card.countryCodeNumeric
+                        ]
+
+                        if properties.allSatisfy({ $0?.isEmpty ?? true }) {
```
jaxdesmarais, line 91:
```suggestion
                        if properties.allSatisfy { $0?.isEmpty ?? true } {
```

Sources/BraintreeCard/CreditCardGraphQLBody.swift
```diff
@@ -0,0 +1,186 @@
+// swiftlint:disable nesting
```
jaxdesmarais, line 1: Can we move this down to above our docstring on the struct to match other files?

Sources/BraintreeCard/CreditCardGraphQLBody.swift
jaxdesmarais, line 91: Just curious, were we running into errors without this?

UnitTests/BraintreeCardTests/BTCard_Tests.swift
```diff
@@ -151,158 +146,120 @@ class BTCard_Tests: XCTestCase {
...
+        let encoder = JSONEncoder()
+        encoder.outputFormatting = .prettyPrinted // Makes the JSON easier to read
+        do {
+            let jsonData = try encoder.encode(object)
+            if let jsonString = String(data: jsonData, encoding: .utf8) {
+                print("Encoded JSON:\n\(jsonString)")
+            }
+        } catch {
+            print("Failed to encode object: \(error)")
+        }
```
jaxdesmarais, line 242: Do we need this?

UnitTests/BraintreeTestShared/MockAPIClient.swift
```diff
@@ -54,7 +54,7 @@ public class MockAPIClient: BTAPIClient {
     
     public override func post(_ path: String, parameters: Encodable, headers: [String: String]? = nil, httpType: BTAPIClientHTTPService = .gateway, completion completionBlock: ((BTJSON?, HTTPURLResponse?, Error?) -> Void)? = nil) {
         lastPOSTPath = path
-        lastPOSTParameters = try? parameters.toDictionary()
+        lastPOSTParameters = try? parameters.toDictionary()
```
jaxdesmarais, line 57:
```suggestion
        lastPOSTParameters = try? parameters.toDictionary()
```

Sources/BraintreeCard/CreditCardGraphQLBody.swift
warmkesselj: I get a warning if I try updating without parens

warmkesselj: Without it? No, we weren't running into errors.

jaxdesmarais: Gotcha. So it looks like this is the equivalent to the if let checks we are doing elsewhere: [LPM example](https://github.com/braintree/braintree_ios/blob/666e6ff2662a608d642607f71c028e055d4e4df7/Sources/BraintreeLocalPayment/Models/LocalPaymentPOSTBody.swift#L59), [PayPal example](https://github.com/braintree/braintree_ios/blob/v7/Sources/BraintreePayPal/Models/PayPalVaultPOSTBody.swift#L40). Would we be okay following that pattern instead of using properties/allSatisfy instead? Then if we feel strongly about the pattern used here we can make a follow up ticket for that work and update all the post bodies.

warmkesselj: I think to maintain the overall contract with GW we'd have to revert back to the other code I had to check for these properties prior to the initialization here. Otherwise, we'd be passing a `BillingAddress` key empty JSON body.

jaxdesmarais: This is Graph QL, so a different contract from the gateway ([schema public here](https://developer.paypal.com/braintree/graphql/reference/#Mutation--tokenizeCreditCard)). Testing locally our contract allows us to pass any of these as nil via graphQL. We do not currently set any of these properties in `CardTokenizationViewController` and updating this struct to the following does not produce any errors confirming that the contract for Graph QL accepts this:
```swift
                struct BillingAddress: Encodable {

                    var firstName: String?
                    var lastName: String?
                    var company: String?
                    var postalCode: String?
                    var streetAddress: String?
                    var extendedAddress: String?
                    var locality: String?
                    var region: String?
                    var countryName: String?
                    var countryCodeAlpha2: String?
                    var countryCodeAlpha3: String?
                    var countryCodeNumeric: String?

                    init?(card: BTCard) {
                        self.firstName = card.firstName
                        self.lastName = card.lastName
                        self.company = card.company
                        self.postalCode = card.postalCode
                        self.streetAddress = card.streetAddress
                        self.extendedAddress = card.extendedAddress
                        self.locality = card.locality
                        self.region = card.region
                        self.countryName = card.countryName
                        self.countryCodeAlpha2 = card.countryCodeAlpha2
                        self.countryCodeAlpha3 = card.countryCodeAlpha3
                        self.countryCodeNumeric = card.countryCodeNumeric
                    }
                }
```

warmkesselj: Can you try with `sandboxClientTokenVersion3`?

jaxdesmarais: Just tried with that client token in the demo app and was successful. Are you seeing an error in the card view controller with that client token?

Sources/BraintreeCard/CreditCardPOSTBody.swift
```diff
@@ -0,0 +1,137 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+// swiftlint:disable nesting
+/// The POST body for "v1/payment_methods/credit_cards"
```
jaxdesmarais, line 8: nit:
```suggestion
/// The POST body for `v1/payment_methods/credit_cards`
```

jaxdesmarais (approved):
Thanks for tackling this work and addressing all of my comments/questions. 🚀 One very minor comment but otherwise looking good.

richherrera (approved):

PR #1491, opened 2025-01-02
[V7] Add `Encodable` protocol for PayPal Checkout

richherrera:
### Summary of changes

- Add `PayPalRequest` protocol to define the contract for the base properties for `BTPayPalCheckoutRequest` and `BTPayPalVaultRequest`
- Conform to the new protocol in the three concrete classes (BTPayPalRequest, BTPayPalCheckoutRequest and BTPayPalVaultRequest) to eliminate inheritance
- Update docstrings for classes BTPayPalCheckoutRequest and BTPayPalVaultRequest
- LineItems now conforms to the Encodable protocol.
- Add encodable types for `post` bodies used in the PayPal Checkout flow.
- PayPal Checkout has been tested in the Demo app to ensure the dictionaries are still constructed as expected.

**Note**: Since we have two different classes (`PayPalCheckoutRequest` and `PayPalVaultRequest`) used in `PayPalClient`, it's currently not possible to use `Encodable` until the same implementation is done for `PayPalVaultRequest`. We tested the flow, but we need to complete task `DTMOBILES-1177` to be able to use `Encodable`.

### Checklist

### Authors

- @richherrera

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
+++ b/Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
@@ -165,23 +165,23 @@ import BraintreeCore
         experienceProfile["no_shipping"] = !isShippingAddressRequired
         experienceProfile["brand_name"] = displayName != nil ? displayName : configuration.json?["paypal"]["displayName"].asString()
 
-        if landingPageType?.stringValue != nil {
-            experienceProfile["landing_page_type"] = landingPageType?.stringValue
+        if let landingPageType = landingPageType?.stringValue {
+            experienceProfile["landing_page_type"] = landingPageType
         }
 
-        if localeCode?.stringValue != nil {
-            experienceProfile["locale_code"] = localeCode?.stringValue
+        if let localeCode = localeCode?.stringValue {
+            experienceProfile["locale_code"] = localeCode
         }
 
         experienceProfile["address_override"] = shippingAddressOverride != nil ? !isShippingAddressEditable : false
 
         var baseParameters: [String: Any] = [:]
 
-        if merchantAccountID != nil {
+        if let merchantAccountID {
             baseParameters["merchant_account_id"] = merchantAccountID
         }
 
-        if riskCorrelationID != nil {
+        if let riskCorrelationID {
             baseParameters["correlation_id"] = riskCorrelationID
         }
         
@@ -210,7 +210,7 @@ import BraintreeCore
 
         let currencyCode = currencyCode != nil ? currencyCode : configuration.json?["paypal"]["currencyIsoCode"].asString()
 
-        if currencyCode != nil {
+        if let currencyCode {
             checkoutParameters["currency_iso_code"] = currencyCode
         }
 
@@ -219,7 +219,7 @@ import BraintreeCore
             baseParameters["experience_profile"] = experienceProfile
         }
 
-        if requestBillingAgreement != false {
+        if requestBillingAgreement {
             checkoutParameters["request_billing_agreement"] = requestBillingAgreement
 
             if billingAgreementDescription != nil {
@@ -227,14 +227,14 @@ import BraintreeCore
             }
         }
 
-        if shippingAddressOverride != nil {
-            checkoutParameters["line1"] = shippingAddressOverride?.streetAddress
-            checkoutParameters["line2"] = shippingAddressOverride?.extendedAddress
-            checkoutParameters["city"] = shippingAddressOverride?.locality
-            checkoutParameters["state"] = shippingAddressOverride?.region
-            checkoutParameters["postal_code"] = shippingAddressOverride?.postalCode
-            checkoutParameters["country_code"] = shippingAddressOverride?.countryCodeAlpha2
-            checkoutParameters["recipient_name"] = shippingAddressOverride?.recipientName
+        if let shippingAddressOverride {
+            checkoutParameters["line1"] = shippingAddressOverride.streetAddress
+            checkoutParameters["line2"] = shippingAddressOverride.extendedAddress
+            checkoutParameters["city"] = shippingAddressOverride.locality
+            checkoutParameters["state"] = shippingAddressOverride.region
+            checkoutParameters["postal_code"] = shippingAddressOverride.postalCode
+            checkoutParameters["country_code"] = shippingAddressOverride.countryCodeAlpha2
+            checkoutParameters["recipient_name"] = shippingAddressOverride.recipientName
         }
 
         return baseParameters.merging(checkoutParameters) { $1 }
--- a/Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
+++ b/Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
@@ -56,17 +56,32 @@ import BraintreeCore
 }
 
 /// Options for the PayPal Checkout flow.
-@objcMembers open class BTPayPalCheckoutRequest: BTPayPalRequest {
-
+@objcMembers public class BTPayPalCheckoutRequest: NSObject, PayPalRequest {
+    
     // MARK: - Internal Properties
+    
+    let hermesPath: String
+    let paymentType: BTPayPalPaymentType
     
     var amount: String
     var intent: BTPayPalRequestIntent
     var userAction: BTPayPalRequestUserAction
     var offerPayLater: Bool
+    var billingAgreementDescription: String?
     var currencyCode: String?
+    var displayName: String?
+    var isShippingAddressEditable: Bool = false
+    var isShippingAddressRequired: Bool = false
+    var landingPageType: BTPayPalRequestLandingPageType?
+    var lineItems: [BTPayPalLineItem]?
+    var localeCode: BTPayPalLocaleCode?
+    var merchantAccountID: String?
     var requestBillingAgreement: Bool
-
+    var riskCorrelationID: String?
+    var shippingAddressOverride: BTPostalAddress?
+    var userAuthenticationEmail: String?
+    var userPhoneNumber: BTPayPalPhoneNumber?
+    
     // MARK: - Initializer
 
     /// Initializes a PayPal Native Checkout request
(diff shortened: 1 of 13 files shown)


Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
```diff
@@ -0,0 +1,176 @@
...
+        let currencyIsoCode = payPalRequest.currencyCode != nil ? payPalRequest.currencyCode : configuration.currencyIsoCode
+        
+        if let currencyIsoCode {
+            self.currencyCode = currencyIsoCode
+        }
+        
+        if payPalRequest.requestBillingAgreement {
+            self.requestBillingAgreement = payPalRequest.requestBillingAgreement
+            
+            if let billingAgreementDescription = payPalRequest.billingAgreementDescription {
```
richherrera, line 53: Currently, this property belongs to class `BTPayPalRequest`. After changing its properties to `internal` in a previous PR, merchants can no longer access these properties, meaning there's no way to configure, for example, "billingAgreementDescription" when creating an instance of `BTPayPalCheckoutRequest` or `BTPayPalVaultRequest`. Do you think we should add the necessary properties to the init of `BTPayPalCheckoutRequest` or `BTPayPalVaultRequest`, or make the properties in `BTPayPalRequest` public again and remove `BTPayPalRequest.init`?

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -75,48 +90,118 @@ import BraintreeCore
...
 
-    override func parameters(
+    // swiftlint:disable cyclomatic_complexity
+    func parameters(
         with configuration: BTConfiguration,
         universalLink: URL? = nil,
         isPayPalAppInstalled: Bool = false
     ) -> [String: Any] {
-        var baseParameters = super.parameters(with: configuration)
+        var experienceProfile: [String: Any] = [:]
```
richherrera, line 163: Since we won't have inheritance, I had to move this code from `BTPayPalRequest.parameters`. In the next PR, this entire method will be removed because we are creating the `PayPalCheckoutPOSTBody` model, which conforms to the Encodable protocol.

Sources/BraintreePayPal/BTPayPalRequest.swift
```diff
@@ -48,9 +48,43 @@ import BraintreeCore
...
+    var riskCorrelationID: String? { get }
+    var shippingAddressOverride: BTPostalAddress? { get }
+    var userAuthenticationEmail: String? { get }
+    var userPhoneNumber: BTPayPalPhoneNumber? { get }
+    
+    // MARK: - Static Properties
+    
+    static var callbackURLHostAndPath: String { get }
+    
+    func parameters(
```
richherrera, line 71: This method is to mimic the inheritance we used to do. Once `BTPayPalVault` has its own `POSTBody` model that conforms to the `Encodable` protocol in the next PR, we will be able to remove this method.

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
```diff
@@ -35,35 +50,109 @@ import BraintreeCore
...
     ///   - userPhoneNumber: Optional: A user's phone number to initiate a quicker authentication flow in the scenario where the user has a PayPal account
     /// identified with the same phone number.
     public init(
         offerCredit: Bool = false,
+        billingAgreementDescription: String? = nil,
+        displayName: String? = nil,
+        isShippingAddressEditable: Bool = false,
+        isShippingAddressRequired: Bool = false,
+        landingPageType: BTPayPalRequestLandingPageType = .none,
+        lineItems: [BTPayPalLineItem]? = nil,
```
richherrera, line 80: Comparing the implementation with Android, on Android, `lineItems` is only used for the `Checkout` flow. What do you think about removing this property from the `PayPalRequest` protocol, so it's only visible in the Checkout flow? On Android, we would need to remove the property from the `AbstractClass` PayPalRequest, and it wouldn't be overridden in `PayPalCheckoutRequest`

jaxdesmarais: Interesting, let's confirm with the pay with paypal team if line items can be passed on vault requests. If not, I agree with you we can remove it from the protocol and make it only available on `PayPalCheckoutRequest`

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -29,7 +29,7 @@ import BraintreeDataCollector
     var clientMetadataID: String?
     
     /// Exposed for testing the intent associated with this request
-    var payPalRequest: BTPayPalRequest?
+    var payPalRequest: PayPalRequest?
```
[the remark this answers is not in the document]
richherrera: Since it's an internal protocol, I think we can omit the `BT` prefix. If we decide to keep the prefix, how about I change it in the next PR? Right now, there would be naming conflicts with the _concrete_ _class_. Another option would be to choose a different name for the protocol 🤔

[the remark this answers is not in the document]
jaxdesmarais: We can change it in the next PR - I feel like it makes sense to keep the prefix for now just based on the class names but can be in the follow up.

Sources/BraintreePayPal/BTPayPalLineItem.swift
```diff
@@ -114,6 +123,39 @@ import Foundation
...
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(quantity, forKey: .quantity)
+        try container.encode(unitAmount, forKey: .unitAmount)
+        try container.encode(name, forKey: .name)
+        try container.encode(kind.stringValue, forKey: .kind)
+        
+        if let unitTaxAmount, !unitTaxAmount.isEmpty {
+            try container.encode(unitAmount, forKey: .unitTaxAmount)
+        }
```
scannillo, line 149: Do we need to implement a custom `encode()` method just for this case?

Does Encodable not by default just omit an optional value that is `nil`?

richherrera: The reason is that on iOS, the value of `unitAmount` is used with the keys `unit_amount`

https://github.com/braintree/braintree_ios/blob/652e69697d4449de5fe509fed5dcafde9f07c1b7/Sources/BraintreePayPal/BTPayPalLineItem.swift#L124

 and `unit_tax_amount`

https://github.com/braintree/braintree_ios/blob/652e69697d4449de5fe509fed5dcafde9f07c1b7/Sources/BraintreePayPal/BTPayPalLineItem.swift#L130

but it seems this has always been an error. On Android it's different 🤔

richherrera: I've updated the PR: [e478437](https://github.com/braintree/braintree_ios/pull/1491/commits/e47843789f07e812d39d8e7f486d1259885ef9a5) let me know what do you think. Unfortunately, the enums need to implement a custom encode because they are of type Int, and we are encoding their string value.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -75,48 +90,118 @@ import BraintreeCore
...
+
+        experienceProfile["no_shipping"] = !isShippingAddressRequired
+        experienceProfile["brand_name"] = displayName != nil ? displayName : configuration.json?["paypal"]["displayName"].asString()
+
+        if landingPageType?.stringValue != nil {
```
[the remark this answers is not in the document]
richherrera: Updated: [e446f03](https://github.com/braintree/braintree_ios/pull/1491/commits/e446f03b33f01e67c7c74af9f64ba0879a245f58)

Sources/BraintreePayPal/BTPayPalClient.swift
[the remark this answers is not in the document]
richherrera: Updated: [f90356b](https://github.com/braintree/braintree_ios/pull/1492/commits/f90356b85bc8d0b0fd5529d41ea4508b61f52df5)

jaxdesmarais (approved):
🚀

Sources/BraintreePayPal/BTPayPalRequest.swift
```diff
@@ -48,9 +48,43 @@ import BraintreeCore
     }
 }
 
+protocol PayPalRequest {
```
scannillo, line 51: A few nits - can some of the enum/class definitions here live in their own file?

Also maybe adding a docstring for PayPalRequest protocol will be useful since `Request` types throughout the SDK aren't protocol types.

scannillo: > Protocols that describe what something is should read as nouns (e.g. Collection).
> Protocols that describe a capability should be named using the suffixes able, ible, or ing (e.g. Equatable, ProgressReporting).

^ From [Apple's Swift style guide](https://www.swift.org/documentation/api-design-guidelines/#:~:text=Protocols%20that%20describe%20what%20something,constants%20should%20read%20as%20nouns.)

Up to you if you think there's a better name for the protocol

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
```diff
@@ -0,0 +1,175 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+/// The POST body for v1/paypal_hermes/create_payment_resource
```
scannillo, line 7:
```suggestion
/// The POST body for `v1/paypal_hermes/create_payment_resource`
```

scannillo (approved):
A few styling nits, otherwise this is awesome!

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
richherrera: Updated: [f3ccfe4](https://github.com/braintree/braintree_ios/pull/1491/commits/f3ccfe4d1eb757034b7925ea4afb5dc743d5f947)

Sources/BraintreePayPal/BTPayPalRequest.swift
richherrera: I moved the enums to their own file. The class will be removed in the next PR. Regarding the protocol name, it was changed to BTPayPalRequest in the next PR, based on the naming recommendations for protocols. I believe it meets the first recommendation since we are describing the "request." [876599d](https://github.com/braintree/braintree_ios/pull/1491/commits/876599dc2b2ea9a8095389d900a1864e98a1d409)

PR #1499, opened 2025-01-14
Bug Fix: Handle Missing 3DS `dfReferenceId`

jaxdesmarais:
### Summary of changes

- Return error if no `dfReferenceId` is returned in the 3D Secure flow
    - Confirmed with Cardinal that we should not continue the flow if we do not get a `dfReferenceId` back, there is also no fallback mechanism on mobile like exists on web

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreeThreeDSecure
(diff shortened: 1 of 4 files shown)


richherrera (approved):

saperi22 (approved):

agedd (approved):
lgtm! 🚀

PR #1503, opened 2025-01-17
Allow Merchant to Only Set Universal Links for Venmo

jaxdesmarais:
### Summary of changes

- A bug was recently brought to the service channel where if a merchant trys to move from the deprecated `BTAppContextSwitcher.sharedInstance.returnURLScheme` for Venmo they get an error if only passing a universal link. We should allow merchants to only use universal links without needing to set both that and a `returnURLScheme`

I tested the following flows on device, please feel free to retest these or call out any use cases I missed:
- `universalLink` and `returnURLScheme` not set (returns an error)
- only `returnURLScheme` set (uses `returnURLScheme` for switch and return)
- only `universalLink` set (uses `universalLink`  for switch and return)

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/Demo/Application/Features/VenmoViewController.swift
+++ b/Demo/Application/Features/VenmoViewController.swift
@@ -8,22 +8,19 @@ class VenmoViewController: PaymentButtonBaseViewController {
 
     let webFallbackToggle = Toggle(title: "Enable Web Fallback")
     let vaultToggle = Toggle(title: "Vault")
+    let universalLinkReturnToggle = Toggle(title: "Use Universal Link Return")
 
     override func viewDidLoad() {
         super.heightConstraint = 150
         super.viewDidLoad()
-        venmoClient = BTVenmoClient(
-            apiClient: apiClient,
-            // swiftlint:disable:next force_unwrapping
-            universalLink: URL(string: "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments")!
-        )
+        venmoClient = BTVenmoClient(apiClient: apiClient)
         title = "Custom Venmo Button"
     }
     
     override func createPaymentButton() -> UIView {
         let venmoButton = createButton(title: "Venmo", action: #selector(tappedVenmo))
 
-        let stackView = UIStackView(arrangedSubviews: [webFallbackToggle, vaultToggle, venmoButton])
+        let stackView = UIStackView(arrangedSubviews: [webFallbackToggle, vaultToggle, universalLinkReturnToggle, venmoButton])
         stackView.axis = .vertical
         stackView.spacing = 15
         stackView.alignment = .fill
@@ -44,6 +41,14 @@ class VenmoViewController: PaymentButtonBaseViewController {
         
         if vaultToggle.isOn {
             venmoRequest.vault = true
+        }
+
+        if universalLinkReturnToggle.isOn {
+            venmoClient = BTVenmoClient(
+                apiClient: apiClient,
+                // swiftlint:disable:next force_unwrapping
+                universalLink: URL(string: "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments")!
+            )
         }
 
         Task {
--- a/Demo/Application/Features/VenmoViewController.swift
+++ b/Demo/Application/Features/VenmoViewController.swift
@@ -8,19 +8,21 @@ class VenmoViewController: PaymentButtonBaseViewController {
 
     let webFallbackToggle = Toggle(title: "Enable Web Fallback")
     let vaultToggle = Toggle(title: "Vault")
-    let universalLinkReturnToggle = Toggle(title: "Use Universal Link Return")
 
     override func viewDidLoad() {
         super.heightConstraint = 150
         super.viewDidLoad()
-        venmoClient = BTVenmoClient(apiClient: apiClient)
+        venmoClient = BTVenmoClient(
+            apiClient: apiClient,
+            universalLink: URL(string: "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments")!
+        )
         title = "Custom Venmo Button"
     }
     
     override func createPaymentButton() -> UIView {
         let venmoButton = createButton(title: "Venmo", action: #selector(tappedVenmo))
 
-        let stackView = UIStackView(arrangedSubviews: [webFallbackToggle, vaultToggle, universalLinkReturnToggle, venmoButton])
+        let stackView = UIStackView(arrangedSubviews: [webFallbackToggle, vaultToggle, venmoButton])
         stackView.axis = .vertical
         stackView.spacing = 15
         stackView.alignment = .fill
@@ -41,14 +43,6 @@ class VenmoViewController: PaymentButtonBaseViewController {
         
         if vaultToggle.isOn {
             venmoRequest.vault = true
-        }
-
-        if universalLinkReturnToggle.isOn {
-            venmoClient = BTVenmoClient(
-                apiClient: apiClient,
-                // swiftlint:disable:next force_unwrapping
-                universalLink: URL(string: "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments")!
-            )
         }
 
         Task {
--- a/Demo/Application/Features/VenmoViewController.swift
+++ b/Demo/Application/Features/VenmoViewController.swift
@@ -14,6 +14,7 @@ class VenmoViewController: PaymentButtonBaseViewController {
         super.viewDidLoad()
         venmoClient = BTVenmoClient(
             apiClient: apiClient,
+            // swiftlint:disable:next force_unwrapping
             universalLink: URL(string: "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments")!
         )
         title = "Custom Venmo Button"
--- a/Sources/BraintreeVenmo/BTVenmoClient.swift
+++ b/Sources/BraintreeVenmo/BTVenmoClient.swift
@@ -93,7 +93,10 @@ import BraintreeCore
                 completion: completion
             )
             return
-        } else if let bundleIdentifier = bundle.bundleIdentifier, !returnURLScheme.hasPrefix(bundleIdentifier) && (universalLink?.absoluteString.isEmpty == true || universalLink?.absoluteString == nil) {
+        } else if
+            let bundleIdentifier = bundle.bundleIdentifier,
+                !returnURLScheme.hasPrefix(bundleIdentifier)
+                && (universalLink?.absoluteString.isEmpty == true || universalLink?.absoluteString == nil) {
             NSLog(
                 // swiftlint:disable:next line_length
                 "%@ Venmo requires [BTAppContextSwitcher setReturnURLScheme:] to be configured to begin with your app's bundle ID (%@). Currently, it is set to (%@)",
--- a/Sources/BraintreeVenmo/BTVenmoClient.swift
+++ b/Sources/BraintreeVenmo/BTVenmoClient.swift
@@ -83,26 +83,24 @@ import BraintreeCore
         apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeStarted, isVaultRequest: shouldVault, linkType: linkType)
         let returnURLScheme = BTAppContextSwitcher.sharedInstance._returnURLScheme
 
-        if universalLink?.absoluteString.isEmpty == true || universalLink?.absoluteString == nil {
-            if returnURLScheme.isEmpty {
-                NSLog(
-                    "%@ Venmo requires a return URL scheme or universal link to be configured.",
-                    BTLogLevelDescription.string(for: .critical)
-                )
-                notifyFailure(
-                    with: BTVenmoError.invalidReturnURL("Venmo requires a return URL scheme or universal link to be configured."),
-                    completion: completion
-                )
-                return
-            } else if let bundleIdentifier = bundle.bundleIdentifier, !returnURLScheme.hasPrefix(bundleIdentifier) {
-                NSLog(
-                    // swiftlint:disable:next line_length
-                    "%@ Venmo requires [BTAppContextSwitcher setReturnURLScheme:] to be configured to begin with your app's bundle ID (%@). Currently, it is set to (%@)",
-                    BTLogLevelDescription.string(for: .critical),
-                    bundleIdentifier,
-                    returnURLScheme
-                )
-            }
+        if universalLink?.absoluteString.isEmpty == true || universalLink?.absoluteString == nil && returnURLScheme.isEmpty {
+            NSLog(
+                "%@ Venmo requires a return URL scheme or universal link to be configured.",
+                BTLogLevelDescription.string(for: .critical)
+            )
+            notifyFailure(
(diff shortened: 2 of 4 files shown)


Demo/Application/Features/VenmoViewController.swift
```diff
@@ -43,14 +45,6 @@ class VenmoViewController: PaymentButtonBaseViewController {
             venmoRequest.vault = true
         }
 
-        if universalLinkReturnToggle.isOn {
-            venmoClient = BTVenmoClient(
-                apiClient: apiClient,
-                // swiftlint:disable:next force_unwrapping
-                universalLink: URL(string: "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments")!
-            )
-        }
```
jaxdesmarais, line 52: This logic seemed pretty odd, is there a usecase where we want a merchant to set a return URL for the switch but use universal links only for the return? That seemed a bit odd to me and what was happening. I would expect if a merchant passes a universal link we want to use that for the switch and return. Maybe we should be setting this for  a new `universalLinkkToggle` instead so we can test both the UL and return URL scheme flows?

scannillo: I think this was added before we supported universal links to the merchant app. Down to consolidate them into 1! I don't know how much value there would be to having 2 separate toggles since _technically_ a merchant can opt into the web-fallback (UL to venmo), but still use schemes back to their app

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -83,21 +83,23 @@ import BraintreeCore
...
+                NSLog(
+                    "%@ Venmo requires a return URL scheme or universal link to be configured.",
+                    BTLogLevelDescription.string(for: .critical)
+                )
+                notifyFailure(with: BTVenmoError.appNotAvailable, completion: completion)
+                return
+            } else if let bundleIdentifier = bundle.bundleIdentifier, !returnURLScheme.hasPrefix(bundleIdentifier) {
+                NSLog(
+                    // swiftlint:disable:next line_length
+                    "%@ Venmo requires [BTAppContextSwitcher setReturnURLScheme:] to be configured to begin with your app's bundle ID (%@). Currently, it is set to (%@)",
```
scannillo, line 97: Do we want to update this to Swift-y syntax instead of ObjC? Would you consider it a breaking change?

Demo/Application/Features/VenmoViewController.swift
jaxdesmarais: > I don't know how much value there would be to having 2 separate toggles since technically a merchant can opt into the web-fallback (UL to venmo), but still use schemes back to their app

I don't think this is true, as the logic exists if we get a universal link we use them for the return: https://github.com/braintree/braintree_ios/blob/f1a6643a4e52e61b215cc484a8fc3d870b857e5e/Sources/BraintreeVenmo/BTVenmoAppSwitchRedirectURL.swift#L60-L62

Is there a reason we'd want to still use URL scheme's to return back to the app if a universal link is passed?

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: Yeah I have always wondered how/if the `NSLog`s are actually useful to merchants. It's also interesting that we return an `appNotAvailable` for a missing return URL scheme and return nothing and instead fail later if the bundle ID doesn't contain the return URL scheme. To me it seems like they should both be handled the same, though not sure what we consider in scope for this PR.

Demo/Application/Features/VenmoViewController.swift
scannillo: Sorry, let me clarify.

A merchant can do venmoRequest.fallbackToWeb = true (use UL into Venmo), but _not_ provide a UL of their own app (merchant.com) for the switch back, therefore using a scheme on the way back.

scannillo: Both directions merchant_app --> venmo and venmo --> merchant_app have the option to do scheme or UL, in V6 of the SDK. So 4 different combinations of uses.

`universalLinkReturnToggle` in the original code snippet talks about the webFallbackFlow or UL _into_ Venmo. It had nothing to do with when we added UL support back into the merchant app. I think that's the confusing part is the naming there!

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -83,21 +83,23 @@ import BraintreeCore
         apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeStarted, isVaultRequest: shouldVault, linkType: linkType)
         let returnURLScheme = BTAppContextSwitcher.sharedInstance._returnURLScheme
 
-        if returnURLScheme.isEmpty {
-            NSLog(
-                "%@ Venmo requires a return URL scheme to be configured via [BTAppContextSwitcher setReturnURLScheme:]",
-                BTLogLevelDescription.string(for: .critical)
-            )
-            notifyFailure(with: BTVenmoError.appNotAvailable, completion: completion)
```
scannillo, line 91: Also, I feel like `VenmoError.appNotAvailable`, which reads `The Venmo app is not installed on this device, or it is not configured or available for app switch.` is not the right error here. I think it should instead be `VenmoError.invalidReturnURL`

scannillo: IMO we can treat that as a bug and fix it in this PR (line 92 below)

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: We don't allow URL schemes at all in v7 ([see code](https://github.com/braintree/braintree_ios/blob/v7/Sources/BraintreeVenmo/BTVenmoClient.swift)) so this is already cleaned up there. 😄

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: Yep, I mentioned that [here](https://github.com/braintree/braintree_ios/pull/1503#discussion_r1924376663), will update this to a more appropriate error.

jaxdesmarais: Updated: https://github.com/braintree/braintree_ios/pull/1503/commits/4f1db1f69af23678ee79b6a591568478f556710d

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -83,21 +83,26 @@ import BraintreeCore
...
-        } else if let bundleIdentifier = bundle.bundleIdentifier, !returnURLScheme.hasPrefix(bundleIdentifier) {
-            NSLog(
-                // swiftlint:disable:next line_length
-                "%@ Venmo requires [BTAppContextSwitcher setReturnURLScheme:] to be configured to begin with your app's bundle ID (%@). Currently, it is set to (%@)",
-                BTLogLevelDescription.string(for: .critical),
-                bundleIdentifier,
-                returnURLScheme
-            )
+        if universalLink?.absoluteString.isEmpty == true || universalLink?.absoluteString == nil {
+            if returnURLScheme.isEmpty {
```
scannillo, line 87: Nit - these nested ifs are a little tricky to read. Can this instead just be combined into 1 if?

Demo/Application/Features/VenmoViewController.swift
jaxdesmarais: Ah you're right, I misremembered that the success/canceled/error universal link was not set for web fallback. Maybe it does make sense to have both buttons. 🤔 

Update here: https://github.com/braintree/braintree_ios/pull/1503/commits/72f133a4b12de9c927beb7e9f345f910a386f0ec and tested the following:
1. universal link set, no return URL scheme set, fallback to web off - switches to app and uses UL return; no longer returns error
2. universal link not set, return URL scheme set, fallback to web off - switches to app and uses return URL scheme return
3. universal link set, no return URL scheme set, fallback to web on - switches to web/app and uses UL return
4. universal link not set, return URL scheme set, fallback to web on - switches to web and uses return URL scheme return

I _think_ those are all the happy paths but let me know if I missed any or if you get different results testing

agedd: ahhh this is interesting! just want to make sure i'm understanding right - would these be the 4 possible cases? 👀

- if fallbackToWeb = true >> merchant_app -> venmo via **UL** and venmo -> merchant_app return via **returnURLScheme** (if merchant does not provide an UL of their own)

- if fallbackToWeb = true >> merchant_app -> venmo via **UL** and venmo -> merchant_app return via **UL** (if merchant does indeed provide an UL of their own)

- if fallbackToWeb = false >> merchant_app -> venmo via **returnURLScheme** and venmo -> merchant_app via **UL** (if merchant provides an UL of their own app)

- if fallbackToWeb = false >> merchant_app -> venmo via **returnURLScheme** and venmo -> merchant_app via **returnURLScheme** (if merchant provides does not provide an UL of their own app)

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: Agree it's a little odd. We could do something like this, not sure if that's easier to read:
```swift
        let universalLinkEmpty: Bool = universalLink?.absoluteString.isEmpty == true || universalLink?.absoluteString == nil
        if universalLinkEmpty && returnURLScheme.isEmpty {
            NSLog(
                "%@ Venmo requires a return URL scheme or universal link to be configured.",
                BTLogLevelDescription.string(for: .critical)
            )
            notifyFailure(
                with: BTVenmoError.invalidReturnURL("Venmo requires a return URL scheme or universal link to be configured."),
                completion: completion
            )
            return
        } else if let bundleIdentifier = bundle.bundleIdentifier, !returnURLScheme.hasPrefix(bundleIdentifier) && universalLinkEmpty {
            NSLog(
                // swiftlint:disable:next line_length
                "%@ Venmo requires [BTAppContextSwitcher setReturnURLScheme:] to be configured to begin with your app's bundle ID (%@). Currently, it is set to (%@)",
                BTLogLevelDescription.string(for: .critical),
                bundleIdentifier,
                returnURLScheme
            )
        }
```
I thought it was a bit more clear to make it apparent that we only care about the return URL scheme/bundle ID if there is no universal link how it was written in the PR, but the above I think achieves the same.

Demo/Application/Features/VenmoViewController.swift
jaxdesmarais: Yep that's correct @agedd!

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -83,21 +83,26 @@ import BraintreeCore
...
+                )
+                return
+            } else if let bundleIdentifier = bundle.bundleIdentifier, !returnURLScheme.hasPrefix(bundleIdentifier) {
+                NSLog(
+                    // swiftlint:disable:next line_length
+                    "%@ Venmo requires [BTAppContextSwitcher setReturnURLScheme:] to be configured to begin with your app's bundle ID (%@). Currently, it is set to (%@)",
+                    BTLogLevelDescription.string(for: .critical),
+                    bundleIdentifier,
+                    returnURLScheme
+                )
```
richherrera, line 104: In this if statement, after `NSLog`, we don't have a `return`. Should we add one, or after displaying the `NSLog`, should we continue with the flow?

Sources/BraintreeVenmo/BTVenmoClient.swift
richherrera: I've tested this approach:

```suggestion
        if universalLink?.absoluteString.isEmpty ?? true {
```

UnitTests/BraintreeVenmoTests/BTVenmoClient_Tests.swift
```diff
@@ -78,6 +78,20 @@ class BTVenmoClient_Tests: XCTestCase {
         let venmoClient = BTVenmoClient(apiClient: mockAPIClient)
         BTAppContextSwitcher.sharedInstance.returnURLScheme = ""
         
+        let expectation = expectation(description: "authorization callback")
+        venmoClient.tokenize(venmoRequest) { venmoAccount, error in
+            guard let error = error as NSError? else {return}
```
agedd, line 83: small nit: 
```suggestion
            guard let error = error as NSError? else { return }
```

agedd (approved):
lgtm! 🚀

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: @richherrera I am not sure I follow. Updating to this and commenting out `BTAppContextSwitcher.sharedInstance.returnURLScheme` in our `AppDelegate` returns an error with that update which is not expected.

richherrera: Sorry let me clarify: 

I was thinking that something like the following might work, instead of 
    `universalLink?.absoluteString.isEmpty == true || universalLink?.absoluteString == nil ` 
we can use 
     `universalLink?.absoluteString.isEmpty ?? true`

final approach suggested:

```swift
        if universalLink?.absoluteString.isEmpty ?? true, returnURLScheme.isEmpty {
            NSLog(
                "%@ Venmo requires a return URL scheme or universal link to be configured.",
                BTLogLevelDescription.string(for: .critical)
            )
            notifyFailure(
                with: BTVenmoError.invalidReturnURL("Venmo requires a return URL scheme or universal link to be configured."),
                completion: completion
            )
            return
        } else if let bundleIdentifier = bundle.bundleIdentifier, !returnURLScheme.hasPrefix(bundleIdentifier) {
            NSLog(
                "%@ Venmo requires [BTAppContextSwitcher setReturnURLScheme:] to be configured to begin with your app's bundle ID (%@). Currently, it is set to (%@)",
                BTLogLevelDescription.string(for: .critical),
                bundleIdentifier,
                returnURLScheme
            )
        }
```
I also removed `swiftlint:disable:next line_length` since it seems to be unnecessary.

jaxdesmarais: Ah gotcha, I think GitHub highlighted the wrong code. Yep we can do that if we feel like it's more clear. I normally add both the nil and empty checks because it makes it a bit more clear that `isEmpty` returns nil if you pass it a nil value. But if we feel like this pattern is explicit enough:
```swift
universalLink?.absoluteString.isEmpty ?? true
```
over this one:
```swift
universalLink?.absoluteString.isEmpty == true || universalLink?.absoluteString == nil 
```
I am okay with that. What do you think?

richherrera: Got it, I'm fine if it's possible to just combine both if statements

richherrera (approved):

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: Sounds good! Updated here: https://github.com/braintree/braintree_ios/pull/1503/commits/07780fc99cd1b08e1784545fde0cf5794a141f1a. Sorry it took me a bit wanted to test the flows again and kept forgetting to bring something to plug my phone into my laptop with to the office. 🤦

PR #1504, opened 2025-01-17
Launch PayPal App For Checkout Flow

agedd:
### Summary of changes

- Update logic to launch the PP app for the 1-time checkout flow.

**Note**: verified app-switch flow works as expected for both flows w/ intercepting the url before app-switch - would need to wait until #1500 is merged in and gateway/other components changes have been deployed.

### Checklist

- @agedd

### DIFF
--- a/Braintree.xcodeproj/xcshareddata/xcschemes/UnitTests.xcscheme
+++ b/Braintree.xcodeproj/xcshareddata/xcschemes/UnitTests.xcscheme
@@ -107,6 +107,11 @@
                BlueprintName = "BraintreePayPalTests"
                ReferencedContainer = "container:Braintree.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
+               <Test
+                  Identifier = "BTPayPalClient_Tests/testTokenizeVaultAccount_whenPayPalAppApprovalURLMissingBAToken_returnsError()">
+               </Test>
+            </SkippedTests>
          </TestableReference>
          <TestableReference
             skipped = "NO"
--- a/Braintree.xcodeproj/xcshareddata/xcschemes/UnitTests.xcscheme
+++ b/Braintree.xcodeproj/xcshareddata/xcschemes/UnitTests.xcscheme
@@ -107,11 +107,6 @@
                BlueprintName = "BraintreePayPalTests"
                ReferencedContainer = "container:Braintree.xcodeproj">
             </BuildableReference>
-            <SkippedTests>
-               <Test
-                  Identifier = "BTPayPalClient_Tests/testTokenizeVaultAccount_whenPayPalAppApprovalURLMissingBAToken_returnsError()">
-               </Test>
-            </SkippedTests>
          </TestableReference>
          <TestableReference
             skipped = "NO"
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -366,7 +366,7 @@ import BraintreeDataCollector
 
                 switch approvalURL.redirectType {
                 case .payPalApp(let url):
-                    guard let _ = self.isVaultRequest ? approvalURL.baToken : approvalURL.ecToken else {
+                    guard (self.isVaultRequest ? approvalURL.baToken : approvalURL.ecToken) != nil else {
                         self.notifyFailure(
                             with: self.isVaultRequest ? BTPayPalError.missingBAToken : BTPayPalError.missingECToken,
                             completion: completion
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -366,6 +366,13 @@ import BraintreeDataCollector
 
                 switch approvalURL.redirectType {
                 case .payPalApp(let url):
+                    guard let token = self.isVaultRequest ? approvalURL.baToken : approvalURL.ecToken else {
+                        self.notifyFailure(
+                            with: self.isVaultRequest ? BTPayPalError.missingBAToken : BTPayPalError.missingECToken,
+                            completion: completion
+                        )
+                        return
+                    }
                     self.launchPayPalApp(with: url, completion: completion)
                 case .webBrowser(let url):
                     self.handlePayPalRequest(with: url, paymentType: request.paymentType, completion: completion)
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -366,7 +366,7 @@ import BraintreeDataCollector
 
                 switch approvalURL.redirectType {
                 case .payPalApp(let url):
-                    guard let token = self.isVaultRequest ? approvalURL.baToken : approvalURL.ecToken else {
+                    guard let _ = self.isVaultRequest ? approvalURL.baToken : approvalURL.ecToken else {
                         self.notifyFailure(
                             with: self.isVaultRequest ? BTPayPalError.missingBAToken : BTPayPalError.missingECToken,
                             completion: completion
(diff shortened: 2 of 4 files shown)


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -393,7 +395,7 @@ import BraintreeDataCollector
 
         var urlComponents = URLComponents(url: payPalAppRedirectURL, resolvingAgainstBaseURL: true)
         urlComponents?.queryItems = [
-            URLQueryItem(name: "ba_token", value: baToken),
+            URLQueryItem(name: isVaultRequest ? "ba_token" : "token", value: token),
```
jaxdesmarais, line 398: On Android Sai and I noticed we don't append this at all there and instead get it from the URL. I am assuming `URLQueryItem` is smart enough to notice a delicate param and doesn't add it, but we may want to check the URL we get here and if the ba or ec token already exists can just append the source and time like on Android.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -381,7 +383,7 @@ import BraintreeDataCollector
 
     private func launchPayPalApp(
         with payPalAppRedirectURL: URL,
-        baToken: String,
+        token: String? = nil,
```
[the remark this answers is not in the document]
scannillo: +1, since we're already doing a guard on line 369 for the token, this prob doesn't need to be nullable?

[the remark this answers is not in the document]
agedd: ahhh yep agreed, addressed this in 8e8275b5d7f414b3ecf4378cc6cc91bbe7dc7a06 and actually got rid of the `token` parameter since this is already being appended in the url returned to us from gw

Sources/BraintreePayPal/BTPayPalClient.swift
agedd: ahh i see! i just tested the vault flow on prod and noticed that this param does indeed get appended to the url - i've addressed this here: 8e8275b5d7f414b3ecf4378cc6cc91bbe7dc7a06

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -366,12 +366,14 @@ import BraintreeDataCollector
 
                 switch approvalURL.redirectType {
                 case .payPalApp(let url):
-                    guard let baToken = approvalURL.baToken else {
-                        self.notifyFailure(with: BTPayPalError.missingBAToken, completion: completion)
+                    guard let token = self.isVaultRequest ? approvalURL.baToken : approvalURL.ecToken else {
+                        self.notifyFailure(
+                            with: self.isVaultRequest ? BTPayPalError.missingBAToken : BTPayPalError.missingECToken,
+                            completion: completion
+                        )
```
agedd, line 373: note**: i wonder if it's still worthwhile to check for the existence of this token value since we're no longer passing it into `launchPayPalApp` 👀

saperi22: IMHO, if you no longer need the token, you can clean up the logic. We expect the backend to pass us the redirect url anyway.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -392,11 +393,11 @@ import BraintreeDataCollector
         )
 
         var urlComponents = URLComponents(url: payPalAppRedirectURL, resolvingAgainstBaseURL: true)
-        urlComponents?.queryItems = [
-            URLQueryItem(name: "ba_token", value: baToken),
+        let additionalQueryItems: [URLQueryItem] = [
             URLQueryItem(name: "source", value: "braintree_sdk"),
             URLQueryItem(name: "switch_initiated_time", value: String(Int(round(Date().timeIntervalSince1970 * 1000))))
         ]
+        urlComponents?.queryItems?.append(contentsOf: additionalQueryItems)
```
jaxdesmarais, line 400:
```suggestion

        urlComponents?.queryItems?.append(contentsOf: additionalQueryItems)
```

Sources/BraintreePayPal/BTPayPalError.swift
```diff
@@ -41,6 +41,9 @@ public enum BTPayPalError: Error, CustomNSError, LocalizedError, Equatable {
 
     /// 12. Missing BA Token for App Switch
     case missingBAToken
+    
+    /// 13. Missing EC Token for App Switch
+    case missingECToken
```
jaxdesmarais, line 46: Let's add this as 14 under the current 13

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: Agree we can probably leave it off since we aren't using it

saperi22 (approved):

Braintree.xcodeproj/xcshareddata/xcschemes/UnitTests.xcscheme
```diff
@@ -107,6 +107,11 @@
                BlueprintName = "BraintreePayPalTests"
                ReferencedContainer = "container:Braintree.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
+               <Test
+                  Identifier = "BTPayPalClient_Tests/testTokenizeVaultAccount_whenPayPalAppApprovalURLMissingBAToken_returnsError()">
```
jaxdesmarais, line 112: We should just delete this test vs skipping it if it's no longer relevant

Sources/BraintreePayPal/BTPayPalError.swift
```diff
@@ -41,9 +41,12 @@ public enum BTPayPalError: Error, CustomNSError, LocalizedError, Equatable {
 
     /// 12. Missing BA Token for App Switch
     case missingBAToken
-
+    
     /// 13. Missing PayPal Request
     case missingPayPalRequest
+    
+    /// 14. Missing EC Token for App Switch
+    case missingECToken
```
jaxdesmarais, line 49: It doesn't look like we return this error anywhere? Also if we removed the `missingBAToken` case we can remove it from our code and update the errors accordingly.

agedd: ahh that's true, updated in 2415a612dc6679830cd04ee4a01ad1765932e291

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -782,35 +782,6 @@ class BTPayPalClient_Tests: XCTestCase {
         }
     }
 
-    func testTokenizeVaultAccount_whenPayPalAppApprovalURLMissingBAToken_returnsError() {
```
scannillo, line 785: Why is this test removed? If the URL doesn't contain either a `token` or `ba_token`, I thought we still want to surface an error to the merchant and not proceed with opening the URL to the PayPal app

agedd: ahhh i thought it'd be ok to remove this test since we're longer checking for the existence of `token` or `ba_token` value before we attempt to app-switch since we expect gw to send us the redirectURL which would contain these params

on second thought, i'm open to re-adding this [check](https://github.com/braintree/braintree_ios/blob/67e0b5d6da3c456d845321a7368bbc6455c80980/Sources/BraintreePayPal/BTPayPalClient.swift#L369) and re-adding this test if we feel this may be a safer approach.

thoughts? @jaxdesmarais @saperi22 @scannillo

jaxdesmarais: Yeah, I feel like it may be safer to check that one of these exists. We should make a ticket to do the same on Android if we are not doing the same already.

agedd: updated in 7d5c59d3b6f4f468796d12212dcb6e4558370c8c

jaxdesmarais (approved):
🚀

PR #1507, opened 2025-01-29
[v7] Remove ObjC annotations from internal classes

agedd:
### Summary of changes

- Remove `@objc` annotations from several internal classes
- Remove and update related tests

### Checklist

- @agedd

### DIFF
--- a/Sources/BraintreeCore/BTAPIClientHTTPType.swift
+++ b/Sources/BraintreeCore/BTAPIClientHTTPType.swift
@@ -2,7 +2,7 @@ import Foundation
 
 /// :nodoc: This enum is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
 @_documentation(visibility: private)
-@objc public enum BTAPIClientHTTPService: Int {
+public enum BTAPIClientHTTPService: Int {
     /// Use the Gateway
     case gateway
 
--- a/Sources/BraintreeCore/BTClientMetadataIntegration.swift
+++ b/Sources/BraintreeCore/BTClientMetadataIntegration.swift
@@ -1,7 +1,6 @@
 /// :nodoc: This enum is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
 /// Integration Types
-@_documentation(visibility: private)
-@objc public enum BTClientMetadataIntegration: Int {
+public enum BTClientMetadataIntegration: Int {
     /// Custom
     case custom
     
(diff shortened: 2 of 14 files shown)


richherrera (approved):

Sources/BraintreeCore/BTAPIClientHTTPType.swift
```diff
@@ -2,7 +2,7 @@ import Foundation
 
 /// :nodoc: This enum is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
 @_documentation(visibility: private)
-@objc public enum BTAPIClientHTTPService: Int {
+public enum BTAPIClientHTTPService: Int {
```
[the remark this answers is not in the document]
agedd: hmm i think so, it seems to be referenced in `BTAPIClient` in the `get` and `post` methods as well as in `BTAPIRequest`

Sources/BraintreeCore/BTClientMetadataIntegration.swift
```diff
@@ -1,7 +1,7 @@
 /// :nodoc: This enum is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
 /// Integration Types
 @_documentation(visibility: private)
-@objc public enum BTClientMetadataIntegration: Int {
+public enum BTClientMetadataIntegration: Int {
```
warmkesselj, line 4: Do we want to keep the @objc annotations for anything marked with public?

agedd: hmm that's a good point - i think it's okay to drop the `objc` support for these public classes/enum because they were only exposed for DropIn SDK usgae, not for merchant use and now that DropIn is not compatible with BT iOS SDK since v6, it shouldn't be an issue.

jaxdesmarais: Correct, anything nodoc'd we can remove obj-c on since it's not meant to be used by merchants anyways.

UnitTests/BraintreeDataCollectorTests/BTDataCollector_Tests.swift
```diff
@@ -19,7 +19,9 @@ class BTDataCollector_Tests: XCTestCase {
                 let json = BTJSON(data: deviceData.data(using: String.Encoding.utf8)!)
                 XCTAssertNil(json["fraud_merchant_id"].asString())
                 XCTAssertNil(json["device_session_id"].asString())
-                XCTAssert((json["correlation_id"] as AnyObject).asString()!.count > 0)
+                if let correlationId = json["correlation_id"].asString() {
+                    XCTAssert(correlationId.count > 0)
```
jaxdesmarais, line 23: nit:
```suggestion
                if let correlationID = json["correlation_id"].asString() {
                    XCTAssert(correlationID.count > 0)
```

jaxdesmarais (approved):
One very minor comment but otherwise looks great!

PR #1510, opened 2025-02-04
Observability app switch

ngupta789:


jaxdesmarais:
We will be closing these PRs as we do not have context on these changes. Please reach out to #service-merchant-sdks in Slack so we can discuss the work to be done for this initiative.

PR #1512, opened 2025-02-04
ECS5 App Switch Feature Bug Fixes

agedd:
### Summary of changes

- Re-added the `token` parameter  and added logic to pass that into the `launchPayPalApp()` function
- Fixed parsing logic to read the `launchPayPalApp` boolean value from GW to determine the redirectType

### Checklist

- @agedd


scannillo (approved):

warmkesselj:
Looks like this unit test is failing: `testTokenizeVaultAccount_whenPayPalAppApprovalURLPresent_attemptsAppSwitchWithParameters`

jaxdesmarais (commented):
In addition to the failing test Justin mentioned, can we verify that there is a test for: `ba_token` present, `token` present, and error if neither exists. We may already have these cases covered but just wanted to make sure.

warmkesselj (approved):
LGTM

PR #1515, opened 2025-02-05
Create threeDSecure encodable

warmkesselj:
### Summary of changes

Transition to use a ThreeDSecure Encodable body for the post parameters.

### Checklist

- [x] Added a changelog entry

### Authors
@warmkesselj 
@jwarmkessel


Sources/BraintreeThreeDSecure/BTThreeDSecurePostBody.swift
```diff
@@ -0,0 +1,96 @@
+import Foundation
+
+// MARK: - BTThreeDSecurePostBody
+struct BTThreeDSecurePostBody: Encodable {
```
scannillo, line 4: Nit - can the docstring for this file include the POST path like the other Encodable types? Will hopefully make searching within the repo easier

Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift
```diff
@@ -352,7 +352,7 @@ import BraintreeCore
                 return
             }
 
-            let requestParameters = self.buildRequestDictionary(with: request)
```
scannillo, line 355: Can we remove the `buildRequestDictionary()` method now?

Sources/BraintreeThreeDSecure/BTThreeDSecurePostBody.swift
```diff
@@ -0,0 +1,96 @@
...
+            self.customFields = customFields
+        } else {
+            self.customFields = nil
+        }
+
+        if request.cardAddChallengeRequested {
+            self.cardAdd = true
+        } else {
+            self.cardAdd = nil
+        }
```
jaxdesmarais, line 39: Are we able to simplify this to something like `self.cardAdd = request.cardAddChallengeRequested ? true : nil`. We may be able to do the same for custom fields above.

Sources/BraintreeThreeDSecure/BTThreeDSecurePostBody.swift
```diff
@@ -0,0 +1,96 @@
...
+            self.billingState = request.billingAddress?.region
+            self.billingSurname = request.billingAddress?.surname
+            self.email = request.email
+            self.mobilePhoneNumber = request.mobilePhoneNumber
+            self.shippingMethod = request.shippingMethod.stringValue
+        }
+    }
+
+    // MARK: - Customer
+    struct Customer: Codable {}
```
jaxdesmarais, line 95:
```suggestion
    struct Customer: Codable { }
```

Sources/BraintreeThreeDSecure/BTThreeDSecurePostBody.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+
+/// The POST body for `v1/payment_methods/\(urlSafeNonce)/three_d_secure/lookup`
+struct BTThreeDSecurePostBody: Encodable {
```
richherrera, line 4: For the POST Encodable model, we are omitting the 'BT' prefix. I think we could omit it here as well.

```suggestion
struct ThreeDSecurePOSTBody: Encodable {
```

Sources/BraintreeThreeDSecure/BTThreeDSecurePostBody.swift
```diff
@@ -0,0 +1,87 @@
...
+        case customer
+        case dataOnlyRequested
+        case dfReferenceID = "dfReferenceId"
+        case exemptionRequested
+        case requestedExemptionType
+        case requestedThreeDSecureVersion
+    }
+    
+    // MARK: - AdditionalInfo
+    struct AdditionalInfo: Codable {
```
richherrera, line 52: I think we're missing quite a few parameters here. We have an object called `BTThreeDSecureAdditionalInformation`. We're missing this logic. Can we double-check?
https://github.com/braintree/braintree_ios/blob/666e6ff2662a608d642607f71c028e055d4e4df7/Sources/BraintreeThreeDSecure/BTThreeDSecureClient.swift#L445

These are all the ones we are missing
https://github.com/braintree/braintree_ios/blob/666e6ff2662a608d642607f71c028e055d4e4df7/Sources/BraintreeThreeDSecure/BTThreeDSecureAdditionalInformation.swift#L214

warmkesselj: Great catch here.

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureAdditionalInformation_Tests.swift
```diff
@@ -1,124 +0,0 @@
-import XCTest
-@testable import BraintreeThreeDSecure
-
-class BTThreeDSecureAdditionalInformation_Tests: XCTestCase {
```
richherrera, line 4: Perhaps we could move these tests to `BTThreeDSecureRequest_Tests`? We might consider having this type of test in some way to avoid losing functionality 🤔 
what do you think?

richherrera (approved):

Demo/Application/Features/ThreeDSecureViewController.swift
```diff
@@ -91,7 +91,7 @@ class ThreeDSecureViewController: PaymentButtonBaseViewController {
         cardFormView.cvvTextField.text = "123"
         cardFormView.postalCodeTextField.text = "12345"
     }
-
+
```
stechiu, line 94: nit: extra space

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureAdditionalInformation_Tests.swift
jaxdesmarais, line 4: I think we still want to check that they key value pairs for this are being encoded correctly.

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecurePostalAddress_Tests.swift
```diff
@@ -1,85 +0,0 @@
-import XCTest
-@testable import BraintreeThreeDSecure
-
-class BTThreeDSecurePostalAddress_Tests: XCTestCase {
```
jaxdesmarais, line 4: We will want to ensure these key/values are being encoded as expected.

warmkesselj: Does [this](https://github.com/braintree/braintree_ios/pull/1515/files/744e06349796628a632d8a6cc3a006cccb14483c#diff-73e63579d2fb724c1cede941c16997d0dbce14f3870e3ae395327ee13948c982R147) satisfy or do you think we need them broken out for each function?

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureAdditionalInformation_Tests.swift
warmkesselj: Does [this](https://github.com/braintree/braintree_ios/pull/1515/files/744e06349796628a632d8a6cc3a006cccb14483c#diff-73e63579d2fb724c1cede941c16997d0dbce14f3870e3ae395327ee13948c982R147) satisfy or do you think we need them broken out for each function?

jaxdesmarais: That just looks to be checking that our new encodable object has the value we expect. Instead we want to check that the key for the encodable object is nested with the correct value. So for example: `parameters["shippingGivenName"]` equals `"Given"`. Right now we aren't checking that the key `shippingGivenName` exists anywhere as far as I can tell.

warmkesselj: I see. 

It does exist here `BTThreeDSecureRequest_Tests`.
 
```
func testAsParameters_parameterizesAllProperties() {
...
XCTAssertEqual(params.additionalInfo.shippingGivenName, "Given")
...
```

However, I'll see if I can integrate more of these tests.

jaxdesmarais: In that test we are not verifying that `params.additionalInfo.shippingGivenName` is equal to the key `shippingGivenName`, which is what is missing. I am fine if you want to update `testAsParameters_parameterizesAllProperties` to actually check that the keys exist and are formatted as expected but currently this is a gap.

Sources/BraintreeThreeDSecure/ThreeDSecurePOSTBody.swift
```diff
@@ -0,0 +1,194 @@
...
+    let customer: Customer
+    let dataOnlyRequested: Bool
+    let dfReferenceID: String?
+    let exemptionRequested: Bool
+    let requestedExemptionType: String?
+    let requestedThreeDSecureVersion: String?
+
+    init(request: BTThreeDSecureRequest) {
+        self.requestedExemptionType = request.requestedExemptionType.stringValue
+        self.requestedThreeDSecureVersion = "2"
```
jaxdesmarais, line 21: Can we hradcode this at the top on like 17? I don't think it's currently optional so can for now have it hardcoded there.

Sources/BraintreeThreeDSecure/ThreeDSecurePOSTBody.swift
```diff
@@ -0,0 +1,194 @@
...
+    }
+    
+    // MARK: - AdditionalInfo
+    struct AdditionalInfo: Encodable {
+
+        let accountAgeIndicator: String?
+        let accountChangeDate: String?
+        let accountChangeIndicator: String?
+        let accountCreateDate: String?
+        let accountId: String?
```
jaxdesmarais, line 59:
```suggestion
        let accountID: String?
```

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecurePostalAddress_Tests.swift
jaxdesmarais, line 4: It looks like we may have missed these test cases. We have `shippingBillingName` but not `givenName` and `billingGivenName` from this test class.

jaxdesmarais: Yep, you're right, looks like we always append only `billing` or `shipping`. Thanks for adding that to the test!

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
```diff
@@ -39,7 +39,6 @@ struct PayPalCheckoutPOSTBody: Encodable {
     private var recipientName: String?
     
     // MARK: - Initializer
-
```
jaxdesmarais, line 42: nit, lets keep this space

UnitTests/BraintreeThreeDSecureTests/BTThreeDSecureRequest_Tests.swift
```diff
@@ -143,4 +143,166 @@ class BTThreeDSecureRequest_Tests: XCTestCase {
...
+        XCTAssertEqual(additionalInfoDict["installment"] as? String, "2")
+        XCTAssertEqual(additionalInfoDict["purchaseDate"] as? String, "20190401123456")
+        XCTAssertEqual(additionalInfoDict["recurringEnd"] as? String, "20201231")
+        XCTAssertEqual(additionalInfoDict["recurringFrequency"] as? String, "30")
+        XCTAssertEqual(additionalInfoDict["sdkMaxTimeout"] as? String, "10")
+        XCTAssertEqual(additionalInfoDict["workPhoneNumber"] as? String, "5551115555")
+    }
+
+    func testAsParameters_parameterizesWithNilProperties() {
+        let info = BTThreeDSecureAdditionalInformation()
```
jaxdesmarais, line 290: nit: can we rename this to `additionalInformation`?

jaxdesmarais (approved):
A couple of very minor comments but approving ahead of time

PR #1516, opened 2025-02-06
Expose `isDeviceToken` on BTApplePayCardNonce

stechiu:
With this PR, Merchants will have the ability to see if their ApplePay payment was tokenized as an MPAN or DPAN.

### Summary of changes

- Exposed `isDeviceToken`
- `false` == MPAN
- `true` == DPAN

### Checklist

- [x] Added a changelog entry

- @stechiu

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreeApplePay
+  * Add `isDeviceToken` to POST `v1/payment_methods/apple_payment_tokens`
+
 ## 6.28.0 (2025-02-05)
 * BraintreeVenmo
   * Allow universal links to be set without a return URL scheme (fixes #1505)
--- a/Sources/BraintreeApplePay/BTApplePayCardNonce.swift
+++ b/Sources/BraintreeApplePay/BTApplePayCardNonce.swift
@@ -10,11 +10,18 @@ import BraintreeCore
     /// The BIN data for the card number associated with this nonce.
     public let binData: BTBinData
 
+    /// This Boolean indicates whether this tokenized card is a device-specific account number (DPAN) or merchant/cloud token (MPAN).
+    /// If `isDeviceToken` is `false`, then token type is MPAN
+    public var isDeviceToken: Bool
+
     init?(json: BTJSON) {
         guard let nonce = json["nonce"].asString() else { return nil }
 
         let cardType = json["details"]["cardType"].asString() ?? "ApplePayCard"
         let isDefault = json["default"].isTrue
+        let isDeviceToken = json["details"]["isDeviceToken"].asBool() ?? false
+
+        self.isDeviceToken = isDeviceToken
 
         binData = BTBinData(json: json["binData"])
         super.init(nonce: nonce, type: cardType, isDefault: isDefault)
--- a/UnitTests/BraintreeApplePayTests/BTApplePayCardNonce_Tests.swift
+++ b/UnitTests/BraintreeApplePayTests/BTApplePayCardNonce_Tests.swift
@@ -12,7 +12,8 @@ class BTApplePayCardNonce_Tests: XCTestCase {
                     "commercial": "yes"
                 ],
                 "details": [
-                    "cardType": "fake-card-type"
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreeApplePay
+  * Add `isDeviceToken` to POST `v1/payment_methods/apple_payment_tokens`
```
stechiu, line 5: Unsure what to write here

scannillo: I would write something like "`Add `BTApplePayCardNonce.isDeviceToken` for MPAN identification"

scannillo (approved):
👏 Thanks for also including details on testing / seeing it working E2E

scannillo:
Actually, is the `false` you're seeing just a default - or is the API giving back false? We were expecting the API to always give back `true` unless the Demo app gets updated to specifically request an MPAN.

Sources/BraintreeApplePay/BTApplePayCardNonce.swift
```diff
@@ -10,11 +10,18 @@ import BraintreeCore
...
+    /// This Boolean indicates whether this tokenized card is a device-specific account number (DPAN) or merchant/cloud token (MPAN).
+    /// If `isDeviceToken` is `false`, then token type is MPAN
+    public var isDeviceToken: Bool
+
     init?(json: BTJSON) {
         guard let nonce = json["nonce"].asString() else { return nil }
 
         let cardType = json["details"]["cardType"].asString() ?? "ApplePayCard"
         let isDefault = json["default"].isTrue
+        let isDeviceToken = json["details"]["isDeviceToken"].asBool() ?? false
```
scannillo, line 22: I think we want to default to `true` instead. DPAN is the standard/current type, and MPAN is the newer type

stechiu: ohh ok no worries! I wasn't sure what the default should be or what the majority of transactions would be. Will update

stechiu:
> ❓ Actually, is the `false` you're seeing just an SDK default - or is the API giving back false? We were expecting the API to always give back `true` until the Demo app gets updated to specifically request an MPAN.
> 
> We will have to decide what value to use if there is NONE returned back from API - either true (is DPAN) or nil.

I believe the value is coming from the SDK. I'm not sure if I exposed the key properly for the value to come from API...

richherrera (approved):
LGTM!!

CHANGELOG.md
stechiu: Thanks! updated

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreeApplePay
+  * Add BTApplePayCardNonce.isDeviceToken for MPAN identification
```
```suggestion
  * Add `BTApplePayCardNonce.isDeviceToken` for MPAN identification
```

Sources/BraintreeApplePay/BTApplePayCardNonce.swift
```diff
@@ -10,11 +10,18 @@ import BraintreeCore
...
+    public var isDeviceToken: Bool
+
     init?(json: BTJSON) {
         guard let nonce = json["nonce"].asString() else { return nil }
 
         let cardType = json["details"]["cardType"].asString() ?? "ApplePayCard"
         let isDefault = json["default"].isTrue
+        let isDeviceToken = json["details"]["isDeviceToken"].asBool() ?? true
+
+        self.isDeviceToken = isDeviceToken
```
jaxdesmarais, line 24: I think we just extract the properties above because it'd make the super init a mess. Here I think we can do:
```suggestion

        self.isDeviceToken = json["details"]["isDeviceToken"].asBool() ?? true
```

jaxdesmarais (approved):
Couple of very minor comments but otherwise looks great!

stechiu:
> Couple of very minor comments but otherwise looks great!

Great thank you!

warmkesselj (approved):

agedd (approved):
lgtm! 🚀

PR #1517, opened 2025-02-07
Apple Pay - Added method to request MPAN in demo app

stechiu:
### Summary of changes

Our merchants will need to start adding additional details to their `PKPaymentRequest` in order to request MPANs. We want to set up our demo app as a merchant would and verify we get the expected response back from the gateway

### Checklist

- [ ] Added a changelog entry

- @stechiu

### DIFF
--- a/Demo/Application/Features/ApplePayViewController.swift
+++ b/Demo/Application/Features/ApplePayViewController.swift
@@ -5,6 +5,7 @@ import PassKit
 class ApplePayViewController: PaymentButtonBaseViewController {
 
     lazy var applePayClient = BTApplePayClient(apiClient: apiClient)
+    // swiftlint:disable:next force_unwrapping
     let managementURL = URL(string: "https://www.merchant.com/update-payment")!
 
     override func viewDidLoad() {
--- a/Demo/Application/Features/ApplePayViewController.swift
+++ b/Demo/Application/Features/ApplePayViewController.swift
@@ -5,6 +5,7 @@ import PassKit
 class ApplePayViewController: PaymentButtonBaseViewController {
 
     lazy var applePayClient = BTApplePayClient(apiClient: apiClient)
+    let managementURL = URL(string: "https://www.merchant.com/update-payment")!
 
     override func viewDidLoad() {
         super.viewDidLoad()
@@ -51,8 +52,6 @@ class ApplePayViewController: PaymentButtonBaseViewController {
 
     @available(iOS 16.0, *)
     private func recurringPaymentRequest() -> PKRecurringPaymentRequest {
-        let managementURL = URL(string: "https://www.merchant.com/update-payment")!
-
         let recurringPaymentRequest = PKRecurringPaymentRequest(
             paymentDescription: "Payment description.",
             regularBilling: PKRecurringPaymentSummaryItem(label: "Payment label", amount: 10.99),
--- a/Demo/Application/Features/ApplePayViewController.swift
+++ b/Demo/Application/Features/ApplePayViewController.swift
@@ -38,8 +38,10 @@ class ApplePayViewController: PaymentButtonBaseViewController {
             }
 
             let paymentRequest = self.constructPaymentRequest(with: request)
-            // swiftlint:disable:next force_unwrapping
-            let paymentAuthorizationViewController = PKPaymentAuthorizationViewController(paymentRequest: paymentRequest)!
+            guard let paymentAuthorizationViewController = PKPaymentAuthorizationViewController(paymentRequest: paymentRequest) else {
+                self.progressBlock("Could not create PKPaymentAuthorizationViewController")
+                return
+            }
             paymentAuthorizationViewController.delegate = self
 
             if #available(iOS 16.0, *) {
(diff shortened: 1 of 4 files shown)


Demo/Application/Features/ApplePayViewController.swift
```diff
@@ -40,6 +40,17 @@ class ApplePayViewController: PaymentButtonBaseViewController {
             let paymentAuthorizationViewController = PKPaymentAuthorizationViewController(paymentRequest: paymentRequest)!
             paymentAuthorizationViewController.delegate = self
 
+            guard let managementURL = URL(string: "https://www.merchant.com/update-payment") else { return }
```
scannillo, line 43: Do we have to do this before constructing the PKPaymentAuthorizationViewController? Might that explain why we aren't seeing `false` come back for `isDeviceToken`?

Demo/Application/Features/ApplePayViewController.swift
```diff
@@ -40,6 +40,17 @@ class ApplePayViewController: PaymentButtonBaseViewController {
...
+
+            if #available(iOS 16, *) {
+                let recurringPaymentRequest = PKRecurringPaymentRequest(
+                    paymentDescription: "Payment description.",
+                    regularBilling: PKRecurringPaymentSummaryItem(label: "Payment label", amount: 10.99),
+                    managementURL: managementURL
+                )
+                paymentRequest.recurringPaymentRequest = recurringPaymentRequest
```
warmkesselj, line 51: Nit: Should the title of PR be related to adding a recurring payment request?

warmkesselj:
Looks good from demo side.I can see a recurring payment "starting now" for 10.99.

Screenshot 2025-02-10 at 9 36 00 AM


Simulator Screenshot - iPhone 16 Pro - 2025-02-10 at 09 35 43

Demo/Application/Features/ApplePayViewController.swift
stechiu: I forgot to fill out the PR info section and I see where the title can be misleading. I've updated it to "Added method to request MPAN in demo app". `recurringPaymentRequest` is one of 3 payment request options that merchants can use

Demo/Application/Features/ApplePayViewController.swift
stechiu: I didn't know what would be the right default URL and didn't want to force unwrap. I just tested in simulator with an AMEX card and I'm able to get back `true`/`YES` now after gateway deployed their ticket that fixed an issue related to MPAN 

Screenshot 2025-02-10 at 2 50 40 PM

stechiu: But I'm open to suggestions if there's a better way to write that line!

Demo/Application/Features/ApplePayViewController.swift
```diff
@@ -40,6 +40,17 @@ class ApplePayViewController: PaymentButtonBaseViewController {
             let paymentAuthorizationViewController = PKPaymentAuthorizationViewController(paymentRequest: paymentRequest)!
```
warmkesselj, line 40: A bit out of scope but do we want to add a guard here to avoid the force unwrap? 

```
        guard let paymentAuthVC = PKPaymentAuthorizationViewController(paymentRequest: paymentRequest) else {
            self.progressBlock("Could not create PKPaymentAuthorizationViewController")
            return
        }
```

Demo/Application/Features/ApplePayViewController.swift
```diff
@@ -40,6 +40,17 @@ class ApplePayViewController: PaymentButtonBaseViewController {
...
             paymentAuthorizationViewController.delegate = self
 
+            guard let managementURL = URL(string: "https://www.merchant.com/update-payment") else { return }
+
+            if #available(iOS 16, *) {
```
warmkesselj, line 45: What do you think about moving the logic to create a PKRecurringPaymentRequest into a separate function?

Demo/Application/Features/ApplePayViewController.swift
stechiu: Ohh ok, sorry, misread. I moved `managementURL`

Demo/Application/Features/ApplePayViewController.swift
stechiu: Thanks good point!

stechiu: Looks like the downside is that `delegate` can't be referenced
Screenshot 2025-02-11 at 1 19 00 PM

Demo/Application/Features/ApplePayViewController.swift
stechiu: That would be cleaner way to write. Updated

Demo/Application/Features/ApplePayViewController.swift
warmkesselj: Looks like `paymentAuthorizationViewController` is being used instead of `paymentAuthVC`. Will that change anything?

warmkesselj (approved):
There are unchecked items on the checklist. I think maybe change log is irrelevant. But approving ahead of time.

PR #1523, opened 2025-02-14
[DTINAPPXO-511] Implemented PayPal App Switch fallback to external browser

abalusupp:
Bug ticket: DTINAPPXO-511

Bug Fix doc: "Top of Funnel App Switch Fail"

### Summary of changes

**Issue**: The PayPal app fails to open the app switch checkout URL, logging the error: "UIApplication failed to perform app switch to PayPal."

**Change**: Makes sure UIApplication only opens universal link only if the URL is a valid universal link and there is an installed app capable of opening that URL and Implemented a fallback mechanism to open the "approvalUrl" in an external browser

### Checklist

- [ ] Added a changelog entry

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -271,9 +271,8 @@ import BraintreeDataCollector
         webSessionReturned = true
     }
     
-    func handlePayPalRequest(
+    func handlePayPalInAppBrowserSwitch(
         with url: URL,
-        paymentType: BTPayPalPaymentType,
         completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void
     ) {
         // Defensive programming in case PayPal returns a non-HTTP URL so that ASWebAuthenticationSession doesn't crash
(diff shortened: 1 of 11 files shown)


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -422,32 +449,52 @@ import BraintreeDataCollector
...
         
-        guard let redirectURL = urlComponents?.url else {
+        guard let appSwitchUrl = urlComponents?.url else {
             self.notifyFailure(with: BTPayPalError.invalidURL("Unable to construct PayPal app redirect URL."), completion: completion)
             return
         }
 
-        application.open(redirectURL) { success in
-            self.invokedOpenURLSuccessfully(success, url: redirectURL, completion: completion)
+        application.open(appSwitchUrl, withOptions: [.universalLinksOnly: true]) { didOpened in
```
scannillo, line 464: IMO we should fallback to the WebView flow in the error case of UIApplication.open() --> false, instead of attempt to open the universal link again

abalusupp: That was our initial thought, but the billing agreement is already configured for the scenario where the app is installed. If we open the universal link in an in-app web session, ModXo tries to open the redirect universal link (which is received by the app/scene delegate) instead of the schema, which breaks the webSession flow by not cancelling/dismissing it

jaxdesmarais:
Closing due to internal discussion around product requirements for this work. Please reach out in internal slack #service-merchant-sdks for planning and scoping this work.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -422,32 +449,52 @@ import BraintreeDataCollector
...
+            if didOpened {
+                self.invokedOpenURLSuccessfully(appSwitchUrl, completion: completion)
+            } else {
+                if let fallbackUrl = config.fallbackUrl {
+                    self.handleFallbackToExternalBrowser(forUrl: fallbackUrl, completion: completion)
```
jaxdesmarais, line 469: If we decide to go forward with a similar PR we should instead use `ASWebAuthenticationSession` for the fallback scenario (configured here for the non app switch flow). If our hypothesis is that `application.open` is failing to open at the OS level I do not think we should just try to open a different URL with the same method as that's likely to also fail.

PR #1527, opened 2025-02-20
[v7] Make APIClient Non-optional & Update `BTAmericanExpressClient` Init

jaxdesmarais:
### Summary of changes

- Update initializer to `BTAmericanExpressClient(authorization:)`
- Add `InvalidAuthorization` type - this matches the [pattern used on Android](https://github.com/braintree/braintree_android/blob/main/BraintreeCore/src/main/java/com/braintreepayments/api/core/InvalidAuthorization.kt)
- Update `BTAPIClient` initializer to require an authorization string
- Add new `BTAPIClientError.invalidAuthorization`
- Update `PaymentButtonBaseViewController` with new authorization string
- Update `AmexViewController` demo app with new initializer
- Update unit and integration tests
- Add `TODO`s for final PR cleanup

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/Demo/Application/Base/PaymentButtonBaseViewController.swift
+++ b/Demo/Application/Base/PaymentButtonBaseViewController.swift
@@ -3,6 +3,7 @@ import BraintreeCore
 
 class PaymentButtonBaseViewController: BaseViewController {
 
+    // TODO: remove API client in final PR
     let apiClient: BTAPIClient
     let authorization: String
 
--- a/Demo/Application/Base/PaymentButtonBaseViewController.swift
+++ b/Demo/Application/Base/PaymentButtonBaseViewController.swift
@@ -4,6 +4,7 @@ import BraintreeCore
 class PaymentButtonBaseViewController: BaseViewController {
 
     let apiClient: BTAPIClient
+    let authorization: String
 
     var heightConstraint: CGFloat?
 
@@ -12,6 +13,7 @@ class PaymentButtonBaseViewController: BaseViewController {
     override init(authorization: String) {
         // swiftlint:disable:next force_unwrapping
         apiClient = BTAPIClient(authorization: authorization)!
+        self.authorization = authorization
         super.init(authorization: authorization)
     }
     
--- a/Demo/Application/Features/AmexViewController.swift
+++ b/Demo/Application/Features/AmexViewController.swift
@@ -4,7 +4,7 @@ import BraintreeCard
 
 class AmexViewController: PaymentButtonBaseViewController {
 
-    lazy var amexClient = BTAmericanExpressClient(apiClient: apiClient)
+    lazy var amexClient = BTAmericanExpressClient(authorization: authorization)
     lazy var cardClient = BTCardClient(apiClient: apiClient)
 
     override func viewDidLoad() {
--- a/SampleApps/CarthageTest/CarthageTest/ViewController.swift
+++ b/SampleApps/CarthageTest/CarthageTest/ViewController.swift
@@ -16,7 +16,7 @@ class ViewController: UIViewController {
     override func viewDidLoad() {
         let apiClient = BTAPIClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")!
 
-        let amexClient = BTAmericanExpressClient(apiClient: apiClient)
+        let amexClient = BTAmericanExpressClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")
         let applePayClient = BTApplePayClient(apiClient: apiClient)
         let cardClient = BTCardClient(apiClient: apiClient)
         let dataCollector = BTDataCollector(apiClient: apiClient)
--- a/SampleApps/CarthageTest/CarthageTest/ViewController.swift
+++ b/SampleApps/CarthageTest/CarthageTest/ViewController.swift
@@ -14,6 +14,7 @@ import BraintreeSEPADirectDebit
 class ViewController: UIViewController {
 
     override func viewDidLoad() {
+        // TODO: remove in the final PR for making authorization internal
         let apiClient = BTAPIClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")!
 
         let amexClient = BTAmericanExpressClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")
--- a/SampleApps/SPMTest/SPMTest/ViewController.swift
+++ b/SampleApps/SPMTest/SPMTest/ViewController.swift
@@ -16,7 +16,7 @@ class ViewController: UIViewController {
     override func viewDidLoad() {
         let apiClient = BTAPIClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")!
 
-        let amexClient = BTAmericanExpressClient(apiClient: apiClient)
+        let amexClient = BTAmericanExpressClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")
         let applePayClient = BTApplePayClient(apiClient: apiClient)
         let cardClient = BTCardClient(apiClient: apiClient)
         let dataCollector = BTDataCollector(apiClient: apiClient)
--- a/SampleApps/SPMTest/SPMTest/ViewController.swift
+++ b/SampleApps/SPMTest/SPMTest/ViewController.swift
@@ -14,6 +14,7 @@ import BraintreeSEPADirectDebit
 class ViewController: UIViewController {
 
     override func viewDidLoad() {
+        // TODO: remove in the final PR for making authorization internal
         let apiClient = BTAPIClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")!
 
         let amexClient = BTAmericanExpressClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")
--- a/Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
+++ b/Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
@@ -11,7 +11,7 @@ import BraintreeCore
     var apiClient: BTAPIClient
     
     ///  Creates an American Express client.
-    /// - Parameter authorization: Your client token or tokenization key
+    /// - Parameter authorization: A valid client token or tokenization key used to authorize API calls
     @objc(initWithAPIClient:)
     public init(authorization: String) {
         self.apiClient = BTAPIClient(newAuthorization: authorization)
--- a/Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
+++ b/Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
@@ -11,7 +11,7 @@ import BraintreeCore
     var apiClient: BTAPIClient
     
     ///  Creates an American Express client.
-    /// - Parameter authorization: A client token or tokenization key
+    /// - Parameter authorization: Your client token or tokenization key
     @objc(initWithAPIClient:)
     public init(authorization: String) {
         self.apiClient = BTAPIClient(newAuthorization: authorization)
--- a/Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
+++ b/Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
@@ -10,10 +10,10 @@ import BraintreeCore
     private let apiClient: BTAPIClient
     
     ///  Creates an American Express client.
-    /// - Parameter apiClient: An instance of `BTAPIClient`
+    /// - Parameter authorization: A client token or tokenization key
     @objc(initWithAPIClient:)
-    public init(apiClient: BTAPIClient) {
-        self.apiClient = apiClient
+    public init(authorization: String) {
+        self.apiClient = BTAPIClient(newAuthorization: authorization)
     }
     
     ///  Gets the rewards balance associated with a Braintree nonce. Only for American Express cards.
--- a/Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
+++ b/Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
@@ -7,7 +7,8 @@ import BraintreeCore
 ///  `BTAmericanExpressClient` enables you to look up the rewards balance of American Express cards.
 @objc public class BTAmericanExpressClient: NSObject {
     
-    private let apiClient: BTAPIClient
+    /// exposed for testing
+    var apiClient: BTAPIClient
     
     ///  Creates an American Express client.
     /// - Parameter authorization: A client token or tokenization key
--- a/Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
+++ b/Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
@@ -1,0 +1,16 @@
+import Foundation
+
+/// An invalid authorization type
+class InvalidAuthorization: ClientAuthorization {
+    
+    let type = AuthorizationType.invalidAuthorization
+    let configURL: URL
+    let bearer: String
+    let originalValue: String
+    
+    init(_ rawValue: String) {
+        self.bearer = rawValue
+        self.originalValue = rawValue
+        self.configURL = URL(string: "https://example.com")!
+    }
+}
--- a/Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
+++ b/Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
@@ -11,6 +11,8 @@ class InvalidAuthorization: ClientAuthorization {
     init(_ rawValue: String) {
         self.bearer = rawValue
         self.originalValue = rawValue
+        
+        // swiftlint:disable:next force_unwrapping
         self.configURL = URL(string: "https://example.com")!
     }
 }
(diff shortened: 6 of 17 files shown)


Demo/Application/Base/PaymentButtonBaseViewController.swift
```diff
@@ -3,7 +3,9 @@ import BraintreeCore
 
 class PaymentButtonBaseViewController: BaseViewController {
 
+    // TODO: remove API client in final PR
```
jaxdesmarais, line 6: Added some TODOs throughout this PR to update in our final PR for this work

Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
```diff
@@ -0,0 +1,16 @@
...
+    
+    let type = AuthorizationType.invalidAuthorization
+    let configURL: URL
+    let bearer: String
+    let originalValue: String
+    
+    init(_ rawValue: String) {
+        self.bearer = rawValue
+        self.originalValue = rawValue
+        self.configURL = URL(string: "https://example.com")!
```
jaxdesmarais, line 14: I went back and fourth on what to do here, the `configURL` is required as part of the `ClientAuthorization` protocol. We will never hit this dummy URL since we throw an error when fetching the config, but open to other ideas if folks feel differently about how we should handle this. We would need to unwrap the `configURL` in several places if we made this optional in the protocol.

Demo/Application/Features/AmexViewController.swift
```diff
@@ -4,7 +4,7 @@ import BraintreeCard
 
 class AmexViewController: PaymentButtonBaseViewController {
 
-    lazy var amexClient = BTAmericanExpressClient(apiClient: apiClient)
+    lazy var amexClient = BTAmericanExpressClient(authorization: authorization)
```
jaxdesmarais, line 7: Changing this to something like this will allow you to see an error returned:
```
lazy var amexClient = BTAmericanExpressClient(authorization: "badAuth")
```

Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
warmkesselj: Down for a pair session to discuss.

jaxdesmarais: Feel free to drop ideas here. Since there is no way for us to ever actually use this URL it seemed safe to hardcode a dummy URL but wanted to add a comment incase folks had thoughts/questions since it seems odd without additional context.

sarahkoop: IMO it should maybe be a domain we own at least - so braintree or paypal or is it a valid url to just use `https://`? Or maybe point to our dev docs?

jaxdesmarais:
UI tests are expected to fail currently, we have ticket DTMOBILES-1372 to address those failures.

Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
jaxdesmarais: I'm down to use one of those, do we have a preference between BT, PayPal, or dev docs? It should never be hit but agree would be good to use one that we own.

Demo/Application/Base/PaymentButtonBaseViewController.swift
```diff
@@ -12,6 +14,7 @@ class PaymentButtonBaseViewController: BaseViewController {
     override init(authorization: String) {
         // swiftlint:disable:next force_unwrapping
         apiClient = BTAPIClient(authorization: authorization)!
+        self.authorization = authorization
```
warmkesselj, line 17: Curious if there's something planned for this string at a later date?

Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
warmkesselj: Frankly, I like your idea of making it an optional.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -65,6 +68,31 @@ import Foundation
...
+    // TODO: rename param to authorization in final PR - set as newAuthorization currently since otherwise the two inits have the same signature
+    /// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    /// Initialize a new API client.
+    /// - Parameter authorization: Your tokenization key or client token.
+    @_documentation(visibility: private)
+    public init(newAuthorization: String) {
+        self.authorization = Self.authorization(from: newAuthorization)
+        self.metadata = BTClientMetadata()
+                
+        let btHTTP = BTHTTP(authorization: self.authorization)
```
warmkesselj, line 81: Take it or leave it: 
```
http = BTHTTP(authorization: self.authorization)
```

Demo/Application/Features/AmexViewController.swift
warmkesselj: Might be a stupid question but why `authorization` versus `clientAuthToken` or `clientToken`.

Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
warmkesselj: Could potentially change ClientAuthorization to a class. And remove InvalidAuthorization,  TokenizationKey, and ClientAuthorization.  And consolidate the logic.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -65,6 +68,31 @@ import Foundation
...
+        http = btHTTP
+        configurationLoader = ConfigurationLoader(http: btHTTP)
+        
+        super.init()
+        
+        analyticsService.setAPIClient(self)
```
warmkesselj, line 87: I wonder for better overall maintenance if some of these were broken out into functions and reused for both `public init(newAuthorization: String) {` and `public init?(authorization: String) {`

Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
jaxdesmarais: I think having the 3 classes is helpful as it allows us to enforce requirements. Since this URL is never hit or used in any way (we return an error if the authorization type is invalid) I am leaning towards hardcoding a URL we own. It will never be used in any way and is just enforced from the protocol. Locally I messed around with making it optional in the protocol and it makes the rest of our code much messier when we need to access the URL (which will always exist for client tokens and tokenization keys). I've updated the URL to `paypal.com` and added a docstring to clarify: https://github.com/braintree/braintree_ios/pull/1527/commits/0ae85a7395d573df4a2da1932996fff6d14924f3

Demo/Application/Features/AmexViewController.swift
jaxdesmarais: The authorization can be a client token or tokenization key so using `clientAuthToken` or `clientToken` would be inaccurate. We are also using the same [string on Android](https://github.com/braintree/braintree_android/blob/8d51b5da585260c740dfbdfeff322face580d81e/PayPal/src/main/java/com/braintreepayments/api/paypal/PayPalClient.kt#L69) and currently construct the `BTAPIClient` with `authorization` so it should be familiar to what merchants do today, just moving the `BTAPIClient` to be constructed by the SDK.

Demo/Application/Base/PaymentButtonBaseViewController.swift
jaxdesmarais: It's being used in this PR! If you put a breakpoint here you will see that this is being set to the authorization string from the init to set the authorization on the amex client.

Sources/BraintreeCore/BTAPIClient.swift
jaxdesmarais: We will be removing `public init?(authorization: String) {` in the final PR for this work as indicated in the TODO.

warmkesselj (approved):

Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+/// An invalid authorization type
+class InvalidAuthorization: ClientAuthorization {
+    
+    let type = AuthorizationType.invalidAuthorization
+    let configURL: URL
+    let bearer: String
+    let originalValue: String
```
scannillo, line 9: None of these values are actually going to be used / we're just setting these properties to conform to the protocol but never using them functionally. 

I crossed this bridge when messing around with this in [this branch](https://github.com/braintree/braintree_ios/compare/v7...spike-api-client-std-init). I wonder if there's a cleaner pattern.

V7_MIGRATION.md
```diff
@@ -68,3 +69,10 @@ For the App Switch flow, you must update your `info.plist` with a simplified URL
 ## PayPal Native Checkout
 The PayPal Native Checkout integration is no longer supported. Please remove it from your app and 
 use the [PayPal (web)](https://developer.paypal.com/braintree/docs/guides/paypal/overview/ios/v6) integration.
+
+## American Express
+Update initializer for `BTAmericanExpressClient`:
+```diff
+-  var amexClient = BTAmericanExpressClient(apiClient: apiClient)
++   var amexClient = BTAmericanExpressClient(authorization: authorization)
+```
```
scannillo, line 78: 🥹 _tears of joy_

Nit:
```suggestion
```diff
-  var amexClient = BTAmericanExpressClient(apiClient: apiClient)
+   var amexClient = BTAmericanExpressClient(authorization: "<CLIENT_AUTHORIZATION>")
```
```

File: Sources/BraintreeCore/BTAPIClient.swift [hunk 9]
```diff
@@ -93,20 +121,28 @@ import Foundation
...
-
-        Task { @MainActor in
-            do {
-                let configuration = try await configurationLoader.getConfig()
-                setupHTTPCredentials(configuration)
-                completion(configuration, nil)
-            } catch {
-                completion(nil, error)
+        
+        if authorization.type == .invalidAuthorization {
```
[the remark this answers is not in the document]
scannillo: We _could_ miss something by nesting it within `fetchOrReturnRemoteConfiguration`. Actually, this PR as-is misses doing auth string validation for `shopperInsightsClient.getRecommendedPaymentMethods()` calls, since we it hits `apiClient.post()` directly.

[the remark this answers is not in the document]
scannillo: IMO it will be better to call the auth string checks explicitly, vs relying on `fetchOrReturnRemoteConfiguration()`, which is already kind of confusing

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -93,20 +121,28 @@ import Foundation
...
+            completion(nil, BTAPIClientError.invalidAuthorization(authorization.originalValue))
+        } else {
+            Task { @MainActor in
```
scannillo, line 128: Can you explain why we're dispatching to `MainActor` in our network layer? I feel like that should be up to the merchant when they get our async result from `tokenize()` with either an error or success

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
```diff
@@ -7,13 +7,14 @@ import BraintreeCore
 ///  `BTAmericanExpressClient` enables you to look up the rewards balance of American Express cards.
 @objc public class BTAmericanExpressClient: NSObject {
     
-    private let apiClient: BTAPIClient
+    /// exposed for testing
+    var apiClient: BTAPIClient
     
     ///  Creates an American Express client.
-    /// - Parameter apiClient: An instance of `BTAPIClient`
+    /// - Parameter authorization: Your client token or tokenization key
```
richherrera, line 14: very nit, take it or leave it:

```suggestion
    /// - Parameter authorization: A valid client token or tokenization key used to authorize API calls
```

richherrera (approved):
LGTM!!! 🚀

Sources/BraintreeCore/BTAPIClient.swift
jaxdesmarais: This is not a change in code, it's just moving this logic into an else. But our code expects fetch or return config to be run on the main thread according to this test: https://github.com/braintree/braintree_ios/blob/258ae49bede22af809e229f9ba6486656955408c/UnitTests/BraintreeCoreTests/BTAPIClient_Tests.swift#L266

Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
jaxdesmarais: Potentially but I do think there is some value in enforcing them for the valid auth types. Since they won't be used it seems harmless. We also do something [similar on Android](https://github.com/braintree/braintree_android/blob/8d51b5da585260c740dfbdfeff322face580d81e/BraintreeCore/src/main/java/com/braintreepayments/api/core/InvalidAuthorization.kt#L4).

Sources/BraintreeCore/BTAPIClient.swift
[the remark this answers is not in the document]
jaxdesmarais: Ah, good catch, yes we do miss that currently. By call the auth string checks explicitly what do you mean? We want to make the network request in `fetchOrReturnRemoteConfiguration` before returning an error and move these to the get/post/etc requests?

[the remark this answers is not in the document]
jaxdesmarais: Ah, nvm we can still do them before that network call. Updated here: https://github.com/braintree/braintree_ios/pull/1527/commits/15bed089b3ea90137feb8ed9e4c3bbb8834ff73c and checked that `getRecommendedPaymentMethods` would hit that block as well.

agedd (approved):
lgtm!!

SampleApps/SPMTest/SPMTest/ViewController.swift
```diff
@@ -16,7 +16,7 @@ class ViewController: UIViewController {
     override func viewDidLoad() {
         let apiClient = BTAPIClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")!
```
scannillo, line 17: Same here, take it or leave it

Sources/BraintreeCore/Authorization/InvalidAuthorization.swift
```diff
@@ -0,0 +1,20 @@
...
+    
+    init(_ rawValue: String) {
+        self.bearer = rawValue
+        self.originalValue = rawValue
+      
+        // swiftlint:disable force_unwrapping
+        /// This URL is never used in the SDK as we always return an error if the authorization type is `.invalidAuthorization`
+        /// before construting or using the `configURL` in any way
+        self.configURL = URL(string: "https://paypal.com")!
```
scannillo, line 18: Nit - should we make a note that we want to re-work this in a minor version? Also could help noting that we're forced to set this URL due to protocol conformance

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 
+// swiftlint:disable type_body_length file_length
```
scannillo, line 3: Random - do we want to revisit some of our lint rules? Or start refactoring to abide by the rules sometime? I notice these coming in more often

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -34,7 +35,7 @@ import Foundation
     public init?(authorization: String) {
```
scannillo, line 35: Do we want a TODO to remove this init?

Sources/BraintreeCore/BTAPIClientError.swift
```diff
@@ -36,6 +50,9 @@ public enum BTAPIClientError: Int, Error, CustomNSError, LocalizedError, Equatab
             
         case .failedBase64Encoding:
             return "Unable to base64 encode the authorization string."
+            
+        case .invalidAuthorization(let authorization):
+            return "Invalid authorization provided: \(authorization)."
```
scannillo, line 55: Is this same error message we surface on Android? Would be nice to align

scannillo (approved):
WOOHOO! Some minor comments, non blocking

Sources/BraintreeCore/BTAPIClient.swift
jaxdesmarais: We _should_ be able to remove both of these once we remove the old init. I think the new init pushed us over the limit, but agree we should have a plan for addressing linter rules we seem to disable often.

SampleApps/CarthageTest/CarthageTest/ViewController.swift
```diff
@@ -16,7 +16,7 @@ class ViewController: UIViewController {
     override func viewDidLoad() {
         let apiClient = BTAPIClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")!
```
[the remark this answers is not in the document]
jaxdesmarais: Added all the TODOs here: https://github.com/braintree/braintree_ios/pull/1527/commits/ef86f3e034dcc4d63bcb0dc66201af458c129082

Sources/BraintreeCore/BTAPIClientError.swift
jaxdesmarais: Updated to align a bit more. I do think it's helpful to pass through the auth the merchant passed in the message so we will diverge a bit there: https://github.com/braintree/braintree_ios/pull/1527/commits/984d65682ec27a56cc1882de4fb4e1f97ac3064f

PR #1533, opened 2025-03-05
Add static framework binaries to GH release

scannillo:
### Summary of changes

- This PR adds an additional zip directory to each GitHub release that contains statically built xcframeworks.
    - Currently our GH releases ([example here](https://github.com/braintree/braintree_ios/releases/tag/6.29.0)) include a `Braintree.xcframework.zip` of dynamic frameworks. This PR adds a `Braintree_Static.xcframework.zip` of static frameworks.
- Why?
    - Static frameworks offer faster launch times since the code is linked directly into the app executable vs linked at runtime, though this comes with the tradeoff of increasing app size. Dynamic frameworks offer the inverse benefit. 
    - We have a request to make static versions of our binaries available for merchants who don't use a package manager.
        - See "Uber Static Framework Solution" on Confluence. 

### Testing

- [X] I ran the added command in terminal to verify that all xcframework binaries properly built
- [X] I then used the instructions in our sdk-knowledge repo to ensure the `file` type of each reads: `current ar archive` (static)
- [X] I imported these xcframeworks into a sample Xcode application, initialized APIClients, + ran on sim

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo


scannillo:
Leaving in draft until group consensus is decided that we do want to add these binaries to our releases. 

IMO I don't see much harm in adding them. It will slow down our release script (they take as long to build as the dynamic/carthage binaries).

jaxdesmarais:
Also went through and ran these steps manually without issue. I assume the zip files from the step `zip -r Braintree_Static.xcframework.zip Braintree_Static` and the directory `Braintree_Static` that showed as a diff locally get cleaned up automatically since they also exist on the carthage build above and we don't currently run into issues with that.

I initialized all of the clients and tested a few basic methods (card tokenize, amex get rewards balance, data collector) in a small sample project to spot check things were working as expected. Built the project for both simulator and on device (iPhone 15) without issue. Moving this out of draft based on mine and Sammy's testing.

If other folks want to test a couple of "gotchas" I ran into that may be helpful:

1. You'll need to delete the comments to run the steps from GitHub Actions via command line
2. I ran into an unrecognized selector error in our analytics after 15 seconds that required me to add `-ObjC` and `-all_load` to the "other linker flags" build setting ([stackoverflow](https://stackoverflow.com/questions/10442653/unrecognized-selector-calling-category-method-in-static-ios-library))
3. Static frameworks need to be set with "Embed and sign" in the "Frameworks, Libraries, and Embedded Content"

scannillo:
> You'll need to delete the comments to run the steps from GitHub Actions via command line

Maybe a longer-term/future PR improvement will be to use separate yml files and have each command be it's own separate `run` block. Same for our reference docs step as well. Currently, if 1 thing fails the CI is harder to read which step it was.

>  I assume the zip files from the step zip -r Braintree_Static.xcframework.zip Braintree_Static and the directory Braintree_Static that showed as a diff locally get cleaned up automatically since they also exist on the carthage build above and we don't currently run into issues with that.

Yeh, this crossed my mind to of do we want to clean up these file leftovers. You're right I don't see us do it for the step above so I don't think it will be an issue. But up to you if you think it's cleaner!

agedd (approved):
this is super cool! lgtm 🚀

jaxdesmarais:
> Maybe a longer-term/future PR improvement will be to use separate yml files and have each command be it's own separate `run` block. Same for our reference docs step as well. Currently, if 1 thing fails the CI is harder to read which step it was.

That's a good idea - I know on Android it's separated out into each step which is helpful, without needing to click into things.

> Yeh, this crossed my mind to of do we want to clean up these file leftovers. You're right I don't see us do it for the step above so I don't think it will be an issue. But up to you if you think it's cleaner!

I also don't think it will be an issue, but we can cross cleaning it up if it becomes one. For now can leave as is since it doesn't exist above.

PR #1535, opened 2025-03-06
Update Cocoapods to End of Release

jaxdesmarais:
### Summary of changes

Move the cocoapods portion of the release to the last step. We’ve been seeing failures with this step recently and the reference docs are the step after. The reference docs are one of the mere tedious steps to release manually. Due to cocoapods being deprecated we can likely expect the cocapods portion of the release to fail more often (as we have seen). This allows us to hopefully have the least amount of steps to do manually if the release fails as cocoapods is often our failing step. We can then release cocoapods via the command line when we see these failures without also needing to do the reference docs.

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -88,11 +88,6 @@ jobs:
           asset_name: Braintree.xcframework.zip
           asset_content_type: application/zip
 
-      - name: Publish to CocoaPods
-        env:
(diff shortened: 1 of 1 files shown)


richherrera (approved):
Thanks!!!

warmkesselj (approved):

agedd (approved):
🚀

PR #1538, opened 2025-03-10
Bug Fix: Universal Links in Same View

jaxdesmarais:
### Summary of changes

- Fix bug where instantiating multiple feature clients with  universal links in the same view causes `handleOpen` to return in the wrong feature client

#### Testing
On main you can see this bug by adding the following in `PayPalWebCheckoutViewController` and tapping the app switch flow:
```swift
/// add under `lazy var payPalClient`
var venmoClient: BTVenmoClient?

/// add to `tappedPayPalAppSwitchForVault`
venmoClient = BTVenmoClient(
    apiClient: apiClient,
    universalLink: URL(string: "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments")!
)

payPalClient.tokenize(request) { nonce, error in
...
```

If you add a breakpoint to `canHandleReturnURL` in `BTVenmoClient` you will see we hit that breakpoint but not the one in `BTPayPalClient`. This is because our logic in `BTAppContextSwitcher.handleOpen()` checks `canHandleReturnURL` first and then proceeds with only checking this on the first client if `true` - since we are currently only checking `https` and `cancel/success/error` it returns `true` and attempts to proceed with `handleReturnURL` in the wrong feature client. 

To resolve this we can add either `braintreeAppSwitchPayPal` or `braintreeAppSwitchVenmo` to the universal link for the respective flows and check for that as part of the `isValid` check run when `canHandleReturnURL` is called. Since merchants are instructed to wildcard the URLs this should not break existing flows. The full URL with our paths will also be passed to upstream services.

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -93,7 +93,10 @@ import BraintreeDataCollector
     @objc(initWithAPIClient:universalLink:)
     public convenience init(apiClient: BTAPIClient, universalLink: URL) {
         self.init(apiClient: apiClient)
-        self.universalLink = universalLink.appendingPathComponent("payPal")
+        
+        /// appending a PayPal app switch specific path to verify we are in the correct flow when
+        /// `canHandleReturnURL` is called
+        self.universalLink = universalLink.appendingPathComponent("braintreeAppSwitchPayPal")
     }
 
     // MARK: - Public Methods
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -93,7 +93,7 @@ import BraintreeDataCollector
     @objc(initWithAPIClient:universalLink:)
     public convenience init(apiClient: BTAPIClient, universalLink: URL) {
         self.init(apiClient: apiClient)
-        self.universalLink = universalLink
+        self.universalLink = universalLink.appendingPathComponent("payPal")
     }
 
     // MARK: - Public Methods
--- a/Sources/BraintreePayPal/BTPayPalReturnURL.swift
+++ b/Sources/BraintreePayPal/BTPayPalReturnURL.swift
@@ -39,9 +39,7 @@ struct BTPayPalReturnURL {
     /// - Parameter url: an app switch or ASWebAuthenticationSession return URL
     /// - Returns: `true` if the url represents a valid PayPal app switch return
     static func isValid(_ url: URL) -> Bool {
-        url.scheme == "https"
-        && (url.path.contains("success") && (url.absoluteString.contains("token") || url.absoluteString.contains("ba_token")))
-        || (url.path.contains("cancel"))
+        url.scheme == "https" && (url.path.contains("cancel") || url.path.contains("success")) && url.path.contains("payPal")
     }
 
     static func isValidURLAction(url: URL, linkType: LinkType?) -> Bool {
--- a/Sources/BraintreePayPal/BTPayPalReturnURL.swift
+++ b/Sources/BraintreePayPal/BTPayPalReturnURL.swift
@@ -39,7 +39,9 @@ struct BTPayPalReturnURL {
     /// - Parameter url: an app switch or ASWebAuthenticationSession return URL
     /// - Returns: `true` if the url represents a valid PayPal app switch return
     static func isValid(_ url: URL) -> Bool {
-        url.scheme == "https" && (url.path.contains("cancel") || url.path.contains("success"))
+        url.scheme == "https"
+        && (url.absoluteString.contains("token") || url.absoluteString.contains("ba_token"))
+        && (url.path.contains("cancel") || url.path.contains("success"))
     }
 
     static func isValidURLAction(url: URL, linkType: LinkType?) -> Bool {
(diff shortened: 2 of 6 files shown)


Sources/BraintreePayPal/BTPayPalReturnURL.swift
```diff
@@ -39,7 +39,9 @@ struct BTPayPalReturnURL {
     /// - Parameter url: an app switch or ASWebAuthenticationSession return URL
     /// - Returns: `true` if the url represents a valid PayPal app switch return
     static func isValid(_ url: URL) -> Bool {
-        url.scheme == "https" && (url.path.contains("cancel") || url.path.contains("success"))
+        url.scheme == "https"
+        && (url.path.contains("cancel") || url.path.contains("success"))
```
agedd, line 43: curious if there's a reason we don't check for `url.path.contains("error")` here

jaxdesmarais: That's a great question. Currently PayPal never sends us an error path, only cancel or success.

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -65,7 +65,10 @@ import BraintreeCore
     @objc(initWithAPIClient:universalLink:)
     public convenience init(apiClient: BTAPIClient, universalLink: URL) {
         self.init(apiClient: apiClient)
-        self.universalLink = universalLink
+        
+        /// appending a PayPal app switch specific path to verify we are in the correct flow when
+        /// `canHandleReturnURL` is called
+        self.universalLink = universalLink.appendingPathComponent("braintreeAppSwitchVenmo")
```
scannillo, line 71: Should we consider the case that a merchant isn't truly "wildcarding" anything after their path and is actually doing a strict path check? If so, this would break their integration

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -93,7 +93,10 @@ import BraintreeDataCollector
     @objc(initWithAPIClient:universalLink:)
     public convenience init(apiClient: BTAPIClient, universalLink: URL) {
         self.init(apiClient: apiClient)
-        self.universalLink = universalLink
+        
+        /// appending a PayPal app switch specific path to verify we are in the correct flow when
+        /// `canHandleReturnURL` is called
+        self.universalLink = universalLink.appendingPathComponent("braintreeAppSwitchPayPal")
```
scannillo, line 99: Maybe you thought through this idea already - but we could instead just tell merchants to set different paths for each Payment method return vs having us manipulating their path.

Ex: `uber.com/venmo-payment-returns` and `uber.com/paypal-payment-returns`

Thoughts?

Sources/BraintreeVenmo/BTVenmoClient.swift
jaxdesmarais: We are pretty explicit in our need to wildcard here: https://developer.paypal.com/braintree/docs/guides/paypal/app-switch/ios/v6#set-up-universalapp-links. We already append details like `success/cancel` and query parameters to the URLs so path checks should already account for custom details. What do you think?

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: That would still run into the same issue. If a merchant sent us `https://uber.com/paypal-payments-returns/success` that would pass our current checks of `https` and path contains `success`: https://github.com/braintree/braintree_ios/blob/57bf620ffacd0829bf368f11bf8ec6261c5ab595/Sources/BraintreeVenmo/BTVenmoAppSwitchReturnURL.swift#L71

This would return true and attempt to call `handleReturnURL` in the wrong feature client under this set up.

scannillo: Interesting! So we never actually compare the returnURL against the URL that the merchant originally passed into the `BTVenmoClient` or `BTPayPalClient`

agedd (approved):
lgtm! 👏🏽

Sources/BraintreeVenmo/BTVenmoAppSwitchReturnURL.swift
```diff
@@ -68,7 +68,9 @@ struct BTVenmoAppSwitchReturnURL {
     /// - Parameter url: an app switch return URL
     /// - Returns: `true` if the url represents a Venmo Touch app switch return
     static func isValid(url: URL) -> Bool {
-        (url.scheme == "https" && (url.path.contains("cancel") || url.path.contains("success") || url.path.contains("error")))
+        (url.scheme == "https"
+            && url.path.contains("braintreeAppSwitchVenmo")
+            && (url.path.contains("cancel") || url.path.contains("success") || url.path.contains("error")))
         || (url.host == "x-callback-url" && url.path.hasPrefix("/vzero/auth/venmo/"))
```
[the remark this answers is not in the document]
jaxdesmarais: That's a good call, we tried to be a bit "smart" here and it's making it way harder than it needs to be to read. I'll make these updates.

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: That's correct, I think merchants _should_ be able to use 1 URL for both if they want. This is also what we do today for our own demo.

Sources/BraintreeVenmo/BTVenmoAppSwitchReturnURL.swift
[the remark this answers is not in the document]
jaxdesmarais: Done: https://github.com/braintree/braintree_ios/pull/1538/commits/8987591904fca64fa24015e4a5027b8cdc5cac7e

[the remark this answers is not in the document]
richherrera: 🚀

PR #1553, opened 2025-04-09
Uncheck "Automatically Manage Signing"

richherrera:
### Summary of changes

- Uncheck "Automatically manage signing" from the `ThreeDSecure` and `PayPal` targets. Automatically managing signing causes our releases to fail at the "Release.Create static pre-built xcframeworks" step.

### Checklist

- [x] Added a changelog entry

### Authors

- @richherrera


jaxdesmarais (commented):
We may also consider unchecking this on PayPal. There is no team listed there but I wonder if it will cause issues in the next release.

richherrera:
> We may also consider unchecking this on PayPal. There is no team listed there but I wonder if it will cause issues in the next release.

I'm unchecking this on PayPal as well. I think the best course of action would be to stay alert during the next release

agedd (approved):

jaxdesmarais (approved):

jaxdesmarais (commented):
We can add a changelog for this similar to this if we want: https://github.com/braintree/braintree_ios/blob/main/CHANGELOG.md#6310-2025-04-09

PR #1559, opened 2025-04-24
[PRV2] Add BTShopperInsightsClientV2

richherrera:
### Summary of changes

- Add skeleton BTShopperInsightsClientV2 class
- Deprecate BTShopperInsightsClient

### Checklist

- [ ] Added a changelog entry

### Authors

- @richherrera

### DIFF
--- a/Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
+++ b/Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
@@ -7,6 +7,7 @@ import BraintreeCore
 ///  Use `BTShopperInsightsClient` to optimize your checkout experience by prioritizing the customer’s preferred payment methods in your UI.
 ///  By customizing each customer’s checkout experience, you can improve conversion, increase sales/repeat buys and boost user retention/loyalty.
 /// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
+@available(*, deprecated, message: "Use BTShopperInsightsClientV2 instead.")
 public class BTShopperInsightsClient {
     
     // MARK: - Internal Properties
--- a/Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
+++ b/Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
@@ -14,9 +14,11 @@ public class BTShopperInsightsClientV2 {
     
     private let apiClient: BTAPIClient
     
+    // MARK: - Initializers
+    
     /// Creates a `BTShopperInsightsClientV2`
     /// - Parameters:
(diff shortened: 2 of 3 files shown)


Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
```diff
@@ -7,6 +7,7 @@ import BraintreeCore
 ///  Use `BTShopperInsightsClient` to optimize your checkout experience by prioritizing the customer’s preferred payment methods in your UI.
 ///  By customizing each customer’s checkout experience, you can improve conversion, increase sales/repeat buys and boost user retention/loyalty.
 /// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
```
stechiu, line 9: If we're deprecating the original `BTShopperInsightsClient`, do we still need this warning about it being in beta?

stechiu:
Out of curiosity, what's the difference between v1 and v2? Why is v1 being deprecated? How long will BTShopperInsightsClient v1 and v2 be in beta?

Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
```diff
@@ -0,0 +1,26 @@
...
+    
+    private let apiClient: BTAPIClient
+    
+    // MARK: - Initializers
+    
+    /// Creates a `BTShopperInsightsClientV2`
+    /// - Parameters:
+    ///    - apiClient: A `BTAPIClient` instance.
+    /// - Warning: This init is beta. It's public API may change or be removed in future releases. This feature only works with a client token.
+    public init(apiClient: BTAPIClient) {
```
tdchow, line 23: Since this is a new Client class, would we be able to internally instantiate the `BTAPIClient`? Or can we only make this change in v7?

richherrera: That’s only possible in V7. On V7 some changes were made that allow us to instantiate `BTAPIClient` internally.

richherrera:
> Out of curiosity, what's the difference between v1 and v2? Why is v1 being deprecated? How long will BTShopperInsightsClient v1 and v2 be in beta?

The main objective of PR v2 is to validate product-market fit before expanding the product to a larger group of merchants. V2 will have new features to support a smooth and quick integration.

Sources/BraintreeShopperInsights/BTShopperInsightsClient.swift
richherrera: Yeah, it does sound a bit redundant, but since most of the logic will be different anyway, i think we can keep the warnings

warmkesselj (approved):

PR #1565, opened 2025-05-06
[v7] Update `release.yml`

jaxdesmarais:
### Summary of changes

- The release for v7 failed on the reference docs step because the GitHub action runner did not have the iPhone type being specified. This bumps to an iPhone 16 which is [supported on the runner](https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md#installed-simulators) we are using.

### Checklist

### Authors

- @jaxdesmarais


agedd (approved):
🚀

PR #1569, opened 2025-05-09
[v7] Venmo GraphQL Encodable

jaxdesmarais:
### Summary of changes

- Convert the following POST requests to Encodable for the Venmo flow:
    - `VenmoAccountsPOSTBody`
    - `VenmoCreatePaymentContextGraphQLBody`
    - `VenmoQueryPaymentContextGraphQLBody`

🚀  **Note:** I used GitHub Copilot to help with some of this logic. While I did need to update some things to match our patterns and the expected Graph QL keys, it was helpful in generating some of the encodable boilerplate.

**Testing:** I have tested both the app switch and universal link fallback flow on device with these changes

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeVenmo/Models/VenmoCreatePaymentContextGraphQLBody.swift
+++ b/Sources/BraintreeVenmo/Models/VenmoCreatePaymentContextGraphQLBody.swift
@@ -1,0 +1,149 @@
+import Foundation
+
+// swiftlint:disable nesting
+/// The POST body for graph QL `mutation CreateVenmoPaymentContext`
+struct VenmoCreatePaymentContextGraphQLBody: Encodable {
+    
+    var query: String
+    var variables: Variables
+    
+    init(request: BTVenmoRequest, merchantProfileID: String?) {
+        // swiftlint:disable:next line_length
+        self.query = "mutation CreateVenmoPaymentContext($input: CreateVenmoPaymentContextInput!) { createVenmoPaymentContext(input: $input) { venmoPaymentContext { id } } }"
+        self.variables = Variables(request: request, merchantProfileID: merchantProfileID)
+    }
+    
+    
+    struct Variables: Encodable {
+        
+        let input: InputParameters
+        
+        init(request: BTVenmoRequest, merchantProfileID: String?) {
+            self.input = InputParameters(
+                paymentMethodUsage: request.paymentMethodUsage.stringValue,
+                merchantProfileID: merchantProfileID,
+                customerClient: "MOBILE_APP",
+                intent: "CONTINUE",
+                isFinalAmount: request.isFinalAmount.description,
+                displayName: request.displayName,
+                paysheetDetails: Variables.InputParameters.PaysheetDetails(
+                    collectCustomerBillingAddress: request.collectCustomerBillingAddress,
+                    collectCustomerShippingAddress: request.collectCustomerShippingAddress,
+                    transactionDetails: Variables.InputParameters.PaysheetDetails.TransactionDetails(
+                        lineItems: request.lineItems?.map { item in
+                            Variables.InputParameters.PaysheetDetails.LineItem(
+                                quantity: item.quantity,
+                                unitAmount: item.unitAmount,
+                                name: item.name,
+                                kind: item.kind.rawValue,
+                                unitTaxAmount: item.unitTaxAmount ?? "0",
+                                itemDescription: item.itemDescription,
+                                productCode: item.productCode,
+                                url: item.url
+                            )
+                        }
+                    )
+                )
+            )
+        }
+        
+        struct InputParameters: Encodable {
+            
+            var paymentMethodUsage: String?
+            var merchantProfileID: String?
+            var customerClient: String = "MOBILE_APP"
+            var intent: String = "intent"
+            var isFinalAmount: String?
+            var displayName: String?
+            var paysheetDetails: PaysheetDetails?
+            
+            enum CodingKeys: String, CodingKey {
+                case paymentMethodUsage
+                case merchantProfileID = "merchantProfileId"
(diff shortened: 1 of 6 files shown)


Sources/BraintreeVenmo/Models/VenmoCreatePaymentContextGraphQLBody.swift
```diff
@@ -0,0 +1,149 @@
...
+                    collectCustomerBillingAddress: request.collectCustomerBillingAddress,
+                    collectCustomerShippingAddress: request.collectCustomerShippingAddress,
+                    transactionDetails: Variables.InputParameters.PaysheetDetails.TransactionDetails(
+                        lineItems: request.lineItems?.map { item in
+                            Variables.InputParameters.PaysheetDetails.LineItem(
+                                quantity: item.quantity,
+                                unitAmount: item.unitAmount,
+                                name: item.name,
+                                kind: item.kind.rawValue,
+                                unitTaxAmount: item.unitTaxAmount ?? "0",
```
jaxdesmarais, line 39: This looked to be the default behavior in `BTVenmoClient` so persisted this logic:
```swift
if item.unitTaxAmount == nil || item.unitTaxAmount?.isEmpty == true {
    item.unitTaxAmount = "0"
}
```

Sources/BraintreeVenmo/Models/VenmoCreatePaymentContextGraphQLBody.swift
```diff
@@ -0,0 +1,149 @@
...
+                case paysheetDetails
+            }
+            
+            struct PaysheetDetails: Encodable {
+                
+                var collectCustomerBillingAddress: Bool?
+                var collectCustomerShippingAddress: Bool?
+                var transactionDetails: TransactionDetails?
+                
+                func encode(to encoder: Encoder) throws {
```
jaxdesmarais, line 76: Custom encoder needed to match this logic:
```swift
if !transactionDetails.isEmpty {
    paysheetDetails["transactionDetails"] = transactionDetails
}
```

Sources/BraintreeVenmo/Models/VenmoCreatePaymentContextGraphQLBody.swift
```diff
@@ -0,0 +1,149 @@
...
+                struct TransactionDetails: Encodable {
+                    
+                    var subTotalAmount: String?
+                    var discountAmount: String?
+                    var taxAmount: String?
+                    var shippingAmount: String?
+                    var totalAmount: String?
+                    let lineItems: [LineItem]?
+                    
+                    func encode(to encoder: Encoder) throws {
```
jaxdesmarais, line 112: Similarly to above, custom encoder needed to match this logic:
```swift
if let lineItems = request.lineItems, !lineItems.isEmpty {
    for item in lineItems {
        if item.unitTaxAmount == nil || item.unitTaxAmount?.isEmpty == true {
            item.unitTaxAmount = "0"
        }
    }
    
    let lineItemsArray = lineItems.compactMap { $0.requestParameters() }
    transactionDetails["lineItems"] = lineItemsArray
}
```

Sources/BraintreeVenmo/Models/VenmoCreatePaymentContextGraphQLBody.swift
```diff
@@ -0,0 +1,149 @@
...
+                )
+            )
+        }
+        
+        struct InputParameters: Encodable {
+            
+            var paymentMethodUsage: String?
+            var merchantProfileID: String?
+            var customerClient: String = "MOBILE_APP"
+            var intent: String = "intent"
```
warmkesselj, line 55: Should the intent value have a different default value besides "intent"?

jaxdesmarais: Ah, good catch, should be `CONTINUE`: https://github.com/braintree/braintree_ios/blob/f5a4c8155425f8626075539b5d759e143ff3b43f/Sources/BraintreeVenmo/BTVenmoClient.swift#L248

jaxdesmarais, line 55:
```suggestion
            var intent: String = "CONTINUE"
```

jaxdesmarais: It was being overridden above but still good to have both the default and init match.

agedd (approved):
lgtm! just added one minor comment 🚀

warmkesselj (approved):

Sources/BraintreeVenmo/Models/VenmoCreatePaymentContextGraphQLBody.swift
```diff
@@ -0,0 +1,149 @@
...
+                    if let transactionDetails = transactionDetails {
+                        let properties = [
+                            transactionDetails.subTotalAmount,
+                            transactionDetails.discountAmount,
+                            transactionDetails.taxAmount,
+                            transactionDetails.shippingAmount,
+                            transactionDetails.totalAmount
+                        ]
+                        
+                        if !properties.allSatisfy({ $0?.isEmpty ?? true }) || !(transactionDetails.lineItems?.isEmpty ?? true) {
```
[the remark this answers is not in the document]
jaxdesmarais: The `lineItems` object is an array of line items where the `transactionDetails` is an object. The logic here is a bit odd, but we only want to construct `transactionDetails` if those fields are present or if `lineItems` is not empty. We have the test `testTokenize_withoutLineItems_createsPaymentContextWithoutTransactionDetails` that was failing before this logic was added.

PR #1572, opened 2025-05-14
Add CreateCustomerSession GraphQL Mutation

agedd:
### Summary of changes

- Create a `BTCreateCustomerSessionApi` class
- Add `BTCustomerSessionRequest` and `BTPurchaseUnit` structs to make the request to `createCustomerSession` GraphQL API.
- Add some sanity unit tests

Tested and verified the API call returns the expected `sessionId` value with unit tests.

### Checklist

- [ ] ~Added a changelog entry~

- @agedd

### DIFF
--- a/Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
+++ b/Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
@@ -6,19 +6,19 @@ import BraintreeCore
 
 /// The API used to create a customer session using the `CreateCustomerSession` GraphQL mutation.
 /// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
-class BTCreateCustomerSessionAPI {
+final class BTCreateCustomerSessionAPI {
     
     // MARK: - Properties
     
     /// Exposed for testing to get the instance of `BTAPIClient`
-    var apiClient: BTAPIClient
+    private var apiClient: BTAPIClient
     
     // MARK: - Initializer
     
     /// Creates a `BTCreateCustomerSessionAPI`
     /// - Parameters:
     ///    - apiClient: A `BTAPIClient` instance
-    public init(apiClient: BTAPIClient) {
+    init(apiClient: BTAPIClient) {
         self.apiClient = apiClient
     }
     
--- a/Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
+++ b/Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
@@ -13,9 +13,6 @@ class BTCreateCustomerSessionAPI {
     /// Exposed for testing to get the instance of `BTAPIClient`
     var apiClient: BTAPIClient
     
-    /// :nodoc: This typealias is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
-    typealias CreateCustomerSessionResult = (String?, Error?) -> Void
-    
     // MARK: - Initializer
     
     /// Creates a `BTCreateCustomerSessionAPI`
@@ -31,7 +28,7 @@ class BTCreateCustomerSessionAPI {
     ///    - completion: This completion will be invoked when the attempt to create a customer session is complete or an error occurs. On success, you will receive a sessionId; on failure you will receive an error.
     func execute(
         _ request: BTCustomerSessionRequest,
-        completion: @escaping CreateCustomerSessionResult
+        completion: @escaping (String?, Error?) -> Void
     ) {
         do {
             let graphQLParams = try self.buildGraphQLDictionary(with: request)
--- a/Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
+++ b/Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
@@ -6,7 +6,7 @@ import BraintreeCore
 
 /// The API used to create a customer session using the `CreateCustomerSession` GraphQL mutation.
 /// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
-class BTCreateCustomerSessionApi {
+class BTCreateCustomerSessionAPI {
     
     // MARK: - Properties
     
@@ -16,12 +16,11 @@ class BTCreateCustomerSessionApi {
     /// :nodoc: This typealias is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     typealias CreateCustomerSessionResult = (String?, Error?) -> Void
     
-    // MARK: - Init
+    // MARK: - Initializer
     
-    /// Creates a `BTCreateCustomerSessionApi`
+    /// Creates a `BTCreateCustomerSessionAPI`
     /// - Parameters:
     ///    - apiClient: A `BTAPIClient` instance
-    /// - Warning: This init is beta. It's public API may change or be removed in future releases. This feature only works with a client token.
     public init(apiClient: BTAPIClient) {
         self.apiClient = apiClient
     }
@@ -30,7 +29,6 @@ class BTCreateCustomerSessionApi {
     /// - Parameters:
     ///    - request: A `BTCustomerSessionRequest`
     ///    - completion: This completion will be invoked when the attempt to create a customer session is complete or an error occurs. On success, you will receive a sessionId; on failure you will receive an error.
-    /// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
     func execute(
         _ request: BTCustomerSessionRequest,
         completion: @escaping CreateCustomerSessionResult
--- a/Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
+++ b/Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
@@ -26,8 +26,6 @@ class BTCreateCustomerSessionApi {
         self.apiClient = apiClient
     }
     
-    // MARK: - Public Methods
-    
     /// This method will call the `CreateCustomerSession` GQL mutation, which returns a sessionId if successful.
     /// - Parameters:
     ///    - request: A `BTCustomerSessionRequest`
@@ -47,7 +45,7 @@ class BTCreateCustomerSessionApi {
                 }
                 
                 guard let body else {
-                    completion(nil, BTHTTPError.dataNotFound)
+                    completion(nil, BTShopperInsightsError.emptyBodyReturned)
                     return
                 }
                 
@@ -65,28 +63,9 @@ class BTCreateCustomerSessionApi {
     // MARK: - Helper Methods
     
     func buildGraphQLDictionary(with request: BTCustomerSessionRequest) throws -> [String: Any] {
-        let customerDetails: [String: Any?] = [
-            "hashedEmail": request.hashedEmail,
-            "hashedPhoneNumber": request.hashedPhoneNumber,
-            "paypalAppInstalled": request.paypalAppInstalled,
-            "venmoAppInstalled": request.venmoAppInstalled
-        ]
-        
-        var purchaseUnitDetails: [[String: Any]] = []
-        if let purchaseUnits = request.purchaseUnits, !purchaseUnits.isEmpty {
-            purchaseUnitDetails = purchaseUnits.map { purchaseUnit in
-                return [
-                    "amount": [
-                        "value": purchaseUnit.amount,
-                        "currencyCode": purchaseUnit.currencyCode
-                    ]
-                ]
-            }
-        }
-        
         let inputParameters: [String: Any?] = [
-            "customer": customerDetails,
-            "purchaseUnits": purchaseUnitDetails
+            "customer": request.customer,
+            "purchaseUnits": request.purchaseUnits
         ]
         let inputDictionary: [String: Any] = ["input": inputParameters]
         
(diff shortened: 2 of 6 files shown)


Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
```diff
@@ -0,0 +1,99 @@
...
+    func execute(
+        _ request: BTCustomerSessionRequest,
+        completion: @escaping CreateCustomerSessionResult
+    ) {
+        do {
+            let graphQLParams = try self.buildGraphQLDictionary(with: request)
+            
+            self.apiClient.post("", parameters: graphQLParams, httpType: .graphQLAPI) { body, _, error in
+                if let error = error as? NSError {
+                    completion(nil, BTHTTPError.httpResponseInvalid)
```
agedd, line 45: question for folks:  i'm returning `BTHTTPError` types for now in places where we may throw errors but wondering if i should create a custom error type for v2 version (i.e. `BTShopperInsightsError`)

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
```diff
@@ -0,0 +1,99 @@
...
+                completion(sessionId, nil)
+            }
+        } catch {
+            completion(nil, error)
+        }
+    }
+    
+    // MARK: - Helper Methods
+    
+    func buildGraphQLDictionary(with request: BTCustomerSessionRequest) throws -> [String: Any] {
```
agedd, line 67: **note**: similar pattern followed from `BTVenmoClient` where construct the graphQL dict as well (i believe eventually we may need to update this to adopt the `Encodable` protocol as we do in v7)

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
```diff
@@ -0,0 +1,106 @@
...
+            let graphQLParams = try self.buildGraphQLDictionary(with: request)
+            
+            self.apiClient.post("", parameters: graphQLParams, httpType: .graphQLAPI) { body, _, error in
+                if let error = error as? NSError {
+                    completion(nil, error)
+                    return
+                }
+                
+                guard let body else {
+                    completion(nil, BTHTTPError.dataNotFound)
```
agedd, line 50: question for folks: i'm returning `BTHTTPError` types here but wondering if i should create a custom error type for v2 version (i.e. BTShopperInsightsError) or this can be handled in a separate PR. would love to get y'all thoughts 👀

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
```diff
@@ -0,0 +1,106 @@
+import UIKit
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+/// The API used to create a customer session using the `CreateCustomerSession` GraphQL mutation.
+/// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
+class BTCreateCustomerSessionApi {
```
tdchow, line 9: I'm not familiar with visibility modifiers in swift, but is this Api class exposed to merchants? Should we ensure it's not visible?

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
```diff
@@ -0,0 +1,106 @@
...
+    
+    /// :nodoc: This typealias is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    typealias CreateCustomerSessionResult = (String?, Error?) -> Void
+    
+    // MARK: - Init
+    
+    /// Creates a `BTCreateCustomerSessionApi`
+    /// - Parameters:
+    ///    - apiClient: A `BTAPIClient` instance
+    /// - Warning: This init is beta. It's public API may change or be removed in future releases. This feature only works with a client token.
```
tdchow, line 24: Since this class should be internal to the SDK, do we need the beta warning for each of these comments?

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
```diff
@@ -0,0 +1,106 @@
...
+import BraintreeCore
+#endif
+
+/// The API used to create a customer session using the `CreateCustomerSession` GraphQL mutation.
+/// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
+class BTCreateCustomerSessionApi {
+    
+    // MARK: - Properties
+    
+    /// Exposed for testing to get the instance of `BTAPIClient`
```
tdchow, line 13: Since apiClient is injected in the `init()` do we still need to expose the internal `apiClient`? Not entirely sure how swift handles this 😅

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
tdchow: I feel that reusing existing errors that we have defined would be okay. I think in most cases, merchants would be handling all returned errors the same, they may just be logging the error type.

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
```diff
@@ -0,0 +1,106 @@
...
+    func execute(
+        _ request: BTCustomerSessionRequest,
+        completion: @escaping CreateCustomerSessionResult
+    ) {
+        do {
+            let graphQLParams = try self.buildGraphQLDictionary(with: request)
+            
+            self.apiClient.post("", parameters: graphQLParams, httpType: .graphQLAPI) { body, _, error in
+                if let error = error as? NSError {
+                    completion(nil, error)
```
tdchow, line 45: Should we add a unit test for this completion that returns an error?

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
tdchow, line 50: Should we add a unit test for this completion that returns an error?

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
```diff
@@ -0,0 +1,85 @@
...
+/// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
+class BTCreateCustomerSessionApi {
+    
+    // MARK: - Properties
+    
+    /// Exposed for testing to get the instance of `BTAPIClient`
+    var apiClient: BTAPIClient
+    
+    /// :nodoc: This typealias is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    typealias CreateCustomerSessionResult = (String?, Error?) -> Void
```
jaxdesmarais, line 17: Do we need a typealias for this?

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
```diff
@@ -0,0 +1,85 @@
...
+    
+    // MARK: - Init
```
jaxdesmarais, line 19: Super nit to match our other files:
```suggestion
    // MARK: - Initializer
```

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
```diff
@@ -0,0 +1,85 @@
...
+    
+    /// Creates a `BTCreateCustomerSessionApi`
```
jaxdesmarais, line 21:
```suggestion
    /// Creates a `BTCreateCustomerSessionAPI`
```

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
```diff
@@ -0,0 +1,85 @@
...
+    /// - Parameters:
+    ///    - apiClient: A `BTAPIClient` instance
+    /// - Warning: This init is beta. It's public API may change or be removed in future releases. This feature only works with a client token.
+    public init(apiClient: BTAPIClient) {
+        self.apiClient = apiClient
+    }
+    
+    /// This method will call the `CreateCustomerSession` GQL mutation, which returns a sessionId if successful.
```
jaxdesmarais, line 29:
```suggestion
    /// This method will call the `CreateCustomerSession` GQL mutation, which returns a session ID if successful.
```

Sources/BraintreeShopperInsights/BTCustomerSessionRequest.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
+
+// swiftlint:disable nesting
+/// The POST body for the `createCustomerSession` GraphQL API.
+/// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
```
jaxdesmarais, line 5: I don't think we need this warning for anything internal

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
agedd: ahhh i think so, this seems to be the pattern in the other client classes as well

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
agedd: yep that's a great callout, addressed in 3125f3eca4f5e95016e30379e6753674dbd1918e

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
agedd: ahhh yep that's true i don't think we need it, addressed in 364600e2cd5aa4829306bb6c114a58e2ea6b5de1

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
agedd: yep updated `BTCustomerSessionRequest` to adopt `Encodable` now, thinking of keeping this helper function as is to build out the graphQL mutation rather than handling this logic in the `execute` func. wdyt?

Sources/BraintreeShopperInsights/BTCreateCustomerSessionApi.swift
agedd: ahh that's right we don't need it - updated in 4b947e866a98bf0b87067d6f6b4301697bdc1a53

Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
```diff
@@ -0,0 +1,80 @@
...
+#endif
+
+/// The API used to create a customer session using the `CreateCustomerSession` GraphQL mutation.
+/// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
+class BTCreateCustomerSessionAPI {
+    
+    // MARK: - Properties
+    
+    /// Exposed for testing to get the instance of `BTAPIClient`
+    var apiClient: BTAPIClient
```
richherrera, line 14: This property can be private, if we inject it through the `init`, our tests can provide whatever `APIClient` we need (like a `mockAPIClient`)

Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
```diff
@@ -0,0 +1,80 @@
...
+    
+    // MARK: - Initializer
+    
+    /// Creates a `BTCreateCustomerSessionAPI`
+    /// - Parameters:
+    ///    - apiClient: A `BTAPIClient` instance
+    public init(apiClient: BTAPIClient) {
```
richherrera, line 21: I think this `init` could be internal, since `ShopperInsightClientV2` is the one responsible for creating this use case.

Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
```diff
@@ -0,0 +1,80 @@
+import UIKit
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+/// The API used to create a customer session using the `CreateCustomerSession` GraphQL mutation.
+/// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
+class BTCreateCustomerSessionAPI {
```
richherrera, line 9: I think we can mark this class as `final`, what do you think?

agedd: yep agreed, addressed in d6d73b794ef809648f85c30e56a1c11290c24967

richherrera (approved):
LGTM!!!

warmkesselj (approved):

Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
```diff
@@ -0,0 +1,80 @@
...
+                completion(sessionId, nil)
+            }
+        } catch {
+            completion(nil, error)
+        }
+    }
+    
+    // MARK: - Helper Methods
+    
+    func buildGraphQLDictionary(with request: BTCustomerSessionRequest) throws -> [String: Any] {
```
jaxdesmarais, line 60: We can do this whole thing as encodable vs just part

jaxdesmarais (commented):
Left some comments though they can go in a follow up PR

PR #1573, opened 2025-05-14
Remove analytics batching

richherrera:
### Summary of changes

Updates the sendAnalyticsEventsImmediately method to leverage UIApplication.shared.beginBackgroundTask in order to ensure that analytics events are successfully delivered even if the app is transitioning to the background.

Key changes:
Wrapped the call to sendAnalyticEvent in a background task to extend the app’s execution time beyond the default 30 seconds allowed by the OS.

Ensured the background task is properly ended in all execution paths to avoid potential app termination.

The update follows Apple’s recommended pattern for background task usage and includes a fallback with MainActor.assumeIsolated to safely call endBackgroundTask.

This change improves the reliability of analytics delivery, especially in critical scenarios like app switch events that may occur just as the app is backgrounded.

Specifically:

The background task is properly ended both via the expirationHandler and after the async operation completes, preventing the OS from terminating the app prematurely.

Screenshot 2025-05-20 at 9 14 35 a m

### Checklist

- [ ] Added a changelog entry

### Authors

- @richherrera

### DIFF
--- a/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
+++ b/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
@@ -55,11 +55,7 @@ final class BTAnalyticsService: AnalyticsSendable {
     /// - Parameter event: A single `FPTIBatchData.Event`
     func sendAnalyticsEvent(_ event: FPTIBatchData.Event, sendImmediately: Bool) {
         Task(priority: .background) {
-            if sendImmediately {
-                await sendAnalyticsEvents(with: event)
-            } else {
-                await performEventRequest(with: event)
-            }
+            sendImmediately ? await sendAnalyticsEvents(with: event) : await performEventRequest(with: event)
         }
     }
     
@@ -102,7 +98,8 @@ final class BTAnalyticsService: AnalyticsSendable {
                 try await postAnalyticsEvents(
                     configuration: configuration,
                     sessionID: sessionID,
-                    events: eventsPerSessionID)
+                    events: eventsPerSessionID
+                )
                 await events.removeFor(sessionID: sessionID)
             }
         } catch {
--- a/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
+++ b/Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
@@ -88,33 +88,35 @@ final class BTAnalyticsService: AnalyticsSendable {
         // The expirationHandler is called if the system’s maximum background execution time
         // (typically around 30 seconds) is reached before the task completes.
         // If we don't explicitly end the task here, the app may be forcefully terminated by the system.
-        Task {
-            backgroundTaskID = await UIApplication.shared.beginBackgroundTask(withName: "BTSendAnalyticEvent") {
-                // We end the task here to avoid the app being terminated.
-                UIApplication.shared.endBackgroundTask(backgroundTaskID)
-                backgroundTaskID = .invalid
-            }
-            
-            await sendAnalyticEvent(event, apiClient: apiClient)
-            
+        backgroundTaskID = UIApplication.shared.beginBackgroundTask(withName: "BTSendAnalyticEvent") {
+            // We end the task here to avoid the app being terminated.
+            UIApplication.shared.endBackgroundTask(backgroundTaskID)
+            backgroundTaskID = .invalid
+        }
+        
+        sendAnalyticEvent(event, apiClient: apiClient) {
             // Explicitly end the background task after the work is completed
-            await UIApplication.shared.endBackgroundTask(backgroundTaskID)
+            UIApplication.shared.endBackgroundTask(backgroundTaskID)
             backgroundTaskID = .invalid
         }
     }
 
     // MARK: - Private Methods
 
-    private func sendAnalyticEvent(_ event: FPTIBatchData.Event, apiClient: BTAPIClient) async {
-        do {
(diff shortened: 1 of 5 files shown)


Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -22,7 +23,8 @@ final class BTAnalyticsService: AnalyticsSendable {
     private let events = BTAnalyticsEventsStorage()
     private let timer = RepeatingTimer()
 
-    private weak var apiClient: BTAPIClient?
+    private var backgroundTaskID: UIBackgroundTaskIdentifier = .invalid
```
agedd, line 26: quick question: i'm curious—how does adding the `backgroundTaskID` assist with triggering the batched analytics events?

richherrera: `UIBackgroundTaskIdentifier` gives the app extra time to run tasks in the background. Some AppSwitch events are triggered when the app is already in the background, which is why we need that extra time, to make sure the event gets sent properly.

[Refetence](https://developer.apple.com/documentation/uikit/uiapplication/beginbackgroundtask(expirationhandler:))

agedd (approved):
lgtm! ty for making this change 🚀

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -69,30 +71,79 @@ final class BTAnalyticsService: AnalyticsSendable {
...
+        endBackgroundTask(identifier: backgroundTaskID)
+    }
+    
     // MARK: - Private Methods
+    
+    private func beginBackgroundTaskIfNeeded() -> UIBackgroundTaskIdentifier {
+        var backgroundTaskID: UIBackgroundTaskIdentifier = .invalid
+        
+        backgroundTaskID = UIApplication.shared.beginBackgroundTask(withName: "BTSendAnalyticEvent") {
+            DispatchQueue.main.async {
```
richherrera, line 92: I had to move this task to the main thread to get rid of the warning: `'backgroundTaskID' mutated after capture by sendable closure`

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -69,30 +70,79 @@ final class BTAnalyticsService: AnalyticsSendable {
...
+        
+        await sendAnalyticEvent(event, apiClient: apiClient)
+        
+        endBackgroundTask(identifier: backgroundTaskID)
+    }
+    
     // MARK: - Private Methods
+    
+    private func beginBackgroundTaskIfNeeded() -> UIBackgroundTaskIdentifier {
+        var backgroundTaskID: UIBackgroundTaskIdentifier = .invalid
```
richherrera, line 88: Originally, there was just one identifier that got overwritten every time an event was sent. But when using this API, we need a separate identifier for each background task, so now, a new one gets created every time an event is sent

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
richherrera: I tried it earlier, and it throws an error: "Closure captures 'backgroundTaskID' before it is declared."

jaxdesmarais: Could we pair on this? I am not seeing the same with the above code, but am probably missing something.

jaxdesmarais: We paired, I was passing `self?.endBackgroundTask(identifier: .invalid)` in the closure which means that our generated `backgroundTaskID` would never get called if the `beginBackgroundTask` closure reached the end of it's life.

jaxdesmarais (approved):
👏 great work and research on this!

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
richherrera: Here we need to pass the `backgroundTaskID` so we can complete tasks that take longer than 30 seconds. There are two possible scenarios when the app goes into the background and we use this API.

In the first scenario, the task finishes within 30 seconds—in that case, the task ends with the next instruction.

In the second scenario, the task takes longer than 30 seconds (let’s say posting to FPTI takes 2 minutes). In this case, we need to end the task manually using the `expirationHandler`, otherwise, the merchant’s app will crash.

That’s why we need a reference to the `backgroundTaskID` inside the `expirationHandler`.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -69,30 +69,87 @@ final class BTAnalyticsService: AnalyticsSendable {
...
         }
     }
+    
+    /// Sends a single analytics event immediately, even if the app is transitioning to the background.
+    ///
+    /// This method initiates a background task using `UIApplication.shared.beginBackgroundTask` to
+    /// ensure that the event has enough time to be sent before the OS suspends the app.
+    /// The background task is safely ended both in the expiration handler and after the event is sent.
+    ///
+    /// Exposed to be able to execute this function synchronously in unit tests
```
jaxdesmarais, line 79: 👏  these docstrings are very helpful!

warmkesselj (approved):

agedd (approved):
looks great to me! 🚀 tyy for taking this on @richherrera

PR #1577, opened 2025-05-20
Cleanup Add `CreateCustomerSession` GraphQL Mutation

agedd:
### Summary of changes

- Address some cleanup and refactor PR comments from this [PR](https://github.com/braintree/braintree_ios/pull/1572)

### Checklist

- @agedd 
- @richherrera


Braintree.xcodeproj/project.pbxproj
```diff
@@ -1613,6 +1623,7 @@
 				804698292B27C4D70090878E /* BraintreeShopperInsights */,
 				A91E1EAC24FEEF21007540E5 /* BraintreeThreeDSecure */,
 				A7C889F01B5F0B30007A0E9C /* BraintreeVenmo */,
+				621964B82DDCF82300684434 /* Helpers */,
```
jaxdesmarais, line 1626: For directory and pattern lets use the same as is being used for `VenmoQueryPaymentContextGraphQLBody`, example here: https://github.com/braintree/braintree_ios/pull/1569/files

We can make a `Models` directory and will want to make sure we update the path for cocoapods, if we havent already. Then naming wise can call the file <FEATURE><GQL REQUEST NAME>GraphQLBody. Happy to pair on this too if you'd like.

Sources/BraintreeShopperInsights/Models/GraphQLBody.swift
```diff
@@ -0,0 +1,16 @@
+/// Generic GraphQL Request Wrapper
+struct BTGraphQLRequest<Variables: Encodable>: Encodable {
```
agedd, line 2: **note**: not sure this is still needed—happy to remove it if others think it's no longer relevant.

Sources/BraintreeShopperInsights/Models/CreateCustomerSessionMutationGraphQLBody.swift
```diff
@@ -0,0 +1,97 @@
+import Foundation
+
+// swiftlint:disable nesting
+/// The POST body for the GraphQL mutation `CreateCustomerSession`
+struct CreateCustomerSessionMutationGraphQLBody: Encodable {
```
jaxdesmarais, line 5: Can we add a few unit tests for this?

agedd: yep added in 9e6acc4ed40ae2ab74a6ff1e180f3f0401834df7

Sources/BraintreeShopperInsights/Models/CreateCustomerSessionMutationGraphQLBody.swift
```diff
@@ -0,0 +1,69 @@
...
```
richherrera, line 5: Since we’ll be using this struct to make the request, I don’t think `BTCustomerSessionRequest` needs to conform to the `Encodable` protocol anymore, or does it? Maybe we could double check that

UnitTests/BraintreeShopperInsightsTests/BTCreateCustomerSessionAPI_Tests.swift
```diff
@@ -102,5 +112,46 @@ class BTCreateCustomerSessionAPI_Tests: XCTestCase {
         }
         waitForExpectations(timeout: 2)
     }
+    
+    func testEncodingCreateCustomerSessionGraphQLBodyWithFullData() throws {
+        let body = try CreateCustomerSessionMutationGraphQLBody(request: createCustomerSessionRequest)
+        let jsonData = try JSONEncoder().encode(body)
+        
+        let jsonObject = try JSONSerialization.jsonObject(with: jsonData) as? [String: Any]
```
richherrera, line 120: we have an extension on the Encodable protocol that converts the object into a dictionary, all of this could be simplified something like this 


``` swift
guard let jsonObject = try? body.toDictionary() else {
   return
}
```


here is an example https://github.com/braintree/braintree_ios/blob/3e6ae46bb2c9f40b364ed17e59005f70ab79775d/UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift#L103

UnitTests/BraintreeShopperInsightsTests/BTCreateCustomerSessionAPI_Tests.swift
```diff
@@ -102,5 +112,46 @@ class BTCreateCustomerSessionAPI_Tests: XCTestCase {
...
+                hashedPhoneNumber: nil,
+                paypalAppInstalled: nil,
+                venmoAppInstalled: nil
+            ),
+            purchaseUnits: nil
+        )
+        let body = try CreateCustomerSessionMutationGraphQLBody(request: request)
+        let jsonData = try JSONEncoder().encode(body)
+        
+        let jsonObject = try JSONSerialization.jsonObject(with: jsonData) as? [String: Any]
```
richherrera, line 146: same here

UnitTests/BraintreeShopperInsightsTests/BTCreateCustomerSessionAPI_Tests.swift
agedd: ahh yep ty for the suggestion, addressed in 8c9195efe07210a1c7bde92a192f07b76120ec80

Sources/BraintreeShopperInsights/BTCustomerSessionRequest.swift
```diff
@@ -1,7 +1,8 @@
 import Foundation
 
-/// The POST body for the `createCustomerSession` GraphQL API.
-struct BTCustomerSessionRequest: Encodable {
+// swiftlint:disable nesting
+/// A `BTCustomerSessionRequest` that specifies options for the Payment Ready v2 flow.
+struct BTCustomerSessionRequest {
```
jaxdesmarais, line 5: I _think_ we want this to be public right? This will be the request merchants ultimately pass to us? If so we should docstring the properties. I think we want to be following this format here vs nesting the objects: https://github.com/braintree/braintree_android/pull/1317/files

Sources/BraintreeShopperInsights/Models/CreateCustomerSessionMutationGraphQLBody.swift
```diff
@@ -0,0 +1,69 @@
...
+        
+        init(request: BTCustomerSessionRequest) {
+            self.input = InputParameters(
+                customer: Variables.InputParameters.Customer(
+                    hashedEmail: request.customer.hashedEmail,
+                    hashedPhoneNumber: request.customer.hashedPhoneNumber,
+                    paypalAppInstalled: request.customer.paypalAppInstalled,
+                    venmoAppInstalled: request.customer.venmoAppInstalled
+                ),
+                purchaseUnits: request.purchaseUnits?.map { purchaseUnit in
```
jaxdesmarais, line 33: Is graphQL okay if we call this map and pass `purchaseUnits` with an empty object? Not sure if we've had a chance to manually test that yet.

Sources/BraintreeShopperInsights/BTCustomerSessionRequest.swift
agedd: ahh yep that's true, addressed this in a57e5266f83742251a4231505231207a99dae06b

Sources/BraintreeShopperInsights/Models/CreateCustomerSessionMutationGraphQLBody.swift
agedd: ahh good catch, i haven't tested that case yet but i've updated to use `compactMap` to be on the safe side and added a unit test to cover it.

richherrera (approved):

Sources/BraintreeShopperInsights/BTCustomerSessionRequest.swift
```diff
@@ -1,22 +1,36 @@
 import Foundation
 
-/// The POST body for the `createCustomerSession` GraphQL API.
-struct BTCustomerSessionRequest: Encodable {
+/// A `BTCustomerSessionRequest` that specifies options for the Payment Ready v2 flow.
+public struct BTCustomerSessionRequest {
```
jaxdesmarais, line 4: I think we want to flatten the public interface and handle the nesting at the SDK level to match the API. Take a look at the Android PR as an example: https://github.com/braintree/braintree_android/pull/1317/files. We will want to have a public init with those properties.

Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
```diff
@@ -22,19 +20,19 @@ final class BTCreateCustomerSessionAPI {
...
         _ request: BTCustomerSessionRequest,
         completion: @escaping (String?, Error?) -> Void
     ) {
         do {
-            let graphQLParams = try self.buildGraphQLDictionary(with: request)
+            let graphQLParams = try CreateCustomerSessionMutationGraphQLBody(request: request)
             
-            self.apiClient.post("", parameters: graphQLParams, httpType: .graphQLAPI) { body, _, error in
-                if let error = error as? NSError {
+            apiClient.post("", parameters: graphQLParams, httpType: .graphQLAPI) { body, _, error in
```
richherrera: I’d like to hear the team’s thoughts on using the async/await version. Since it’s an internal class, we could definitely take advantage of it

Sources/BraintreeShopperInsights/BTCustomerSessionRequest.swift
richherrera: Updated, let me know what you think!!!

Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
jaxdesmarais: I think we should use async/await for sure if it's internal! I don't see a compelling reason not to move to that internally. What do you think?

Sources/BraintreeShopperInsights/BTCustomerSessionRequest.swift
```diff
@@ -1,22 +1,34 @@
...
-/// The POST body for the `createCustomerSession` GraphQL API.
-struct BTCustomerSessionRequest: Encodable {
+/// A `BTCustomerSessionRequest`for creating a customer session.
+/// - Warning: This feature is in beta. It's public API may change or be removed in future releases.
+public struct BTCustomerSessionRequest {
     
-    let customer: Customer
+    let hashedEmail: String?
+    let hashedPhoneNumber: String?
+    let paypalAppInstalled: Bool?
```
jaxdesmarais:
```suggestion
    let payPalAppInstalled: Bool?
```

warmkesselj (approved):

Sources/BraintreeShopperInsights/BTCustomerSessionRequest.swift
richherrera: Done

Sources/BraintreeShopperInsights/BTCreateCustomerSessionAPI.swift
richherrera: Awesome, changes have been made!

Sources/BraintreeShopperInsights/BTCustomerSessionRequest.swift
```diff
@@ -1,22 +1,34 @@
...
-    struct BTPurchaseUnit: Encodable {
-        
-        let amount: String
-        let currencyCode: String
+    /// Creates a BTCustomerSessionRequest
+    /// - Parameters:
+    ///   - hashedEmail: Optional: The customer's email address hashed via SHA256 algorithm.
+    ///   - hashedPhoneNumber: Optional: The customer's phone number hased via SHA256 algorithm.
+    ///   - payPalAppInstalled: Optional: Checks whether the PayPal app is installed on the device.
+    ///   - venmoAppInstalled: Optional: Checks whether the Venmo app is installed on the device
```
jaxdesmarais:
```suggestion
    ///   - venmoAppInstalled: Optional: Checks whether the Venmo app is installed on the device.
```

jaxdesmarais (approved):

PR #1587, opened 2025-06-13
Set `prefersEphemeralWebBrowserSession` Flag Based On Query Param Experiment

agedd:
### Summary of changes

- Read the `experiment` query param that Billing Backend passes within the checkout URL to determine whether to enable/disable the WAS popup within the web fallback flow. 
- If this value is present, set the `prefersEphemeralWebBrowserSession` flag to true, disabling the privacy WAS popup; otherwise, the property is defaulted to false.

**Note**: Waiting on changes from Billing team to be deployed to test this flow thoroughly. 

### Checklist

- [x] Added a changelog entry

- @agedd

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -63,6 +63,10 @@ import BraintreeDataCollector
     /// In the PayPal flow this will be either an EC token or a Billing Agreement token
     private var payPalContextID: String?
     
+    /// Used to determine whether or not to render the WAS popup.
+    /// If the experiement is enabled, set the `prefersEphemeralWebBrowserSession` flag to true.
+    private var experiment: String?
+    
     /// Used for analytics purposes, to determine if brower-presentation event is associated with a locally cached, or remotely fetched `BTConfiguration`
     private var isConfigFromCache: Bool?
 
@@ -401,6 +405,8 @@ import BraintreeDataCollector
                 }
                 
                 self.payPalContextID = approvalURL.baToken ?? approvalURL.ecToken
+                
+                self.experiment = approvalURL.experiment
 
                 let dataCollector = BTDataCollector(apiClient: self.apiClient)
                 self.clientMetadataID = self.payPalRequest?.riskCorrelationID ?? dataCollector.clientMetadataID(self.payPalContextID)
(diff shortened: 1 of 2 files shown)


warmkesselj (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -463,6 +463,11 @@ import BraintreeDataCollector
         approvalURL = appSwitchURL
         webSessionReturned = false
         
+        let urlComponents = URLComponents(url: appSwitchURL, resolvingAgainstBaseURL: true)
+        if let experiment = urlComponents?.queryItems?.first(where: { $0.name == "experiment" })?.value, !experiment.isEmpty {
+            webAuthenticationSession.prefersEphemeralWebBrowserSession = true
+        }
```
richherrera: We could move this into a separate method, I think this new logic would be easier to maintain if it's isolated in its own function.

richherrera: we could also consider following this approach https://github.com/braintree/braintree_ios/blob/fb0f006059216a41c449c24782398f8ae96a605d/Sources/BraintreePayPal/BTPayPalApprovalURLParser.swift#L24?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -463,6 +463,11 @@ import BraintreeDataCollector
         approvalURL = appSwitchURL
         webSessionReturned = false
         
+        let urlComponents = URLComponents(url: appSwitchURL, resolvingAgainstBaseURL: true)
+        if let experiment = urlComponents?.queryItems?.first(where: { $0.name == "experiment" })?.value, !experiment.isEmpty {
```
tdchow: I believe we'll need to check for this key/value pair: `experiment=InAppBrowserNoPopup` on when to set `prefersEphemeralWebBrowserSession` to true. It sounded like they may want to run a number of experiments, so we may need to add a condition to check if the value of the query param is `InAppBrowserNoPopup`.

agedd: ahh yep ty for calling this out, addressed in 21373462341054edf92a4cf6a2c412c7a5a967e7

Sources/BraintreePayPal/BTPayPalClient.swift
agedd: ahh yep, cleaned up and addressed in 21373462341054edf92a4cf6a2c412c7a5a967e7

richherrera (approved):

tdchow (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -463,6 +469,8 @@ import BraintreeDataCollector
         approvalURL = appSwitchURL
         webSessionReturned = false
         
+        webAuthenticationSession.prefersEphemeralWebBrowserSession = experiment == "InAppBrowserNoPopup"
```
jaxdesmarais, line 472: Are we capturing this in our analytics somewhere on the PayPal side? Just want to make sure we are able to track if this is making a difference or not in terms of conversion. I am sure this was already discussed so am just catching up. :)

tdchow: Yep! We discussed if we should be removing the query param in the checkout URL before we open it. We decided to keep it in the URL so the tracking of the experiment can live on the web side.

PR #1590, opened 2025-06-23
Added analytics

baubrey91:

### Summary of changes

- Added FTPI analytics for when app is backgrounded

### Checklist

- [ ] Added a changelog entry

- @baubrey91

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -84,6 +84,13 @@ import BraintreeDataCollector
             name: UIApplication.didBecomeActiveNotification,
             object: nil
         )
+        
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(applicationDidEnterBackground),
+            name: UIApplication.didEnterBackgroundNotification,
+            object: nil
+        )
     }
 
     /// Initialize a new PayPal client instance for the PayPal App Switch flow.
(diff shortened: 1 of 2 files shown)


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -281,6 +288,23 @@ import BraintreeDataCollector
     
     @objc func applicationDidBecomeActive(notification: Notification) {
         webSessionReturned = true
+        apiClient.sendAnalyticsEvent(
```
jaxdesmarais, line 291: Both of these events will be sent any time the app is sent to the background/foreground for any reason. How do we expect to use these events since both will be sent any time the merchant app is opened/closed by the customer? We will have these 2 events for practically every interaction. Have we tested this with breakpoints to ensure this is the behavior we expect?

baubrey91: I believe the ask is to ensure the user is actually leaving the Merchant app after attempting to deep link to the PP app.

jaxdesmarais: I am seeing both of these events sent for every attempt (app switch or legacy fallback web flow). Have you done manual testing with this change and adding breakpoints? `applicationDidEnterBackground` will be sent for every session ID as the customer would navigate away from the merchant app at the end of the flow or when switching to the PayPal app. `applicationDidBecomeActive` is sent when we present the web based flow as well as when returning back to the app after a successful app switch.

PR #1591, opened 2025-06-24
Fix PayPal Phone Number

jaxdesmarais:
### Summary of changes

- Fix bug where `BTPayPalRequest.userPhoneNumber` could be passed as an empty string resulting in an error

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais


jaxdesmarais:
> In the demo app, the API was returning this error, but only in the sandbox environment.

@richherrera that's correct. This feature has been rolled out in Sandbox so far, so we will also see this error in Production once released there as well.

PR #1592, opened 2025-06-24
Add GenerateCustomerRecommendations GraphQL Query

agedd:
### Summary of changes

- Add a new feature to generate customer recommendations using the `GenerateCustomerRecommendations` GraphQL mutation.
- Implement the `BTCustomerRecommendationsApi` class, its corresponding GraphQL POST body, and unit tests for both the class and model

### Checklist

- [ ]~Added a changelog entry~

- @agedd

### DIFF
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -103,6 +103,9 @@
 		57D94370296CC79B0079EAB1 /* BraintreePayPal_IntegrationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 57D9436F296CC79B0079EAB1 /* BraintreePayPal_IntegrationTests.swift */; };
 		57D94372296CCA2F0079EAB1 /* String+NonceValidation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 57D94371296CCA2F0079EAB1 /* String+NonceValidation.swift */; };
 		620C3C652BA21E5700C40A03 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 620C3C642BA21E5700C40A03 /* PrivacyInfo.xcprivacy */; };
+		621149DA2E01E9A5006D7687 /* BTCustomerRecommendationsAPI.swift in Sources */ = {isa = PBXBuildFile; fileRef = 621149D92E01E999006D7687 /* BTCustomerRecommendationsAPI.swift */; };
+		621149DC2E01EC7F006D7687 /* GenerateCustomerRecommendationsGraphQLBody.swift in Sources */ = {isa = PBXBuildFile; fileRef = 621149DB2E01EC74006D7687 /* GenerateCustomerRecommendationsGraphQLBody.swift */; };
+		621149DE2E0201C7006D7687 /* Variables.swift in Sources */ = {isa = PBXBuildFile; fileRef = 621149DD2E0201C3006D7687 /* Variables.swift */; };
 		62236E1A2B98CFB000CDCC37 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 62236E192B98CFB000CDCC37 /* PrivacyInfo.xcprivacy */; };
 		624454DE2DDE399D00A3383B /* CreateCustomerSessionMutationGraphQLBody.swift in Sources */ = {isa = PBXBuildFile; fileRef = 624454DD2DDE398200A3383B /* CreateCustomerSessionMutationGraphQLBody.swift */; };
 		624B27F72B6AE0C2000AC08A /* BTShopperInsightsError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 624B27F62B6AE0C2000AC08A /* BTShopperInsightsError.swift */; };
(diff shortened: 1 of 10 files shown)


Braintree.xcodeproj/project.pbxproj
```diff
@@ -67,7 +67,6 @@
 		455AA27C2DEE38ED008D99AB /* UpdateCustomerSessionMutationGraphQLBody.swift in Sources */ = {isa = PBXBuildFile; fileRef = 455AA27A2DEE38ED008D99AB /* UpdateCustomerSessionMutationGraphQLBody.swift */; };
 		455AA27F2DEF8E4C008D99AB /* BTPurchaseUnit.swift in Sources */ = {isa = PBXBuildFile; fileRef = 455AA27E2DEF8E4C008D99AB /* BTPurchaseUnit.swift */; };
 		455AA2822DF0E5DC008D99AB /* BTUpdateCustomerSessionAPI_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 455AA2812DF0E5DC008D99AB /* BTUpdateCustomerSessionAPI_Tests.swift */; };
-		455AA27F2DEF8E4C008D99AB /* BTPurchaseUnit.swift in Sources */ = {isa = PBXBuildFile; fileRef = 455AA27E2DEF8E4C008D99AB /* BTPurchaseUnit.swift */; };
```
agedd, line 70: **note**: not sure why this untouched file has been removed from the project's graph - re-added this file to enable its reference

Sources/BraintreeShopperInsights/Models/CreateCustomerSessionMutationGraphQLBody.swift
```diff
@@ -15,68 +14,6 @@ struct CreateCustomerSessionMutationGraphQLBody: Encodable {
                 }
             }
             """
-        variables = Variables(request: request)
-    }
-    
-    struct Variables: Encodable {
```
agedd, line 21: **note**: moved the `Variables` struct into its own file as it is used commonly across all three GraphQL PR v2 mutations.

Sources/BraintreeShopperInsights/BTCustomerRecommendationsAPI.swift
```diff
@@ -0,0 +1,67 @@
...
+                    paymentOptions?.append(
+                        BTPaymentOptions(
+                            paymentOption: recommendation["paymentOption"].asString() ?? "",
+                            recommendedPriority: recommendation["recommendedPriority"].asIntegerOrZero()
+                        )
+                    )
+                }
+            }
+            
+            guard let sessionID = sessionID, let isInPayPalNetwork = isInPayPalNetwork, let paymentOptions = paymentOptions else {
```
jaxdesmarais:
```suggestion
            guard let sessionID, let isInPayPalNetwork, let paymentOptions else {
```

jaxdesmarais: Will `paymentOptions` ever be empty or is it always returned?

Sources/BraintreeShopperInsights/Models/CreateCustomerSessionMutationGraphQLBody.swift
```diff
@@ -15,68 +14,6 @@ struct CreateCustomerSessionMutationGraphQLBody: Encodable {
...
-                }
-                
-                struct Amount: Encodable {
-                    
-                    let value: String?
-                    let currencyCode: String?
-                }
-            }
-        }
+        variables = Variables(request: request, sessionID: nil)
```
jaxdesmarais: Will the session ID always be nil with this? Or should we be passing a value in?

jaxdesmarais: Ah, I see this is the create session so yes should be nil here. I wonder if we should default it to nil in the init so we can omit it here. May make it a bit more clear. What do you think?

UnitTests/BraintreeShopperInsightsTests/BTGenerateCustomerRecommendationAPI_Tests.swift
```diff
@@ -0,0 +1,122 @@
+import XCTest
+@testable import BraintreeTestShared
+@testable import BraintreeCore
+@testable import BraintreeShopperInsights
+
+class BTGenerateCustomerRecommendationApi_Tests: XCTestCase {
```
jaxdesmarais:
```suggestion
class BTGenerateCustomerRecommendationAPI_Tests: XCTestCase {
```

Sources/BraintreeShopperInsights/BTCustomerRecommendationsAPI.swift
agedd: ahh yep i just checked and realized `paymentOptions` along with `sessionID` and `isInPayPalNetwork` are all optional, so removed the empty check here - addressed in e2df581fa71c70ae79670606164357d82c06bfc2

Sources/BraintreeShopperInsights/Models/CreateCustomerSessionMutationGraphQLBody.swift
agedd: yep agreed, addressed in e2df581fa71c70ae79670606164357d82c06bfc2

saperi22 (approved):

Sources/BraintreeShopperInsights/BTCustomerRecommendationsAPI.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// This method will call the `GenerateCustomerRecommendations` GQL query, which returns a `BTCustomerRecommendationsResult` if successful.
+    /// - Parameters:
+    ///    - request: A `BTCustomerSessionRequest`
+    ///    - sessionID: The session ID to update.
+    ///    - Returns: A `BTCustomerRecommendationsResult` which determines what payment options to render.
+    ///    - Throws: An error if the request fails or if the response is invalid.
+    func execute(
+        _ request: BTCustomerSessionRequest,
+        sessionID: String
+    ) async throws -> BTCustomerRecommendationsResult? {
```
jaxdesmarais: Does the `BTCustomerRecommendationsResult` need to be optional? It looks like we always throw an error or return a result.

agedd: ahh yep that's right, it doesn't need to be - updated in bc4d51fd25ed14c3b9d001ac724655eea34d44e8

jaxdesmarais (approved):

PR #1594, opened 2025-06-26
Visa checkout feature nonce

warmkesselj:
### Summary of changes

Adding Visa Checkout Nonce and UserData. 

### Checklist

- [X] Added a changelog entry
- [X ] Tested and confirmed payment flows affected by this change are functioning as expected

### Authors
@jwarmkessel @warmkesselj


Sources/BraintreeVisaCheckout/BTVisaCheckoutAddress.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+/// A model representing a user's address information returned by Visa Checkout.
+/// Parses and stores address values from a BTJSON object.
+@objc public class BTVisaCheckoutAddress: NSObject {
```
stechiu: Are all of these parameters supposed to be optional? I'd think that certain values can't be blank, like `firstName` and address information

Sources/BraintreeVisaCheckout/BTVisaCheckoutCardNonce.swift
```diff
@@ -0,0 +1,91 @@
...
+public class BTVisaCheckoutCardNonce: BTCardNonce {
+
+    public let shippingAddress: BTVisaCheckoutAddress?
+    public let billingAddress: BTVisaCheckoutAddress?
+    public let userData: BTVisaCheckoutUserData?
+
+    @objc public convenience init(
+        nonce: String,
+        type: String,
+        lastTwo: String,
```
stechiu: out of curiosity, what's `lastTwo`? like last 2 digits of visa card number?

warmkesselj: I believe it's a string representing the last two digits of a payment card number, typically used in display or masked representations. Versus last four of a cc.

Sources/BraintreeVisaCheckout/BTVisaCheckoutAddress.swift
warmkesselj: I think these properties can't be guaranteed so they are optional. Below is a sample of the `BTVisaCheckoutAddress.m` which we are using as template to translate to the swift version.

https://github.com/braintree/braintree-ios-visa-checkout/blob/main/BraintreeVisaCheckout/BTVisaCheckoutAddress.m

```
- (instancetype)initWithJSON:(BTJSON *)addressJSON {
    if (self = [super init]) {
        _firstName = [addressJSON[@"firstName"] asString];
```

jaxdesmarais: For this module work it's already moved into the Android SDK, so we can always compare there. Looks like these are optional there as well: https://github.com/braintree/braintree_android/blob/main/VisaCheckout/src/main/java/com/braintreepayments/api/visacheckout/VisaCheckoutAddress.kt

Sources/BraintreeVisaCheckout/BTVisaCheckoutCardNonce.swift
jaxdesmarais: We can use the docstrings from android for these: https://github.com/braintree/braintree_android/blob/main/VisaCheckout/src/main/java/com/braintreepayments/api/visacheckout/VisaCheckoutNonce.kt

Sources/BraintreeVisaCheckout/BTVisaCheckoutCardNonce.swift
```diff
@@ -0,0 +1,91 @@
...
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+#if canImport(BraintreeCard)
+import BraintreeCard
+#endif
+
+public class BTVisaCheckoutCardNonce: BTCardNonce {
```
jaxdesmarais: We should be subclassing `BTPaymentMethodNonce` here instead, and can remove the `open` change in the `BTCardNonce`: https://github.com/braintree/braintree_android/blob/356ed7229e16e19d2873ab01e8cd715dc3c9ac40/VisaCheckout/src/main/java/com/braintreepayments/api/visacheckout/VisaCheckoutNonce.kt#L37

jaxdesmarais: Rename to `BTVisaCheckoutNonce`

Sources/BraintreeVisaCheckout/BTVisaCheckoutUserData.swift
```diff
@@ -0,0 +1,31 @@
...
+import BraintreeCard
+#endif
+
+@objc public class BTVisaCheckoutUserData: NSObject {
+
+    @objc public let firstName: String?
+    @objc public let lastName: String?
+    @objc public let fullName: String?
+    @objc public let username: String?
+    @objc public let email: String?
```
jaxdesmarais: Lets match the names used in Android for these properties: https://github.com/braintree/braintree_android/blob/main/VisaCheckout/src/main/java/com/braintreepayments/api/visacheckout/VisaCheckoutUserData.kt

Sources/BraintreeVisaCheckout/BTVisaCheckoutUserData.swift
```diff
@@ -0,0 +1,31 @@
...
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+#if canImport(BraintreeCard)
+import BraintreeCard
+#endif
+
+@objc public class BTVisaCheckoutUserData: NSObject {
```
jaxdesmarais: Can use `@objcMembers` instead

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutCardNonce_Tests.swift
```diff
@@ -0,0 +1,65 @@
+import XCTest
+@testable import BraintreeCore
+@testable import BraintreeVisaCheckout
+
+final class BTVisaCheckoutCardNonceTests: XCTestCase {
+
+    func testVisaCheckoutCardNonce_parsesValidJSON() {
+        let json = BTJSON(value: [
```
jaxdesmarais, line 8: Can we pass and test on all the values?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutCardNonce_Tests.swift
```diff
@@ -0,0 +1,65 @@
+import XCTest
+@testable import BraintreeCore
+@testable import BraintreeVisaCheckout
+
+final class BTVisaCheckoutCardNonceTests: XCTestCase {
```
jaxdesmarais:
```suggestion
final class BTVisaCheckoutNonce_Tests: XCTestCase {
```

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutUserData_Tests.swift
```diff
@@ -0,0 +1,44 @@
+import XCTest
+
+@testable import BraintreeCore
+@testable import BraintreeVisaCheckout
+
+final class BTVisaCheckoutUserDataTests: XCTestCase {
```
jaxdesmarais:
```suggestion
final class BTVisaCheckoutUserData_Tests: XCTestCase {
```

stechiu (approved):

stechiu:
> @jwarmkessel or @stechiu are one of y'all able to resolve the failing tests?

I fixed the unit test. UI test failures were resolved by pulling in latest from `main`. For the integration test, it's coming from `BraintreeAmexExpress_IntegrationTests` - `testGetRewardsBalance_returnsResult, failed - Unexpected error: Authorization fingerprint is invalid`. I haven't resolved integration tests before so it will require some time tomorrow.

Sources/BraintreeVisaCheckout/BTVisaCheckoutNonce.swift
```diff
@@ -0,0 +1,43 @@
...
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+#if canImport(BraintreeCard)
+import BraintreeCard
+#endif
+
+@objc public class BTVisaCheckoutNonce: BTPaymentMethodNonce {
```
jaxdesmarais: We can use `@objcMembers` instead of all the `@objc` annotations. Take a look at the other nonce classes as a reference.

Sources/BraintreeVisaCheckout/BTVisaCheckoutNonce.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    @objc public let lastTwo: String
+    @objc public let cardType: String
+    @objc public let billingAddress: BTVisaCheckoutAddress
+    @objc public let shippingAddress: BTVisaCheckoutAddress
+    @objc public let userData: BTVisaCheckoutUserData
+    @objc public let callId: String
```
jaxdesmarais:
```suggestion
    @objc public let callID: String
```

jaxdesmarais: Lets add docstrings to all of these: https://github.com/braintree/braintree_android/blob/main/VisaCheckout/src/main/java/com/braintreepayments/api/visacheckout/VisaCheckoutNonce.kt

Sources/BraintreeVisaCheckout/BTVisaCheckoutNonce.swift
```diff
@@ -0,0 +1,43 @@
...
+    @objc public let binData: BTBinData
+
+    @objc public init(
```
jaxdesmarais: This can be internal. We can also just init with `BTJSON` and extract each of these from the JSON. You can take a look at other nonce classes as a reference.

jaxdesmarais: We will also need to add the parsing logic to the init. Can use Android as a reference and unit tests to verify.

Sources/BraintreeVisaCheckout/BTVisaCheckoutUserData.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+#if canImport(BraintreeCard)
+import BraintreeCard
+#endif
```
jaxdesmarais: Do we need this import?

Sources/BraintreeVisaCheckout/BTVisaCheckoutNonce.swift
```diff
@@ -0,0 +1,43 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+#if canImport(BraintreeCard)
+import BraintreeCard
+#endif
```
jaxdesmarais: Is card set up as a dependency of this module?

Sources/BraintreeVisaCheckout/BTVisaCheckoutUserData.swift
```diff
@@ -0,0 +1,31 @@
...
+
+@objcMembers public class BTVisaCheckoutUserData: NSObject {
+
+    public let userFirstName: String?
+    public let userLastName: String?
+    public let userFullName: String?
+    public let username: String?
+    public let userEmail: String?
+
+    @objc public init(json: BTJSON) {
```
jaxdesmarais: This doesn't need to be public or have `@objc`

Sources/BraintreeVisaCheckout/BTVisaCheckoutUserData.swift
```diff
@@ -0,0 +1,31 @@
...
+        self.userLastName = json["userLastName"].asString()
+        self.userFullName = json["userFullName"].asString()
+        self.username = json["userName"].asString()
+        self.userEmail = json["userEmail"].asString()
+        super.init()
+    }
+
+    public static func userData(with json: BTJSON) -> BTVisaCheckoutUserData {
+        return BTVisaCheckoutUserData(json: json)
+    }
```
jaxdesmarais: Is this helper method being used?

Sources/BraintreeVisaCheckout/BTVisaCheckoutAddress.swift
```diff
@@ -37,7 +37,7 @@ import BraintreeCore
 
     /// Initializes a `BTVisaCheckoutAddress` from a BTJSON object.
     /// - Parameter json: A BTJSON object containing the address fields.
-    @objc public init(json: BTJSON) {
+    public init(json: BTJSON) {
```
jaxdesmarais: I don't think this needs to be public, right?

jaxdesmarais: I think this class can also use `objcMembers` but that can be a follow up PR.

Sources/BraintreeVisaCheckout/BTVisaCheckoutNonce.swift
```diff
@@ -0,0 +1,55 @@
...
+    public let userData: BTVisaCheckoutUserData
+    
+    // The Call ID from the VisaPaymentSummary.
+    public let callID: String
+    
+    // The BIN data for the card number associated with [VisaCheckoutNonce]
+    public let binData: BTBinData
+
+    init?(json: BTJSON) {
+
```
jaxdesmarais, line 33:
```suggestion
```

Sources/BraintreeVisaCheckout/BTVisaCheckoutNonce.swift
```diff
@@ -0,0 +1,55 @@
...
+        let json = json["visaCheckoutCards"].isArray ? json["visaCheckoutCards"][0] : json
```
jaxdesmarais:
```suggestion
        let visaCheckoutCards = json["visaCheckoutCards"].isArray ? json["visaCheckoutCards"][0] : json
```

Sources/BraintreeVisaCheckout/BTVisaCheckoutNonce.swift
```diff
@@ -0,0 +1,55 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+@objcMembers public class BTVisaCheckoutNonce: BTPaymentMethodNonce {
```
jaxdesmarais, line 8: Lets add a docstring to the top of this.

Sources/BraintreeVisaCheckout/BTVisaCheckoutNonce.swift
```diff
@@ -0,0 +1,55 @@
...
+        }
+        
+        self.lastTwo = lastTwo
+        self.billingAddress = BTVisaCheckoutAddress(json: json["billingAddress"])
+        self.shippingAddress = BTVisaCheckoutAddress(json: json["shippingAddress"])
+        self.userData = BTVisaCheckoutUserData(json: json["userData"])
+        self.binData = BTBinData(json: json["binData"])
+        self.cardType = cardType
+        self.callID = callID
+        let isDefault = json["default"].isTrue
```
jaxdesmarais: Should this be in the guard?

Sources/BraintreeVisaCheckout/BTVisaCheckoutUserData.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+@objcMembers public class BTVisaCheckoutUserData: NSObject {
```
jaxdesmarais, line 8: Lets add docstrings here. Can use what's used on Android: https://github.com/braintree/braintree_android/blob/9a20611391e7cc4f7ce33717dfacad4e47b45143/VisaCheckout/src/main/java/com/braintreepayments/api/visacheckout/VisaCheckoutUserData.kt#L11-L17

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutCardNonce_Tests.swift
```diff
@@ -0,0 +1,62 @@
+import XCTest
+@testable import BraintreeCore
+@testable import BraintreeVisaCheckout
+
+final class BTVisaCheckoutCardNonce_Tests: XCTestCase {
+    
+    func testInitFromJSON_parsesAllFieldsCorrectly() {
```
jaxdesmarais, line 7: It looks like this may be missing some fields compared to the Android SDK: https://github.com/braintree/braintree_android/blob/9a20611391e7cc4f7ce33717dfacad4e47b45143/VisaCheckout/src/test/java/com/braintreepayments/api/visacheckout/VisaCheckoutNonceUnitTest.java#L25

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutCardNonce_Tests.swift
```diff
@@ -0,0 +1,62 @@
+import XCTest
+@testable import BraintreeCore
+@testable import BraintreeVisaCheckout
+
+final class BTVisaCheckoutCardNonce_Tests: XCTestCase {
```
jaxdesmarais, line 5: It looks like Android has an additional test that we may be missing: https://github.com/braintree/braintree_android/blob/9a20611391e7cc4f7ce33717dfacad4e47b45143/VisaCheckout/src/test/java/com/braintreepayments/api/visacheckout/VisaCheckoutNonceUnitTest.java#L80C17-L80C78

Sources/BraintreeVisaCheckout/BTVisaCheckoutNonce.swift
```diff
@@ -0,0 +1,55 @@
...
+        
+        guard
+            let lastTwo = json["details"]["lastTwo"].asString(),
+            let cardType = json["details"]["cardType"].asString(),
+            let nonce = json["nonce"].asString(),
+            let type = json["type"].asString(),
+            let callID = json["callId"].asString()
+        else {
+            return nil
+        }
```
jaxdesmarais: It looks like in the rest of the SDK we just return nil of the nonce is not returned. Do we want to do the same here?

Sources/BraintreeVisaCheckout/BTVisaCheckoutUserData.swift
```diff
@@ -0,0 +1,33 @@
...
+    // The user's full name.
+    public let userFullName: String?
+    
+    // The user's username.
+    public let username: String?
+    
+    // The user's email.
+    public let userEmail: String?
+
+    init(json: BTJSON) {
```
jaxdesmarais: nit: since these are all optional we should make `BTJSON` optional in the init.

Sources/BraintreeVisaCheckout/BTVisaCheckoutUserData.swift
```diff
@@ -0,0 +1,33 @@
...
+        self.userFirstName = json["userFirstName"].asString()
+        self.userLastName = json["userLastName"].asString()
+        self.userFullName = json["userFullName"].asString()
+        self.username = json["userName"].asString()
+        self.userEmail = json["userEmail"].asString()
+        super.init()
```
jaxdesmarais: Can we remove this? It doesn't look like other classes require us to call the super init like `BTBinData`

Sources/BraintreeVisaCheckout/BTVisaCheckoutNonce.swift
```diff
@@ -0,0 +1,54 @@
...
+    init?(json: BTJSON) {
+        let visaCheckoutCards = json["visaCheckoutCards"].isArray ? json["visaCheckoutCards"][0] : json
+
+        guard let nonce = visaCheckoutCards["nonce"].asString() else {
+            return nil
+        }
+
+        let lastTwo = visaCheckoutCards["details"]["lastTwo"].asString() ?? ""
+        let cardType = visaCheckoutCards["details"]["cardType"].asString() ?? "Unknown"
+        let type = visaCheckoutCards["type"].asString() ?? "VisaCheckout"
```
jaxdesmarais: Looks like this isn't being used, so we should be able to remove it. This will resolve the cocoapods error according to the CI output.

Sources/BraintreeVisaCheckout/BTVisaCheckoutNonce.swift
```diff
@@ -0,0 +1,54 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+// A `nonce` representing a Visa Checkout card.
```
jaxdesmarais: nit: looks like we use `///` for comments in the rest of the SDK. Can we update this class and properties to match that? Same with `UserData`

Sources/BraintreeVisaCheckout/BTVisaCheckoutAddress.swift
```diff
@@ -4,40 +4,41 @@ import Foundation
 import BraintreeCore
 #endif
 
-/// A model representing a user's address information returned by Visa Checkout.
-/// Parses and stores address values from a BTJSON object.
-@objc public class BTVisaCheckoutAddress: NSObject {
+
+// A class containing Visa Checkout information about the user's address.
+// Parses and stores address values from a `BTJSON` object.
```
jaxdesmarais: To match the rest of the SDK docstrings:
```suggestion
/// A class containing Visa Checkout information about the user's address.
/// Parses and stores address values from a `BTJSON` object.
```

Sources/BraintreeVisaCheckout/BTVisaCheckoutAddress.swift
```diff
@@ -4,40 +4,41 @@ import Foundation
 import BraintreeCore
 #endif
 
-/// A model representing a user's address information returned by Visa Checkout.
-/// Parses and stores address values from a BTJSON object.
-@objc public class BTVisaCheckoutAddress: NSObject {
+
+/// A class containing Visa Checkout information about the user's address.
+/// Parses and stores address values from a `BTJSON` object.
```
jaxdesmarais: This is done in the background and doesn't need to be exposed to merchants.
```suggestion
```

Sources/BraintreeVisaCheckout/BTVisaCheckoutAddress.swift
```diff
@@ -53,7 +54,7 @@ import BraintreeCore
     /// Convenience method for creating an address object from JSON.
     /// - Parameter json: A BTJSON object containing the address fields.
     /// - Returns: An initialized `BTVisaCheckoutAddress` instance.
-    @objc public static func address(with json: BTJSON) -> BTVisaCheckoutAddress {
+    public static func address(with json: BTJSON) -> BTVisaCheckoutAddress {
```
jaxdesmarais: Doesn't look like we are using this. Can it be removed?

jaxdesmarais (approved):

PR #1595, opened 2025-06-30
Add  `createCustomerSession` function to BTShopperInsightsClientV2

agedd:
### Summary of changes

- Add the `createCustomerSession` func to the `BTShopperInsightsClientV2` class

### Checklist

- [x] Added a changelog entry
- ~[] Tested and confirmed payment flows affected by this change are functioning as expected~

- @agedd

### DIFF
--- a/Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
+++ b/Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
@@ -17,8 +17,6 @@ public class BTShopperInsightsClientV2 {
     /// Defaults to `UIApplication.shared`, but exposed for unit tests to mock calls to `canOpenURL`.
     var application: URLOpener = UIApplication.shared
     
-    lazy var createCustomerSessionAPI = BTCreateCustomerSessionAPI(apiClient: apiClient)
-
     // MARK: - Private Properties
     
     private let apiClient: BTAPIClient
@@ -41,7 +39,7 @@ public class BTShopperInsightsClientV2 {
     /// - Returns: A `String` representing a session ID if successful
     /// - Throws: An error if the request fails for some reason or if the response is invalid.
     public func createCustomerSession(request: BTCustomerSessionRequest) async throws -> String {
-        try await createCustomerSessionAPI.execute(request)
+        try await BTCreateCustomerSessionAPI(apiClient: apiClient).execute(request)
     }
     
     /// Call this method when the PayPal or Venmo button has been successfully displayed to the buyer.
--- a/Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
+++ b/Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
@@ -17,6 +17,8 @@ public class BTShopperInsightsClientV2 {
     /// Defaults to `UIApplication.shared`, but exposed for unit tests to mock calls to `canOpenURL`.
     var application: URLOpener = UIApplication.shared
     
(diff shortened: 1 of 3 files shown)


Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
```diff
@@ -17,6 +17,8 @@ public class BTShopperInsightsClientV2 {
     /// Defaults to `UIApplication.shared`, but exposed for unit tests to mock calls to `canOpenURL`.
     var application: URLOpener = UIApplication.shared
     
+    lazy var createCustomerSessionAPI = BTCreateCustomerSessionAPI(apiClient: apiClient)
```
jaxdesmarais: Can we instead create this in the method? I am a bit hesitant to lazily instantiate the API layer since that's something we haven't done before. This would always be called after we have the `apiClient` in the init so should be fine to create there. What do you think?

agedd: ahh yep i agree, also feels it matches the pattern we use everywhere else

richherrera: I was thinking about instantiating the API in the init so we can inject mocks in the future, and avoid creating a new instance every time the method is called. However, this API seems pretty lightweight, stateless, and driven by the parameters passed in, so I think this works great as is. Nice work and great suggestion!

Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
```diff
@@ -33,6 +33,15 @@ public class BTShopperInsightsClientV2 {
     
     // MARK: - Public Methods
     
+    /// This method creates a new customer session.
+    /// - Parameters:
+    ///    - request: A `BTCustomerSessionRequest`
+    /// - Returns: A `String` representing a session ID if successful
+    /// - Throws: An error if the request fails for some reason or if the response is invalid.
+    public func createCustomerSession(request: BTCustomerSessionRequest) async throws -> String {
```
richherrera, line 41: Should we only expose the async/await method, or should we also give merchants the option to use it with closures/completions?

jaxdesmarais: I looks like for the current method we only expose async/await. Should be fine to not expose a completion since that's the direction we are (slowly) moving.

Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
jaxdesmarais: Gotcha, that makes sense. Yeah I think it may be overkill for a more lightweight method, but totally understand that thinking for more complex requests in the SDK.

Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
```diff
@@ -33,6 +33,15 @@ public class BTShopperInsightsClientV2 {
...
+        try await BTCreateCustomerSessionAPI(apiClient: apiClient).execute(request)
```
jaxdesmarais: Take it or leave it:
```suggestion
        let createCustomerSessionAPI = BTCreateCustomerSessionAPI(apiClient: apiClient)
        try await createCustomerSessionAPI.execute(request)
```

jaxdesmarais (approved):

richherrera (approved):

PR #1596, opened 2025-07-02
Added BTVisaCheckoutClient to Visa Checkout

stechiu:
Converted VisaCheckoutClient form ObjC to Swift. **This PR depends on merging in https://github.com/braintree/braintree_ios/pull/1602 first**

### Summary of changes

- Converted these from these 2 files from `braintree-ios-visa-checkout`: [BTVisaCheckoutClient.m](https://github.com/braintree/braintree-ios-visa-checkout/blob/main/BraintreeVisaCheckout/BTVisaCheckoutClient.m) + [BTVisaCheckoutClient.h](https://github.com/braintree/braintree-ios-visa-checkout/blob/main/BraintreeVisaCheckout/Public/BraintreeVisaCheckout/BTVisaCheckoutClient.h)

### Checklist

- [x] Added a changelog entry

- @stechiu


jaxdesmarais:
The base branch should be the feature branch, not main

Sources/BraintreeVisaCheckout/BTVisaCheckoutError.swift
```diff
@@ -0,0 +1,39 @@
...
+    /// 2. Braintree SDK is integrated incorrectly.
+    case integration
+
+    /// 3. Visa Checkout SDK responded with an unsuccessful status code.
+    case checkoutUnsuccessful
+
+    /// 4. Visa Checkout was cancelled by the user.
+    case cancelled
+
+    public var errorCode: String {
```
richherrera, line 25: In our SDK, most flow errors include an `errorCode` and an `errorDescription`. Typically, the `errorCode` is a numeric identifier for the error, and the `errorDescription` is a brief explanation of it.
Do you think it would be a good idea to follow the same approach?

Reference https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreePayPal/BTPayPalError.swift

Sources/BraintreeVisaCheckout/BTVisaCheckoutConfiguration.swift
```diff
@@ -0,0 +1,49 @@
+import BraintreeCore
+import VisaCheckoutSDK
+
+extension BTConfiguration {
```
richherrera, line 4: Just out of curiosity, do we really need all these variables to be public?

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -1,6 +1,180 @@
...
+    /// Tokenizes a Visa checkout result. Exposed for testing properties of the Visa `CheckoutResult`
+    /// - Parameters:
+    ///   - statusCode: The result code indicating the status of a completed Visa Checkout transaction.
+    ///   - callId: The unique identifier for the Visa Checkout transaction.
+    ///   - encryptedKey: The encrypted key associated with the Visa Checkout transaction.
+    ///   - encryptedPaymentData: The encrypted payment data for the Visa Checkout transaction.
+    ///   - completion: A completion block that is invoked when tokenization has completed.
+    @objc public func tokenizeVisaCheckoutResult(
+        statusCode: CheckoutResultStatus,
+        callId: String?,
```
richherrera:
```suggestion
        callID: String?,
```

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -1,6 +1,180 @@
...
+            encryptedKey: encryptedKey,
+            encryptedPaymentData: encryptedPaymentData,
+            completion: completion
+        )
+    }
+
+    /// Tokenizes a Visa checkout result. Exposed for testing properties of the Visa `CheckoutResult`
+    /// - Parameters:
+    ///   - statusCode: The result code indicating the status of a completed Visa Checkout transaction.
+    ///   - callId: The unique identifier for the Visa Checkout transaction.
```
richherrera, line 83:
```suggestion
    ///   - callID: The unique identifier for the Visa Checkout transaction.
```

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -1,6 +1,180 @@
...
+                return
+            }
+
+            guard let configuration = configuration, configuration.isVisaCheckoutEnabled else {
+                let error = BTVisaCheckoutError.unsupported
+                completion(nil, error)
+                return
+            }
+
+            let environmentString = configuration.json?["environment"].asString()
```
richherrera: What if we add `configuration.json?["environment"].asString()` as a variable in BTVisaCheckoutConfiguration?

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -1,6 +1,180 @@
...
 import Foundation
 
-public class BTVisaCheckoutClient {
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
 
-    // TODO: - Add implementation
+// Creates a Visa Checkout client for processing Visa Checkout payments.
+@objc public class BTVisaCheckoutClient: NSObject {
```
richherrera: Just out of curiosity, why does this client need to be of type `NSObject`? Do its methods need to be accessible from `Objective-C` for any particular reason?

stechiu: I'm not sure because this is the first time I'm working on an ObjC file and converting it to Swift. Should I remove NSObject? The methods wouldn't need to be accessible since they don't reference anything ObjC

Podfile
```diff
@@ -8,12 +8,14 @@ inhibit_all_warnings!
 target 'Demo' do
   project 'Demo/Demo'
   pod 'InAppSettingsKit'
+  pod 'VisaCheckoutSDK', '7.0.0'
```
richherrera: I think we need to add the VisaCheckout framework manually, since this Podfile belongs to the demo app. I believe our SDK can’t access the Visa SDK from there 🤔

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
richherrera: From what I can see on the Visa SDK page, it has full compatibility with Swift. I think we could skip inheriting from _`NSObject`,_ I don’t see anything in the client that would require it. We could look into it a bit more and determine whether it’s actually necessary to inherit from Objective-C or not. What do you think?

Podfile
stechiu: I figured out where to download the VisaCheckoutSDK framework so I've updated it to remove the pod

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: Thanks for explaining! I'll test it out. I don't think it's necessary to use NSObject but all the files in `VisaCheckoutSDK` are in ObjC, and there are some essential properties used in `BTVisaCheckoutClient` that's inherited from `VisaCheckoutSDK`

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: I'll update it to `callID`. Heads up, I tried changing `checkoutResult.callId` to have the same syntax but `CheckoutResult` is called from `VisaCheckoutSDK` and Visa uses `callId`

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: thanks for the suggestion

Sources/BraintreeVisaCheckout/BTVisaCheckoutConfiguration.swift
stechiu: Would merchants ever use them? These were public in the original repo and capture information from the Visa Checkout SDK

PR #1598, opened 2025-07-02
Fix Multiple ASWeb Sessions

richherrera:
### Summary of changes

Prevents duplicate session starts, and refining analytics reporting. Additionally, test cases have been updated to reflect these changes.

### Enhancements to Token Handling and Session Management:

* `BTWebAuthenticationSession.swift`:
  - Introduced a `BAToken` type alias and a `currentSession` property to manage active sessions.
  - Added logic to extract `ba_token` from URLs using a helper method `getBAToken(from:)`.
  - Updated the `start` method to prevent duplicate sessions via a new `sessionDidDuplicate` callback and reset the session upon completion.

* `BTPayPalClient.swift`:
  - Replaced `clientMetadataID` with `clientMetadataIDs` to manage multiple correlation IDs tied to tokens.
  - Added a method `extractToken(from:)` to extract `ba_token` or `ec_token` from URLs.
  - Updated session callbacks (`sessionDidAppear`, `sessionDidCancel`, `sessionDidDuplicate`) to handle tokens and send appropriate analytics events, including a new event for duplicate requests. [[1]](diffhunk://#diff-7593900bd413ee81d48fae61aeeab899a00cbfdc0d2a688e21d3acf3f7ef3360R203-R207) [[2]](diffhunk://#diff-7593900bd413ee81d48fae61aeeab899a00cbfdc0d2a688e21d3acf3f7ef3360L501-R511) [[3]](diffhunk://#diff-7593900bd413ee81d48fae61aeeab899a00cbfdc0d2a688e21d3acf3f7ef3360R549-R567)

### Updates to Analytics Tracking:

* `BTPayPalAnalytics.swift`:
  - Added a new analytics event `tokenizeDuplicateRequest` to track duplicate PayPal tokenization attempts.

* `BTPayPalClient.swift`:
  - Updated analytics calls to use `payPalContextID` and its associated correlation ID for more accurate tracking. [[1]](diffhunk://#diff-7593900bd413ee81d48fae61aeeab899a00cbfdc0d2a688e21d3acf3f7ef3360L412-R418) [[2]](diffhunk://#diff-7593900bd413ee81d48fae61aeeab899a00cbfdc0d2a688e21d3acf3f7ef3360L549-R578)

### Checklist

- [ ] Added a changelog entry

### Authors

-  @richherrera

### DIFF
--- a/Sources/BraintreeCore/BTWebAuthenticationSession.swift
+++ b/Sources/BraintreeCore/BTWebAuthenticationSession.swift
@@ -8,6 +8,8 @@ public class BTWebAuthenticationSession: NSObject {
     /// :nodoc: This property is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     public var prefersEphemeralWebBrowserSession: Bool?
 
+    private var currentSession: ASWebAuthenticationSession?
+
     /// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     public func start(
         url: URL,
@@ -16,10 +18,12 @@ public class BTWebAuthenticationSession: NSObject {
         sessionDidAppear: @escaping (Bool) -> Void,
         sessionDidCancel: @escaping () -> Void
     ) {
+        guard currentSession == nil else { return }
         let authenticationSession = ASWebAuthenticationSession(
             url: url,
             callbackURLScheme: BTCoreConstants.callbackURLScheme
         ) { url, error in
+            self.currentSession = nil
             if let error = error as? NSError, error.code == ASWebAuthenticationSessionError.canceledLogin.rawValue {
                 sessionDidCancel()
             } else {
@@ -27,9 +31,9 @@ public class BTWebAuthenticationSession: NSObject {
             }
         }
 
-        authenticationSession.prefersEphemeralWebBrowserSession = prefersEphemeralWebBrowserSession ?? false
-
-        authenticationSession.presentationContextProvider = context
+        currentSession?.prefersEphemeralWebBrowserSession = prefersEphemeralWebBrowserSession ?? false
+        currentSession?.presentationContextProvider = context
+        currentSession = authenticationSession
         DispatchQueue.main.async {
             sessionDidAppear(authenticationSession.start())
         }
--- a/Sources/BraintreeCore/BTWebAuthenticationSession.swift
+++ b/Sources/BraintreeCore/BTWebAuthenticationSession.swift
@@ -9,18 +9,22 @@ public class BTWebAuthenticationSession: NSObject {
     public var prefersEphemeralWebBrowserSession: Bool?
 
     private var currentSession: ASWebAuthenticationSession?
+    
+    public typealias BAToken = String
 
     /// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     public func start(
         url: URL,
         context: ASWebAuthenticationPresentationContextProviding,
-        sessionDidComplete: @escaping (URL?, Error?) -> Void,
+        sessionDidComplete: @escaping (URL?, Error?, BAToken?) -> Void,
         sessionDidAppear: @escaping (Bool) -> Void,
         sessionDidCancel: @escaping () -> Void,
-        sessionDidDuplicate: @escaping () -> Void = { }
+        sessionDidDuplicate: @escaping (BAToken?) -> Void = { _ in }
     ) {
+        let baToken = getBAToken(from: url)
+        
         guard currentSession == nil else {
-            sessionDidDuplicate()
+            sessionDidDuplicate(baToken)
             return
         }
         
@@ -32,7 +36,7 @@ public class BTWebAuthenticationSession: NSObject {
             if let error = error as? NSError, error.code == ASWebAuthenticationSessionError.canceledLogin.rawValue {
                 sessionDidCancel()
             } else {
-                sessionDidComplete(url, error)
+                sessionDidComplete(url, error, baToken)
             }
         }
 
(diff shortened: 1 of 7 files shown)


Sources/BraintreeCore/BTWebAuthenticationSession.swift
```diff
@@ -5,33 +5,58 @@ import AuthenticationServices
...
     /// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     public func start(
         url: URL,
         context: ASWebAuthenticationPresentationContextProviding,
         sessionDidComplete: @escaping (URL?, Error?) -> Void,
-        sessionDidAppear: @escaping (Bool) -> Void,
-        sessionDidCancel: @escaping () -> Void
+        sessionDidAppear: @escaping (Bool, BAToken?) -> Void,
+        sessionDidCancel: @escaping (BAToken?) -> Void,
+        sessionDidDuplicate: @escaping (BAToken?) -> Void = { _ in }
```
jaxdesmarais: Instead of the BA token what do we think about passing the URL? That way extracting the token from the URL can all live in the client vs in this more abstract class. Could also consider having 2 methods in this class, 1 with the duplicate check and 1 without. What do you think?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -26,7 +26,7 @@ import BraintreeDataCollector
 
     /// Exposed for testing the clientMetadataID associated with this request.
     /// Used in POST body for FPTI analytics & `/paypal_account` fetch.
-    var clientMetadataID: String?
+    var clientMetadataIDs: [String: String] = [:]
```
jaxdesmarais, line 30: It may be helpful to add some docstrings on why we need this as a dictionary

jaxdesmarais (commented):
I think we will likely want a changelog entry for this

Sources/BraintreeCore/BTWebAuthenticationSession.swift
jaxdesmarais: Actually we may not need to pass anything around since we have the URL in the client. Could we instead have `sessionDidDuplicate` return a bool and then in each of the above methods in the feature client call `payPalContextID = extractToken(from: appSwitchURL)`? Happy to sync up on this if it doesn't make sense, kinda hard to describe.

richherrera: What would be the purpose of having `sessionDidDuplicate` return a boolean?

Sources/BraintreePayPal/BTPayPalClient.swift
richherrera: Updated: [da58d59](https://github.com/braintree/braintree_ios/pull/1598/commits/da58d59fe6f082e16cac7f6b2699e0d3314b8319)

Sources/BraintreeCore/BTWebAuthenticationSession.swift
richherrera: Done, what do you think?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -472,6 +479,8 @@ import BraintreeDataCollector
         webAuthenticationSession.prefersEphemeralWebBrowserSession = experiment == "InAppBrowserNoPopup"
 
         webAuthenticationSession.start(url: appSwitchURL, context: self) { [weak self] url, error in
+            self?.payPalContextID = self?.extractToken(from: url)
```
agedd, line 482: nit: take it or leave it - can we do something like `guard let self = self else { return }` before this line

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -535,8 +548,27 @@ import BraintreeDataCollector
...
+                didEnablePayPalAppSwitch: payPalRequest?.enablePayPalAppSwitch,
+                didPayPalServerAttemptAppSwitch: didPayPalServerAttemptAppSwitch,
+                isVaultRequest: isVaultRequest,
+                payPalContextID: payPalContextID,
+                shopperSessionID: payPalRequest?.shopperSessionID
+            )
         }
     }
+    
+    /// extract BA or EC token from the URL to set `payPalContextID` correctly
```
agedd, line 565: nit (optional):
```suggestion
    /// This helper method extracts the BA or EC token from the URL to set `payPalContextID` correctly
```

UnitTests/BraintreeCoreTests/BTWebAuthenticationSession_Tests.swift
```diff
@@ -0,0 +1,33 @@
+import XCTest
+import BraintreeCore
+
+class BTWebAuthenticationSession_Tests: XCTestCase {
+
+    func testStart_whenSessionAlreadyExists_callsSessionDidDuplicate() {
```
agedd, line 6: 🚀

agedd (approved):
this looks great to me! tyy for the helpful pr description 🎉

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## unreleased
 * BraintreePayPal
   * Bug fix: `BTPayPalRequest.userPhoneNumber` could be passed as an empty string resulting in an error
+  * Bug fix: Ensure that ASWebAuthenticationSession.start is only called with one URL at a time by blocking additional start calls until the current one completes.
```
```suggestion
  * Bug fix: Ensure that `ASWebAuthenticationSession.start` is only called with one URL at a time by blocking additional start calls until the current one completes.
```

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: It looks like we have that on line 484 right below this!

UnitTests/BraintreeCoreTests/MockPresentationContext.swift
```diff
@@ -0,0 +1,7 @@
+import AuthenticationServices
+
+class MockPresentationContext: NSObject, ASWebAuthenticationPresentationContextProviding {
+    func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
+        return ASPresentationAnchor()
```
jaxdesmarais: I think we can just use the implicit return here:
```suggestion
        ASPresentationAnchor()
```

richherrera: Updated: [dc4e780](https://github.com/braintree/braintree_ios/pull/1598/commits/dc4e780fa5f50b955561bc99a11adfb020fda333)

CHANGELOG.md
richherrera: Updated: [dc4e780](https://github.com/braintree/braintree_ios/pull/1598/commits/dc4e780fa5f50b955561bc99a11adfb020fda333)

PR #1599, opened 2025-07-03
Prevent Multiple Attempts to App Switch to a URL

jaxdesmarais:
The goal of this PR is to ensure that merchants who do not disable their buttons properly do not potentially run into issues where they are displaying their customer both a success an error message. While best practices indicate that merchants should disable buttons after tap as we do in our demo app this is not always the case. In these cases we want to ensure that the customer does not see a confusing experience where they get both a success and error result displayed. We will not manage state of the entire tokenize method as this is a major breaking change, but managing state of methods leading to poor customer and merchant experiences is within the scope of this PR.

### Summary of changes

- Prevent multiple attempts to open a URL while a current attempt is in progress, this currently results in us returning a success as well as the error `appSwitchFailed` to the merchant and customer potentially
- Add event `tokenizeDuplicateRequest` - this will allow the analytics team to remove the duplicate requests from the reporting funnel
- Add `hasOpenedURL` and reset this property where appropriate - docstring have been added to clarify why and where it's being reset
- Add unit test

### Notes for Reviewers
I have gone through manual testing on this branch and things appear to be working as expected in my tests. That said, this PR introduces the behavior of managing state at a low level in the SDK. Please pull down this branch before reviewing so that you can go through edge cases and ensure you are seeing the expected events in FPTI as well as the expected behavior when killing the PayPal app, using back buttons, canceling, etc. This PR only touches app switch, so no testing needs to be done on the fallback web based flow.

To test these changes you need to update the following methods in `PayPalWebCheckoutViewController` to test the Checkout and Vault flows:
```swift
    @objc func tappedPayPalAppSwitchForCheckout(_ sender: UIButton) {
//        sender.setTitle("Processing...", for: .disabled)
//        sender.isEnabled = false
        
        let request = BTPayPalCheckoutRequest(
            userAuthenticationEmail: emailTextField.text,
            enablePayPalAppSwitch: true,
            amount: "10.00"
        )

        payPalClient.tokenize(request) { nonce, error in
//            sender.isEnabled = true
...

// AND
    @objc func tappedPayPalAppSwitchForVault(_ sender: UIButton) {
//        sender.setTitle("Processing...", for: .disabled)
//        sender.isEnabled = false

        let request = BTPayPalVaultRequest(
            userAuthenticationEmail: emailTextField.text,
            enablePayPalAppSwitch: true
        )

        payPalClient.tokenize(request) { nonce, error in
//            sender.isEnabled = true
...
```
After allowing multiple button clicks you should now be able to tap the app switch buttons multiple times to see this new behavior. This replicates a merchant not disabling their buttons properly.

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -61,6 +61,8 @@
 		45227FC52C330FDE00A15018 /* MockURLSessionTask.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45227FC32C330FDE00A15018 /* MockURLSessionTask.swift */; };
 		45227FC72C33104100A15018 /* MockBTHTTPNetworkTiming.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45227FC62C33104100A15018 /* MockBTHTTPNetworkTiming.swift */; };
 		454722AC2CEFB291000DCF4E /* BTContactInformation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 454722AB2CEFB291000DCF4E /* BTContactInformation.swift */; };
+		454BCF752E1C5FAA0099E674 /* BTWebAuthenticationSession_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 454BCF742E1C5FA50099E674 /* BTWebAuthenticationSession_Tests.swift */; };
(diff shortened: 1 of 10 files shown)


richherrera (approved):

agedd (approved):
this is looking great! the detailed doc strings are really helpful 🚀

PR #1601, opened 2025-07-07
Added Application State to analytics

baubrey91:

### Summary of changes

When sending up analytics we will also record the application state ie background, active, inactive.
- 

### Checklist

- [ ] Added a changelog entry

### FPTI

image


Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -153,16 +156,28 @@ final class BTAnalyticsService: AnalyticsSendable {
...
+            applicationState: applicationState.asString,
             sessionID: sessionID,
             tokenizationKey: apiClient?.authorization.type == .tokenizationKey ? apiClient?.authorization.originalValue : nil
         )
         
         return FPTIBatchData(metadata: batchMetadata, events: events)
     }
 }
+
+extension UIApplication.State {
```
baubrey91: So we have three options for this.
1. Keep as is, enum var
2. Remove enum var and just send raw value (harder to read in analytics with just 0,1,2)
3. Remove enum var and just add an helper function.
Please advise

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -203,6 +203,8 @@ struct FPTIBatchData: Codable {
         let payPalInstalled: Bool = application.isPayPalAppInstalled()
 
         let platform = "iOS"
+        
+        let applicationState: String
```
baubrey91: We could just grab Application.State here instead of passing through each level
Please advise

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -142,9 +144,10 @@ final class BTAnalyticsService: AnalyticsSendable {
     }
     
     /// Posts analytics events to the endpoint.
-    private func postAnalyticsEvents(configuration: BTConfiguration, sessionID: String, events: [FPTIBatchData.Event]) async throws {
+    private func postAnalyticsEvents(configuration: BTConfiguration, applicationState: UIApplication.State, sessionID: String, events: [FPTIBatchData.Event]) async throws {
         let payload = createAnalyticsEvent(
             config: configuration,
+            applicationState: applicationState,
```
agedd: nit: does this also need to be:
```suggestion
            applicationState: applicationState.asString
```

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
agedd: i feel since this value is dynamic and can change at any moment, imo it makes sense to add a helper function that returns the `UIApplication.shared.applicationState` and that way we also don't introduce state management.

agedd: unrelated question: could also add some more context on why we're needing to send this value for every event? i'm curious if it may only be relevant to certain flows - that way we wouldn't need to send it as part of each event we post to fpti

baubrey91: Addressed value of event in slack message, changed to computed var(function) but needed to fetch app state on main thread.

baubrey91:
The two tests failing for me are `testThreeDSecurePaymentFlowV2_challengeFlow_andFails` and `testGetRewardsBalance_returnsResult`. Are theses generally flaky tests? `testGetRewardsBalance_returnsResult` passes for me locally and `testThreeDSecurePaymentFlowV2_challengeFlow_andFails` fails on main.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -26,6 +26,16 @@ final class BTAnalyticsService: AnalyticsSendable {
     private let timer = RepeatingTimer()
 
     private var apiClient: BTAPIClient?
+    
+    @MainActor
```
jaxdesmarais: Does this need to be called on the main thread? Are we concerned about blocking merchants mean thread for our internal analytics?

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
jaxdesmarais: Don't we want to pass this per event? `Metadata` is for properties that won't change. I think we instead want this nested in the event.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
baubrey91: image

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
baubrey91: Your absolutely right, so after further investigation it looks like we need to fetch the app state on the main thread thus we can't just inject it into an analytic event initializer as those are non-isolate. One option is to add the app state at each sendAnalytic call which would require an update to all calls (adds more code, pollutes codebase more)
Another option is to set a global variable and update it with notification center when app state changes.

stechiu:
> The two tests failing for me are `testThreeDSecurePaymentFlowV2_challengeFlow_andFails` and `testGetRewardsBalance_returnsResult`. Are theses generally flaky tests? `testGetRewardsBalance_returnsResult` passes for me locally and `testThreeDSecurePaymentFlowV2_challengeFlow_andFails` fails on main.

Yes, these tests are flaky in CI. Try "re-run failed tests"

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
jaxdesmarais: Does it need to be global? Can't we make an extension on `UIApplication` to have access to it across the SDK? Where we need this we should already have access to `application`.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -26,7 +27,17 @@ import Foundation
     
     /// Exposed for testing analytics
     weak var analyticsService: AnalyticsSendable? = BTAnalyticsService.shared
-
+    
+    // MARK: - Private Properties
+    private var applicationStateString: String {
```
jaxdesmarais:
```suggestion
    // MARK: - Private Properties
    
    private var applicationStateString: String {
```

jaxdesmarais: Though I think this can be an extension on `UIApplication` and live in our extension file

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -203,6 +207,7 @@ struct FPTIBatchData: Codable {
         let payPalInstalled: Bool = application.isPayPalAppInstalled()
 
         let platform = "iOS"
+
```
jaxdesmarais: Don't need this newline:
```suggestion
```

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -71,6 +71,7 @@ struct FPTIBatchData: Codable {
         let startTime: Int?
         let timestamp = String(Date().utcTimestampMilliseconds)
         let tenantName: String = "Braintree"
+        let applicationState: String?
```
jaxdesmarais: This file is in alphabetical order, lets follow that pattern

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -334,7 +345,8 @@ import Foundation
                 merchantExperiment: merchantExperiment,
                 pageType: pageType,
                 payPalContextID: payPalContextID,
-                shopperSessionID: shopperSessionID
+                shopperSessionID: shopperSessionID,
+                applicationState: applicationStateString
```
jaxdesmarais: Should we be sending this anywhere?

Sources/BraintreeCore/BTAPIClient.swift
baubrey91: So I did see the extension but then we would have to call it like so `UIApplication.shared.applicationStateString` wasn't sure that was the best approach but will update.

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
baubrey91: I really appreciate how clean the code in BT is compared to venice

Sources/BraintreeCore/BTAPIClient.swift
baubrey91: Sorry I don't follow. Now that it is part of the event isn't it sent up? Was there something else I need to configure? When I test I see application state in FPTI 
image

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -28,6 +28,8 @@ struct FPTIBatchData: Codable {
 
     /// Encapsulates a single event by it's name and timestamp.
     struct Event: Codable {
+        /// state application is in ie active, inactive, background
```
stechiu: Nit: `The app's state within the device's lifecycle (ie active, inactive, background)`

Sources/BraintreeCore/BTAPIClient.swift
baubrey91: @jaxdesmarais

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -142,7 +142,11 @@ final class BTAnalyticsService: AnalyticsSendable {
     }
     
     /// Posts analytics events to the endpoint.
-    private func postAnalyticsEvents(configuration: BTConfiguration, sessionID: String, events: [FPTIBatchData.Event]) async throws {
+    private func postAnalyticsEvents(
+        configuration: BTConfiguration,
+        sessionID: String,
+        events: [FPTIBatchData.Event]
+    ) async throws {
```
jaxdesmarais: I think we can keep this as is

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
```diff
@@ -153,7 +157,11 @@ final class BTAnalyticsService: AnalyticsSendable {
     }
 
     /// Constructs POST params to be sent to FPTI
-    private func createAnalyticsEvent(config: BTConfiguration, sessionID: String, events: [FPTIBatchData.Event]) -> Codable {
+    private func createAnalyticsEvent(
+        config: BTConfiguration,
+        sessionID: String,
+        events: [FPTIBatchData.Event]
+    ) -> Codable {
```
jaxdesmarais: Same here

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -139,6 +144,7 @@ struct FPTIBatchData: Codable {
             case startTime = "start_time"
             case endTime = "end_time"
             case endpoint = "endpoint"
+            case applicationState = "application_state"
```
jaxdesmarais: Lets have this alphabetical

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -114,6 +118,7 @@ struct FPTIBatchData: Codable {
             self.requestStartTime = requestStartTime
             self.shopperSessionID = shopperSessionID
             self.startTime = startTime
+            self.applicationState = applicationState
```
jaxdesmarais: Lets have this alphabetical

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -320,6 +321,7 @@ import Foundation
     ) {
         analyticsService?.sendAnalyticsEvent(
             FPTIBatchData.Event(
+                applicationState: UIApplication.shared.applicationStateString,
```
jaxdesmarais: I think we only want to send this in the PayPal flow. Do we see an advantage to tracking the state in all the other modules?

Sources/BraintreeCore/Analytics/BackgroundTaskManaging.swift
```diff
@@ -10,4 +10,13 @@ extension UIApplication: BackgroundTaskManaging {
     func beginBackgroundTask(named: String?, expirationHandler handler: (() -> Void)?) -> UIBackgroundTaskIdentifier {
         beginBackgroundTask(withName: named, expirationHandler: handler)
     }
+    
+    var applicationStateString: String {
```
jaxdesmarais: We should move this into a new file. It doesn't appear to related to sending analytics as a background task. We can name the new file in Core `UIApplication+ApplicationState`

jaxdesmarais:
Also missing unit tests

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
baubrey91: No problem but the rest is not alphabetical ie endTime/endpoint

Sources/BraintreeCore/BTAPIClient.swift
baubrey91: Sorry little confused I thought this was the PayPal flow hence the payPalContextID parameter. When does the venmo and pp flow diverge or are you suggesting we pass in the application state through every analytic event needed?

jaxdesmarais: No, `BTAPIClient` is the base level for all feature client to send analytics through. The `BraintreeCore` module is shared by all SDK feature clients. This method in particular is used to send analytics anywhere we call `apiClient.sendAnalytics` in any feature client (venmo, card, amex, etc). We only want to pass this new tag where we call `apiClient.sendAnalytics` in `BTPayPalClient`.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -1,4 +1,5 @@
 import Foundation
+import UIKit
```
jaxdesmarais, line 1: iirc if we import `UIKit` we can remove the `Foundation` import since it should be included.

Sources/BraintreeCore/UIApplication+ApplicationState.swift
```diff
@@ -0,0 +1,19 @@
+import UIKit
+
```
jaxdesmarais, line 2:
```suggestion
```

Sources/BraintreeCore/UIApplication+ApplicationState.swift
```diff
@@ -0,0 +1,19 @@
...
+extension UIApplication.State {
+    var asString: String {
+        switch self {
+        case .active: "active"
+        case .inactive: "inactive"
+        case .background: "background"
+        @unknown default: "unknown"
+        }
+    }
+}
```
jaxdesmarais: I believe this can be simplified to:
```swift
extension UIApplication {
    
    public var applicationStateString: String {
        switch applicationState {
        case .active:
            return "active"
        case .background:
            return "background"
        case .inactive:
            return "inactive"
        @unknown default:
            return "unknown"
        }
    }
}
```

baubrey91: So that's how I had it initially but I did this to make it easier for testing. Would you like me to revert and remove tests?

baubrey91: I reverted an also removed this test file. Not positive how to test without DI. 
```
import XCTest

final class MockApplicationStateProvider: ApplicationStateProviding {
    
    let applicationState: UIApplication.State
    
    init(applicationState: UIApplication.State) {
        self.applicationState = applicationState
    }
}
    
class UIApplicationApplicationStateTests: XCTestCase {
    
    func testApplicationStateString_Active() {
        let mockProvider = MockApplicationStateProvider(applicationState: .active)
        XCTAssertEqual(mockProvider.applicationState.asString, "active")
    }
    
    func testApplicationStateString_Inactive() {
        let mockProvider = MockApplicationStateProvider(applicationState: .inactive)
        XCTAssertEqual(mockProvider.applicationState.asString, "inactive")
    }
    
    func testApplicationStateString_Background() {
        let mockProvider = MockApplicationStateProvider(applicationState: .background)
        XCTAssertEqual(mockProvider.applicationState.asString, "background")
    }
    
    func testApplicationStateString_Unknown() {
        let mockProvider = MockApplicationStateProvider(applicationState: UIApplication.State(rawValue: 999)!)
        XCTAssertEqual(mockProvider.applicationState.asString, "unknown")
    }
}
```

Sources/BraintreeCore/BTAPIClient.swift
baubrey91: Oops I should have caught that

Sources/BraintreeCore/BTAPIClient.swift
baubrey91: Got it ok I think everything is updated properly

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -28,6 +28,9 @@ struct FPTIBatchData: Codable {
 
     /// Encapsulates a single event by it's name and timestamp.
     struct Event: Codable {
+        
+        /// The app's state within the device's lifecycle (ie active, inactive, background)
```
agedd: small nit: 
```suggestion
        /// The app's state within the device's lifecycle (i.e active, inactive, background)
```

baubrey91:
> nit: can we also add a unit test in `BTPayPalClient_Tests` to make sure that the `applicationState` in the analytics event is being sent

I added a test in BTPayPalClient_Tests but am not positive if its the right naming convention or if you wanted me to just add an assert in another test. May need help. Thanks!

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -213,6 +213,7 @@ import BraintreeDataCollector
 
         apiClient.sendAnalyticsEvent(
             BTPayPalAnalytics.handleReturnStarted,
+            applicationState: UIApplication.shared.applicationStateString,
```
agedd, line 216: small nit: think we can use `application` in place of `UIApplication.shared` like below and replace it in all other places below too

```suggestion
            applicationState: application.applicationStateString,
```

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -1247,4 +1247,12 @@ class BTPayPalClient_Tests: XCTestCase {
 
         XCTAssertEqual(mockAPIClient.postedShopperSessionID, "fake-shopper-session-id")
     }
+    
+    func testTokenize_applicationStateAnalytics() async {
```
agedd: small nit: 
```suggestion
    func testTokenize_whenSuccess_ sendsapplicationStateInAnalytics() async {
```

Sources/BraintreePayPal/BTPayPalClient.swift
baubrey91: Discussed offline: Would require URLOpener to have applicationStateString in protocol causes logic in different files.

baubrey91:
> Lets add a changelog entry for this. There should be several similar entries when adding new tags to reference.
> 
> Also to confirm we have tested that doing something like backgrounding the app while sending an event fetches the correct state and does not cause a crash?

@jaxdesmarais  I updated the changelog, wasn't sure if we wanted to do that as this is an internal feature for FPTI. For testing I set a repeated task to continue send up an analytic and checked to see if it was sent while background. Are you asking to test if the state changes while we are sending? Do you have an idea how to test that? Would this require setting up an async task in the app delegate. Thanks!

```
        Task {
            repeat {
                apiClient.sendAnalyticsEvent(
                    BTPayPalAnalytics.tokenizeStarted,
                    applicationState: UIApplication.shared.applicationStateString,
                    contextType: contextType,
                    didEnablePayPalAppSwitch: payPalRequest?.enablePayPalAppSwitch,
                    isVaultRequest: isVaultRequest,
                    shopperSessionID: payPalRequest?.shopperSessionID
                )
                try? await Task.sleep(nanoseconds: 5_000_000_000) // Sleep for 5 second
            } while !Task.isCancelled //
        }
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
   * Analytics updates for PayPal's analytics service (FPTI)
     * Add `space_key` and `product_name` to `batch_params`
     * Add `context_type` to `event_params`
+    * Add `application_state` to PayPal analytics for tracking UIApplication state
```
jaxdesmarais: This needs to be moved under `## unreleased`

UnitTests/BraintreeCoreTests/Analytics/FPTIBatchData_Tests.swift
```diff
@@ -17,6 +17,7 @@ final class FPTIBatchData_Tests: XCTestCase {
 
     let eventParams = [
         FPTIBatchData.Event(
+            applicationState: UIApplication.shared.applicationStateString,
```
jaxdesmarais: I think we can hardcode this value for the test

baubrey91:
> > I updated the changelog, wasn't sure if we wanted to do that as this is an internal feature for FPTI.
> 
> It's generally helpful for us to know what version a FPTI change was made in, so still worth having one.
> 
> > For testing I set a repeated task to continue send up an analytic and checked to see if it was sent while background. Are you asking to test if the state changes while we are sending? Do you have an idea how to test that? Would this require setting up an async task in the app delegate. Thanks!
> 
> Sorry should have been more clear. I more meant backgrounding the app and ensuring the application state is correct since it's sent on main. So something like:
> 
> 1. Open app
> 2. Click on paypal and immediately background the app
> 3. Ensure the application state is correct (should be background) and does not crash the app when attempting to fetch the state on main as the app is backgrounded

So I did test this previously, when vaulting the app is backgrounded regardless, but now I have also tested with web checkout. Attached is a video and FTPI however out of curiosity do you guys use feature switches. We could wrap this behind one and ramp it up incase there is an issue.

https://github.com/user-attachments/assets/34e3fc4c-2497-40b6-a88c-4538ad28332f

image

CHANGELOG.md
baubrey91: Aw man I did that initially, must have broke it when I rebased

baubrey91: Aw man I did that initially, must have broke it when I rebased

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## Unreleased
```
jaxdesmarais: The CI job relies on lowercase, should be able to just accept this suggestion:
```suggestion
## unreleased
```

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPal
+  * Add `application_state` to PayPal analytics for tracking UIApplication state
```
agedd: small nit: 
```suggestion
  * Send `application_state` to FPTI for tracking UIApplication state
```

PR #1602, opened 2025-07-08
Dependencies added to Visa Checkout 

stechiu:
`VisaCheckoutSDK` framework was added to Visa Checkout

### Summary of changes

- Added `VisaCheckoutSDK` framework

### Checklist

- @stechiu


Braintree.podspec
```diff
@@ -110,6 +110,8 @@ Pod::Spec.new do |s|
   s.subspec "VisaCheckout" do |s|
     s.source_files = "Sources/BraintreeVisaCheckout/*.swift"
     s.dependency "Braintree/Core"
+    s.dependency "VisaCheckoutSDK", "6.6.0"
+    s.vendored_frameworks = 'Pods/VisaCheckoutSDK/DigitalTerminal-iOS-SDK-7.3.1/VisaCheckoutSDK.xcframework'
```
jaxdesmarais, line 114: We shouldn't need the vendored frameworks for cocoapods. Should be similar to `PayPalMessaging` above with just the dependency.

Podfile
```diff
@@ -8,12 +8,14 @@ inhibit_all_warnings!
 target 'Demo' do
   project 'Demo/Demo'
   pod 'InAppSettingsKit'
+  pod 'VisaCheckoutSDK'
```
jaxdesmarais, line 11: We shouldn't need this in our Podfile or Podfile.lock

Package.swift
```diff
@@ -154,8 +154,12 @@ let package = Package(
         ),
         .target(
             name: "BraintreeVisaCheckout",
-            dependencies: ["BraintreeCore"],
+            dependencies: ["BraintreeCore", "VisaCheckoutSDK"],
             resources: [.copy("PrivacyInfo.xcprivacy")]
         ),
+        .binaryTarget(
+            name: "VisaCheckoutSDK",
+            path: "Frameworks/XCFrameworks/VisaCheckout/DigitalTerminal-iOS-SDK-7.3.1/VisaCheckoutSDK.xcframework"
```
jaxdesmarais, line 162: Is this the right framework? This looks different than what is here: https://github.com/braintree/braintree-ios-visa-checkout/tree/main/Frameworks/VisaCheckoutSDK.xcframework

Braintree.podspec
stechiu: I keep trying new things, adding `s.dependency` only or with `s.vendored_frameworks`. Nothing is able to get rid of the build errors

jaxdesmarais: It looks like v7.3.1 is the latest: https://github.com/CocoaPods/Specs/blob/master/Specs/7/7/b/VisaCheckoutSDK/7.3.1/VisaCheckoutSDK.podspec.json

If we use `vendored_frameworks` we would want to use something like `Frameworks/VisaCheckoutSDK.xcframework` vs pointing to our local pod.

PR #1607, opened 2025-07-10
Payment ready v2 generate customer recommendations

warmkesselj:
### Summary of changes

Add generateCustomerRecommendation() 

### Checklist

- [X] Tested and confirmed payment flows affected by this change are functioning as expected
- [X] Update Changelog
### Authors
@jwarmkessel 
@warmkesselj

### DIFF
--- a/Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
+++ b/Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
@@ -64,9 +64,7 @@ public class BTShopperInsightsClientV2 {
     public func generateCustomerRecommendations(
         request: BTCustomerSessionRequest?,
         sessionID: String?
-    ) async throws -> BTCustomerRecommendationsResult? {
-        guard let request = request, let sessionID = sessionID else { return nil }
-
+    ) async throws -> BTCustomerRecommendationsResult {
         let customerRecommendationsAPI = BTCustomerRecommendationsAPI(apiClient: apiClient)
         return try await customerRecommendationsAPI.execute(request, sessionID: sessionID)
     }
--- a/Sources/BraintreeShopperInsights/BTCustomerRecommendationsAPI.swift
+++ b/Sources/BraintreeShopperInsights/BTCustomerRecommendationsAPI.swift
@@ -21,13 +21,13 @@ final class BTCustomerRecommendationsAPI {
     
     /// This method will call the `GenerateCustomerRecommendations` GQL query, which returns a `BTCustomerRecommendationsResult` if successful.
     /// - Parameters:
-    ///    - request: A `BTCustomerSessionRequest`
-    ///    - sessionID: The session ID to update.
+    ///    - request: Optional. A `BTCustomerSessionRequest`
+    ///    - sessionID: Optional. The session ID to update.
     ///    - Returns: A `BTCustomerRecommendationsResult` which determines what payment options to render.
     ///    - Throws: An error if the request fails or if the response is invalid.
     func execute(
-        _ request: BTCustomerSessionRequest,
-        sessionID: String
+        _ request: BTCustomerSessionRequest?,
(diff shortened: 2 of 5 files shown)


Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
```diff
@@ -101,4 +101,23 @@ public class BTShopperInsightsClientV2 {
     public func isVenmoAppInstalled() -> Bool {
         application.isVenmoAppInstalled()
     }
+    
+    /// Generates customer recommendations
+    /// - Warning: This method is currently in beta and may change or be removed in future releases.
```
jaxdesmarais: Lets add the parameters to the docstring

Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
```diff
@@ -101,4 +101,23 @@ public class BTShopperInsightsClientV2 {
...
+    public func generateCustomerRecommendations(
```
jaxdesmarais: Can we move this up with the other create/update methods in the file? Also I think these should be using async/await vs completion.

Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
```diff
@@ -56,6 +56,19 @@ public class BTShopperInsightsClientV2 {
         return try await updateCustomerSessionAPI.execute(request, sessionID: sessionID)
     }
     
+    /// Generates customer recommendations
+    /// - Parameters:
+    ///    - request: request type `BTCustomerSessionRequest`
+    ///    - sessionID: The shopper session ID
+    /// - Warning: This method is currently in beta and may change or be removed in future releases.
+    public func generateCustomerRecommendations(
```
jaxdesmarais, line 64: Lets add a changelog entry for this as well

Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
```diff
@@ -56,6 +56,19 @@ public class BTShopperInsightsClientV2 {
...
+        request: BTCustomerSessionRequest,
+        sessionID: String
```
jaxdesmarais: It looks like both of these should be optional according to the HLD and Android PR: https://github.com/braintree/braintree_android/pull/1343/files#diff-c35d2fc4471a0a14a4c65a01d06b4e50890a01b01eb3ffa732134fac9a990add

Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
```diff
@@ -56,6 +56,19 @@ public class BTShopperInsightsClientV2 {
...
+    ) async throws -> BTCustomerRecommendationsResult {
+        let api = BTCustomerRecommendationsAPI(apiClient: apiClient)
```
jaxdesmarais: To match the rest of this module:
```suggestion
        let customerRecommendationsAPI = BTCustomerRecommendationsAPI(apiClient: apiClient)
```

Sources/BraintreeShopperInsights/BTShopperInsightsClientV2.swift
warmkesselj: Yeah, I noticed in the HLD these are optional properties. `BTCustomerRecommendationsAPI` and `GenerateCustomerRecommendationsGraphQLBody` both take these properties and they are no optional. Does the guard make sense here?

warmkesselj: I just updated `BTCustomerRecommendationsAPI`, `GenerateCustomerRecommendationsGraphQLBody` including `Variables` which is the encodable part of the query and seems like I'm still getting a successful response with a nil request and nil sessionId.

PR #1608, opened 2025-07-10
Added new VisaCheckoutSDK framework

stechiu:
`VisaCheckoutSDK` framework was added to Visa Checkout

### Summary of changes

- Added `VisaCheckoutSDK` framework

### Checklist

- @stechiu


Braintree.podspec
```diff
@@ -110,6 +110,7 @@ Pod::Spec.new do |s|
   s.subspec "VisaCheckout" do |s|
     s.source_files = "Sources/BraintreeVisaCheckout/*.swift"
     s.dependency "Braintree/Core"
+    s.vendored_frameworks = "Frameworks/XCFrameworks/VisaCheckoutSDK.xcframework"
```
jaxdesmarais: I think we can use the `dependency` flag pointing to the correct version vs cocoapods like we looked at. Since we were able to find the framework that correlated with the version we should be able to use that same version here: https://github.com/CocoaPods/Specs/blob/bd9e6c13b5b174f4a3faf2d29e0d69fda3715882/Specs/7/7/b/VisaCheckoutSDK/7.3.1/VisaCheckoutSDK.podspec.json

Cartfile.resolved
```diff
@@ -0,0 +1,4 @@
+binary "https://assets.braintreegateway.com/mobile/ios/carthage-frameworks/cardinal-mobile/CardinalMobile.json" "2.2.5-9"
+binary "https://assets.braintreegateway.com/mobile/ios/carthage-frameworks/pp-risk-magnes/PPRiskMagnes.json" "5.5.0-static-Xcode15-MinOSVersion100"
+binary "https://github.com/paypal/paypal-messages-ios/raw/main/Carthage/PayPalMessages.json" "1.0.0"
+binary "https://github.com/paypal/paypalcheckout-ios/raw/main/Carthage/PayPalCheckout.json" "1.3.0"
```
jaxdesmarais: We should revert this file to what's on main

stechiu: Done

Braintree.podspec
stechiu: Good point, thanks

PR #1609, opened 2025-07-11
Fix bug where `browserLoginAlertCanceled` is fired in the None-WAS popup flow

agedd:
### Summary of changes

- Context: The `browserLoginAlertCanceled` was being fired even though the experiment=InAppBrowserNoPopup was being passed in the approval url, which would essentially bypass the WAS popup.
- This fix ensures that we add a check for the experiment and accordingly set the `webSessionReturned` and `prefersEphemeralWebBrowserSession` flag. Abstracted this logic into a separate helper function to check if the experiment is enabled and set these properties accordingly (and to allow us to edit in the future in case we enable more experiments)

### Checklist

- [ ] Added a changelog entry

- @agedd

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -510,8 +510,8 @@ import BraintreeDataCollector
         approvalURL = appSwitchURL
         webSessionReturned = false
         
-        webAuthenticationSession.prefersEphemeralWebBrowserSession = experiment == "InAppBrowserNoPopup"
-
+        configureSessionIfNeeded(for: experiment)
+        
         webAuthenticationSession.start(url: appSwitchURL, context: self) { [weak self] url, error in
             self?.payPalContextID = self?.extractToken(from: url)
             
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
   * Bug fix: Pass `unitTaxAmount` as expected in `BTPayPalLineItem`
   * Make `BTPayPalLineItem` parameters settable
   * Bug fix: only attempt to call `UIApplication.shared.open` on one URL at a time blocking other `open` calls until the current one is finished
+  * Bug fix: Ensure that the `browserLoginAlertCanceled` event is not fired off when the WAS popup is disabled.
```
jaxdesmarais: Since this is an external changlog using the proper name makes more sense:
```suggestion
  * Bug fix: Ensure that the `browserLoginAlertCanceled` event is not fired off when the `ASWebAutheniticationSession` popup is disabled.
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -604,6 +604,13 @@ import BraintreeDataCollector
         let ecToken = BTURLUtils.queryParameters(for: url)["token"]
         return baToken ?? ecToken
     }
+    
+    private func configureSessionIfNeeded(for experiment: String? = nil) {
+        if experiment == "InAppBrowserNoPopup" {
+            webSessionReturned = true
```
jaxdesmarais, line 610: Are we able to add a test to check that the old event isn't sent? Assuming you've tested manually as well.

agedd: yep updated an already existing unit test in d66df24168cd5f6a9ea592593e424555f13945bc and tested manually as well to make sure it's working as expected - added another test to check for the regular user cancel flow scenario as well

warmkesselj (approved):

jaxdesmarais (approved):
Love the new test so we can prevent this from slipping through again. 🚀

PR #1611, opened 2025-07-15
[v7] Revert Open Override

jaxdesmarais:
### Summary of changes

- Revert this override change needed when supporting older Xcode versions: https://github.com/braintree/braintree_ios/pull/1360#discussion_r1672339486

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Sources/BraintreeCore/UIApplication+URLOpener.swift
+++ b/Sources/BraintreeCore/UIApplication+URLOpener.swift
@@ -7,9 +7,13 @@ import UIKit
 public protocol URLOpener {
 
     func canOpenURL(_ url: URL) -> Bool
-    func open(_ url: URL, completionHandler completion: ((Bool) -> Void)?)
     func isPayPalAppInstalled() -> Bool
     func isVenmoAppInstalled() -> Bool
+    func open(
+        _ url: URL,
+        options: [UIApplication.OpenExternalURLOptionsKey: Any],
+        completionHandler completion: (@MainActor @Sendable (Bool) -> Void)?
+    )
 }
 
 extension UIApplication: URLOpener {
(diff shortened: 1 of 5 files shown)


UnitTests/BraintreeVenmoTests/BTVenmoClient_Tests.swift
```diff
@@ -339,7 +339,9 @@ class BTVenmoClient_Tests: XCTestCase {
             ]
         ])
 
-        BTVenmoClient.handleReturnURL(URL(string: "scheme://x-callback-url/vzero/auth/venmo/success?resource_id=12345")!)
+        DispatchQueue.main.async {
```
jaxdesmarais, line 342: This is needed for our tests because with unit tests there is a race condition with the open mock being called on `@MainActor`. There were some workarounds I found with adding an expectation to the completion, fulfilling that here in a new completion handler, then calling `handleReturnURL` within that. I think instead calling `handleReturnURL` on the main queue is much cleaner in the long run.

stechiu (commented):
Why do we need to revert the open override changes?

Sources/BraintreeCore/UIApplication+URLOpener.swift
```diff
@@ -7,9 +7,13 @@ import UIKit
...
 
     func canOpenURL(_ url: URL) -> Bool
-    func open(_ url: URL, completionHandler completion: ((Bool) -> Void)?)
     func isPayPalAppInstalled() -> Bool
     func isVenmoAppInstalled() -> Bool
+    func open(
+        _ url: URL,
+        options: [UIApplication.OpenExternalURLOptionsKey: Any],
+        completionHandler completion: (@MainActor @Sendable (Bool) -> Void)?
+    )
```
richherrera, line 16: Is there any particular reason why this method needs to match the one in `UIApplication`? I think this method could be simpler, once it conforms, it would look something like this.

``` swift
    public func open(_ url: URL, completionHandler completion: ((Bool) -> Void)?) {
        open(url, options: [:]) { success in
            completion?(success)
        }
    }
```
I think it would be similar to what was done in `BackgroundTaskManaging`.

jaxdesmarais: I think generally we've opted to match the signature vs providing an override of the method where possible. Mainly to make it more clear that we are using the apple method and the protocol only exists for mocking. But if folks feel that having the empty options is more confusing I am open to providing our own signature! What do you think?

richherrera: That makes sense, and I agree that matching the Apple method signature helps make the intent clearer

agedd (approved):
🚀

PR #1621, opened 2025-07-30
Bug Fix: Print Error Message Rather Than Memory Address

agedd:
### Summary of changes

- Fix a minor bug where we are printing the error's object reference rather than the actual error message for the `BTPayPalError.httpPostRequestError` error type.

### Checklist

- @agedd


stechiu (approved):

jaxdesmarais (approved):
🎉 thanks for fixing this!

PR #1631, opened 2025-08-06
Add BTConfiguration+VisaCheckout_Tests

agedd:
### Summary of changes

- Add `BTConfiguration_VisaCheckoutTests` unit tests

### Checklist

- ~[] Added a changelog entry~

- @agedd

### DIFF
--- a/Sources/BraintreeVisaCheckout/BTConfiguration+VisaCheckout.swift
+++ b/Sources/BraintreeVisaCheckout/BTConfiguration+VisaCheckout.swift
@@ -14,7 +14,10 @@ extension BTConfiguration {
 
     /// Determines if the Visa Checkout flow is available to be used. This can be used to determine if UI components should be shown or hidden.
     var isVisaCheckoutEnabled: Bool {
-        json?["visaCheckout"]["apiKey"].isTrue ?? false
+        guard let visaCheckout = json?["visaCheckout"], visaCheckout.isObject else {
+            return false
+        }
+        return (visaCheckoutAPIKey?.isEmpty == false)
     }
 
     /// The Visa Checkout API Key associated with this merchant's Visa Checkout configuration.
--- a/UnitTests/BraintreeVisaCheckoutTests/BTConfiguration+VisaCheckout_Tests.swift
+++ b/UnitTests/BraintreeVisaCheckoutTests/BTConfiguration+VisaCheckout_Tests.swift
@@ -1,0 +1,112 @@
+import XCTest
+import VisaCheckoutSDK
+import BraintreeTestShared
+@testable import BraintreeCore
+@testable import BraintreeVisaCheckout
+
+class BTConfiguration_VisaCheckoutTests: XCTestCase {
+    
+    func testIsVisaCheckoutEnabled_whenVisaCheckoutEmptyJSONExists_returnsFalse() {
+        let configurationJSON = BTJSON(value: [
+            "visaCheckout": [:]
+        ])
+        let configuration = BTConfiguration(json: configurationJSON)
+        
+        XCTAssertFalse(configuration.isVisaCheckoutEnabled)
+    }
+    
+    func testIsVisaCheckoutEnabled_whenVisaCheckoutFromConfigurationJSONExists_returnsTrue() {
+        let configurationJSON = BTJSON(value: [
+            "visaCheckout": [
+                "apiKey": "apikey" // key must be 'apiKey' not 'apikey' to match Swift logic
+            ]
+        ])
+        let configuration = BTConfiguration(json: configurationJSON)
+        
+        XCTAssertTrue(configuration.isVisaCheckoutEnabled)
+    }
+    
+    func testIsVisaCheckoutEnabled_whenVisaCheckoutFromConfigurationJSONIsMissing() {
+        let configurationJSON = BTJSON(value: [:])
+        let configuration = BTConfiguration(json: configurationJSON)
+        
+        XCTAssertFalse(configuration.isVisaCheckoutEnabled)
+    }
+    
+    func testVisaCheckoutSupportedNetworks_returnsSupportedNetworks() {
+        let configurationJSON = BTJSON(value: [
+            "visaCheckout": [
+                "supportedCardTypes": [
+                    "Visa",
+                    "MasterCard",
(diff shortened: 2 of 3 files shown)


Sources/BraintreeVisaCheckout/BTConfiguration+VisaCheckout.swift
```diff
@@ -14,7 +14,10 @@ extension BTConfiguration {
 
     /// Determines if the Visa Checkout flow is available to be used. This can be used to determine if UI components should be shown or hidden.
     var isVisaCheckoutEnabled: Bool {
-        json?["visaCheckout"]["apiKey"].isTrue ?? false
+        guard let visaCheckout = json?["visaCheckout"], visaCheckout.isObject else {
+            return false
+        }
+        return (visaCheckoutAPIKey?.isEmpty == false)
```
agedd: **note:** edited this logic to add check for if `visaCheckout isObject` && if `visaCheckoutAPIKey` is not nil to match original obj-c logic [here](https://github.com/braintree/braintree-ios-visa-checkout/blob/81afb7f3c8c08031125c3d707e17260db8a26902/BraintreeVisaCheckout/BTConfiguration%2BVisaCheckout.m#L7C1-L9C2)

jaxdesmarais: 🤔 I am not sure we need the `isObject` check unless you were running into test failures? We use the linked objects with `isTrue` in other places of the SDK. I believe that would default to false if for some reason it wasn't an object.

UnitTests/BraintreeVisaCheckoutTests/BTConfiguration+VisaCheckout_Tests.swift
```diff
@@ -0,0 +1,112 @@
...
+        ])
+        let configuration = BTConfiguration(json: configurationJSON)
+        
+        XCTAssertFalse(configuration.isVisaCheckoutEnabled)
+    }
+    
+    func testIsVisaCheckoutEnabled_whenVisaCheckoutFromConfigurationJSONExists_returnsTrue() {
+        let configurationJSON = BTJSON(value: [
+            "visaCheckout": [
+                "apiKey": "apikey" // key must be 'apiKey' not 'apikey' to match Swift logic
```
jaxdesmarais: I'm not sure I follow this comment. Can you provide more details?

UnitTests/BraintreeVisaCheckoutTests/BTConfiguration+VisaCheckout_Tests.swift
```diff
@@ -0,0 +1,112 @@
...
+                    "American Express",
+                    "Discover",
+                    "MysteryCard",
+                    "Braintree Express"
+                ]
+            ]
+        ])
+        let configuration = BTConfiguration(json: configurationJSON)
+        
+        XCTAssertEqual(configuration.acceptedCardBrands, [CardBrand.visa.rawValue, CardBrand.mastercard.rawValue, CardBrand.amex.rawValue, CardBrand.discover.rawValue])
```
jaxdesmarais, line 67: Should we be checking this against the literal strings instead? Something like this with the actual values:
```suggestion
        XCTAssertEqual(configuration.acceptedCardBrands, ["visa" , "mastercard", "amex", "discover"])
```

Sources/BraintreeVisaCheckout/BTConfiguration+VisaCheckout.swift
agedd: ahh yep agreed, removed the `isObject` check (it wasn't affecting any unit tests either)

UnitTests/BraintreeVisaCheckoutTests/BTConfiguration+VisaCheckout_Tests.swift
agedd: ahh this comment was copy/pasted from the original obj-c code, removed it as it is no longer needed

UnitTests/BraintreeVisaCheckoutTests/BTConfiguration+VisaCheckout_Tests.swift
agedd: hmm i think it's okay since it looks like `acceptedCardBrands` is of type`[Int]` and `supportedCardTypesToAcceptedCardBrands` takes in the string array of these card types and converts them into the accepted card brands [here](https://github.com/braintree/braintree_ios/blob/76adaf3db0873df2dda53f86837ef1fa976748e4/Sources/BraintreeVisaCheckout/BTConfiguration%2BVisaCheckout.swift#L42)

richherrera (approved):

UnitTests/BraintreeVisaCheckoutTests/BTConfiguration+VisaCheckout_Tests.swift
jaxdesmarais: Ah, gotcha. Sorry didn't realize the raw values were integers and just assumed they were strings. That makes sense then.

UnitTests/BraintreeVisaCheckoutTests/BTConfiguration+VisaCheckout_Tests.swift
```diff
@@ -18,7 +18,7 @@ class BTConfiguration_VisaCheckoutTests: XCTestCase {
     func testIsVisaCheckoutEnabled_whenVisaCheckoutFromConfigurationJSONExists_returnsTrue() {
         let configurationJSON = BTJSON(value: [
             "visaCheckout": [
-                "apiKey": "apikey" // key must be 'apiKey' not 'apikey' to match Swift logic
+                "apiKey": "apikey"
```
warmkesselj: I think the expected resulting _key_ may need to be `apikey` with a lower case K. cc: @stechiu

agedd: ahh gotcha ty - updated in 7b6c010aa47ce09ac70eddee592babdf736dd09e

warmkesselj (approved):

PR #1633, opened 2025-08-07
[v7] LocalPayment Encodable

richherrera:
### Summary of changes

Several minor improvements and fixes related to local payment handling and documentation.

**Local payment POST body changes:**

* Changed the encoding of the `options` property in `LocalPaymentPayPalAccountsPOSTBody` from an array to a single `Option` object for better alignment with expected API format.
* Added the `response` property to the list of encoded keys in the `CodingKeys` enum for `LocalPaymentPayPalAccountsPOSTBody`.

**Demo application change:**

* Updated the authorization value in `IdealViewController.swift` to use a hardcoded sandbox token instead of a variable, likely for testing purposes.

| Main branch | 🐛 | 🧼 |
| ------- | ------- | ------- |
| Screenshot 2025-08-07 at 3 58 17 p m | Screenshot 2025-08-07 at 3 59 51 p m | Screenshot 2025-08-07 at 3 59 01 p m |

### Checklist

- [ ] Added a changelog entry

### Authors

- @richherrera

### DIFF
--- a/Demo/Application/Features/IdealViewController.swift
+++ b/Demo/Application/Features/IdealViewController.swift
@@ -47,7 +47,7 @@ class IdealViewController: PaymentButtonBaseViewController {
     }
 
     private func startPaymentWithBank() {
-        localPaymentClient = BTLocalPaymentClient(authorization: authorization)
+        localPaymentClient = BTLocalPaymentClient(authorization: "sandbox_f252zhq7_hh4cpc39zq4rgjcg")
 
         let postalAddress = BTPostalAddress()
         postalAddress.countryCodeAlpha2 = "NL"
--- a/Sources/BraintreeLocalPayment/Models/LocalPaymentAccountsPOSTBody.swift
+++ b/Sources/BraintreeLocalPayment/Models/LocalPaymentAccountsPOSTBody.swift
@@ -4,7 +4,7 @@ import Foundation
 import BraintreeCore
 #endif
 
-/// The POST body for v1/payment_methods/paypal_accounts
+/// The POST body for `v1/payment_methods/paypal_accounts`
 struct LocalPaymentPayPalAccountsPOSTBody: Encodable {
     
     // MARK: - Private Properties
@@ -41,7 +41,7 @@ extension LocalPaymentPayPalAccountsPOSTBody {
         let intent = "sale"
         let response: Response
         let responseType = "web"
-        let options = [Option()]
+        let options = Option()
         
         var correlationID: String?
         
@@ -70,6 +70,7 @@ extension LocalPaymentPayPalAccountsPOSTBody {
             case correlationID = "correlation_id"
             case intent
             case options
+            case response
             case responseType = "response_type"
         }
     }
--- a/Sources/BraintreeLocalPayment/Models/LocalPaymentPOSTBody.swift
+++ b/Sources/BraintreeLocalPayment/Models/LocalPaymentPOSTBody.swift
@@ -4,7 +4,7 @@ import Foundation
 import BraintreeCore
 #endif
 
-/// The POST body for v1/local_payments/create
+/// The POST body for `v1/local_payments/create`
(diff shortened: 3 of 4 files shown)


Sources/BraintreeLocalPayment/Models/LocalPaymentAccountsPOSTBody.swift
```diff
@@ -41,7 +41,7 @@ extension LocalPaymentPayPalAccountsPOSTBody {
         let intent = "sale"
         let response: Response
         let responseType = "web"
-        let options = [Option()]
+        let options = Option()
```
warmkesselj: Since it's just an option should we update to `option`?

Demo/Application/Features/IdealViewController.swift
```diff
@@ -47,7 +47,7 @@ class IdealViewController: PaymentButtonBaseViewController {
     }
 
     private func startPaymentWithBank() {
-        localPaymentClient = BTLocalPaymentClient(authorization: authorization)
+        localPaymentClient = BTLocalPaymentClient(authorization: "sandbox_f252zhq7_hh4cpc39zq4rgjcg")
```
warmkesselj, line 50: I'm guessing this is intentional. Should any of the below be updated to reflect these changes in the `ContainmentViewController`? 

```
    private func instantiateViewController(with authorization: String) -> BaseViewController? {
        guard let integrationName = UserDefaults.standard.string(forKey: "BraintreeDemoSettingsIntegration") else {
            return PayPalWebCheckoutViewController(authorization: authorization)
        }

        switch integrationName {
        case "AmexViewController":
            return AmexViewController(authorization: authorization)
        case "ApplePayViewController":
            return ApplePayViewController(authorization: authorization)
        case "CardTokenizationViewController":
            return CardTokenizationViewController(authorization: authorization)
        case "DataCollectorViewController":
            return DataCollectorViewController(authorization: authorization)
        **case "IdealViewController":
            return IdealViewController(authorization: authorization)**
```

richherrera: yes, this is intentional, this is the only module that should use a different authorization in our demo app, im basically using the same key thats used in the main branch https://github.com/braintree/braintree_ios/blob/d0e3397573b5ab0214e847aaacd7fc1c7b2b7e4c/Demo/Application/Features/IdealViewController.swift#L51

Sources/BraintreeLocalPayment/Models/LocalPaymentAccountsPOSTBody.swift
richherrera: 🤔 `Option` is an object, it can have multiple properties, though right now it only has one, what do you think about renaming the object from `Option` to `Options`?

warmkesselj: That works

richherrera: Thanks, updated: [dc785a6](https://github.com/braintree/braintree_ios/pull/1633/commits/dc785a6f5eed5bc1a5c956242ff69150a7733bb1)

Demo/Application/Features/IdealViewController.swift
stechiu: What's the intention behind hardcoding `authorization`?

agedd (approved):

Demo/Application/Features/IdealViewController.swift
richherrera: that needs to be used to enable local payments, we even have the same authorization on Android.

https://github.com/braintree/braintree_android/blob/c53faec7f918c4d391587719d0917615036f39b1/Demo/src/main/java/com/braintreepayments/demo/Settings.kt#L22

If we use a different authorization key, the flow cannot be completed.
I will create a ticket to investigate why this flow only works with this authorization key.

PR #1636, opened 2025-08-13
Added Visa Checkout to Demo App

stechiu:
Added Visa Checkout SDK to Demo App.

Screenshot 2025-08-13 at 11 52 53 AM

### Summary of changes

- Added Visa Checkout button to Demo App

### Checklist

- [ ] Added a changelog entry

- @stechiu


Demo/Application/Base/PaymentButtonBaseViewController.swift
```diff
@@ -11,7 +11,7 @@ class PaymentButtonBaseViewController: BaseViewController {
 
     override init(authorization: String) {
         // swiftlint:disable:next force_unwrapping
-        apiClient = BTAPIClient(authorization: authorization)!
+        apiClient = BTAPIClient(authorization: "sandbox_9dbg82cq_dcpspy2brwdjr3qn")!
```
jaxdesmarais: Is this change intentional? This will change every view controller to use this tokenization key.

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+import VisaCheckoutSDK
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+class VisaCheckoutViewController: PaymentButtonBaseViewController {
+
+    // swiftlint:disable:next implicitly_unwrapped_optional
+    var visaCheckoutClient: BTVisaCheckoutClient!
```
jaxdesmarais: I think we can make this a lazy var instead to avoid the `implicitly_unwrapped_optional`

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+    var launchHandler: LaunchHandle?
+    let visaCheckoutButton = VisaCheckoutButton()
+
+    // MARK: - Public Methods
```
jaxdesmarais: This does not appear to be correct

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+            visaCheckoutButton.centerYAnchor.constraint(equalTo: view.centerYAnchor),
+            visaCheckoutButton.widthAnchor.constraint(equalToConstant: 215),
+            visaCheckoutButton.heightAnchor.constraint(equalToConstant: 45)
+        ])
+    }
+
+    /// Creates a Visa profile and initiates the checkout process.
+    func createVisaProfileAndCheckout() {
+        visaCheckoutClient.createProfile { profile, error in
+            if let error = error {
```
jaxdesmarais: In this file all of these can be updated to the following anywhere the property name is being defined as the same name:
```suggestion
            if let error {
```

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+                print("Failed to create Visa profile: \(error)")
+                return
+            }
+            guard let profile = profile else { return }
```
jaxdesmarais: We likely want to return an error if a profile isn't created in our demo. What do you think?
```suggestion
            }
            
            guard let profile = profile else { return }
```

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+            self.visaCheckoutButton.onCheckout(
+                profile: profile,
+                purchaseInfo: purchaseInfo,
+                presenting: self,
+                onReady: { launchHandle in
+                    self.launchHandler = launchHandle
+                },
+                onButtonTapped: {
+                    guard let launchHandler = self.launchHandler else {
+                        return
```
jaxdesmarais, line 66: Do we want to return an error if we don't get a launch handler here?

Podfile.lock
```diff
@@ -37,4 +37,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 224cc2d6666b79d3d6adb45cccd6dfbc6fe74d18
 
-COCOAPODS: 1.16.2
+COCOAPODS: 1.14.3
```
jaxdesmarais: Lets revert this

Sources/BraintreeVisaCheckout/BTConfiguration+VisaCheckout.swift
```diff
@@ -14,12 +14,18 @@ extension BTConfiguration {
...
     /// Determines if the Visa Checkout flow is available to be used. This can be used to determine if UI components should be shown or hidden.
     var isVisaCheckoutEnabled: Bool {
-        json?["visaCheckout"]["apiKey"].isTrue ?? false
+        let visaCheckout = json?["visaCheckout"]
+        let apiKey = visaCheckout?["apikey"].asString()
+
+        if visaCheckout == nil && ((apiKey?.isEmpty) == nil) {
+            return false
+        }
+        return true
```
jaxdesmarais: I think this can be simplified to:
```suggestion
        let apiKey = json?["visaCheckout"]?["apikey"].asString()
        return apiKey?.isEmpty == false
```

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -28,7 +28,6 @@ import BraintreeCore
     @objc public func createProfile(completion: @escaping (Profile?, Error?) -> Void) {
         apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
             if let error {
-                /// TODO: Add failure analytics event
```
jaxdesmarais: Should we be removing this TODO? It doesn't look like this has been done?

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -65,7 +64,7 @@ import BraintreeCore
             let profile = Profile(
                 environment: environment,
                 apiKey: visaCheckoutAPIKey,
-                profileName: nil
+                profileName: " "
```
jaxdesmarais: Is an empty profile name valid? Why the change to not pass nil?

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -90,17 +89,17 @@ import BraintreeCore
         let encryptedPaymentData = checkoutResult.encryptedPaymentData
 
         if statusCode == .statusUserCancelled {
-            completion(nil, BTVisaCheckoutError.canceled)
+            completion(nil, "\(BTVisaCheckoutError.canceled)" as? Error)
```
jaxdesmarais: I don't think we need to cast any of these. They are already an error type.

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -128,21 +127,23 @@ import BraintreeCore
                 return
             }
 
-            completion(visaCheckoutCardNonce, nil)
+            self.notifySuccess(with: visaCheckoutCardNonce, completion: completion)
```
jaxdesmarais: This should be done in the analytics ticket/PR

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -128,21 +127,23 @@ import BraintreeCore
...
     }
 
+    // MARK: - Analytics Helper Methods
+
     /// Notifies the success of the Visa Checkout tokenization.
     private func notifySuccess(
         with result: BTVisaCheckoutNonce?,
         completion: @escaping (BTVisaCheckoutNonce?, Error?) -> Void
     ) {
-        /// TODO: Send success analytics event
```
jaxdesmarais: This TODO still needs to be done

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -128,21 +127,23 @@ import BraintreeCore
...
         completion(result, nil)
     }
 
+    /// Notifies the failure of the Visa Checkout tokenization.
     private func notifyFailure(with error: Error, completion: @escaping (BTVisaCheckoutNonce?, Error?) -> Void) {
-        /// TODO: Send failure analytics event
```
jaxdesmarais: This TODO still needs to be done

Sources/BraintreeVisaCheckout/BTVisaCheckoutConfiguration.swift
```diff
@@ -0,0 +1,50 @@
+import VisaCheckoutSDK
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+/// Contains the remote Visa Checkout configuration for the Braintree SDK.
+extension BTConfiguration {
```
jaxdesmarais: I don't think this should be added?

jaxdesmarais:
Adding this line visa checkout works without issue: https://github.com/braintree/braintree-ios-visa-checkout/blob/81afb7f3c8c08031125c3d707e17260db8a26902/Demo/Demo%20Base/BraintreeDemoAppDelegate.m#L17

Demo/Application/Base/PaymentButtonBaseViewController.swift
stechiu: No, there are several known issues in this PR and hardcoding different tokenization keys has been part of the testing

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: I know, I'm leaving analytics work for another PR

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: I know, I'm leaving analytics work for another PR

Sources/BraintreeVisaCheckout/BTVisaCheckoutConfiguration.swift
stechiu: The properties in `BTConfiguration` extension do get used in `BTVisaCheckoutClient`

jaxdesmarais: Yes but this file already exists and is being re-added in this PR. Existing file: https://github.com/braintree/braintree_ios/blob/visa-checkout-feature/Sources/BraintreeVisaCheckout/BTConfiguration%2BVisaCheckout.swift

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: I was testing something out for analytics and missed removing a few things, like this typecasting. Thanks

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: I have the analytics ticket assigned to me so no worries, I know where to add in analytic events

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: When there were issues with loading Visa Checkout, Justin and I were testing out different values in `Profile`. As shared in the description, `profileName is a required property to create a Visa Profile but the Visa Checkout button won't render if profileName is NOT nil or an empty string`. This PR is tagged with "do not review" and I was going to clean it up after figuring out the Visa Checkout loading part first. Then, "do not review" tag would be removed and ready for review.

jaxdesmarais: Sure, you can ignore my comments for now. 😄  Was just reviewing since you said you were blocked/Visa Checkout wasn't working.

Demo/Application/Features/VisaCheckoutViewController.swift
stechiu: True. I've added this `self.progressBlock("Error: launchHandler is not set.")`

Demo/Application/Features/VisaCheckoutViewController.swift
stechiu: Can you be more specific?

Sources/BraintreeVisaCheckout/BTConfiguration+VisaCheckout.swift
stechiu: Thanks!

Demo/Application/Features/VisaCheckoutViewController.swift
stechiu: Good point. I'm not sure what's a good error message here. What do you think of this? I'm open to suggestions
```
guard let profile = profile else {
                self.progressBlock("Failed to create Visa profile: \(error ?? BTVisaCheckoutError.integration)")
                return
            }
```

Demo/Application/Features/VisaCheckoutViewController.swift
jaxdesmarais: We may not need a mark but these are internal so can update this mark to be accurate to the methods below it.

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+
+    lazy var visaCheckoutClient = BTVisaCheckoutClient(apiClient: apiClient)
+    var launchHandler: LaunchHandle?
+    let visaCheckoutButton = VisaCheckoutButton()
+
+    // MARK: - Public Methods
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        title = "Visa Checkout Button"
```
jaxdesmarais: nit (take it or leave it):
```suggestion
        title = "Visa Checkout"
```

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+        createVisaCheckoutButton()
+        createVisaProfileAndCheckout()
+    }
+
+    /// Creates and displays the Visa Checkout button.
+    func createVisaCheckoutButton() {
+        visaCheckoutButton.translatesAutoresizingMaskIntoConstraints = false
+        visaCheckoutButton.style = .standard
+        view.addSubview(visaCheckoutButton)
+        NSLayoutConstraint.activate([
```
jaxdesmarais: nit to maybe add some spacing for readability here:
```suggestion
        visaCheckoutButton.style = .standard
        
        view.addSubview(visaCheckoutButton)
        
        NSLayoutConstraint.activate([
```

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+
+    /// Creates a Visa profile and initiates the checkout process.
+    func createVisaProfileAndCheckout() {
+        visaCheckoutClient.createProfile { profile, error in
+            if let error {
+                self.progressBlock("Failed to create Visa profile: \(error)")
+                return
+            }
+
+            guard let profile = profile else {
```
jaxdesmarais:
```suggestion
            guard let profile else {
```

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+                        return
+                    }
+                    launchHandler()
+                },
+                completion: { result in
+                    self.progressBlock("Tokenizing VisaCheckoutResult...")
+                    self.visaCheckoutClient.tokenize(result) { tokenizedVisaCheckoutCard, error in
+                        if let error {
+                            self.progressBlock("Error tokenizing Visa Checkout card: \(error.localizedDescription)")
+                        } else if let tokenizedVisaCheckoutCard = tokenizedVisaCheckoutCard {
```
jaxdesmarais:
```suggestion
                        } else if let tokenizedVisaCheckoutCard {
```

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+                            self.completionBlock(tokenizedVisaCheckoutCard)
+                        } else {
+                            self.progressBlock("\(BTVisaCheckoutError.canceled)")
```
jaxdesmarais: I think canceled is one of our error cases, right? If so here we can instead do something like:
```suggestion
                            self.progressBlock("No error or nonce returned from the Visa Checkout flow.")
```

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -256,6 +259,14 @@
 /* End PBXFrameworksBuildPhase section */
 
 /* Begin PBXGroup section */
+		045C0C912E53C80B002A6FFC /* Recovered References */ = {
```
jaxdesmarais: Do we need the recovered references additions in this file? Looks like there are a few that can be removed.

Sources/BraintreeVisaCheckout/BTConfiguration+VisaCheckout.swift
```diff
@@ -14,7 +14,8 @@ extension BTConfiguration {
 
     /// Determines if the Visa Checkout flow is available to be used. This can be used to determine if UI components should be shown or hidden.
     var isVisaCheckoutEnabled: Bool {
-        visaCheckoutAPIKey?.isEmpty == false
+        let apiKey = json?["visaCheckout"]["apikey"].asString()
+        return apiKey?.isEmpty == false
```
jaxdesmarais: It looks like this logic may not be needed. `visaCheckoutAPIKey` does the same check we do on line 17 here. I think we may be able to revert this change.

Demo/Application/Features/VisaCheckoutViewController.swift
stechiu: Thanks and yes, `.canceled` is one of the error cases

Demo/Demo.xcodeproj/project.pbxproj
stechiu: Thanks, that might be what's causing the weird Carthage failure

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+class VisaCheckoutViewController: PaymentButtonBaseViewController {
+
+    lazy var visaCheckoutClient = BTVisaCheckoutClient(apiClient: apiClient)
+    var launchHandler: LaunchHandle?
+    let visaCheckoutButton = VisaCheckoutButton()
```
agedd: small nit: can we add spaces in-between these properties

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -0,0 +1,88 @@
+import UIKit
+import BraintreeVisaCheckout
+import VisaCheckoutSDK
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
```
jaxdesmarais: This is only needed for modules we deploy. Can just do a normal import in the VC:
```suggestion
import BraintreeCore
```

PR #1641, opened 2025-08-20
Add required query params

NastassiaRodzik:

### Summary of changes

- Added new query parameters that will be required from the PayPal app switch universal link.
- Updated the Demo App so that we'd be able to test the Pay Later functionality.

### Checklist

- [x] Added a changelog entry
- [x] Relevant test coverage

### Video
Pay later functionality on One time checkout flow:
https://github.com/user-attachments/assets/9ab4fe2f-9908-40e5-beaf-7ce725bc95b1

### DIFF
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -145,7 +145,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         sender.setTitle("Processing...", for: .disabled)
         sender.isEnabled = false
 
-        let request = BTPayPalCheckoutRequest(amount: "5.00")
+        let request = BTPayPalCheckoutRequest(amount: "5.00", offerPayLater: payLaterToggle.isOn)
         request.userAuthenticationEmail = emailTextField.text
         request.userPhoneNumber = BTPayPalPhoneNumber(
             countryCode: countryCodeTextField.text ?? "",
@@ -247,7 +247,8 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
             userAuthenticationEmail: emailTextField.text,
             enablePayPalAppSwitch: true,
             amount: "10.00",
-            userAction: .payNow
+            userAction: .payNow,
+            offerPayLater: payLaterToggle.isOn
         )
 
         payPalClient.tokenize(request) { nonce, error in
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -145,7 +145,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         sender.setTitle("Processing...", for: .disabled)
         sender.isEnabled = false
 
-        let request = BTPayPalCheckoutRequest(amount: "5.00", offerPayLater: payLaterToggle.isOn)
+        let request = BTPayPalCheckoutRequest(amount: "50.00", offerPayLater: payLaterToggle.isOn)
         request.userAuthenticationEmail = emailTextField.text
         request.userPhoneNumber = BTPayPalPhoneNumber(
             countryCode: countryCodeTextField.text ?? "",
@@ -246,7 +246,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         let request = BTPayPalCheckoutRequest(
             userAuthenticationEmail: emailTextField.text,
             enablePayPalAppSwitch: true,
-            amount: "10.00",
+            amount: "50.00",
             userAction: .payNow,
             offerPayLater: payLaterToggle.isOn
         )
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -246,7 +246,8 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         let request = BTPayPalCheckoutRequest(
             userAuthenticationEmail: emailTextField.text,
             enablePayPalAppSwitch: true,
-            amount: "10.00"
+            amount: "10.00",
+            userAction: .payNow
         )
 
         payPalClient.tokenize(request) { nonce, error in
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -145,7 +145,9 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         sender.setTitle("Processing...", for: .disabled)
         sender.isEnabled = false
 
-        let request = BTPayPalCheckoutRequest(amount: "50.00", offerPayLater: payLaterToggle.isOn)
+        // pay later is only available on amounts greater than or equal to 35
+        let amount = payLaterToggle.isOn ? "35.00" : "5.00"
+        let request = BTPayPalCheckoutRequest(amount: amount, offerPayLater: payLaterToggle.isOn)
         request.userAuthenticationEmail = emailTextField.text
         request.userPhoneNumber = BTPayPalPhoneNumber(
             countryCode: countryCodeTextField.text ?? "",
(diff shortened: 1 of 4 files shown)


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -465,6 +466,7 @@ import BraintreeDataCollector
 
     private func launchPayPalApp(
         with payPalAppRedirectURL: URL,
+        merchantAccountId: String?,
```
agedd: small nit: 
```suggestion
        merchantAccountId: String? = nil,
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -500,7 +502,9 @@ import BraintreeDataCollector
         var urlComponents = URLComponents(url: payPalAppRedirectURL, resolvingAgainstBaseURL: true)
         let additionalQueryItems: [URLQueryItem] = [
             URLQueryItem(name: "source", value: "braintree_sdk"),
-            URLQueryItem(name: "switch_initiated_time", value: String(Int(round(Date().timeIntervalSince1970 * 1000))))
+            URLQueryItem(name: "switch_initiated_time", value: String(Int(round(Date().timeIntervalSince1970 * 1000)))),
+            URLQueryItem(name: "flow_type", value: isVaultRequest ? "va" : "ecs"),
+            URLQueryItem(name: "merchant", value: merchantAccountId ?? "unknown")
```
agedd: nit: can probably remove the "unknown" default value here
```suggestion
            URLQueryItem(name: "merchant", value: merchantAccountId)
```

agedd: also looks like we're not sending in an "unknown" in the android pr 👀

agedd (approved):
some small nit comments, but otherwise looks good!

Sources/BraintreePayPal/BTPayPalClient.swift
NastassiaRodzik: That's right: in Android `merchantID` is non optional, so they don't need to provide placeholder value.
In case of iOS when `merchantAccountId` is nil, merchant query item will result in 
`.../app-switch-checkout?/token=ABCD&merchant`

Vs after providing default value:
.../app-switch-checkout?/token=ABCD&merchant=unknown

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -145,7 +145,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         sender.setTitle("Processing...", for: .disabled)
         sender.isEnabled = false
 
-        let request = BTPayPalCheckoutRequest(amount: "5.00")
+        let request = BTPayPalCheckoutRequest(amount: "50.00", offerPayLater: payLaterToggle.isOn)
```
jaxdesmarais: Why the change to $50? $5 is reasonable since this will put a temp auth on cards for prod testing, they will fall of but $50 seems unnecessary even if no money is actually moved.

Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -246,7 +246,9 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         let request = BTPayPalCheckoutRequest(
             userAuthenticationEmail: emailTextField.text,
             enablePayPalAppSwitch: true,
-            amount: "10.00"
+            amount: "50.00",
+            userAction: .payNow,
```
jaxdesmarais, line 254: is `payNow` required to test pay later? Same question on the amount change as above.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -454,7 +454,8 @@ import BraintreeDataCollector
                         )
                         return
                     }
-                    self.launchPayPalApp(with: url, completion: completion)
+                    let merchantAccountId = json["merchantId"].asString()
```
jaxdesmarais: To match style guide:
```suggestion
                    let merchantAccountID = json["merchantId"].asString()
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -465,6 +466,7 @@ import BraintreeDataCollector
 
     private func launchPayPalApp(
         with payPalAppRedirectURL: URL,
+        merchantAccountId: String? = nil,
```
jaxdesmarais:
```suggestion
        merchantAccountID: String? = nil,
```

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: Have we confirmed these values are coming through as expected?

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -874,6 +874,10 @@ class BTPayPalClient_Tests: XCTestCase {
         } else {
             XCTFail("Expected integer value for query param `switch_initiated_time`")
         }
+        XCTAssertEqual(urlComponents?.queryItems?[3].name, "flow_type")
+        XCTAssertEqual(urlComponents?.queryItems?[3].value, "va")
+        XCTAssertEqual(urlComponents?.queryItems?[4].name, "merchant")
+        XCTAssertEqual(urlComponents?.queryItems?[4].value, "unknown")
```
jaxdesmarais: Can we test an actual value? It should exist in our unit tests vs unknown.

jaxdesmarais (changes requested):

Demo/Application/Features/PayPalWebCheckoutViewController.swift
NastassiaRodzik: This is for merchants to be able to test Pay Later since it's only working on amounts greater than 35$ 
As an alternative, I can change the amount based on `pay later toggle` for prod testing

Demo/Application/Features/PayPalWebCheckoutViewController.swift
NastassiaRodzik: `payNow` is required to app switch into the app, otherwise it always opens browser instead

Sources/BraintreePayPal/BTPayPalClient.swift
NastassiaRodzik: @jaxdesmarais yes merchantID is coming as expected from configuration

Demo/Application/Features/PayPalWebCheckoutViewController.swift
jaxdesmarais: Yes, lets bump down to $35 and change the amount only based on toggle

Demo/Application/Features/PayPalWebCheckoutViewController.swift
jaxdesmarais: For `offerPayLater` specifically to app switch?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -454,7 +454,8 @@ import BraintreeDataCollector
                         )
                         return
                     }
-                    self.launchPayPalApp(with: url, completion: completion)
+                    let merchantAccountID = json["merchantId"].asString()
+                    self.launchPayPalApp(with: url, merchantAccountId: merchantAccountId, completion: completion)
```
jaxdesmarais: You're going to need to change all of these to `merchantAccountID`. I expect the current commit doesn't even build.

NastassiaRodzik: my bad, using auto git fix suggestion first time today - fixed
[Style guide fix](https://github.com/braintree/braintree_ios/pull/1641/commits/6846d3d3cd85d42a2437e9a3a61da494fa800a77)

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
NastassiaRodzik: Updated here 
[Update tests to pass merchant id in response](https://github.com/braintree/braintree_ios/pull/1641/commits/7f567ff2f101eba1a6af2a094576feeefd00fae2)

Demo/Application/Features/PayPalWebCheckoutViewController.swift
NastassiaRodzik: Fixed here 
[Amount fix](https://github.com/braintree/braintree_ios/pull/1641/commits/7a43d00684b1ed36f0dbcdd478222da84a7b30a3)

Demo/Application/Features/PayPalWebCheckoutViewController.swift
NastassiaRodzik: Regardless of `offerPayLater` parameter: 
without userAction set to `payNow` which translates into "user_action": "commit"  parameter of`v1/paypal_hermes/create_payment_resource` request, it will return `launchPayPalApp = 0` in response which always launches browser instead of the app

PR #1644, opened 2025-08-21
Visa Checkout Analytics

stechiu:
### Summary of changes

- Added analytics to Visa Checkout

### Checklist

- [ ] Added a changelog entry

- @stechiu

### DIFF
--- a/Sources/BraintreeVisaCheckout/BTVisaCheckoutAnalytics.swift
+++ b/Sources/BraintreeVisaCheckout/BTVisaCheckoutAnalytics.swift
@@ -1,0 +1,7 @@
+import Foundation
+
+enum BTVisaCheckoutAnalytics {
+    static let tokenizeStarted = "visa-checkout:tokenize:started"
+    static let tokenizeFailed = "visa-checkout:tokenize:failed"
+    static let tokenizeSucceeded = "visa-checkout:tokenize:succeeded"
+}
--- a/Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
+++ b/Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
@@ -28,7 +28,6 @@ import BraintreeCore
     @objc public func createProfile(completion: @escaping (Profile?, Error?) -> Void) {
         apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
             if let error {
-                /// TODO: Add failure analytics event
                 completion(nil, error)
                 return
             }
@@ -90,8 +89,11 @@ import BraintreeCore
             return
         }
 
+        let analyticEvent = visaCheckoutAnalyticsEvent(for: statusCode)
+        apiClient.sendAnalyticsEvent(BTVisaCheckoutAnalytics.tokenizeStarted)
+
         guard statusCode == .statusSuccess else {
-            completion(nil, BTVisaCheckoutError.checkoutUnsuccessful)
+            completion(nil, "\(BTVisaCheckoutError.checkoutUnsuccessful): \(analyticEvent)" as? Error)
             return
         }
 
@@ -100,7 +102,7 @@ import BraintreeCore
             let encryptedKey = checkoutResult.encryptedKey,
             let encryptedPaymentData = checkoutResult.encryptedPaymentData
         else {
-            completion(nil, BTVisaCheckoutError.integration)
+            completion(nil, "\(BTVisaCheckoutError.integration): \(analyticEvent)" as? Error)
             return
         }
 
@@ -119,12 +121,12 @@ import BraintreeCore
             }
 
             guard let visaCheckoutCards = body?["visaCheckoutCards"].asArray()?.first else {
-                self.notifyFailure(with: BTVisaCheckoutError.failedToCreateNonce, completion: completion)
+                self.notifyFailure(with: "\(BTVisaCheckoutError.failedToCreateNonce): \(analyticEvent)" as! Error, completion: completion)
                 return
             }
 
             guard let visaCheckoutCardNonce = BTVisaCheckoutNonce(json: visaCheckoutCards) else {
-                self.notifyFailure(with: BTVisaCheckoutError.failedToCreateNonce, completion: completion)
+                self.notifyFailure(with: "\(BTVisaCheckoutError.failedToCreateNonce): \(analyticEvent)" as! Error, completion: completion)
                 return
             }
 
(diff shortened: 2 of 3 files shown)


Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -28,7 +28,6 @@ import BraintreeCore
     @objc public func createProfile(completion: @escaping (Profile?, Error?) -> Void) {
         apiClient.fetchOrReturnRemoteConfiguration { configuration, error in
             if let error {
-                /// TODO: Add failure analytics event
```
warmkesselj, line 31: I see that there isn't a failure event for this but do we need one?

Demo/Application/Features/VisaCheckoutViewController.swift
```diff
@@ -73,6 +73,8 @@ class VisaCheckoutViewController: PaymentButtonBaseViewController {
                         if let error {
                             self.progressBlock("Error tokenizing Visa Checkout card: \(error.localizedDescription)")
                         } else if let tokenizedVisaCheckoutCard {
+                            let nonce = tokenizedVisaCheckoutCard.nonce
+                            let shippingAddress = tokenizedVisaCheckoutCard.shippingAddress
```
jaxdesmarais: Are these changes needed?

Sources/BraintreeVisaCheckout/BTVisaCheckoutAnalytics.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+enum BTVisaCheckoutAnalytics {
+    static let tokenizeStarted = "visa-checkout:tokenize:started"
```
jaxdesmarais: nit:
```suggestion
enum BTVisaCheckoutAnalytics {

    static let tokenizeStarted = "visa-checkout:tokenize:started"
```

jaxdesmarais (commented):
Missing calling `notifyFailure` for `completion` on like 118 in the `if let error` block.

Demo/Application/Features/VisaCheckoutViewController.swift
stechiu: They're extra as part of something I was testing. Removed

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: I thought about it and created another `notifyFailure` function for `Profile` so analytics get passed for it

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -28,26 +28,26 @@ import BraintreeCore
...
+                self.notifyFailure(with: error, completion: completion)
```
richherrera: Do we need the completion on line 32 `completion(nil, error)`? I think we can remove it, what do you think?

richherrera (commented):
On line 119 we have this completion: 
`completion(nil, error)`

 I think we could replace it with:
`self.notifyFailure(with: error, completion: completion)`

 can we double-check that?

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: Oops! `completion(nil, error)` was supposed to be removed. Thanks for catching that

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
jaxdesmarais: I think we can omit the analytics events for the profile method to match android: https://github.com/braintree/braintree_android/blob/66e57b03649cdffd3380d3e310a865718192151b/VisaCheckout/src/main/java/com/braintreepayments/api/visacheckout/VisaCheckoutClient.kt#L52-L95. This entire method can just call the completion without analytics.

jaxdesmarais: To clarify, we should only send analytics events in the `tokenize` method.

stechiu: Gotcha. I've reverted

stechiu:
> Looks like we are missing 1 notify failure on line 118

Don't know how I missed that, thanks

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -113,8 +114,8 @@ import BraintreeCore
         ]
 
         apiClient.post("v1/payment_methods/visa_checkout_cards", parameters: parameters) { body, _, error in
-            if let error {
-                completion(nil, error)
+            if error != nil {
+                self.notifyFailure(with: BTVisaCheckoutError.failedToCreateNonce, completion: completion)
```
```suggestion
            if let error {
                self.notifyFailure(with: error, completion: completion)
```

Sources/BraintreeCore/BTClientMetadataIntegration.swift
```diff
@@ -1,4 +1,5 @@
-/// :nodoc: This enum is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+/// :nodoc: This enum is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning
+/// and may change or be removed at any time
```
agedd: small nit: 
```suggestion
/// and may change or be removed at any time.
```

agedd (approved):
lgtm!! 🚀

stechiu:
> LGTM!!! 🎉 Will the client side tests to validate that we’re sending the events correctly be added in a follow up PR?

thanks! yes I'll be pushing up a PR for client unit tests today after I merge in this PR

Sources/BraintreeCore/BTClientMetadataIntegration.swift
```diff
@@ -1,4 +1,5 @@
-/// :nodoc: This enum is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+/// :nodoc: This enum is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning
```
jaxdesmarais, line 1: Looks like this and the BTClientMetadataSource file were updated by accident.

stechiu: * Discussed via slack that we can keep the change as is

PR #1615, opened 2025-07-18
Added BTVisaCheckoutClient to Visa Checkout

stechiu:
Converted `VisaCheckoutClient` form ObjC to Swift.

### Summary of changes

- Converted these from these 2 files from braintree-ios-visa-checkout: [BTVisaCheckoutClient.m](https://github.com/braintree/braintree-ios-visa-checkout/blob/main/BraintreeVisaCheckout/BTVisaCheckoutClient.m) + [BTVisaCheckoutClient.h](https://github.com/braintree/braintree-ios-visa-checkout/blob/main/BraintreeVisaCheckout/Public/BraintreeVisaCheckout/BTVisaCheckoutClient.h)

### Checklist

- [x] Added a changelog entry

- @stechiu

PR #1649, opened 2025-09-03
Visa Checkout - Unit Test

stechiu:
### Summary of changes

- Added unit tests for Visa Checkout Client 

### Checklist

- [ ] Added a changelog entry

- @stechiu


UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -44,4 +58,453 @@ final class BTVisaCheckoutClient_Tests: XCTestCase {
...
+            }
+
+            guard let nonce = nonce else {
+                XCTFail("Expected a nonce")
+                return
+            }
+
+            XCTAssertEqual(nonce.type, "Visa")
+            XCTAssertEqual(nonce.nonce, "123456-12345-12345-a-adfa")
+//            XCTAssertEqual(nonce.cardType, BTCardNetwork.visa.rawValue)
```
richherrera: Can we validate the card type? Or should we just remove this line of code?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,8 +1,22 @@
...
+    var encryptedKey = "b"
+    var encryptedPaymentData = "c"
+
+    override func setUp() {
+        super.setUp()
+        mockAPIClient = MockAPIClient(authorization: "development_tokenization_key")!
+        visaCheckoutClient = BTVisaCheckoutClient(apiClient: mockAPIClient)
+    }
 
     func testBTVisaCheckoutAddress_initializesAllPropertiesCorrectly() {
```
richherrera: I think the `BTVisaCheckoutAddress` tests might be better off in a separate file, maybe creating a `BTVisaCheckoutAddress_Tests` file would work better, what do you think?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,8 +1,22 @@
 import XCTest
+import BraintreeTestShared
+import VisaCheckoutSDK
 @testable import BraintreeCore
 @testable import BraintreeVisaCheckout
 
 final class BTVisaCheckoutClient_Tests: XCTestCase {
+    var mockAPIClient: MockAPIClient = MockAPIClient(authorization: "development_tokenization_key")!
+    var visaCheckoutClient: BTVisaCheckoutClient!
```
richherrera: In each of the tests, we’re creating a new instance of `BTVisaCheckoutClient`. Should we use this one that’s instantiated in `setUp`, or should we remove this line of code?

Braintree.xcodeproj/project.pbxproj
```diff
@@ -9,7 +9,6 @@
 /* Begin PBXBuildFile section */
 		0453F7AD2E3D5E7900849AD8 /* VisaCheckoutSDK.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 04C652832E2B049A00C879A0 /* VisaCheckoutSDK.xcframework */; };
 		0453F7B02E3D5E8200849AD8 /* VisaCheckoutSDK.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 04C652832E2B049A00C879A0 /* VisaCheckoutSDK.xcframework */; };
-		0453F7B12E3D5E8200849AD8 /* VisaCheckoutSDK.xcframework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 04C652832E2B049A00C879A0 /* VisaCheckoutSDK.xcframework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
```
jaxdesmarais, line 12: Should these be removed? Is everything still working as expected without this?

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -86,6 +86,27 @@ import BraintreeCore
         apiClient.sendAnalyticsEvent(BTVisaCheckoutAnalytics.tokenizeStarted)
 
         let statusCode = checkoutResult.statusCode
+        let callID = checkoutResult.callId
+        let encryptedKey = checkoutResult.encryptedKey
+        let encryptedPaymentData = checkoutResult.encryptedPaymentData
+
+        tokenize(
```
jaxdesmarais, line 103: Can you share details on why this change is needed?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -44,4 +58,453 @@ final class BTVisaCheckoutClient_Tests: XCTestCase {
...
+        client.createProfile { (profile, error) in
+            let err = error! as NSError
+            XCTAssertNil(profile)
+            XCTAssertEqual(err.domain, "MyError")
+            XCTAssertEqual(err.code, 123)
+
+            expecation.fulfill()
+        }
+
+        waitForExpectations(timeout: 1, handler: nil)
```
jaxdesmarais: I think a lot of the syntax is still using some older coding styles here. This file should be updated and look something like the following:
```suggestion
        client.createProfile { profile, error in
            guard let error = error as? NSError else {
                XCTFail("Error expected to be returned in this tests")
                return
            }
                        
            XCTAssertNil(profile)
            XCTAssertEqual(error.domain, "MyError")
            XCTAssertEqual(error.code, 123)

            expecation.fulfill()
        }

        waitForExpectations(timeout: 1)
```

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: I tried not breaking up `tokenize` to test it. Tried different ways and this was what worked best

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -44,4 +58,453 @@ final class BTVisaCheckoutClient_Tests: XCTestCase {
...
+                    "Discover"
+                ]
+            ]
+        ])
+
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let expecation = expectation(description: "profile success")
+
+        client.createProfile { profile, error in
+            guard let profile = profile else {
```
jaxdesmarais: Anytime we see this syntax we can shorten to: `guard let profile else`

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
jaxdesmarais: Are you able to share what wasn't working for the tests before this change? Just want to make sure I am understanding the change.

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: That commented out line is supposed to be removed. We validate card type through `type`

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: That's a good idea

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: I believe I removed all the older syntax that I saw. The one you pointed out is the only one I saw that I overlooked. If there's a lot, then it's better we hop on a call to figure it out

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: I dove into it with Rich on Tuesday to see how we can pass in `CheckoutResult`. `VisaCheckoutResult` has no initiators and we can't change the values of `callID`, `encryptedKey`, etc directly. 

Ex: 
```
CheckoutResult.callID = "callID"

// or

let callID = CheckoutResult.callID
callID = "abc"
```
both result in the error "Type 'CheckoutResult' has no member 'callID'"

I think it's why the original BT Visa Checkout broke up tokenize function, for testing purposes

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: I'll remove the line in this case

Braintree.xcodeproj/project.pbxproj
stechiu: The change here was that the VisaCheckoutSDK framework was set to Embed & Sign. I changed it to Do Not Embed and everything's working as expected

jaxdesmarais: See comment here: https://github.com/braintree/braintree_ios/pull/1615#discussion_r2248861251

Embed and sign is correct for the type of framework this is.

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
jaxdesmarais: Ah gotcha, so the `CheckoutResult .init` is not accessible. We could technically create a mock but likely would be overcomplicating things for merchants to take that approach. In the future I think it's helpful to include a comment on your own PR for this to ensure folks that didn't pair have the same context.

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -96,10 +96,41 @@ import BraintreeCore
...
+    }
+
+    func tokenize(
+        statusCode: CheckoutResultStatus,
+        callID: String?,
+        encryptedKey: String?,
+        encryptedPaymentData: String?,
+        completion: @escaping (BTVisaCheckoutNonce?, Error?) -> Void
+    ) {
+
```
```suggestion
```

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -96,10 +96,41 @@ import BraintreeCore
...
+            return
+        }
+
         guard
-            let callID = checkoutResult.callId,
-            let encryptedKey = checkoutResult.encryptedKey,
-            let encryptedPaymentData = checkoutResult.encryptedPaymentData
+            let callID = callID,
+            let encryptedKey = encryptedKey,
+            let encryptedPaymentData = encryptedPaymentData
```
jaxdesmarais: Can likely single line this too with the guard as the line length will be shorter:
```suggestion
            let callID,
            let encryptedKey,
            let encryptedPaymentData
```

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutAddress_Tests.swift
```diff
@@ -0,0 +1,48 @@
+import XCTest
+import BraintreeTestShared
+import VisaCheckoutSDK
+@testable import BraintreeCore
+@testable import BraintreeVisaCheckout
+
+final class BTVisaCheckoutAddress_Tests: XCTestCase {
```
jaxdesmarais, line 7: 👏  thanks for putting these in their own file!

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
-            "streetAddress": "123 Main St",
-            "extendedAddress": "Apt 4B",
-            "locality": "San Francisco",
-            "region": "CA",
-            "postalCode": "94105",
-            "countryCode": "US",
-            "phoneNumber": "1234567890"
+    override func setUp() {
+        super.setUp()
+        mockAPIClient = MockAPIClient(authorization: "development_tokenization_key")!
```
jaxdesmarais: This is the same as line 8. Do we need both or can we remove this method?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+            }
+
+            XCTAssertNil(profile)
+            XCTAssertEqual(error.domain, "MyError")
+            XCTAssertEqual(error.code, 123)
+
+            expecation.fulfill()
+        }
+
+        waitForExpectations(timeout: 1, handler: nil)
```
jaxdesmarais:
```suggestion
        waitForExpectations(timeout: 1)
```

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+
+    func testCreateProfile_whenVisaCheckoutIsNotEnabled_callsBackWithError() {
+        mockAPIClient.cannedConfigurationResponseBody = BTJSON(value: {})
+
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let expecation = expectation(description: "profile error")
+
+        client.createProfile { profile, error in
+            XCTAssertNil(profile)
+            XCTAssertEqual(error as! BTVisaCheckoutError, BTVisaCheckoutError.disabled)
```
jaxdesmarais: This should be testing the specific error. Take a look at `testHandleReturn_whenURLIsUnknown_returnsError` as an example for how to test errors.

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+
+            expecation.fulfill()
+        }
+
+        waitForExpectations(timeout: 1, handler: nil)
```
jaxdesmarais:
```suggestion
        waitForExpectations(timeout: 1)
```

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+
+            if let acceptedCardBrands = profile.acceptedCardBrands {
+                XCTAssertTrue(acceptedCardBrands.count == 4)
+            } else {
+                XCTFail("Expected 4 accepted card brands.")
+            }
+            expecation.fulfill()
+        }
+
+        waitForExpectations(timeout: 1, handler: nil)
```
jaxdesmarais:
```suggestion
        waitForExpectations(timeout: 1)
```

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+                XCTFail("Expected nonce")
+                return
+            }
+            
+            guard let error = error as? NSError else {
+                XCTFail("Error expected to be returned in this tests")
+                return
+            }
+
+            XCTAssertEqual(error as! BTVisaCheckoutError, BTVisaCheckoutError.integration)
```
jaxdesmarais: Same comment here on updating to test a specific error.

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
-        XCTAssertNil(address.phoneNumber)
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let encryptedKey = "b"
+        let encryptedPaymentData = "c"
+
+        let expecation = expectation(description: "Tokenization error due to malformed CheckoutResult")
+        
+        client.tokenize(statusCode: .statusSuccess, callID: nil, encryptedKey: encryptedKey, encryptedPaymentData: encryptedPaymentData) { nonce, error in
+            if nonce != nil {
+                XCTFail("Expected nonce")
```
jaxdesmarais: It looks like this comment is incorrect based on the logic. Should we instead of this bock just be asserting that the nonce is nil?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+            
+            guard let error = error as? NSError else {
+                XCTFail("Error expected to be returned in this tests")
+                return
+            }
+
+            XCTAssertEqual(error as! BTVisaCheckoutError, BTVisaCheckoutError.integration)
+            expecation.fulfill()
+        }
+        waitForExpectations(timeout: 1, handler: nil)
```
jaxdesmarais:
```suggestion
        waitForExpectations(timeout: 1)
```

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+    }
+
+    func testTokenize_whenStatusCodeIndicatesCancellation_callsCompletionWithNilNonceAndError() {
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let expectation = self.expectation(description: "Callback invoked")
+
+        
+        client.tokenize(statusCode: .statusUserCancelled, callID: nil, encryptedKey: encryptedKey, encryptedPaymentData: encryptedPaymentData) { result, error in
+
```
jaxdesmarais:
```suggestion
```

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+            XCTAssertNil(result)
+            XCTAssertEqual(error as! BTVisaCheckoutError, BTVisaCheckoutError.canceled)
```
jaxdesmarais: Update to test specifics of error

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+            expectation.fulfill()
+        }
+
+        self.waitForExpectations(timeout: 1, handler: nil)
```
jaxdesmarais:
```suggestion
        waitForExpectations(timeout: 1)
```

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+    func testTokenize_whenStatusCodeIndicatesError_callsCompletionWithError() {
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let expectation = self.expectation(description: "Callback invoked")
+
+        client.tokenize(statusCode: .statusInternalError, callID: callID, encryptedKey: encryptedKey, encryptedPaymentData: encryptedPaymentData) { _, error in
+            guard let error = error as NSError? else {
+                XCTFail("Expected an error to be returned")
+                return
+            }
+            XCTAssertEqual(error.code, BTVisaCheckoutError.checkoutUnsuccessful.rawValue)
```
jaxdesmarais: Can we test description and domain here too?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+            expectation.fulfill()
+        }
+
+        self.waitForExpectations(timeout: 1, handler: nil)
```
jaxdesmarais:
```suggestion
        waitForExpectations(timeout: 1)
```

jaxdesmarais: We should do this on all the tests.

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+    }
+
+    func testTokenize_whenStatusCodeIndicatesCancellation_callsAnalyticsWithCancelled() {
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let expectation = self.expectation(description: "Analytic sent")
+        
+        client.tokenize(statusCode: .statusUserCancelled, callID: callID, encryptedKey: encryptedKey, encryptedPaymentData: encryptedPaymentData) { _, _ in
+            XCTAssertEqual(self.mockAPIClient.postedAnalyticsEvents.last, BTVisaCheckoutAnalytics.tokenizeFailed.description)
```
jaxdesmarais: nit to just be a bit more explicit on this assertion:
```suggestion
            XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains("EVENT_NAME_HERE"))
```

jaxdesmarais: Same in below tests

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+                XCTFail("Expected nonce to be nil")
+                return
+            }
+
+            guard let error = error as? NSError else {
+                XCTFail("Error expected to be returned in this tests")
+                return
+            }
+
+            XCTAssertEqual(error, self.mockAPIClient.cannedResponseError)
```
jaxdesmarais: Should we be throwing a custom error here?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+        mockAPIClient.cannedResponseError = NSError(domain: "https://braintree.com", code: 123, userInfo: nil)
+
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let expecation = expectation(description: "tokenization error")
+        
+        client.tokenize(statusCode: .statusSuccess, callID: callID, encryptedKey: encryptedKey, encryptedPaymentData: encryptedPaymentData) { nonce, error in
+            if nonce != nil {
+                XCTFail("Expected nonce to be nil")
+                return
+            }
```
jaxdesmarais: Can we instead assert that the nonce is nil?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+        )
+
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let expecation = expectation(description: "tokenization success")
+
+        client.tokenize(statusCode: .statusSuccess, callID: callID, encryptedKey: encryptedKey, encryptedPaymentData: encryptedPaymentData) { nonce, error in
+            if error != nil {
+                XCTFail("Expected no error")
+                return
+            }
```
jaxdesmarais: Can we instead assert that the error is nil?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+            ])
+
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let expecation = expectation(description: "tokenization success")
+
+        client.tokenize(statusCode: .statusSuccess, callID: callID, encryptedKey: encryptedKey, encryptedPaymentData: encryptedPaymentData) { nonce, error in
+            if error != nil {
+                XCTFail("Expected no error")
+                return
+            }
```
jaxdesmarais: Can we instead assert that the error is nil?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,470 @@
...
+            guard let nonce else {
+                XCTFail("Expected a nonce")
+                return
+            }
+
+            XCTAssertEqual(nonce.type, "Visa")
+            XCTAssertEqual(nonce.nonce, "123456-12345-12345-a-adfa")
+            XCTAssertEqual(nonce.lastTwo, "11")
+
+            [(nonce.shippingAddress, "shipping"), (nonce.billingAddress, "billing")].forEach { address, type in
```
jaxdesmarais: Should we have a test for the type at all?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: Why remove `handler`?

jaxdesmarais: It's no longer required and we are passing it as nil. This helps make our code cleaner to omit unused params that are not required.

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
jaxdesmarais: `XCTAssertNil` is specifically designed for validating the absence of a value in an optional type, which is the expected behavior of our tests. `XCTFail`  is used to unconditionally fail a test at a specific point in the code. It does not evaluate any expression; it simply signals a test failure. In these cases where we check for `!= nil` we should actually be testing that the optional value we expect to be nil is nil instead, so `XCTAssertNil` is more appropriate.

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: I'm thinking yes to check if the shipping and billing addresses have the same value. Sounds like `type` wouldn't be the right name so I'll change it

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: Got it, I've updated them all. FYI there are other unit test files that hav `handler: nil`. like in 3DS, Amex, ApplePay

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: Noted, I'll include a comment in the future

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: Ah got it, thank you. There are multiple ways to write tests so that that makes sense

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: I was experimenting with test setup and overlooked the duplicate `mockAPIClient`. I'll remove `setUp()` since it's not used

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: Ok, I've updated to
```
client.createProfile { profile, error in
            guard let error = error as NSError? else {
                XCTFail("Expected error")
                return
       }
XCTAssertNil(profile)
XCTAssertEqual(error.code, BTVisaCheckoutError.disabled.errorCode)
XCTAssertEqual(error.localizedDescription, BTVisaCheckoutError.disabled.localizedDescription)

expecation.fulfill()
}
```

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: Hmm yea it would be cleaner to just have `XCTAssertNil(profile)`. Thanks, I'll update other parts of the test with this

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,452 @@
...
+                XCTFail("Expected a nonce")
+                return
+            }
+
+            XCTAssertNil(error)
+            XCTAssertEqual(nonce.type, "Visa")
+            XCTAssertEqual(nonce.nonce, "123456-12345-12345-a-adfa")
+            XCTAssertEqual(nonce.lastTwo, "11")
+
+            [(nonce.shippingAddress, "shipping"), (nonce.billingAddress, "billing")].forEach { address, value in
```
jaxdesmarais: If we aren't unit testing the `value` property we can likely just `_` it here.

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: Do you mean like assertions checking the error domain, code, localized description? I made changes and added this 
```
XCTAssertEqual(error.domain, BTVisaCheckoutError.errorDomain)
XCTAssertEqual(error.code, BTVisaCheckoutError.failedToCreateNonce.errorCode)
XCTAssertEqual(error.localizedDescription, BTVisaCheckoutError.failedToCreateNonce.errorDescription)
```

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,452 @@
...
+    }
+
+    func testTokenize_whenStatusCodeIndicatesCancellation_callsCompletionWithNilNonceAndError() {
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let expectation = self.expectation(description: "Callback invoked")
+
+        
+        client.tokenize(statusCode: .statusUserCancelled, callID: nil, encryptedKey: encryptedKey, encryptedPaymentData: encryptedPaymentData) { result, error in
+            XCTAssertNil(result)
+            XCTAssertEqual(error as! BTVisaCheckoutError, BTVisaCheckoutError.canceled)
```
jaxdesmarais: I think we can update this one to match the other format where we use the code, domain, and localized description.

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: I see what you mean. Will make the updates

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: I'll update the other errors as well to include domain (if applicable), code, and localized description

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,464 @@
...
+                    "MasterCard",
+                    "American Express",
+                    "Discover"
+                ]
+            ]
+        ])
+
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let encryptedKey = "b"
+        let encryptedPaymentData = "c"
```
richherrera: Can we reuse the ones we have at the beginning of the file, on lines 11 and 12?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,464 @@
...
+                XCTFail("Expected 4 accepted card brands.")
+            }
+            expecation.fulfill()
+        }
+
+        waitForExpectations(timeout: 1)
+    }
+
+    func testTokenize_whenCheckoutResultMissingValues_callsCompletionWithIntegrationError() {
+        mockAPIClient.cannedConfigurationResponseBody = BTJSON(value: [
```
richherrera: Do we need to set up a configuration on the mock API when testing the `tokenize` method? I think it might not be necessary, worth double-checking?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,464 @@
...
+
+        client.tokenize(statusCode: .statusInternalError, callID: callID, encryptedKey: encryptedKey, encryptedPaymentData: encryptedPaymentData) { _, _ in
+            XCTAssertTrue(self.mockAPIClient.postedAnalyticsEvents.contains("visa-checkout:tokenize:failed"))
+            expectation.fulfill()
+        }
+        waitForExpectations(timeout: 1)
+    }
+
+    func testTokenize_whenTokenizationErrorOccurs_callsCompletionWithError() {
+        mockAPIClient.cannedConfigurationResponseBody = BTJSON(value: [
```
richherrera: same here, can we double check?

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,464 @@
...
+            XCTAssertEqual(error.code, BTVisaCheckoutError.failedToCreateNonce.errorCode)
+            XCTAssertEqual(error.localizedDescription, BTVisaCheckoutError.failedToCreateNonce.errorDescription)
+            expecation.fulfill()
+        }
+
+        waitForExpectations(timeout: 1)
+    }
+
+    func testTokenize_whenTokenizationErrorOccurs_sendsAnalyticsEvent() {
+        mockAPIClient.cannedConfigurationResponseBody = BTJSON(value: [
```
richherrera: same here, can we double check please

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: Thanks for the catch. I had meant to remove these after I moved the parameters from local to global

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: Ah yes, looks like it's not needed. I'll check the other tests as well, thanks

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: Removed

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
stechiu: Removed

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,416 @@
...
+            XCTAssertEqual(error.localizedDescription, BTVisaCheckoutError.integration.localizedDescription)
+            expecation.fulfill()
+        }
+        waitForExpectations(timeout: 1)
+    }
+
+    func testTokenize_whenStatusCodeIndicatesCancellation_callsCompletionWithNilNonceAndError() {
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let expectation = self.expectation(description: "Callback invoked")
+
```
jaxdesmarais: Looks like there is a double space here:
```suggestion
```

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -1,47 +1,416 @@
...
+            XCTAssertEqual(error.localizedDescription, BTVisaCheckoutError.canceled.errorDescription)
+            expectation.fulfill()
+        }
+
+        waitForExpectations(timeout: 1)
+    }
+
+    func testTokenize_whenStatusCodeIndicatesError_callsCompletionWithError() {
+        let client = BTVisaCheckoutClient(apiClient: mockAPIClient)
+        let expectation = self.expectation(description: "Callback invoked")
```
jaxdesmarais: Not a blocker but we should be able to access `expectation` directly without using self here. As an example for all the tests:
```suggestion
        let expectation = expectation(description: "Callback invoked")
```

jaxdesmarais: ^ actually looks like you did this is most of the tests so may have just missed one or two.

stechiu: ah thanks, will remove those before merging!

PR #1652, opened 2025-09-09
Visa Checkout UI Test

stechiu:
# Important

UI tests for New User was originally written, but due to inconsistencies in the Visa Checkout SDK and flakiness when testing, Jax and I decided to remove this from the test. I tried the Returning User flow and I'm not able to get past the OTP screen, so that test isn't included either.

### Summary of changes

- Added Visa Checkout UI tests
- Moved 3DS UI test extensions to BTUITest

### Checklist

- [ ] Added a changelog entry

- @stechiu


Demo/UI Tests/Helpers/BTUITest.swift
```diff
@@ -30,3 +30,142 @@ extension XCUIElement {
         }
     }
 }
+
+internal extension XCUIApplication {
```
stechiu: The `ThreeDSecure_UITests_Extensions` file was moved into `BTUITest` because the extensions can be used for other files since it extends to `XCUIApplication` and not `ThreeDSecure_UITests`

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -116,17 +116,11 @@ import BraintreeCore
...
     ) {
-        if statusCode == .statusUserCancelled {
-            notifyFailure(with: BTVisaCheckoutError.canceled, completion: completion)
-            return
-        }
-
-        guard statusCode == .statusSuccess else {
-            notifyFailure(with: BTVisaCheckoutError.checkoutUnsuccessful, completion: completion)
-            return
-        }
```
jaxdesmarais, line 127: Should this be removed? I think we still want this to notify specifically of cancel using the status code. Not sure if this was removed for testing something else.

stechiu: I needed to move this section of code, otherwise I run into the issue where the Visa XO SDK crashes when I try to cancel upon first starting the Visa SDK. I've added an `if statusCode == .statusUserCancelled` block in `tokenize` to account for how to handle the `error` if `statusCode == `.statusUserCancelled`

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -117,16 +112,23 @@ import BraintreeCore
         completion: @escaping (BTVisaCheckoutNonce?, Error?) -> Void
     ) {
         if statusCode == .statusUserCancelled {
+            _ = NSError(
```
stechiu: This `NSError` is necessary to ensure that the `BTVisaCheckoutError` is `.canceled`. Without this block, `BTVisaCheckoutError` is `.checkoutUnsuccessful`. If there are suggestions on what to have here besides `_` or `let error`, I'm open to them

UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift
```diff
@@ -141,9 +142,10 @@ final class BTVisaCheckoutClient_Tests: XCTestCase {
                 XCTFail("Expected error")
                 return
             }
+
             XCTAssertEqual(error.domain, BTVisaCheckoutError.errorDomain)
-            XCTAssertEqual(error.code, BTVisaCheckoutError.checkoutUnsuccessful.errorCode)
-            XCTAssertEqual(error.localizedDescription, BTVisaCheckoutError.checkoutUnsuccessful.localizedDescription)
+            XCTAssertEqual(error.code, BTVisaCheckoutError.failedToCreateNonce.errorCode)
+            XCTAssertEqual(error.localizedDescription, BTVisaCheckoutError.failedToCreateNonce.localizedDescription)
```
stechiu: `.failedToCreateNonce` is the correct `BTVisaCheckoutError` that gets returned

Demo/UI Tests/Visa Checkout UI Tests/VisaCheckout_UITests.swift
```diff
@@ -0,0 +1,30 @@
+import XCTest
+
+class BraintreeVisaCheckout_UITests: XCTestCase {
```
richherrera: Can we make the name of this class match the name of the file?

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -117,16 +112,23 @@ import BraintreeCore
...
+                domain: BTVisaCheckoutError.errorDomain,
+                code: BTVisaCheckoutError.canceled.errorCode,
+                userInfo: [
+                    NSLocalizedDescriptionKey:
+                        BTVisaCheckoutError.canceled.errorDescription ?? "Visa Checkout flow was canceled by the user."
```
richherrera: `errorDescription` is currently an optional `String`, could we make it non-optional? I think we’re covering all the cases in the error enum. Is there a specific reason it needs to be optional?

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
richherrera: Question: what’s the purpose of instantiating an NSError here? From what I can tell, it’s not being used, which is why the compiler suggests using _. Maybe we could remove it, what do you think?

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -117,16 +112,23 @@ import BraintreeCore
...
-        guard statusCode == .statusSuccess else {
-            notifyFailure(with: BTVisaCheckoutError.checkoutUnsuccessful, completion: completion)
-            return
-        }
-
-        guard let callID, let encryptedKey, let encryptedPaymentData else {
+        guard
+            let callID = callID,
+            let encryptedKey = encryptedKey,
+            let encryptedPaymentData = encryptedPaymentData
```
richherrera: Suggestion: since we’re using Swift 5.7+, we could simplify the optional binding with the shorthand syntax to keep the code cleaner:

```suggestion
guard let callID,
      let encryptedKey,
      let encryptedPaymentData else { return }

File: UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift (hunk 4 above)
Comment (richherrera) [reply UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift]: Question: are we sure that `BTVisaCheckoutClient` is returning the correct error? Looking at the guards around lines _142_ and `147`, it seems we’re returning the same error in both cases.

Could we double-check that? Maybe the guard at line 142 should return a different error so we can better distinguish the two scenarios. What do you think?

File: Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift (hunk 3 above)
Comment (stechiu) [reply Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift]: It gets used in the unit test `testTokenize_whenStatusCodeIndicatesCancellation_callsCompletionWithNilResultAndError()`. I tried a few other ways so that `XCTAssertEqual(error.code, BTVisaCheckoutError.canceled.errorCode)` and `XCTAssertEqual(error.localizedDescription, BTVisaCheckoutError.canceled.errorDescription)` are equal. Without explicitly stating what the localized description and code are for `.canceled`, the test return back with the error as `.checkoutUnsuccessful`.

If you have other suggestions that are cleaner or what I can do instead, I'd be happy to hear them

File: Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift (hunk 6 above)
Comment (stechiu) [reply Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift]: I could use force unwrap or use `as Any`. The `NSError` init contains `userInfo dict: [String : Any]?`

File: Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift (hunk 7 above)
Comment (stechiu) [reply Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift]: Thanks! What about the `notifyFailure`?

File: Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift (hunk 3 above)
Comment (stechiu) [reply Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift]: Rich and I paired on this and determined we don't need this `NSError` block

File: UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift (hunk 4 above)
Comment (stechiu) [reply UnitTests/BraintreeVisaCheckoutTests/BTVisaCheckoutClient_Tests.swift]: Rich and I paired on this and used a switch statement in `VisaCheckoutClient` to address the issue

File: Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift (hunk 6 above)
Comment (stechiu) [reply Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift]: Removed `NSError` block

File: Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift (hunk 7 above)
Comment (stechiu) [reply Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift]: I kept it in

**Comment by stechiu** [comment]
> I'm running into the following error when building the Demo on this branch but not on main. Are you seeing the same:
> 
> ```
> Multiple commands produce '/Users/jdesmarais/Library/Developer/Xcode/DerivedData/Braintree-crkoxshsemsgendudzaascsjktzc/Build/Products/Debug-iphonesimulator/Demo.app/Info.plist'
> ```
> 
> It's also building as expected on the `visa-checkout-feature` branch

Strange... I just ran my branch and I don't have issues. I've seen the error before. I'm wondering if you have an extra `Info.plist` under Build Resources > Copy Bundle Resources?

File: Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift [hunk 8]
```diff
@@ -86,11 +86,6 @@ import BraintreeCore
         apiClient.sendAnalyticsEvent(BTVisaCheckoutAnalytics.tokenizeStarted)
 
         let statusCode = checkoutResult.statusCode
-        if statusCode == .statusUserCancelled {
-            notifyFailure(with: BTVisaCheckoutError.canceled, completion: completion)
-            return
-        }
-
         guard statusCode == .statusSuccess else {
```
richherrera: I think the statusCode checks on checkoutResult.statusCode should be handled inside the internal tokenize method.

The tests pass because we’re testing the internal method, not this public one. Ideally, this public method should have as little logic as possible and just delegate to the internal method.

jaxdesmarais:
> Strange... I just ran my branch and I don't have issues. I've seen the error before. I'm wondering if you have an extra `Info.plist` under Build Resources > Copy Bundle Resources?

🤔 that should be determined by the pbxproj on the branch and since other branches are working it seemed to be specific to this. But if it's working for @richherrera and @agedd I can dig into it more on my side.

agedd:
> > Strange... I just ran my branch and I don't have issues. I've seen the error before. I'm wondering if you have an extra `Info.plist` under Build Resources > Copy Bundle Resources?
> 
> 🤔 that should be determined by the pbxproj on the branch and since other branches are working it seemed to be specific to this. But if it's working for @richherrera and @agedd I can dig into it more on my side.

hmm i haven't seen that error before - everything is building normally on my end. i wonder if it may be related to the deletion in the `.pbxproj` file

richherrera:
> > Strange... I just ran my branch and I don't have issues. I've seen the error before. I'm wondering if you have an extra `Info.plist` under Build Resources > Copy Bundle Resources?
> 
> 🤔 that should be determined by the pbxproj on the branch and since other branches are working it seemed to be specific to this. But if it's working for @richherrera and @agedd I can dig into it more on my side.

In this branch I don’t see any issues on my end, I was able to run the tests successfully.

There’s another branch with a similar(ish) name, and in *that* branch, the error you mentioned does occur

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -86,17 +86,8 @@ import BraintreeCore
...
-        }
-
-        guard statusCode == .statusSuccess else {
-            notifyFailure(with: BTVisaCheckoutError.checkoutUnsuccessful, completion: completion)
-            return
-        }
-
-        let callID = checkoutResult.callId
+        let checkoutResultCallID: Any = checkoutResult.callId as NSString? as Any
+        let callID = checkoutResultCallID as? String
```
richherrera: The purpose of this is that when the flow is canceled, `checkoutResult.callId` doesn’t exist, instead of returning Swift `nil`. I’d like to see if we can improve this, I think we could look into how to avoid doing multiple casts

richherrera: Curious to hear what the team thinks about this

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: Rich and I paired on this

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: Adding to what Rich said, `checkoutResult.callId` is an `NSString` so multiple castings were used here

jaxdesmarais:
Ah - ignore my error, I was checking out the wrong branch. If you don't mind cleaning up/deleting the old branches you pushed up that would be great. 😄

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
jaxdesmarais: Is there a reason we need this casting? It looks like this compiles as expected: `let callID = checkoutResult.callId`

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -116,14 +107,20 @@ import BraintreeCore
...
     ) {
-        if statusCode == .statusUserCancelled {
-            notifyFailure(with: BTVisaCheckoutError.canceled, completion: completion)
-            return
-        }
-
-        guard statusCode == .statusSuccess else {
-            notifyFailure(with: BTVisaCheckoutError.checkoutUnsuccessful, completion: completion)
-            return
+        if statusCode != .statusSuccess {
```
```suggestion
      guard statusCode == .statusSuccess else {

```

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -116,14 +107,20 @@ import BraintreeCore
...
+                notifyFailure(with: BTVisaCheckoutError.checkoutUnsuccessful, completion: completion)
+                return
+            case .statusUserCancelled:
+                notifyFailure(with: BTVisaCheckoutError.canceled, completion: completion)
+                return
+            case .statusNetworkError:
+                notifyFailure(with: BTVisaCheckoutError.failedToCreateNonce, completion: completion)
+                return
+            default:
+                break
```
jaxdesmarais: should we consider a new error for this:
```suggestion
                notifyFailure(with: BTVisaCheckoutError.unknownStatus, completion: completion)
```

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
richherrera: Yes, it compiles correctly, in fact, `checkoutResult.callId` returns a nullable String (Swift `String?`).

But for example, when the user cancels the flow, this property doesn’t exist, and we get an error like the one below.

Screenshot 2025-09-17 at 5 34 42 p m

We haven’t found another way to evaluate `callId`

jaxdesmarais: Ah, remembering now, this is why we were checking status first. Which we only do because we cannot mock the checkout result. Technically this could happen with both the encrypted key and encrypted payment data, so we may want to handle those too. We could use a helper function to extract these vs doing it inline:
```swift
func safeStringValue(from object: NSObject, key: String) -> String? {
    guard let value = object.value(forKey: key) else { return nil }
    if value is NSNull { return nil }

    guard let string = value as? String, !string.isEmpty else { return nil }
    return string
}
```

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -116,14 +107,20 @@ import BraintreeCore
...
+        if statusCode != .statusSuccess {
+            switch statusCode {
+            case .statusDuplicateCheckoutAttempt, .statusNotConfigured, .statusInternalError:
+                notifyFailure(with: BTVisaCheckoutError.checkoutUnsuccessful, completion: completion)
+                return
+            case .statusUserCancelled:
+                notifyFailure(with: BTVisaCheckoutError.canceled, completion: completion)
+                return
+            case .statusNetworkError:
+                notifyFailure(with: BTVisaCheckoutError.failedToCreateNonce, completion: completion)
```
agedd, line 118: qq here - do you think we would need to fire off another errorType here that explicitly says it's a network error rather than `.failedToCreateNonce` or is it okay to leave as is

stechiu:
> Ah - ignore my error, I was checking out the wrong branch. If you don't mind cleaning up/deleting the old branches you pushed up that would be great. 😄

oops sorry about that!! The old branch is removed now

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
richherrera: I think that’s a good idea, CheckoutResult is a JSON-type object, so we can extract `callId`, `encryptedKey`, and `encryptedPaymentData` directly from that JSON.

It’d definitely be cleaner than doing inline casting.

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
stechiu: Yea I think it would make the most sense to have additional error messages specifically for network error and unknown errors.

Sources/BraintreeVisaCheckout/BTVisaCheckoutClient.swift
```diff
@@ -159,6 +156,17 @@ import BraintreeCore
         }
     }
 
+    func safeStringValue(from checkoutResult: CheckoutResult, key: String) -> String? {
```
richherrera, line 159: Can we add some unit tests for this method?
I think it’d be a good idea to validate that it returns the correct value based on the shared JSON models, what do you think?

stechiu: Since Visa Checkout SDK will be sunsetted by January and Aisha said we should shelve the work for now, I think we can have the unit test done in a separate ticket if Aisha decides we should merge in the feature branch

richherrera (approved):
LGTM!!! 🚀

PR #1655, opened 2025-09-15
[v7] PRV2 BTApiClient refactor

richherrera:
### Summary of changes

Updates the initialization and usage of key client classes to use an authorization string instead of a `BTAPIClient` instance directly. Additionally, it refactors constructors and test setups to align with these changes.

**Client Initialization Refactor**

* Updated `BTShopperInsightsClientV2`, `BTPayPalClient`, and `BTVenmoClient` initializers in `ShopperInsightsViewControllerV2.swift` to accept an `authorization` string instead of a `BTAPIClient`, and provided a universal link for Venmo.
* Changed the internal storage of `apiClient` in `BTShopperInsightsClientV2` to be internally accessible and initialized via the `authorization` string, rather than requiring an external `BTAPIClient` instance.
* Refactored the creation of `BTPayPalVaultRequest` to use a new initializer that takes `shopperSessionID` and `userAuthenticationEmail` as parameters

**Testing Updates**

* Updated unit tests for `BTShopperInsightsClientV2` to construct the client using the new `authorization`-based initializer and then inject the mock `apiClient` for testing purposes.

**Code Cleanup**

* Removed unnecessary Mark section in `BTShopperInsightsClient.swift`

### Checklist

- [ ] Added a changelog entry

### Authors

- @richherrera

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -436,8 +436,8 @@ import BraintreeDataCollector
                         )
                         return
                     }
-                    let merchantAccountID = json["merchantId"].asString()
-                    self.launchPayPalApp(with: url, merchantAccountID: merchantAccountID, completion: completion)
+                    let merchantID = json["merchantId"].asString()
+                    self.launchPayPalApp(with: url, merchantID: merchantID, completion: completion)
                 case .webBrowser(let url):
                     self.didPayPalServerAttemptAppSwitch = false
                     self.handlePayPalRequest(with: url, paymentType: request.paymentType, completion: completion)
@@ -448,7 +448,7 @@ import BraintreeDataCollector
 
     private func launchPayPalApp(
         with payPalAppRedirectURL: URL,
-        merchantAccountID: String? = nil,
+        merchantID: String? = nil,
         completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void
     ) {
         /// Prevent multiple calls to open the app
@@ -488,7 +488,7 @@ import BraintreeDataCollector
             URLQueryItem(name: "source", value: "braintree_sdk"),
             URLQueryItem(name: "switch_initiated_time", value: String(Int(round(Date().timeIntervalSince1970 * 1000)))),
             URLQueryItem(name: "flow_type", value: isVaultRequest ? "va" : "ecs"),
-            URLQueryItem(name: "merchant", value: merchantAccountID ?? "unknown")
+            URLQueryItem(name: "merchant", value: merchantID ?? "unknown")
         ]
         
         urlComponents?.queryItems?.append(contentsOf: additionalQueryItems)


Demo/Application/Features/ShopperInsightsViewControllerV2.swift
```diff
@@ -8,9 +8,13 @@ import CryptoKit
...
     
-    lazy var shopperInsightsClient = BTShopperInsightsClientV2(apiClient: apiClient)
-    lazy var payPalClient = BTPayPalClient(apiClient: apiClient)
-    lazy var venmoClient = BTVenmoClient(apiClient: apiClient)
+    lazy var shopperInsightsClient = BTShopperInsightsClientV2(authorization: authorization)
+    lazy var payPalClient = BTPayPalClient(authorization: authorization)
+    lazy var venmoClient = BTVenmoClient(
+        authorization: authorization,
+        // swiftlint:disable:next force_unwrapping
+        universalLink: URL(string: "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments")!
```
stechiu, line 16: Out of curiosity, since this is force unwrapped, is the link something that you're sure won't break in the future?

Sources/BraintreePayPal/BTConfiguration+PayPal.swift
```diff
@@ -28,7 +28,8 @@ extension BTConfiguration {
         json?["paypal"]["currencyIsoCode"].asString()
     }
     
-    var merchantID: String? {
+    /// The merchant ID
```
stechiu: Should the description here be merchant account ID? I know there's merchant ID and merchant account ID, so describing it as `The merchant ID` can be confusing when the variable is `merchantAccountID`

Demo/Application/Features/ShopperInsightsViewControllerV2.swift
richherrera: This is a common pattern we use in different parts of our Demo app, for example, [here](https://github.com/braintree/braintree_ios/blob/20ab65ebc3bc1c9a22150229d05b891ae7328e8b/Demo/Application/Features/PayPalWebCheckoutViewController.swift#L11). It shouldn’t cause any issues in the future.

Sources/BraintreePayPal/BTConfiguration+PayPal.swift
richherrera: make sense, updated: [e69da75](https://github.com/braintree/braintree_ios/pull/1655/commits/e69da7592b15fe17baa778d43bde71265c133795)

stechiu (approved):

Sources/BraintreePayPal/BTConfiguration+PayPal.swift
```diff
@@ -28,7 +28,8 @@ extension BTConfiguration {
         json?["paypal"]["currencyIsoCode"].asString()
     }
     
-    var merchantID: String? {
+    /// The merchant account ID
+    var merchantAccountID: String? {
```
jaxdesmarais: On the JSON it looks like both of these should be `merchantID` right?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -436,7 +436,7 @@ import BraintreeDataCollector
                         )
                         return
                     }
-                    let merchantAccountID = configuration.merchantID
+                    let merchantAccountID = configuration.merchantAccountID
```
jaxdesmarais: I think these are maybe wrong and should both be `merchantID`? Not totally sure, we should look into which is correct and make sure it's consistent across the JSON, docstrings, params, etc.

Sources/BraintreePayPal/BTConfiguration+PayPal.swift
richherrera: Yes, that’s correct. Using `merchantID` creates ambiguity, `merchantID` already exists in `BTConfiguration+Core`, and since it also exists here, CocoaPods detects a [conflict](https://github.com/braintree/braintree_ios/actions/runs/17737968399/job/50405518117#step:5:15). That’s why I chose to use a different name. Any other suggestion?

agedd (approved):
lgtm 🚀

Sources/BraintreePayPal/BTConfiguration+PayPal.swift
jaxdesmarais: Ah, can we just access the property on `BTConfiguration+Core`? Wondering why we created a duplicate property for this if it extracts the same value from the JSON. Seems like it was a mistake.

richherrera: We’d need the property in `BTConfiguration+Core` to be public so it’s accessible from the PayPal module.
It’s been in Core for a long time, we use it when sending our events. It was recently added to the PayPal module to be passed as a parameter when launching the PayPal app.

What I can do is remove the property from `BTConfiguration+PayPal` and just use `json?["merchantId"]`, like it’s currently done in the main branch, what do you think?

jaxdesmarais: Oh I was not saying remove it from core, I was saying remove it from `BTConfiguration+PayPal`. If it's in core we have access to `configuration.merchantID` anywhere we need it in the SDK.

richherrera: I removed the computed property from `BTConfiguration+PayPal` to avoid duplication.

Use json["merchantId"].asString to get the value in PayPalClient, just like it’s currently done in the main branch.

Main Branch: 
https://github.com/braintree/braintree_ios/blob/20ab65ebc3bc1c9a22150229d05b891ae7328e8b/Sources/BraintreePayPal/BTPayPalClient.swift#L402

https://github.com/braintree/braintree_ios/blob/20ab65ebc3bc1c9a22150229d05b891ae7328e8b/Sources/BraintreePayPal/BTPayPalClient.swift#L461

Updated: [84745ee](https://github.com/braintree/braintree_ios/pull/1655/commits/84745ee19fcd0279c3148ba8fc0190709047ea5a)

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -436,7 +436,7 @@ import BraintreeDataCollector
                         )
                         return
                     }
-                    let merchantAccountID = configuration.merchantID
+                    let merchantAccountID = json["merchantId"].asString()
```
jaxdesmarais: Should we rename this property and the `launchPayPalApp` property to `merchantID` to properly represent the value?

richherrera: Update: [66f3675](https://github.com/braintree/braintree_ios/pull/1655/commits/66f3675029cb90477959983ed7cb88391509639b)

PR #1657, opened 2025-09-18
Include RP fields in App Switch flow for BT iOS SDK

karthikeyan-2907:
### Summary of changes

- Identified an issue where Recurring Payment (RP) fields (plan_type and plan_metadata) were not being appended during the App Switch flow due to an early return in BTPayPalRequest.parameters. The flow breaks starting from this [commit](https://github.com/braintree/braintree_ios/commit/abb4adc3570f471ba82b5c141275de938bb147fc#diff-12a919f5b0b117b30354b85383385d0c8885df582c4974647e7407f0804518cf) 
- Moved the logic for setting RP fields above the enableAppSwitch condition, ensuring they are included for both App Switch and non-App Switch flows.
- Verified that the RP fields are now correctly passed in the final parameters during App Switch.

### Checklist

- [ ] Verified that RP fields are included in both App Switch and non-App Switch flows
- [ ] Tested with **temp-rp4.1-and-rp8-testing** branch for RP without purchase

### Authors
@karthikeyan-2907


agedd:
hi @karthikeyan-2907 -  thanks for catching this and opening up a PR. would you be able to move it from draft to ready to review so CI can ruth through?

PR #1660, opened 2025-09-23
Include RP fields in App Switch flow for BT iOS SDK

karthikeyan-2907:
### Summary of changes

- Identified an issue where Recurring Payment (RP) fields (plan_type and plan_metadata) were not being appended during the App Switch flow due to an early return in BTPayPalRequest.parameters. The flow breaks starting from this [commit](https://github.com/braintree/braintree_ios/commit/abb4adc3570f471ba82b5c141275de938bb147fc#diff-12a919f5b0b117b30354b85383385d0c8885df582c4974647e7407f0804518cf) .
- Moved the logic for setting RP fields above the enableAppSwitch condition, ensuring they are included for both App Switch and non-App Switch flows.
- Verified that the RP fields are now correctly passed in the final parameters during App Switch.

### Checklist

- [x] Verified that RP fields are included in both App Switch and non-App Switch flows
- [x] Tested with **main** branch for RP without purchase

### Authors
@karthikeyan-2907

### DIFF
--- a/UnitTests/BraintreePayPalTests/BTPayPalVaultRequest_Tests.swift
+++ b/UnitTests/BraintreePayPalTests/BTPayPalVaultRequest_Tests.swift
@@ -166,4 +166,67 @@ class BTPayPalVaultRequest_Tests: XCTestCase {
         XCTAssertEqual(pricingScheme["price"], "test-price")
         XCTAssertEqual(pricingScheme["reload_threshold_amount"], "test-threshold")
     }
+    
+    func testParameters_enableAppSwitch_withRecurringBillingDetails_returnsAllParams() {
+        let billingPricing = BTPayPalBillingPricing(
+            pricingModel: .autoReload,
+            amount: "test-price",
+            reloadThresholdAmount: "test-threshold"
+        )
+        
+        let billingCycle = BTPayPalBillingCycle(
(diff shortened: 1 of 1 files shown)


Sources/BraintreePayPal/BTPayPalRequest.swift
```diff
@@ -218,6 +218,14 @@ import BraintreeCore
...
         parameters["cancel_url"] = BTCoreConstants.callbackURLScheme + "://\(BTPayPalRequest.callbackURLHostAndPath)cancel"
         parameters["experience_profile"] = experienceProfile
+        
+        if let recurringBillingPlanType {
+            parameters["plan_type"] = recurringBillingPlanType.rawValue
+        }
+
+        if let recurringBillingDetails {
+            parameters["plan_metadata"] = recurringBillingDetails.parameters()
+        }
```
richherrera, line 228: Just wondering if we can include a simple test for this, just to help prevent the issue this PR fixes from happening again.

Thanks

karthikeyan-2907: Hi @richherrera 
Added the UT for this change.

richherrera (approved):

stechiu (approved):

PR #1667, opened 2025-09-30
Update BTHTTP with `func data(for:)`

stechiu:
### Summary of changes

- Replace existing `dataTask(with request: URLRequest, completionHandler: @escaping @Sendable (Data?, URLResponse?, (any Error)?) -> Void) -> URLSessionDataTask` with the async counterpart: `func data(for request: URLRequest) async throws -> (Data, URLResponse)`
- 

### Checklist

- [ ] Added a changelog entry

- @stechiu

### DIFF
--- a/Sources/BraintreeCore/BTHTTP.swift
+++ b/Sources/BraintreeCore/BTHTTP.swift
@@ -69,12 +69,14 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
     // MARK: - HTTP Methods
 
     func get(_ path: String, configuration: BTConfiguration? = nil, parameters: Encodable? = nil, completion: @escaping RequestCompletion) {
-        do {
-            let dict = try parameters?.toDictionary()
-            
-            httpRequest(method: "GET", path: path, configuration: configuration, parameters: dict, completion: completion)
-        } catch let error {
-            completion(nil, nil, error)
+        Task {
+            do {
+                let dict = try parameters?.toDictionary()
+                let (json, response) = try await httpRequest(method: "GET", path: path, configuration: configuration, parameters: dict)
+                completion(json, response, nil)
+            } catch {
+                completion(nil, nil, error)
+            }
         }
     }
     
@@ -102,14 +104,20 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
         headers: [String: String]? = nil,
         completion: @escaping RequestCompletion
     ) {
-        httpRequest(
-            method: "POST",
-            path: path,
-            configuration: configuration,
-            parameters: parameters,
-            headers: headers,
-            completion: completion
-        )
+        Task {
+            do {
+                let (json, response) = try await httpRequest(
+                    method: "POST",
+                    path: path,
+                    configuration: configuration,
+                    parameters: parameters,
+                    headers: headers
+                )
+                completion(json, response, nil)
+            } catch {
+                completion(nil, nil, error)
+            }
+        }
     }
     
     func post(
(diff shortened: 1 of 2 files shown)


Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -151,30 +159,34 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
-        do {
-            let request = try createRequest(
-                method: method,
-                path: path,
-                configuration: configuration,
-                parameters: parameters,
-                headers: headers
-            )
+        headers: [String: String]? = nil
+    ) async throws -> (BTJSON?, HTTPURLResponse?) {
```
tdchow: Are we able to make `BTJSON` and `HTTPURLResponse` return types non null? They might have been optional before since we could return an Error in the completion handler.

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -151,30 +159,34 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
 
-            self.session.dataTask(with: request) { [weak self] data, response, error in
-                guard let self else {
-                    completion?(nil, nil, BTHTTPError.deallocated("BTHTTP"))
-                    return
-                }
+        let (data, response) = try await session.data(for: request)
+        let json: BTJSON = data.isEmpty ? BTJSON() : BTJSON(data: data)
 
-                handleRequestCompletion(data: data, request: request, response: response, error: error, completion: completion)
```
tdchow, line 172: It looks like we there's logic in `handleRequestCompletion()` that we'd still want to use. That function also looks like it's creating the `BTJSON` instance, so we may not need to do that in this function on line 173.

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -151,30 +159,34 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
+        }
+
+        guard let data = data as Data? else {
+            throw BTHTTPError.dataNotFound
+        }
+
+        if json.isError {
+            if let error = try await handleJSONResponseError(json: json, response: response) {
+                throw error
+            }
```
tdchow: This logic seems to be duplicated from `handleRequestCompletion()`. Can we update the signature of that function to remove the completion handler, and invoke that function instead of the duplication?

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -288,7 +300,7 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
         response: URLResponse?,
         error: Error?,
         completion: RequestCompletion?
-    ) {
+    ) async {
```
tdchow: I don't believe this function needs to be marked with `async`. There are no network calls happening in here, so it should be a synchronous function.

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -393,19 +405,18 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
 
     func handleJSONResponseError(
         json: BTJSON,
-        response: URLResponse,
-        completion: @escaping (Error?) -> Void
-    ) {
+        response: URLResponse
+    ) async throws -> Error? {
```
tdchow: Same for this function. There would be no async work, so we shouldn't need the async keyword.

We could also refactor this function to `throw` an error instead of having it as a return type. This would follow the async/await pattern a bit closer, since we'd eventually have to throw the returned error.

Sources/BraintreeCore/BTHTTP.swift
stechiu: Thanks, I see why it doesn't need to be marked as `async`. I'll update the function

Sources/BraintreeCore/BTHTTP.swift
stechiu: I have them as optional here since `handleRequestCompletion` function takes in optional parameters, including `Data`, `Response`, and `RequestCompletion`, so I assume there are still instances where they can be optional here, but I'm open to discussing otherwise

tdchow: I think we can make some changes to `handleRequestCompletion()` to return non null values. For the cases where these properties are null, we could throw an error. This will make the code less complex since we'd only return non null types.

Sources/BraintreeCore/BTHTTP.swift
stechiu: I made the updates in `handleRequestCompletion` and now this function returns `return try await handleRequestCompletion(data: data, request: request, response: httpResponse)`

Sources/BraintreeCore/BTHTTP.swift
stechiu: Gotcha, I'll make the update

Sources/BraintreeCore/BTHTTP.swift
stechiu: I've removed `let json: BTJSON = data.isEmpty ? BTJSON() : BTJSON(data: data)` and replaced with `return try await handleRequestCompletion(data: data, request: request, response: httpResponse)` at the end of the function

PR #1674, opened 2025-10-08
[v7] Update `BTVenmoLineItem` Parameters

agedd:
### Summary of changes

- Update `BTVenmoLineItem` properties to have everything in the init vs using dot syntax

### Checklist

- [x] Added a changelog entry

- @agedd

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Braintree iOS SDK Release Notes
 
-##unreleased 
+## unreleased 
+* BraintreePayPal
+  * Update `BTPayPalLineItem` to make all properties accessible on the initializer only vs via the dot syntax.   
 * BraintreeVenmo
   *  Update `BTVenmoLineItem` to make all properties accessible on the initializer only vs via the dot syntax.
 
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -149,10 +149,15 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         sender.setTitle("Processing...", for: .disabled)
         sender.isEnabled = false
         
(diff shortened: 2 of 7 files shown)


Sources/BraintreeVenmo/BTVenmoLineItem.swift
```diff
@@ -12,45 +12,46 @@ import Foundation
...
-    public init(quantity: Int, unitAmount: String, name: String, kind: BTVenmoLineItemKind) {
+    ///   - quantity: Required. Number of units of the item purchased. Can include up to 4 decimal places. This value must be a whole number and can't be negative or zero.
+    ///   - unitAmount: Required. Per-unit price of the item. Can include up to 2 decimal places. This value can't be negative or zero.
+    ///   - name: Required. Item name. Maximum 127 characters.
+    ///   - kind: Required. Indicates whether the line item is a debit (sale) or credit (refund) to the customer.
+    ///   - unitTaxAmount: Optional: Per-unit tax price of the item. Can include up to 2 decimal places. This value can't be negative or zero.
+    ///   - itemDescription: Optional: Item description. Maximum 127 characters.
+    ///   - productCode: Optional: Product or UPC code for the item. Maximum 127 characters.
+    ///   - url: Optional: The URL to product information.
+    public init(
```
jaxdesmarais, line 38: General question, have we been alphabetizing the required than non required params? I cannot remember what we decided on there or if we cared.

agedd: ah yeah, checked old PRs — we alphabetized events but not the params in init(:). imo might be cleaner to alphabetize all of them - i can make a ticket for it and add it to our backlog if we want to address this in a follow-up PR

jaxdesmarais: Ah, we don't need to then. Don't want to create more work than we need since we are targeting end of the month for release. We can leave as is.

PR #1675, opened 2025-10-08
update BTAnalyticsService test to pass offline

buzzamus:

### Summary of changes

- When running all unit tests offline all tests pass except `testSendAnalyticsEvent_whenMultipleSessionIDs_sendsMultiplePOSTs`. This seems to be due to the shared Singleton of BTAnalyticsService and a previously running unit test not sending the events it creates after the test runs. I initially tried re-working the test class to do teardown and more extensive setup, but due to the asynchronous calls this caused more issues. 
- 
- Instead, I created a straightforward solution of just sending an event to include any queued events and then capturing the value of the `currentPOSTRequestCount`. However, I created a more readable solution (second commit) that moves the reset logic to a separate helper function.

### Checklist

- [ ] Added a changelog entry

- buzzamus

### DIFF
--- a/UnitTests/BraintreeCoreTests/Analytics/BTAnalyticsService_Tests.swift
+++ b/UnitTests/BraintreeCoreTests/Analytics/BTAnalyticsService_Tests.swift
@@ -275,6 +275,7 @@ final class BTAnalyticsService_Tests: XCTestCase {
         return eventParams?[index]["event_name"] as? String
     }
     
+    /// Sends a dummy event to flush queued analytics and reset state before subsequent unit tests.
     private func resetAnalyticsState(analyticsService: BTAnalyticsService, analyticsHTTP: FakeHTTP) async {
         await analyticsService.performEventRequest(with: FPTIBatchData.Event(eventName: "clear-queue-event"))
         analyticsHTTP.POSTRequestCount = 0


UnitTests/BraintreeCoreTests/Analytics/BTAnalyticsService_Tests.swift
```diff
@@ -54,6 +54,8 @@ final class BTAnalyticsService_Tests: XCTestCase {
         sut.setAPIClient(stubAPIClient)
         sut.http = mockAnalyticsHTTP
         
+        await resetAnalyticsState(analyticsService: sut, analyticsHTTP: mockAnalyticsHTTP)
```
jaxdesmarais, line 57: Is it any specific event that isn't being reset or is it random depending on the order the tests are run in?

I also notice in this test we have `shouldBypassTimerQueue` after we send the first event but before the second. I wonder if that's intentional or is missing from another test which is causing the extra event to get stuck.

buzzamus: @jaxdesmarais  The other unit test that causes this issue in the failing test is `testSendAnalyticsEvent_whenConfigFetchCompletes_setsUpAnalyticsHTTPToUseBaseURL`. Whenever that test is skipped and doesn't run then the failing test passes. I tried adding `shouldBypassTimerQueue` into that test, but it didn't make a difference. The other option is to remove the `performEventRequest` call from the first test (`setsUpAnalyticsHTTPToUseBaseURL`)since the test isn't actually testing anything from that event. I also tried resolving the issue by using the same fakeHTTP mock in the first test as the other tests use it, but that causes the URL assertion to fail since the fakeHTTP has a hard coded url value.

As far as why this occurs only offline I am not sure. I suspect it may be due to the apiClient.fetchConfiguration() making a real network call and since the first test doesn't specify a mockHTTP that may be causing issues since this call fails.

tl;dr - `testSendAnalyticsEvent_whenConfigFetchCompletes_setsUpAnalyticsHTTPToUseBaseURL` not being set up the same way as the other tests is somehow causing an event to stay in the queue when offline and a more straightforward solution is to remove the `performEventRequest` line from that test if that is preferred since it doesn't care about the call. Another (probably better) solution would also be to either add the mockHTTP and allow a value to be injected into the fakeHTTP() initializer to  [override the default URL here](https://github.com/braintree/braintree_ios/blob/58755c0553e4ad97468d5b9d75f2e23afe5f2072/UnitTests/BraintreeTestShared/FakeHTTP.swift#L20)

jaxdesmarais: Ah, yeah you're right. Fetch config is being called in the `MockAPIClient`. When we initialize that API Client (even in the mock) we call `fetchOrReturnRemoteConfiguration` here: https://github.com/braintree/braintree_ios/blob/58755c0553e4ad97468d5b9d75f2e23afe5f2072/Sources/BraintreeCore/BTAPIClient.swift#L65

We could probably go down the rabbit hole of mocking that init and skipping the network call but that's likely overcomplicating things. I also assume you've tested setting `sut.shouldBypassTimerQueue = true` in the test `testSendAnalyticsEvent_whenConfigFetchCompletes_setsUpAnalyticsHTTPToUseBaseURL` to see if sending the event immediately clears the queue? We could also consider making `BTAnalyticsService.events` internal and removing all events in a teardown method in our tests. Not a blocker, just throwing out some ideas that you may have already thought through.

buzzamus: Yep I tried setting 'sut.shouldBypassTimerQueue = true', but it didn't make a difference (I think due to not using fakeHTTP, but not positive...that part stumped me a bit). I also tried creating better setup and teardown logic to clear events, but that actually caused more issues due to the async calls and led to more failures, but I didn't exhaust that path so I can look into that a bit more

jaxdesmarais: Gotcha - it doesn't seem to have to deal with using `FakeHTTP.fakeHTTP()`  instead it's an issue with `MockAPIClient` calling the method above in the init. If you add a breakpoint to that line it seems to be hit on all the tests. Not a blocker, was more just curious what was going on.

UnitTests/BraintreeCoreTests/Analytics/BTAnalyticsService_Tests.swift
```diff
@@ -272,4 +274,9 @@ final class BTAnalyticsService_Tests: XCTestCase {
         let eventParams = topLevelEvent?[0]["event_params"] as? [[String: Any]]
         return eventParams?[index]["event_name"] as? String
     }
+    
+    private func resetAnalyticsState(analyticsService: BTAnalyticsService, analyticsHTTP: FakeHTTP) async {
```
jaxdesmarais, line 279: Can we add some docstrings so we don't forget in the future why this was added?

UnitTests/BraintreeCoreTests/Analytics/BTAnalyticsService_Tests.swift
buzzamus: Ahh Yeah you are right I was conflating the fakeHTTP issue with another issue around the first test failing when making changes to use that

PR #1676, opened 2025-10-13
Update `BTSEPADirectDebitRequest` Parameters

agedd:
### Summary of changes

- Update `BTSEPADirectDebitRequest` properties to have everything in the init vs using dot syntax

**Note:** A `CHANGELOG` and `V7_MIGRATION` guide entry already exists for this change

### Checklist

- @agedd

### DIFF
--- a/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
+++ b/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
@@ -9,9 +9,9 @@ import BraintreeCore
 
     // MARK: - Internal Properties
     
-    let accountHolderName: String?
-    let iban: String?
-    let customerID: String?
+    let accountHolderName: String
+    let iban: String
+    let customerID: String
     let billingAddress: BTPostalAddress?
     let mandateType: BTSEPADirectDebitMandateType?
     let merchantAccountID: String?
@@ -19,9 +19,9 @@ import BraintreeCore
 
     /// Initialize a new SEPA Direct Debit request.
     /// - Parameters:
-    ///   - accountHolderName: Optional. The account holder name.
-    ///   - iban: Optional. The full IBAN.
-    ///   - customerID: Optional. The customer ID.
+    ///   - accountHolderName: Required. The account holder name.
+    ///   - iban: Required. The full IBAN.
+    ///   - customerID: Required. The customer ID.
     ///   - billingAddress: Optional. The user's billing address.
     ///   - mandateType: Optional. The `BTSEPADebitMandateType`. If not set, defaults to `.oneOff
     ///   - merchantAccountID: Optional. A non-default merchant account to use for tokenization.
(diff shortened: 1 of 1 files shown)


Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
```diff
@@ -7,54 +7,41 @@ import BraintreeCore
...
+    let iban: String
+    let customerID: String
+    let billingAddress: BTPostalAddress
+    let mandateType: BTSEPADirectDebitMandateType?
+    let merchantAccountID: String?
+    let locale: String?
 
     /// Initialize a new SEPA Direct Debit request.
     /// - Parameters:
     ///   - accountHolderName:Required. The account holder name.
```
jaxdesmarais:
```suggestion
    ///   - accountHolderName: Required. The account holder name.
```

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
```diff
@@ -7,54 +7,41 @@ import BraintreeCore
...
     ///   See https://developer.paypal.com/reference/locale-codes/ for a list of possible values.
     ///   Locale code should be supplied as a BCP-47 formatted locale code.
     public init(
-        accountHolderName: String? = nil,
-        iban: String? = nil,
-        customerID: String? = nil,
+        accountHolderName: String,
+        iban: String,
+        customerID: String,
+        billingAddress: BTPostalAddress,
```
jaxdesmarais: Is this required? It looks like it was optional before but hard to tell since the docstrings says it's required. Can we check on Android and Web?

agedd: ahh yep that's a good callout, checking on both [android](https://github.com/braintree/braintree_android/blob/6f411f3150dd53b2e7b36c91f8b49968fda1f6c5/SEPADirectDebit/src/main/java/com/braintreepayments/api/sepadirectdebit/SEPADirectDebitRequest.kt#L19) and [web](https://github.com/braintree/braintree-web/blob/65c4e14c28624283fd0cb8bebb92975aacbb2d78/src/sepa/external/sepa.js#L52), i noticed that all the properties are marked as optional, so updated in 96d7c7502c9b447ee47a6f0722f8572d7a33cac0 to reflect the same

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
```diff
@@ -7,36 +7,23 @@ import BraintreeCore
...
+    let locale: String?
 
     /// Initialize a new SEPA Direct Debit request.
     /// - Parameters:
-    ///   - accountHolderName:Required. The account holder name.
-    ///   - iban: Required. The full IBAN.
-    ///   - customerID: Required. The customer ID.
-    ///   - mandateType: Optional. The `BTSEPADebitMandateType`. If not set, defaults to `.oneOff`
-    ///   - billingAddress: Required. The user's billing address.
+    ///   - accountHolderName: Optional. The account holder name.
```
saperi22: Is`accountHolderName` required or optional?
If every field is optional, and the merchant sends us nothing, what does it mean? Should we validate that not everything is unset?

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
agedd: update: checked with the bt-sepa team and confirmed that `billingAddress` is optional and it seems the only required parameters that we pass are `accountHolderName`, `iban` and `customerID`

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitRequest.swift
agedd: ahh yep - thanks for the callout! double-checked with the sepa team and confirmed this is required. updated in 87db9936878528e20726060a6761a9933bff710c

PR #1678, opened 2025-10-14
Update Codescanning for `v7`

jaxdesmarais:
### Summary of changes

- Update codescanning for the iOS v7 branch to use the correct Xcode and Swift version to avoid older versions throwing an incorrect error

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -18,8 +18,9 @@ on:
 
 jobs:
   code-scanning:
-    uses: braintree/security-workflows/.github/workflows/codeql-ios.yml@main
+    uses: braintree/security-workflows/.github/workflows/codeql-ios.yml@addIOS7Support
     with:
+      mac-runner-version: 15
       project: Demo/Demo.xcodeproj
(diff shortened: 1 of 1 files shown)


.github/workflows/security.yml
```diff
@@ -18,11 +18,12 @@ on:
 
 jobs:
   code-scanning:
-    uses: braintree/security-workflows/.github/workflows/codeql-ios.yml@main
+    uses: braintree/security-workflows/.github/workflows/codeql-ios.yml@addIOS7Support
```
tdchow, line 21: Is `addIOS7Support` a branch in a repo that contains `codeql-ios.yml`? Would we need to switch this back to main at some point?

JLESUS: Exactly, this points to our branch with the fix in our shared-workflows repo. Once validated we will merge that branch to main then we can just revert back to @main. 😄

tdchow: Great, thanks for the confirmation!

PR #1680, opened 2025-10-15
[v7] Update `BTPostalAddress` Parameters

agedd:
### Summary of changes

- Update `BTPostalAddress` Parameters

### Checklist

- [x] Added a changelog entry

- @agedd

### DIFF
--- a/Sources/BraintreeCore/BTJSON.swift
+++ b/Sources/BraintreeCore/BTJSON.swift
@@ -267,22 +267,14 @@ public class BTJSON: NSObject {
     public func asAddress() -> BTPostalAddress? {
         guard self.isObject else { return nil }
         
-        let recipientName = self["recipientName"].asString() ?? self["fullName"].asString()
-        let streetAddress = self["street1"].asString() ?? self["line1"].asString() ?? self["addressLine1"].asString()
-        let extendedAddress = self["street2"].asString() ?? self["line2"].asString() ?? self["addressLine2"].asString()
-        let locality = self["city"].asString() ?? self["adminArea2"].asString()
-        let region = self["state"].asString() ?? self["adminArea1"].asString()
-        let postalCode = self["postalCode"].asString()
-        let countryCodeAlpha2 = self["country"].asString() ?? self["countryCode"].asString()
-        
         return BTPostalAddress(
-            recipientName: recipientName,
-            streetAddress: streetAddress,
-            extendedAddress: extendedAddress,
-            locality: locality,
-            countryCodeAlpha2: countryCodeAlpha2,
-            postalCode: postalCode,
-            region: region
+            recipientName: self["recipientName"].asString() ?? self["fullName"].asString(),
+            streetAddress: self["street1"].asString() ?? self["line1"].asString() ?? self["addressLine1"].asString(),
+            extendedAddress: self["street2"].asString() ?? self["line2"].asString() ?? self["addressLine2"].asString(),
+            locality: self["city"].asString() ?? self["adminArea2"].asString(),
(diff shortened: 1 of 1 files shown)


Sources/BraintreeCore/BTPostalAddress.swift
```diff
@@ -37,4 +53,33 @@ import Foundation
         case recipientName = "recipient_name"
         case streetAddress = "line1"
     }
+    
+    // MARK: - Helper Method
+    
+    /// Returns address components as a dictionary for accessing internal properties across modules
+    public func addressComponents() -> [String: String] {
+        var components: [String: String] = [:]
+        if let recipientName = recipientName {
```
jaxdesmarais: nit, for all of these we can do:

```suggestion
        if let recipientName {
```

Sources/BraintreeCore/BTPostalAddress.swift
```diff
@@ -37,4 +53,33 @@ import Foundation
         case recipientName = "recipient_name"
         case streetAddress = "line1"
     }
+    
+    // MARK: - Helper Method
+    
+    /// Returns address components as a dictionary for accessing internal properties across modules
+    public func addressComponents() -> [String: String] {
```
jaxdesmarais, line 60: 🤔 I wonder if we could do something like this here instead for the whole method implementation:
```swift
[
          "recipientName": recipientName,
          "streetAddress": streetAddress,
          "extendedAddress": extendedAddress,
          "locality": locality,
          "countryCodeAlpha2": countryCodeAlpha2,
          "postalCode": postalCode,
          "region": region
      ].compactMapValues { $0 }
```

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
   * Update `BTPayPalLineItem` to make all properties accessible on the initializer only vs via the dot syntax.   
 * BraintreeVenmo
   *  Update `BTVenmoLineItem` to make all properties accessible on the initializer only vs via the dot syntax.
+* BraintreeCore
```
saperi22, line 8: nit: do we care about alphabetical order here?
`BraintreeCore` before `BraintreePayPal`?

Sources/BraintreeCore/BTPostalAddress.swift
agedd: ahh yep this reads much cleaner, ty updated in a2cc2fad0eee0b47aad211f0f62bae943439b5b9

Sources/BraintreeCore/BTJSON.swift
```diff
@@ -267,14 +267,22 @@ public class BTJSON: NSObject {
...
+        let recipientName = self["recipientName"].asString() ?? self["fullName"].asString()
+        let streetAddress = self["street1"].asString() ?? self["line1"].asString() ?? self["addressLine1"].asString()
+        let extendedAddress = self["street2"].asString() ?? self["line2"].asString() ?? self["addressLine2"].asString()
+        let locality = self["city"].asString() ?? self["adminArea2"].asString()
+        let region = self["state"].asString() ?? self["adminArea1"].asString()
+        let postalCode = self["postalCode"].asString()
+        let countryCodeAlpha2 = self["country"].asString() ?? self["countryCode"].asString()
+        
+        return BTPostalAddress(
+            recipientName: recipientName,
```
jaxdesmarais: not a blocker, very much a  take it or leave it. For these instead of making them all stored properties we could also just this inline:

```suggestion
            recipientName: self["recipientName"].asString() ?? self["fullName"].asString(),
```

CHANGELOG.md
agedd: yeah it seems like order doesn't really matter here (based on previous entries too)

PR #1687, opened 2025-11-04
Update README, CHANGELOG, and Fix Docs Release

jaxdesmarais:
### Summary of changes

- Update package.resolved
    - Fixes docs release failure: `Your local changes to the following files would be overwritten by checkout: Braintree.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved`
    - Based on PR where we needed to fix this previously: https://github.com/braintree/braintree_ios/pull/1303
- Add `Note` to v7 CHANGELOG
- Update README `Versions` section
- Remove `V6_MIGRATION` from main (it will continue to exist on the `6.x` branch

### Checklist

- [x] Added a changelog entry

### Authors

- @jaxdesmarais

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 * BraintreeLocalPayment
   * Rename `BTLocalPaymentClient.startPaymentFlow(with:completion:)` to `BTLocalPaymentClient.start(with:completion:)`
 
+**Note:** Includes all changes in [7.0.0-beta1](#700-beta1-2025-04-30) and [7.0.0-beta2](#700-beta2-2025-07-29)
+
 ## 6.39.0 (2025-10-01)
 * BraintreeCore
   * Fix crash related to data race in `BTWebAuthenticationSession` (fixes #1653)


Braintree.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -8,16 +9,7 @@
...
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/paypal/paypalcheckout-ios/",
-      "state" : {
-        "revision" : "da1b998dcab91011b05df139a6819cc58dd45ccd",
-        "version" : "1.3.0"
-      }
     }
   ],
-  "version" : 2
+  "version" : 3
```
tdchow, line 14: For my learning - what is this file used for? Does it contain all of the project's transitive dependencies for documentation purposes? Is it also managed by xcode (we don't manually make changes to it)?

jaxdesmarais: That's a great question. This file is used as similar package.lock type file to lock in our dependency versions for the project. It doesn't need updating often, but since we removed MXO we need to remove it here. For us, we work out of the `xcworkspace` since we use cocoapods locally which does not use this file since it's for SPM. It was not updated manually but instead is updated automatically when the `xcodeproj` is opened.

Essentially what CI is doing is making this change when it uses the `xcodeproj` as part of the docs release process. Then it's failing because this file does not match what's on main since it wants these changes to exist but they don't. I linked a previous PR where we had to do this as well but essentially is an out of site out of mind since we use the cocoapods based `xcworkspace` for development vs the `xcodeproj`.

Issue #1671, opened 2025-10-06
`BTConfiguration.canMakeApplePayPayments` and `applePaySupportedNetworks `are inaccessible (internal) in v6


Smertowing:
### Braintree SDK Version

6.39.0

### Environment

Both

### Xcode Version

16.4

### Integration type

Prebuilt Binaries / Other

### Development Processor

Both

### Describe the bug

While migrating our app from an older Objective-C implementation to v6 of the SDK with Swift, I discovered that `BTConfiguration.canMakeApplePayPayments` is now `internal`. This prevents us from checking whether Apple Pay is configured and available for the current merchant account and device.

A potential workaround using Apple's native `PKPaymentAuthorizationViewController.canMakePayments(usingNetworks:)` directly is also blocked because the required `BTConfiguration.applePaySupportedNetworks` property is also `internal`. This is a regression from Objective-C versions where these properties were `public`.

We were using this property to determine if the user's Apple Pay method is supported by Braintree for our merchant account, and we want to avoid hardcoding payment networks on the client side. This leads to two main questions:

1. **Was this an intentional API change in v6?** I noticed `canMakeApplePayPayments` does not appear to be used anywhere within the SDK itself, which led me to wonder if its removal from the public API was an oversight.

2. **Is it safe to use a local patch?** As a temporary workaround, I'm considering forking the repository and changing the protection level of `canMakeApplePayPayments` to `public`. Are there any potential risks or side effects we should be aware of with this approach? 

### Steps to reproduce the behavior:
- Integrate Braintree iOS SDK v6.x into a Swift project.
- Attempt to access `configuration.canMakeApplePayPayments`.
- The compiler will produce the error: `'canMakeApplePayPayments' is inaccessible due to 'internal' protection level`.

### Example Code:

```
import Braintree

// ...

let braintreeApiClient = BTApplePayClient(apiClient:)

braintreeApiClient.fetchOrReturnRemoteConfiguration { (configuration, error) in
    guard let configuration = configuration, error == nil else {
        // Handle error
        return
    }

    // 🛑 This line fails to compile in v6
    let isApplePayAvailable = configuration.canMakeApplePayPayments
    
    // We use this value to decide whether we have an Apple Pay option available or not, and render the UI accordingly
}
```

### Expected behavior

[...]

PR #1688, opened 2025-11-06
Apple Pay - Expose `canMakeApplePayPayments` property

stechiu:
This PR addresses a [GH issue](https://github.com/braintree/braintree_ios/issues/1671) that exposes the internal property `canMakeApplePayPayments` property so that a Merchant can check whether Apple Pay is configured and available for the current merchant account and device.

### Summary of changes

- Created `isApplePaySupported()` method in `BTApplePayClient`
- Updated unit tests

### Checklist

- [ ] Added a changelog entry

- @stechiu

### DIFF
--- a/Sources/BraintreeApplePay/BTConfiguration+ApplePay.swift
+++ b/Sources/BraintreeApplePay/BTConfiguration+ApplePay.swift
@@ -17,7 +17,7 @@ extension BTConfiguration {
 
     /// Indicates if the Apple Pay merchant enabled payment networks are supported on this device.
     var canMakeApplePayPayments: Bool {
-        guard let applePaySupportedNetworks else { return false }
+        guard let applePaySupportedNetworks, !applePaySupportedNetworks.isEmpty else { return false }
         return PKPaymentAuthorizationViewController.canMakePayments(usingNetworks: applePaySupportedNetworks)
     }
 
@@ -40,13 +40,9 @@ extension BTConfiguration {
     var applePaySupportedNetworks: [PKPaymentNetwork]? {
         let gatewaySupportedNetworks: [String]? = json?["applePay"]["supportedNetworks"].asStringArray()
 
-        guard let networks = gatewaySupportedNetworks, !networks.isEmpty else {
-            return nil
-        }
-
         var supportedNetworks: [PKPaymentNetwork] = []
 
-        networks.forEach { gatewaySupportedNetwork in
+        gatewaySupportedNetworks?.forEach { gatewaySupportedNetwork in
             if gatewaySupportedNetwork.localizedCaseInsensitiveCompare("visa") == .orderedSame {
                 supportedNetworks.append(.visa)
             } else if gatewaySupportedNetwork.localizedCaseInsensitiveCompare("mastercard") == .orderedSame {
@@ -62,6 +58,6 @@ extension BTConfiguration {
             }
         }
 
-        return supportedNetworks.isEmpty ? nil : supportedNetworks
+        return supportedNetworks
     }
 }
--- a/UnitTests/BraintreeApplePayTests/BTConfiguration+ApplePay_Tests.swift
+++ b/UnitTests/BraintreeApplePayTests/BTConfiguration+ApplePay_Tests.swift
@@ -96,6 +96,6 @@ class BTConfiguration_ApplePay_Tests : XCTestCase {
         ])
         let configuration = BTConfiguration(json: configurationJSON)
 
-        XCTAssertEqual(configuration.applePaySupportedNetworks, nil)
+        XCTAssertEqual(configuration.applePaySupportedNetworks!, [])
     }
 }


Demo/Application/Features/ApplePayViewController.swift
```diff
@@ -28,7 +28,7 @@ class ApplePayViewController: PaymentButtonBaseViewController {
         return applePayButton
     }
 
-    @objc func tappedApplePayButton() {
+    @objc func tappedApplePayButton() async {
```
saralvasquez: I feel like I missed something. Why does this have to change to `async`?

saralvasquez (approved):
I'm not an iOS dev but this looks good to me! 🙂 Just had a question for my info

Demo/Application/Features/ApplePayViewController.swift
jaxdesmarais: Do we instead want to call the new method we are exposing here?

Demo/Application/Features/ApplePayViewController.swift
jaxdesmarais: ➕ 1 - I had the same question

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -69,6 +69,19 @@ import BraintreeCore
         }
     }
 
+    /// Checks if the Apple Pay merchant enabled payment networks are supported on the user's device.
+    /// - Returns: A `Bool`
+    public func isApplePaySupported() async -> Bool {
```
jaxdesmarais: Does this need to be async?

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -69,6 +69,19 @@ import BraintreeCore
         }
     }
 
+    /// Checks if the Apple Pay merchant enabled payment networks are supported on the user's device.
+    /// - Returns: A `Bool`
```
jaxdesmarais: small nit:

```suggestion
    /// - Returns: A `Bool` that returns true if Apple Pay is supported for the customer
```

Sources/BraintreeApplePay/BTApplePayClient.swift
jaxdesmarais, line 84: Wondering if we also want to expose a completion handler vs just the async method. Since this module exposes both elsewhere. What do you think?

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -69,6 +69,19 @@ import BraintreeCore
...
+    public func isApplePaySupported() async -> Bool {
+        await withCheckedContinuation { continuation in
+            apiClient.fetchOrReturnRemoteConfiguration { configuration, _ in
+                let isConfigEnabled = configuration?.isApplePayEnabled ?? false
```
jaxdesmarais: We check the config elsewhere and this check is more for if Apple Pay is configured and available for the current merchant account and device. I think we can just use `canMakeApplePayPayments` in this method.

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -69,6 +69,19 @@ import BraintreeCore
         }
     }
 
+    /// Checks if the Apple Pay merchant enabled payment networks are supported on the user's device.
```
jaxdesmarais:
```suggestion
    /// Checks if Apple Pay is configured and available for the current merchant account and device.
```

UnitTests/BraintreeApplePayTests/BTApplePay_Tests.swift
```diff
@@ -14,6 +14,40 @@ class BTApplePay_Tests: XCTestCase {
 
     // MARK: - Payment Request
 
+    func testCanMakeApplePayPayments_ReturnsTrueWhenEnabled() async {
```
jaxdesmarais:
```suggestion
    func testIsApplePaySupported_ReturnsTrueWhenEnabled() async {
```

UnitTests/BraintreeApplePayTests/BTApplePay_Tests.swift
```diff
@@ -14,6 +14,40 @@ class BTApplePay_Tests: XCTestCase {
...
+            ] as [String: Any]
+        ])
+
+        applePayClient.apiClient = mockClient
+
+        let result = await applePayClient.isApplePaySupported()
+        XCTAssertTrue(result)
+    }
+
+    func testCanMakeApplePayPayments_ReturnsFalseWhenDisabled() async {
```
jaxdesmarais:
```suggestion
    func testIsApplePaySupported_ReturnsFalseWhenDisabled() async {
```

Sources/BraintreeApplePay/BTApplePayClient.swift
stechiu: Added, thanks!

Demo/Application/Features/ApplePayViewController.swift
stechiu: Thanks, `async` shouldn't be here. I was testing code that required using `async` and forgot to move it earlier when I removed the test code

Sources/BraintreeApplePay/BTApplePayClient.swift
stechiu: ah I see, like how `makePaymentRequest` exposes the completion handler. I'll make the update

Sources/BraintreeApplePay/BTApplePayClient.swift
stechiu: Gotcha, I made the updates so that it's only `canMakeApplePayPayments`. I was trying to do that before but unit test kept failing when no supported networks were provided, so I updated `applePaySupportedNetworks` in `BTConfiguration+ApplePay` to handle `nil` instances.

Sources/BraintreeApplePay/BTConfiguration+ApplePay.swift
```diff
@@ -39,9 +39,14 @@ extension BTConfiguration {
     /// The Apple Pay payment networks supported by your Braintree merchant account.
     var applePaySupportedNetworks: [PKPaymentNetwork]? {
         let gatewaySupportedNetworks: [String]? = json?["applePay"]["supportedNetworks"].asStringArray()
+
+        guard let networks = gatewaySupportedNetworks, !networks.isEmpty else {
+            return nil
+        }
```
jaxdesmarais: Do we want to return nil if this is empty? Or was this returning an error in our testing. Just curious because on line 47 it looks like we'd just return an empty array for supportedNetworks but I might be missing something.

Sources/BraintreeApplePay/BTConfiguration+ApplePay.swift
```diff
@@ -57,6 +62,6 @@ extension BTConfiguration {
             }
         }
 
-        return supportedNetworks
+        return supportedNetworks.isEmpty ? [] : supportedNetworks
```
jaxdesmarais: It looks like we set `supportedNetworks` to `[]` earlier. Do we need this logic or should that catch this case?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreeApplePay
+  * Expose `BTConfiguration.canMakeApplePayPayments` publicly using `BTApplePayClient.isApplePaySupported(completion:)`
```
```suggestion
  * Expose `BTConfiguration.canMakeApplePayPayments` publicly using `BTApplePayClient.isApplePaySupported(completion:) or `BTApplePayClient.isApplePaySupported() async -> Bool` (fixes #1671)`
```

stechiu: Thanks!

Sources/BraintreeApplePay/BTConfiguration+ApplePay.swift
stechiu: The latter, where the test `testIsApplePaySupported_ReturnsFalseWhenDisabled` was erroring out when asserting that `isApplePaySupported` is false. I believe why it failed when returning `[]` instead of `nil` when `networks.isEmpty` is because `[]` could mean Apple Pay is configured with no networks. On the contrary, `nil` would indicate the Apple Pay configuration is unavailable.

The guard statement in `canMakeApplePayPayments` (`guard let applePaySupportedNetworks else { return false }`) checks for `nil` which indicates that Apple Pay is unavailable.

I wrote a new test `testIsApplePaySupported_ReturnsFalseWhenNetworkIsNotSupported` where a non-supported network is used and `isApplePaySupported` returns `false` like how we'd want it to.

Sources/BraintreeApplePay/BTConfiguration+ApplePay.swift
stechiu: After more testing on my end, I've updated it to `return supportedNetworks.isEmpty ? nil : supportedNetworks`. We'll need this logic to check whether the `gatewaySupportedNetworks` are in the list of networks that we support. If not, then `nil` ensures that `canMakeApplePayPayments` returns `false`

Sources/BraintreeApplePay/BTConfiguration+ApplePay.swift
jaxdesmarais: Gotcha, thanks for the explanation. I am a bit hesitant to change the logic here since we use this method elsewhere in the SDK. What do you think about reverting this methods changes and instead doing a check in `canMakeApplePayPayments` that the result from here is not empty instead? That way we aren't accidentally impacting other parts of the SDK with the changes here.

agedd: +1 - i think reverting these changes (and the guard check on line 43) might make things cleaner - wdyt?

UnitTests/BraintreeApplePayTests/BTConfiguration+ApplePay_Tests.swift
```diff
@@ -96,6 +96,6 @@ class BTConfiguration_ApplePay_Tests : XCTestCase {
         ])
         let configuration = BTConfiguration(json: configurationJSON)
 
-        XCTAssertEqual(configuration.applePaySupportedNetworks!, [])
+        XCTAssertEqual(configuration.applePaySupportedNetworks, nil)
```
agedd: nit: can probably revert this as well

Sources/BraintreeApplePay/BTConfiguration+ApplePay.swift
stechiu: Since `applePaySupportedNetworks` does get used elsewhere, it makes sense to revert. I'll make the updates and check that `appPaySupportedNetworks` isn't empty

jaxdesmarais (approved):
🎉  thanks for addressing feedback and helping get this fix out for our merchants!

PR #1689, opened 2025-11-08
Create Shell for UI Components

stechiu:
### Summary of changes

- Added new UI Components module to the SDK
- Added new UI Components Tests target to the SDK
- Added UI Components module to all Package managers - SPM, Carthage, Cocoapods

### Checklist

- [x] Added a changelog entry

- @stechiu

### DIFF
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -15,6 +15,8 @@
 		045241652EBED39400C25EA6 /* BraintreeCore.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 570B93AC285397520041BAFE /* BraintreeCore.framework */; platformFilter = ios; };
 		0452416B2EC4F80F00C25EA6 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 045241692EC4F80F00C25EA6 /* PrivacyInfo.xcprivacy */; };
 		0452416C2EC4F80F00C25EA6 /* BTPaymentButtons.swift in Sources */ = {isa = PBXBuildFile; fileRef = 045241682EC4F80F00C25EA6 /* BTPaymentButtons.swift */; };
+		045241762EC7F14D00C25EA6 /* BraintreePayPal.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 2D941D381B59C76A0016EFB4 /* BraintreePayPal.framework */; platformFilter = ios; };
+		045241772EC7F15400C25EA6 /* BraintreeVenmo.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = A77AA29C1B618C7700217B73 /* BraintreeVenmo.framework */; platformFilter = ios; };
 		04AA31182D07974D0043ACAB /* BTButtonType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 04AA31172D0797460043ACAB /* BTButtonType.swift */; };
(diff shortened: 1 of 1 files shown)


Braintree.podspec
```diff
@@ -99,4 +99,9 @@ Pod::Spec.new do |s|
     s.resource_bundle = { "BraintreeVenmo_PrivacyInfo" => "Sources/BraintreeVenmo/PrivacyInfo.xcprivacy" }
   end
 
+  s.subspec "UIComponents" do |s|
+    s.source_files = "Sources/BraintreeUIComponents/*.swift"
+    s.dependency "Braintree/Core"
```
jaxdesmarais, line 104: We will also want to add the PayPal and Venmo dependencies

Package.swift
```diff
@@ -137,6 +141,11 @@ let package = Package(
         .binaryTarget(
             name: "PPRiskMagnes",
             path: "Frameworks/XCFrameworks/PPRiskMagnes.xcframework"
+        ),
+        .target(
+            name: "BraintreeUIComponents",
+            dependencies: ["BraintreeCore"],
```
jaxdesmarais: Same here, will want to add Venmo and PayPal as dependencies.

jaxdesmarais (commented):
For the carthage failure, it's likely an issue with one of the settings on the schemes. You should be able to run that command locally to take a look at the output log to resolve if it doesn't immediately jump out in the build settings.

Braintree.podspec
stechiu: Added the dependencies

Specs/BraintreeUIComponents/Supporting Files/BraintreeUIComponents-Info.plist
```diff
@@ -0,0 +1,14 @@
...
+    <key>CFBundleIdentifier</key>
+    <string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+    <key>CFBundleName</key>
+    <string>BraintreeUIComponents</string>
+    <key>CFBundleVersion</key>
+    <string>$(CURRENT_PROJECT_VERSION)</string>
+    <key>CFBundleShortVersionString</key>
+    <string>$(MARKETING_VERSION)</string>
+</dict>
+</plist>
```
stechiu: Carthage kept failing because it was looking for an Info.plist file in this specific folder structure, so I added this file

jaxdesmarais: It looks like the individual modules point to the `Info.plist` in the core module vs having their own. We should likely do the same for this module.

stechiu: I see, I'll remove this file and try that

Braintree.xcodeproj/project.pbxproj
```diff
@@ -1127,6 +1175,23 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
+		045241162EBEBAC100C25EA6 /* Frameworks */ = {
+			isa = PBXFrameworksBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
+				045241302EBEBAD200C25EA6 /* BraintreeCore.framework in Frameworks */,
```
jaxdesmarais, line 1186: Can we add PayPal and Venmo as frameworks here too?

jaxdesmarais (commented):
Just a minor comment to add the PayPal and Venmo dependency to the framework and dependency target for this module as well.

Braintree.xcodeproj/project.pbxproj
stechiu: Done

agedd (approved):

PR #1673, opened 2025-10-08
Added fallback URL scheme for merchants

baubrey91:
Thank you for your contribution to Braintree. 

> Before submitting this PR, note we cannot accept language translation PRs. We support the same languages that are supported by PayPal, and have a dedicated localization team to provide the translations. If there is an error in a specific translation, you may open an issue and we will escalate it to the localization team.

### Summary of changes
We will give merchants a field to add a uri scheme for paypal to use as a backup if deeplink to merchant fails.
Changes made include 
 1. Adding the new parameter
 2. Sending up the value
 3. Checking if uri scheme value in BT validation on return.

- 

### Checklist

- [x] Added a changelog entry
- [ ] Tested and confirmed payment flows affected by this change are functioning as expected

-

PR #1690, opened 2025-11-10
[V6.x] Add fallbackURLScheme support to PayPal app switch flow 

ndrsKloss:
### Summary of changes

We will give merchants a field to add a uri scheme for paypal to use as a backup if deeplink to merchant fails.
Changes made include

1. Adding the new parameter
2. Sending up the value
3. Checking if uri scheme value in BT validation on return.

Original PR: https://github.com/braintree/braintree_ios/pull/1673

## Videos

Setup: XOSphere with SPM pointing to my branch

Fallback with universal link activated:
https://github.com/user-attachments/assets/fe111b18-98d9-42e4-b201-6ee8078bef65

Fallback with universal link deactivated:
https://github.com/user-attachments/assets/7b95cd48-307d-4478-8478-f4b64fa40bff

### Checklist

- [x] Added a changelog entry

- @ndrsKloss

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
+++ b/Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
@@ -171,7 +171,12 @@ import BraintreeCore
         fallbackURLScheme: String? = nil,
         isPayPalAppInstalled: Bool = false
     ) -> [String: Any] {
-        var baseParameters = super.parameters(with: configuration, universalLink: universalLink, isPayPalAppInstalled: isPayPalAppInstalled)
+        var baseParameters = super.parameters(
+            with: configuration,
+            universalLink: universalLink,
+            fallbackURLScheme: fallbackURLScheme,
+            isPayPalAppInstalled: isPayPalAppInstalled
+        )
         
         var checkoutParameters: [String: Any] = [
             "intent": intent.stringValue,
--- a/UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift
+++ b/UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift
@@ -284,6 +284,31 @@ class BTPayPalCheckoutRequest_Tests: XCTestCase {
         XCTAssertTrue((parameters["os_type"] as! String).matches("iOS|iPadOS"))
         XCTAssertEqual(parameters["merchant_app_return_url"] as? String, "some-url")
     }
+    
+    func testParametersWithConfiguration_whenFallbackURLSchemeProvided_includesFallbackURLScheme() {
+        let request = BTPayPalCheckoutRequest(enablePayPalAppSwitch: true, amount: "1")
+        
+        let parameters = request.parameters(
+            with: configuration,
+            universalLink: URL(string: "https://example.com")!,
+            fallbackURLScheme: "com.my-app.payments",
+            isPayPalAppInstalled: true
+        )
+        
+        XCTAssertEqual(parameters["merchant_app_fallback_url_scheme"] as? String, "com.my-app.payments")
+    }
+    
+    func testParametersWithConfiguration_whenFallbackURLSchemeNotProvided_doesNotIncludeFallbackURLScheme() {
+        let request = BTPayPalCheckoutRequest(enablePayPalAppSwitch: true, amount: "1")
+        
+        let parameters = request.parameters(
+            with: configuration,
+            universalLink: URL(string: "https://example.com")!,
+            isPayPalAppInstalled: true
+        )
+        
+        XCTAssertNil(parameters["merchant_app_fallback_url_scheme"])
+    }
 
     func testCreateRequestBody_setsAmountBreakdown() {
         let amountBreakdown = BTAmountBreakdown(


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -102,13 +104,14 @@ import BraintreeDataCollector
     ///   - apiClient: The API Client
     ///   - universalLink: The URL to use for the PayPal app switch flow. Must be a valid HTTPS URL dedicated to Braintree app switch returns. This URL must be allow-listed in your Braintree Control Panel.
     /// - Warning: This initializer should be used for merchants using the PayPal App Switch flow. This feature is currently in beta and may change or be removed in future releases.
-    @objc(initWithAPIClient:universalLink:)
-    public convenience init(apiClient: BTAPIClient, universalLink: URL) {
+    @objc(initWithAPIClient:universalLink:fallbackURLScheme:)
+    public convenience init(apiClient: BTAPIClient, universalLink: URL, fallbackURLScheme: String? = nil) {
```
jaxdesmarais, line 109: We are missing a docstring in the init for this param

Sources/BraintreePayPal/BTPayPalReturnURL.swift
```diff
@@ -38,11 +38,11 @@ struct BTPayPalReturnURL {
     /// Evaluates whether the url represents a valid PayPal return URL.
     /// - Parameter url: an app switch or ASWebAuthenticationSession return URL
     /// - Returns: `true` if the url represents a valid PayPal app switch return
-    static func isValid(_ url: URL) -> Bool {
-        let isHTTPSScheme = url.scheme == "https"
-        let containsAppSwitchPath = url.path.contains("braintreeAppSwitchPayPal")
+    static func isValid(_ url: URL, fallbackURLScheme: String?) -> Bool {
+        let isValidScheme = ["https", fallbackURLScheme].contains(url.scheme)
+        let containsAppSwitchPath = url.path.contains("braintreeAppSwitchPayPal") || fallbackURLScheme == url.scheme
```
jaxdesmarais: This property is no longer doing what we say it does. We should have 2 sections here 1 for checking the `fallbackURLScheme` values and 1 for checking the HTTPS values. That way our properties can be properly named and we can do the checks as expected for each.

CHANGELOG.md
```diff
@@ -12,6 +16,7 @@
   * Add `userAction` property to `BTPayPalVaultRequest`
   * Add `paypal:tokenize:default-browser-switch:started` event for when the appSwitch fails for universalLink and `openURLInDefaultBrowser` is triggered.
   * Add `paypal:tokenize:default-browser-switch:succeeded` and `paypal:tokenize:default-browser-switch:failed` events for when the browser switch succeeds or fails.
+  * Add `fallbackURLScheme` to `BTPayPalClient` initializer for merchants to have fallback return url if the universal link fails
```
jaxdesmarais: I think we can remove this one

Sources/BraintreePayPal/BTPayPalReturnURL.swift
ndrsKloss: Good point! Refactored to separate the HTTPS universal link checks from the fallback URL scheme checks, with properly named variables for each section.

CHANGELOG.md
ndrsKloss: Done

Sources/BraintreePayPal/BTPayPalClient.swift
ndrsKloss: Done

UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift
```diff
@@ -276,7 +276,7 @@ class BTPayPalCheckoutRequest_Tests: XCTestCase {
     func testParametersWithConfiguration_setsAppSwitchParameters_WithoutUserAuthenticationEmail() {
         let request = BTPayPalCheckoutRequest(enablePayPalAppSwitch: true, amount: "1")
         
-        let parameters = request.parameters(with: configuration, universalLink: URL(string: "some-url")!, isPayPalAppInstalled: true)
+        let parameters = request.parameters(with: configuration, universalLink: URL(string: "some-url")!, fallbackURLScheme: "paypal", isPayPalAppInstalled: true)
```
jaxdesmarais: Do we need any new tests in this file similar to the Vault tests?

ndrsKloss: Hey Jax, we don't need, fallbackURLScheme is only for Vault flows per the doc, so no new tests needed in Checkout.

jaxdesmarais: It's added to the checkout request and as this PR is written will be available to both Checkout and Vault flows. Additionally there were tests for this added in the v7 PR. Lets check on this since as both PRs are written it can be used for both request types.

ndrsKloss: Removed the additional paramter, but we can't remove it from the method signature because it overrides `BTPayPalRequest.parameters()`, which is shared by both Vault and Checkout requests.

Sources/BraintreePayPal/BTPayPalReturnURL.swift
```diff
@@ -36,15 +36,23 @@ struct BTPayPalReturnURL {
...
-    /// - Parameter url: an app switch or ASWebAuthenticationSession return URL
+    /// - Parameters:
+    ///   - url: an app switch or ASWebAuthenticationSession return URL
+    ///   - fallbackURLScheme: Optional fallback URL scheme to check against if universal link check fails
     /// - Returns: `true` if the url represents a valid PayPal app switch return
-    static func isValid(_ url: URL) -> Bool {
-        let isHTTPSScheme = url.scheme == "https"
+    static func isValid(_ url: URL, fallbackURLScheme: String? = nil) -> Bool {
+        let containsExpectedAction = url.path.contains("cancel") || url.path.contains("success")
         let containsAppSwitchPath = url.path.contains("braintreeAppSwitchPayPal")
```
jaxdesmarais, line 45: Will the PayPal API be appending this portion of the path for the fallback URL use case? Right now this is appended by the SDK only for universal links unless we intended to update URL fallbacks to do the same. Right now I don't think this will pass for merchants unless we are appending this for them upstream.

UnitTests/BraintreePayPalTests/BTPayPalVaultRequest_Tests.swift
```diff
@@ -104,12 +104,13 @@ class BTPayPalVaultRequest_Tests: XCTestCase {
...
         )
 
-        let parameters = request.parameters(with: configuration, universalLink: URL(string: "some-url")!, isPayPalAppInstalled: true)
+        let parameters = request.parameters(with: configuration, universalLink: URL(string: "some-url")!, fallbackURLScheme: "myapp", isPayPalAppInstalled: true)
 
         XCTAssertEqual(parameters["launch_paypal_app"] as? Bool, true)
         XCTAssertTrue((parameters["os_version"] as! String).matches("\\d+\\.\\d+"))
         XCTAssertTrue((parameters["os_type"] as! String).matches("iOS|iPadOS"))
         XCTAssertEqual(parameters["merchant_app_return_url"] as? String, "some-url")
+        XCTAssertEqual(parameters["merchant_app_fallback_url_scheme"] as? String, "myapp")
```
jaxdesmarais, line 113: Can we add a test case that this is nil when expected?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* PayPal
+  * Add optional `fallbackURLScheme` parameter to support custom URL scheme fallback in PayPal App Switch
```
```suggestion
  * Add optional `fallbackURLScheme` parameter to support custom URL scheme fallback in the PayPal App Switch flow
```

Sources/BraintreePayPal/BTPayPalReturnURL.swift
ndrsKloss: I tested the fallback URL flow and it's working. The return URL from the PayPal backend already includes `/braintreeAppSwitchPayPal/success` in the path. When we build the fallback URL, we just swap the scheme but keep the same path. So the `isValid` check passes because `braintreeAppSwitchPayPal` is preserved in the path. Some evidences from my logs:

`> com.paypal.XOSphereApp.payments://braintree-payments/braintreeAppSwitchPayPal/success?token=..."`

UnitTests/BraintreePayPalTests/BTPayPalVaultRequest_Tests.swift
ndrsKloss: Done!

CHANGELOG.md
ndrsKloss: Done!

Sources/BraintreePayPal/BTPayPalVaultBaseRequest.swift
```diff
@@ -66,6 +72,10 @@ import BraintreeCore
 
             vaultParameters["shipping_address"] = shippingAddressParameters
         }
+        
+        if let fallbackURLScheme {
+            vaultParameters["merchant_app_fallback_url_scheme"] = fallbackURLScheme
```
jaxdesmarais: I think we discussed internally setting this on the `BTPayPalRequest`. That way when deeplinks are enabled there a second merchant upgrade isn't needed. If that's the case we may only need it on the base request class since it's set in the init, but not 100% sure.

ndrsKloss: You're correct, updting it

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -100,15 +102,17 @@ import BraintreeDataCollector
     /// Initialize a new PayPal client instance for the PayPal App Switch flow.
     /// - Parameters:
     ///   - apiClient: The API Client
-    ///   - universalLink: The URL to use for the PayPal app switch flow. Must be a valid HTTPS URL dedicated to Braintree app switch returns. This URL must be allow-listed in your Braintree Control Panel.
+    ///   - universalLink: The URL to use for the PayPal app switch flow. Must be a valid HTTPS URL dedicated to Braintree app switch returns. This URL must be allowlisted in your Braintree Control Panel.
+    ///   - fallbackURLScheme: Optional: A custom URL scheme to use as a fallback if the universal link fails. This scheme must be registered in your app's Info.plist and allowlisted in the Braintree Control Panel.
```
jaxdesmarais: We should update this docstring to include that mercahnts _do not_ need need to include the full scheme. Most merchants will assume that this includes `://` since that's typically how URL schemes are implemented. Also is the allowlist piece correct? Is that something that needs to be done for this?

Sources/BraintreePayPal/BTPayPalReturnURL.swift
```diff
@@ -36,15 +36,23 @@ struct BTPayPalReturnURL {
     // MARK: - Static Methods
 
     /// Evaluates whether the url represents a valid PayPal return URL.
-    /// - Parameter url: an app switch or ASWebAuthenticationSession return URL
+    /// - Parameters:
+    ///   - url: an app switch or ASWebAuthenticationSession return URL
+    ///   - fallbackURLScheme: Optional fallback URL scheme
```
agedd:
```suggestion
    ///   - url: An App Switch or `ASWebAuthenticationSession` return URL
    ///   - fallbackURLScheme: Optional. A fallback URL scheme if the universal link fails.
```

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* PayPal
```
agedd:
```suggestion
* BraintreePayPal
```

ndrsKloss: Done

Sources/BraintreePayPal/BTPayPalReturnURL.swift
ndrsKloss: Done

Sources/BraintreePayPal/BTPayPalClient.swift
ndrsKloss: Done, please check if it makes more sense now

> Also is the allowlist piece correct?

Yes, I will ask Semih to clarify it in the docs

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPal
+  * Add optional `fallbackURLScheme` parameter to support custom URL scheme fallback in PayPal App Switch flow
```
jaxdesmarais: Small nit: 
```suggestion
  * Add optional `fallbackURLScheme` parameter to `BTPayPalClient` to support custom URL scheme fallback in PayPal App Switch flow
```

ndrsKloss: Done

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -765,6 +770,6 @@ extension BTPayPalClient: BTAppContextSwitchClient {
     /// :nodoc:
     @_documentation(visibility: private)
     @objc public static func canHandleReturnURL(_ url: URL) -> Bool {
-        BTPayPalReturnURL.isValid(url)
+        BTPayPalReturnURL.isValid(url, fallbackURLScheme: payPalClient?.fallbackURLScheme)
```
jaxdesmarais, line 773: small nit: I believe we should just be able to access the property directly here vs using the optional client:

```suggestion
        BTPayPalReturnURL.isValid(url, fallbackURLScheme: fallbackURLScheme)
```

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -168,9 +168,11 @@ import BraintreeCore
     public override func parameters(
         with configuration: BTConfiguration,
         universalLink: URL? = nil,
+        fallbackURLScheme: String? = nil,
         isPayPalAppInstalled: Bool = false
     ) -> [String: Any] {
         var baseParameters = super.parameters(with: configuration, universalLink: universalLink, isPayPalAppInstalled: isPayPalAppInstalled)
```
jaxdesmarais: shouldn't we be passing `fallbackURLScheme` here?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -100,15 +102,17 @@ import BraintreeDataCollector
     /// Initialize a new PayPal client instance for the PayPal App Switch flow.
     /// - Parameters:
     ///   - apiClient: The API Client
```
agedd, line 104:
```suggestion
    ///   - apiClient: Required. The API Client
```

UnitTests/BraintreePayPalTests/BTPayPalVaultRequest_Tests.swift
```diff
@@ -104,12 +104,26 @@ class BTPayPalVaultRequest_Tests: XCTestCase {
...
+    }
+    
+    func testParameters_withEnablePayPalAppSwitchTrue_withoutFallbackURLScheme_doesNotIncludeFallbackURLScheme() {
+        let request = BTPayPalVaultRequest(
+            userAuthenticationEmail: "sally@gmail.com",
+            enablePayPalAppSwitch: true
+        )
+
```
agedd: nit: can remove the empty line here
```suggestion
```

Sources/BraintreePayPal/BTPayPalClient.swift
ndrsKloss: As `canHandleReturnURL` is a static method and `fallbackURLScheme` is an instance property, I think we need to access it through the static `payPalClient` reference. Otherwise: "Instance member 'fallbackURLScheme' cannot be used on type 'BTPayPalClient'"

Sources/BraintreePayPal/BTPayPalClient.swift
ndrsKloss: Is that right? We have other three clients and they don explicity use the wording `Required.`
image

UnitTests/BraintreePayPalTests/BTPayPalVaultRequest_Tests.swift
ndrsKloss: Done

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: Ah, gotcha. Forgot this method was static. Thanks for clarifying.

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: We moved to marking as required as part of v7. I would say it's helpful for merchants, but not a blocker.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
jaxdesmarais: Once `fallbackURLScheme` is added to the super here and we've added a few tests for this, we are good to go. We confirmed yesterday the GW will ignore this, but allows merchants to not need to make another update when this is released for 1-time.

ndrsKloss: Changes done

PR #1698, opened 2025-11-20
PayPal buttons interaction states

stechiu:
https://github.com/user-attachments/assets/2cc76edf-3a98-4bc5-b889-bc5e2bdf22a1

### Summary of changes

- Added tapped/pressed button state

### Checklist

- [ ] Added a changelog entry

- @stechiu

### DIFF
--- a/Sources/BraintreeUIComponents/VenmoButtonColor.swift
+++ b/Sources/BraintreeUIComponents/VenmoButtonColor.swift
@@ -50,13 +50,6 @@ public enum VenmoButtonColor: PaymentButtonColorProtocol {
 
     /// Background color when button is tapped
     var tappedButtonColor: Color {
-        switch self {
-        case .primary:
-            return Color(hex: "#0074FF")
-        case .black:
-            return Color(hex: "#696969")
-        case .white:
-            return Color(hex: "#E9E9E9")
-        }
+        .black
     }
 }


Sources/BraintreeUIComponents/BTPayPalButtons.swift
```diff
@@ -41,40 +41,52 @@ public struct PayPalButton: View {
...
-                PayPalButton(color: .black, width: 350) {}
-                
-                // White Button
-                PayPalButton(color: .white, width: 100) {}
-            }
-        }
+}
+
+/// Different styles of the PayPal payment buttons. Available in the colors primary (PayPal blue), black, and white.
+public struct PayPalButtonStyle: ButtonStyle {
```
tdchow: Does this class need to be public?

buzzamus (approved):
nit: I think we discussed we are not using the BT prefix for the files, so BTPayPalButtonColor can just be renamed `BTPayPalButtonColor` and then also I think this PR just needs to be merged after the other PR that introduces the shared state logic

Sources/BraintreeUIComponents/VenmoButtonColor.swift
```diff
@@ -47,4 +47,16 @@ public enum VenmoButtonColor: PaymentButtonColorProtocol {
...
+    var tappedButtonColor: Color {
+        switch self {
+        case .primary:
+            return Color(hex: "#0074FF")
+        case .black:
+            return Color(hex: "#696969")
+        case .white:
+            return Color(hex: "#E9E9E9")
+        }
+    }
```
stechiu, line 54: Ignore this block of code. It was added in because `VenmoButtonColor` needs to conform to `PaymentButtonColorProtocol`, otherwise the code won't compile

Sources/BraintreeUIComponents/BTPayPalButtons.swift
stechiu: We'll need it public since it's how Merchants will access the PayPal button

Sources/BraintreeUIComponents/VenmoButtonColor.swift
jaxdesmarais: We can update this to just:

```suggestion
    /// Background color when button is tapped
    var tappedButtonColor: Color {
        .black
    }
```
That way whichever PR goes in first between this and venmo, there can still be a small Venmo PR with the implementation for that feature.

stechiu: thanks!

PR #1699, opened 2025-11-20
UI Components - Venmo Payment Button

buzzamus:
### Summary of changes
* Created Venmo button class with light, dark, and blue background and corresponding logo

This PR is dependent on https://github.com/braintree/braintree_ios/pull/1696, so will need to be rebased and merged after that PR is perged

Screenshot 2025-11-20 at 1 56 44 PM

### Checklist

- [x] Added a changelog entry
- [x] Tested and confirmed payment flows affected by this change are functioning as expected

PR #1701, opened 2025-11-21
Venmo Button UI Interaction States

buzzamus:
### Summary of changes

- Added Venmo Button background color state changes

Currently pointing PR to the Venmo button branch, but once https://github.com/braintree/braintree_ios/pull/1699 is merged I will point to the feature branch

https://github.com/user-attachments/assets/4c66ccd0-e9ac-41ac-b2c9-43f60aae387d

### Checklist

- [ ] Added a changelog entry

-

### DIFF
--- a/Sources/BraintreeUIComponents/PaymentButtonStyle.swift
+++ b/Sources/BraintreeUIComponents/PaymentButtonStyle.swift
@@ -5,6 +5,7 @@ struct PaymentButtonStyle<Color: PaymentButtonColorProtocol>: ButtonStyle {
     
     let color: Color
     let width: CGFloat?
+    let logoHeight: CGFloat
     
     /// This is the width range for the payment button.
     private var widthRange: CGFloat {
@@ -18,7 +19,7 @@ struct PaymentButtonStyle<Color: PaymentButtonColorProtocol>: ButtonStyle {
                 Image(logoImageName, bundle: .uiComponents)
                     .resizable()
                     .scaledToFit()
-                    .frame(height: 14)
+                    .frame(height: logoHeight)
             }
         }
         .frame(width: widthRange)
--- a/Sources/BraintreeUIComponents/PaymentButtonView.swift
+++ b/Sources/BraintreeUIComponents/PaymentButtonView.swift
@@ -15,7 +15,7 @@ struct PaymentButtonView<Color: PaymentButtonColorProtocol>: View {
         Button(action: action) {
             EmptyView()
         }
-        .buttonStyle(PaymentButtonStyle(color: color, width: width))
+        .buttonStyle(PaymentButtonStyle(color: color, width: width, logoHeight: logoHeight))
         .accessibilityLabel(accessibilityLabel)
         .accessibilityHint(accessibilityHint)
     }
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -19,6 +19,7 @@
 		045241772EC7F15400C25EA6 /* BraintreeVenmo.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = A77AA29C1B618C7700217B73 /* BraintreeVenmo.framework */; platformFilter = ios; };
 		046F006C2ECBF46800F08512 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 046F006B2ECBF46800F08512 /* Assets.xcassets */; };
 		046F007C2ED6406000F08512 /* PayPalButtonColor.swift in Sources */ = {isa = PBXBuildFile; fileRef = 046F007B2ED6404C00F08512 /* PayPalButtonColor.swift */; };
+		046F007E2ED64E8D00F08512 /* PaymentButtonStyle.swift in Sources */ = {isa = PBXBuildFile; fileRef = 046F007D2ED64E8800F08512 /* PaymentButtonStyle.swift */; };
 		04AA31182D07974D0043ACAB /* BTButtonType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 04AA31172D0797460043ACAB /* BTButtonType.swift */; };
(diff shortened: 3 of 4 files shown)


Sources/BraintreeUIComponents/PaymentButtonView.swift
```diff
@@ -38,6 +40,9 @@ struct PaymentButtonView<Color: PaymentButtonColorProtocol>: View {
                     lineWidth: color.hasOutline ? 1 : 0
                 )
         )
+        .onLongPressGesture(minimumDuration: 0, maximumDistance: .infinity, pressing: { pressing in
```
jaxdesmarais: It looks like this method signature may have changed slightly: https://developer.apple.com/documentation/swiftui/view/onlongpressgesture(minimumduration:maximumdistance:perform:onpressingchanged:)

buzzamus: Updated to use current version

Sources/BraintreeUIComponents/PaymentButtonView.swift
```diff
@@ -38,6 +40,9 @@ struct PaymentButtonView<Color: PaymentButtonColorProtocol>: View {
                     lineWidth: color.hasOutline ? 1 : 0
                 )
         )
+        .onLongPressGesture(minimumDuration: 0, maximumDistance: .infinity, perform: {}, onPressingChanged: { pressing in
+            isPressed = pressing
```
tdchow: Instead of having to manually keep track of the pressed state, is `ButtonStyle` something we could utilize?

https://developer.apple.com/documentation/swiftui/buttonstyle

AI generated example:
```swift
struct PressableButtonStyle: ButtonStyle {
    func makeBody(configuration: Configuration) -> some View {
        configuration.label
            .padding()
            .background(configuration.isPressed ? Color.blue.opacity(0.5)
                                                : Color.blue)
            .cornerRadius(10)
            .animation(.easeInOut(duration: 0.1), value: configuration.isPressed)
    }
}

struct ExampleView: View {
    var body: some View {
        Button("Tap Me") {
            print("Button tapped!")
        }
        .buttonStyle(PressableButtonStyle())
    }
}
```

It looks like the `ButtonStyle` will provide a `Configuration` property that contains `isPressed`.

tdchow: It looks like @stechiu has a variant of this in the PayPal UI interaction states PR: https://github.com/braintree/braintree_ios/pull/1698/files#diff-8e2782c4db6a12abd17a7530a75809f0ddb6331222e6c2ed7ffc48e2482bb769R63-R67

buzzamus: I can take a look at that! Not super familiar with ButtonStyle, but i'll take a look

buzzamus: Updated this to use ButtonStyle

Sources/BraintreeUIComponents/PaymentButtonStyle.swift
```diff
@@ -0,0 +1,36 @@
...
+        }
+        .frame(width: widthRange)
+        .frame(height: 45)
+        .background(configuration.isPressed ? color.tappedButtonColor : color.backgroundColor)
+        .cornerRadius(4)
+        .overlay(
+            RoundedRectangle(cornerRadius: 4)
+                .stroke(
+                    color.hasOutline ? .black : .clear,
+                    lineWidth: color.hasOutline ? 1 : 0
```
jaxdesmarais, line 33: Are we able to double check the figma for this to make sure 1px is correct or should we be using like `0.75` or something smaller? It looks a little big currently but my eyes may be playing tricks.

Sources/BraintreeUIComponents/PaymentButtonView.swift
```diff
@@ -10,34 +10,11 @@ struct PaymentButtonView<Color: PaymentButtonColorProtocol>: View {
...
-            HStack {
-                if let logoImageName = color.logoImageName {
-                    Image(logoImageName, bundle: .uiComponents)
-                        .resizable()
-                        .scaledToFit()
-                        .frame(height: 14)
-                }
-            }
-            .frame(width: widthRange)
+            EmptyView()
```
jaxdesmarais, line 16: Do we need an empty view here or can we omit it?

buzzamus: We actually don't need to put anything there and can just do this:
```
        Button(action: action) { }
```

Since the view is actually being populated via PaymentButtonStyle's `makeBody()` function I just put EmptyView there since previously in situations like this that is what was used, but I can definitely switch it up to just being empty or add a comment in there instead

Sources/BraintreeUIComponents/PaymentButtonStyle.swift
buzzamus: I also thought it looked heavier than it looks in Figma, but it is showing as 1px in there as well

jaxdesmarais: Cool - thanks for confirming! Our eyes are playing tricks on us.

Sources/BraintreeUIComponents/PaymentButtonView.swift
jaxdesmarais: 🤔 seems like what you have is the best practice:

> When creating a Button in SwiftUI and applying a ButtonStyle, you must still provide a view as the button's label in its initializer, even if that view is conceptually "empty" and the visual representation comes entirely from the ButtonStyle.

buzzamus: TIL. I always wondered what EmptyView() was _actually_ for and didn't realize this was legit the use 🤣

jaxdesmarais (approved):

stechiu (approved):

PR #1704, opened 2025-11-26
Added BTPayPalCheckoutRequest in PayPal Button

stechiu:
### Summary of changes

- Added `BTPayPalCheckoutRequest`
- Separated UI Component extensions to 2 files

### Checklist

- [x] Added a changelog entry

- @stechiu

### DIFF
--- a/Sources/BraintreeUIComponents/PayPalButton.swift
+++ b/Sources/BraintreeUIComponents/PayPalButton.swift
@@ -11,6 +11,12 @@ import BraintreeCore
 /// PayPal payment button. Available in the colors PayPal blue, black, and white.
 public struct PayPalButton: View {
 
+    /// The PayPal Checkout request.
+    let checkoutRequest: BTPayPalCheckoutRequest?
+
+    /// The PayPal Vault request.
+    let vaultRequest: BTPayPalVaultRequest?
+
     /// The style of the PayPal payment button. Available in the colors PayPal blue, black, and white.
     let color: PayPalButtonColor?
     
@@ -29,11 +35,13 @@ public struct PayPalButton: View {
     ///  - width: Optional. The width of the button. Defaults to 300 px.
     ///  - completion: The completion handler to handle client tokenize request success or failure on button press.
     public init(
-        request: BTPayPalCheckoutRequest? = nil,
+        request: BTPayPalCheckoutRequest,
         color: PayPalButtonColor? = .blue,
         width: CGFloat? = 300,
         completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void
     ) {
+        self.checkoutRequest = request
+        self.vaultRequest = nil
         self.color = color
         self.width = width
         self.completion = completion
@@ -41,25 +49,28 @@ public struct PayPalButton: View {
 
     /// Creates a Vault PayPal payment button.
     /// - Parameters:
-    ///  - vaultRequest: Optional. The PayPal Vault request.
+    ///  - request: Optional. The PayPal Vault request.
     ///  - color: Optional. The color of the button. Defaults to `.blue`.
     ///  - width: Optional. The width of the button. Defaults to 300 px.
     ///  - completion: The completion handler to handle client tokenize request success or failure on button press.
     public init(
-        request: BTPayPalVaultRequest? = nil,
+        request: BTPayPalVaultRequest,
         color: PayPalButtonColor? = .blue,
         width: CGFloat? = 300,
         completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void
     ) {
+        self.checkoutRequest = nil
+        self.vaultRequest = request
         self.color = color
         self.width = width
         self.completion = completion
     }
 
     public var body: some View {
+        let clampedWidth = min(max(width ?? 300, 131), 300)
         PaymentButtonView(
             color: color ?? .blue,
-            width: width,
+            width: clampedWidth,
             logoHeight: 24,
             accessibilityLabel: "Pay with PayPal",
             accessibilityHint: "Complete payment using PayPal",


Sources/BraintreeUIComponents/UIComponents+Bundle.swift
```diff
@@ -0,0 +1,22 @@
...
+            }
+        }
+        return frameworkBundle
+        #else
+        return Bundle(for: BundleToken.self)
+        #endif
+    }
+}
+
+private final class BundleToken {}
```
buzzamus, line 22: I think this was already done on a previous PR too right? Might just need a rebase now

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -1,26 +1,41 @@
 import SwiftUI
+import BraintreePayPal
```
jaxdesmarais, line 4: You'll need to do the `canImport` thing with PayPal too.

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -1,26 +1,41 @@
...
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
 
 /// PayPal payment button. Available in the colors PayPal blue, black, and white.
 public struct PayPalButton: View {
 
+    /// PayPal request
+    let request: BTPayPalCheckoutRequest
```
jaxdesmarais: We will also need the ability to pass in a vault request. So likely a second property here and a second init below that takes in a vault request.

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -30,8 +45,11 @@ public struct PayPalButton: View {
             logoHeight: 24,
             accessibilityLabel: "Pay with PayPal",
             accessibilityHint: "Complete payment using PayPal",
-            action: action
-        )
+        ) {
+            // TODO: Implement Venmo flow when button is tapped
+            // This will create BTVenmoClient and call tokenize
```
jaxdesmarais: I am guessing this TODO should say PayPal?

Sources/BraintreeUIComponents/UIComponents+Bundle.swift
jaxdesmarais: It got moved back out in the PayPal Button PR, so this is just moving it back to it's own file. I had left a comment in the other PR.

stechiu: I thought Bundle was moved to a different file so that's what I did here. @buzzamus Which PR has the right file for extensions?

buzzamus: No, you are right! I was confused on what the preference ended up being

Sources/BraintreeUIComponents/PayPalButton.swift
stechiu: I can make the update but what's the difference between using `canImport` and not using it? I checked the codebase beforehand and some files import the former way and others import the latter way

Sources/BraintreeUIComponents/PayPalButton.swift
stechiu: thanks, I've added it in

Sources/BraintreeUIComponents/PayPalButton.swift
stechiu: True, I've updated it now

Sources/BraintreeUIComponents/PayPalButton.swift
buzzamus: I think it has something to do with how one of the package managers handles importing works. I was getting random CI failures until I added both since I am guessing it was failing if a specific package wasn't added in a certain order or something

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -1,26 +1,50 @@
 import SwiftUI
 
+#if canImport(BraintreeCore)
```
jaxdesmarais:
```suggestion
#if canImport(BraintreePayPal)
```

Sources/BraintreeUIComponents/PayPalButton.swift
jaxdesmarais: Yeah, it's used to check that the dependencies have been loaded correctly.

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -1,26 +1,50 @@
...
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
 /// PayPal payment button. Available in the colors PayPal blue, black, and white.
 public struct PayPalButton: View {
 
+    /// PayPal Checkout request
+    let checkoutRequest: BTPayPalCheckoutRequest
```
jaxdesmarais: I think we can make this optional too, then required in the init.

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -1,26 +1,50 @@
...
+    let completion: (BTPayPalAccountNonce?, Error?) -> Void
 
     /// Creates a PayPal payment button.
     /// - Parameters:
-    ///   - color: Optional. The color of the button. Defaults to `.blue`.
-    ///   - width: Optional. The width of the button. Defaults to 300 px.
-    ///   - action: Button action to handle the result of the PayPal flow.
-    public init(color: PayPalButtonColor? = .blue, width: CGFloat? = 300, action: @escaping () -> Void) {
+    ///  - checkoutRequest: Required. The PayPal Checkout request.
+    ///  - vaultRequest: Optional. The PayPal Vault request.
```
jaxdesmarais: We will want 2 inits, 1 with Checkout, 1 with Vault. That way merchants can initialize whichever they need depending on the flow they are using.

Sources/BraintreeUIComponents/PayPalButton.swift
stechiu: Ohh so that's why cocoapods isn't happy. Thanks!

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -9,18 +17,43 @@ public struct PayPalButton: View {
...
+    }
 
-    /// Creates a PayPal payment button.
+    /// Creates a Vault PayPal payment button.
     /// - Parameters:
-    ///   - color: Optional. The color of the button. Defaults to `.blue`.
-    ///   - width: Optional. The width of the button. Defaults to 300 px.
-    ///   - action: Button action to handle the result of the PayPal flow.
-    public init(color: PayPalButtonColor? = .blue, width: CGFloat? = 300, action: @escaping () -> Void) {
+    ///  - vaultRequest: Optional. The PayPal Vault request.
```
jaxdesmarais:
```suggestion
    ///  - request: Required. The PayPal Vault request.
```

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -9,18 +17,43 @@ public struct PayPalButton: View {
...
+    ///  - color: Optional. The color of the button. Defaults to `.blue`.
+    ///  - width: Optional. The width of the button. Defaults to 300 px.
+    ///  - completion: The completion handler to handle client tokenize request success or failure on button press.
+    public init(
+        request: BTPayPalVaultRequest? = nil,
```
jaxdesmarais:
```suggestion
        request: BTPayPalVaultRequest,
```

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -9,18 +17,43 @@ public struct PayPalButton: View {
...
+    // MARK: - Initializers
+
+    /// Creates a PayPal Checkout payment button.
+    /// - Parameters:
+    ///  - checkoutRequest: Optional. The PayPal Checkout request.
+    ///  - color: Optional. The color of the button. Defaults to `.blue`.
+    ///  - width: Optional. The width of the button. Defaults to 300 px.
+    ///  - completion: The completion handler to handle client tokenize request success or failure on button press.
+    public init(
+        request: BTPayPalCheckoutRequest? = nil,
```
jaxdesmarais:
```suggestion
        request: BTPayPalCheckoutRequest,
```

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -9,18 +17,43 @@ public struct PayPalButton: View {
...
-    /// The PayPal payment button action.
-    let action: () -> Void
+    /// The completion handler to handle PayPal tokenization request success or failure.
+    let completion: (BTPayPalAccountNonce?, Error?) -> Void
+
+    // MARK: - Initializers
+
+    /// Creates a PayPal Checkout payment button.
+    /// - Parameters:
+    ///  - checkoutRequest: Optional. The PayPal Checkout request.
```
jaxdesmarais, line 33:
```suggestion
    ///  - request: Required. The PayPal Checkout request.
```

jaxdesmarais (commented):
With the suggestion changes, at the top of this file we can have properties for:
```swift
let checkoutRequest: BTPayPalCheckoutRequest?
let vaultRequest: BTPayPalVaultRequest?
```
We can then set only the property we need in the init with `self.chekoutRequest = request` in the first init and `self.vaultRequest = request` in the 2nd init. Since the constants are optional at the top of the file and each init has them defined as required, there will not be any compile time issues. Plus it will enforce the request as required for the merchants to use, which will keep things type safe.

Sources/BraintreeUIComponents/PayPalButton.swift
tdchow: Would we be able to have merchants pass in a required `BTPayPalRequest`? Since that protocol would cover both checkout and vault request types?

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -1,37 +1,82 @@
...
     /// - Parameters:
-    ///   - color: Optional. The color of the button. Defaults to `.blue`.
-    ///   - width: Optional. The width of the button. Defaults to 300 px.
-    ///   - action: Button action to handle the result of the PayPal flow.
-    public init(color: PayPalButtonColor? = .blue, width: CGFloat? = 300, action: @escaping () -> Void) {
+    ///  - request: Optional. The PayPal Vault request.
+    ///  - color: Optional. The color of the button. Defaults to `.blue`.
+    ///  - width: Optional. The width of the button. Defaults to 300 px.
+    ///  - completion: The completion handler to handle client tokenize request success or failure on button press.
+    public init(
```
tdchow, line 56: Would a single init work if a merchant passed in a `BTPayPalRequest`? I'm not sure if this is possible in Swift, but in the current Android PR, the generic `PayPalRequest` object is used: https://github.com/braintree/braintree_android/pull/1483/files#diff-9ebca10c41b157ad9cc4d2c3bfe5119876bd34ae2fcfef60cdb5d34cb85cf8aeR43

Sources/BraintreeUIComponents/PayPalButton.swift
stechiu: It's doable and would be cleaner but requires typecasting to checkout request and vault request, a pattern we aren't using in iOS. Thoughts @buzzamus ?

```
private func invokePayPalFlow(authorization: String) {
        let payPalClient = BTPayPalClient(authorization: authKey)
        if let checkoutRequest = request as? BTPayPalCheckoutRequest {
            payPalClient.tokenize(checkoutRequest) { nonce, error in
                completion(nonce, error)
            }
        } else if let vaultRequest = request as? BTPayPalVaultRequest {
            payPalClient.tokenize(vaultRequest) { nonce, error in
                completion(nonce, error)
            }
        }
    }
```

tdchow: I got clarification from @jaxdesmarais that we should avoid typecasting. So the 2 initializers is the best path forward!

Sources/BraintreeUIComponents/PayPalButton.swift
tdchow: Scratch that, we should avoid typecasting, so we'll need the 2 inits.

Sources/BraintreeUIComponents/PayPalButton.swift
buzzamus: Sorry I don't know how I missed this, but glad it got resolved!

PR #1708, opened 2025-12-02
Add VenmoButton to Demo App

buzzamus:
### Summary of changes

- Added Venmo button to demo app

### Checklist

- [ ] Added a changelog entry

https://github.com/user-attachments/assets/fc01e78a-c9a4-49f6-92ee-1401e829c34e

Simulator Screenshot - iPhone 16 Pro - 2025-12-02 at 10 14 17

-

### DIFF
--- a/Demo/Application/Features/PaymentButtonViewController.swift
+++ b/Demo/Application/Features/PaymentButtonViewController.swift
@@ -34,15 +34,15 @@ class PaymentButtonViewController: PaymentButtonBaseViewController {
     }
     
     @objc private func colorChanged(_ sender: UISegmentedControl) {
-        switch sender.selectedSegmentIndex {
+        selectedColor = switch sender.selectedSegmentIndex {
         case 0:
-            selectedColor = .blue
+            .blue
         case 1:
-            selectedColor = .black
+            .black
         case 2:
-            selectedColor = .white
+            .white
         default:
-            selectedColor = .blue
+            .blue
         }
 
         setupVenmoButton()
@@ -57,22 +57,9 @@ class PaymentButtonViewController: PaymentButtonBaseViewController {
             universalLink: URL(string: "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments")!,
             request: venmoRequest,
             color: selectedColor,
-            width: 300
-        ) { [weak self] nonce, error in
-            DispatchQueue.main.async {
-                if let nonce {
-                    self?.progressBlock("Got a nonce 💎!")
-                    self?.completionBlock(nonce)
-                } else if let error {
-                    if error as? BTVenmoError == .canceled {
-                        self?.progressBlock("Canceled 🔰")
-                    } else {
-                        self?.progressBlock(error.localizedDescription)
-                    }
-                }
-            }
-        }
-
+            width: 300,
+            completion: venmoCompletionHandler
+        )
         if let existingHostingController = hostingController {
             existingHostingController.willMove(toParent: nil)
             existingHostingController.view.removeFromSuperview()
@@ -96,5 +83,20 @@ class PaymentButtonViewController: PaymentButtonBaseViewController {
         ])
         
         hostingController.didMove(toParent: self)
+    }
+
+    private func venmoCompletionHandler(nonce: BTVenmoAccountNonce?, error: Error?) {
+        DispatchQueue.main.async { [weak self] in
+            if let nonce {
+                self?.progressBlock("Got a nonce 💎!")
+                self?.completionBlock(nonce)
+            } else if let error {
+                if error as? BTVenmoError == .canceled {
+                    self?.progressBlock("Canceled 🔰")
+                } else {
+                    self?.progressBlock(error.localizedDescription)
+                }
+            }
+        }
     }
 }


Demo/Application/Features/VenmoPaymentButtonViewController.swift
```diff
@@ -0,0 +1,139 @@
+import UIKit
+import BraintreeUIComponents
+import BraintreeVenmo
+import SwiftUI
+
+class VenmoPaymentButtonViewController: BaseViewController {
```
jaxdesmarais: I think if we use `PaymentButtonBaseViewController` instead we can remove `authorization`, and the 2 inits.

Demo/Application/Features/VenmoPaymentButtonViewController.swift
```diff
@@ -0,0 +1,139 @@
...
+
+    let authorization: String
+    private var hostingController: UIHostingController<VenmoButton>?
+    private var currentColor: VenmoButtonColor = .blue
```
jaxdesmarais: Would `selectedColor` be a better name for this?

Demo/Application/Features/VenmoPaymentButtonViewController.swift
```diff
@@ -0,0 +1,139 @@
...
+        let venmoButtonView = VenmoButton(
+            authentication: authorization,
+            // swiftlint:disable:next force_unwrapping
+            universalLink: URL(string: "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments")!,
+            request: venmoRequest,
+            color: currentColor,
+            width: 300
+        ) { [weak self] nonce, error in
+            DispatchQueue.main.async {
+                if let nonce = nonce {
```
jaxdesmarais:
```suggestion
                if let nonce {
```

Demo/Application/Features/VenmoPaymentButtonViewController.swift
```diff
@@ -0,0 +1,139 @@
...
+                    self?.progressBlock("Got a nonce 💎!")
+                    self?.completionBlock(nonce)
+                } else if let error = error {
```
jaxdesmarais:
```suggestion
                } else if let error {
```

Demo/Application/Features/VenmoPaymentButtonViewController.swift
buzzamus: Yep! That makes more sense

Demo/Application/Features/VenmoPaymentButtonViewController.swift
buzzamus: Oh nice yep just switched out the base class, removed the inits and authorization reference and all is working

Demo/Application/Features/PaymentButtonViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+                    if error as? BTVenmoError == .canceled {
+                        self?.progressBlock("Canceled 🔰")
+                    } else {
+                        self?.progressBlock(error.localizedDescription)
+                    }
+                }
+            }
+        }
+
+        if let existingHostingController = hostingController {
```
stechiu, line 63: Can you explain the difference between existing vs hosting? Is existing embedded in hosting?

buzzamus: existing is checking to see if there is already a hostingController and if one does exist `existingHostingController` is handling cleanup before a new hostingController is created to render the SwiftUI view in UIKit. In this case I didn't do name shadowing here since I wanted to make sure teardown happens before creating a new one

Demo/Application/Base/Settings/Settings.bundle/Root.plist
```diff
@@ -20,6 +20,7 @@
 				<string>PayPal - Web Checkout</string>
 				<string>Credit Card Tokenization</string>
 				<string>Venmo - Custom Button</string>
+				<string>Venmo Payment Button</string>
```
jaxdesmarais: Do we want to call this "Payment Buttons" and also add the PayPal button here? iirc that's what we did on Android.

Demo/Application/Features/VenmoPaymentButtonViewController.swift
```diff
@@ -0,0 +1,129 @@
+import UIKit
+import BraintreeUIComponents
+import BraintreeVenmo
+import SwiftUI
```
jaxdesmarais: Small nit to group these by BT vs not BT:

```suggestion
import UIKit
import SwiftUI
import BraintreeUIComponents
import BraintreeVenmo
```

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -8,6 +8,8 @@
 
 /* Begin PBXBuildFile section */
 		08222B07C311489F90D08D41 /* Pods_Demo.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 6D23244B5E9EE59BAB3F3003 /* Pods_Demo.framework */; };
+		08A83F592EDE268E0076BD28 /* VenmoPaymentButtonViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 08A83F582EDE268E0076BD28 /* VenmoPaymentButtonViewController.swift */; };
```
jaxdesmarais: Depending on our other decision on naming lets rename this to just PaymentButtonViewController.

Demo/Application/Features/VenmoPaymentButtonViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+            colorSegmentedControl.widthAnchor.constraint(equalToConstant: 250)
+        ])
+
+        let colorLabel = UILabel()
+        colorLabel.text = "Button Color:"
+        colorLabel.textAlignment = .center
+        colorLabel.font = .systemFont(ofSize: 16, weight: .medium)
+        colorLabel.textColor = .label
+        colorLabel.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(colorLabel)
```
stechiu: I don't think `colorLabel` is necessary since the segmented controller themselves have colors names in them, so it's self explanatory. Also would save us room on the screen when PayPal button and segmented controller is added

Demo/Application/Features/VenmoPaymentButtonViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+        createVenmoButton()
+    }
+    
+    private func setupVenmoButton() {
+        let descriptionLabel = UILabel()
+        descriptionLabel.text = "VenmoButton from BraintreeUIComponents"
+        descriptionLabel.textAlignment = .center
+        descriptionLabel.font = .systemFont(ofSize: 16, weight: .medium)
+        descriptionLabel.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(descriptionLabel)
```
stechiu: Same here, we can remove this label

Demo/Application/Features/PaymentButtonViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+            existingHostingController.willMove(toParent: nil)
+            existingHostingController.view.removeFromSuperview()
+            existingHostingController.removeFromParent()
+        }
+
+        hostingController = UIHostingController(rootView: venmoButtonView)
+        guard let hostingController = hostingController else { return }
```
jaxdesmarais: I think we can use the shorthand syntax here too:

```suggestion
        guard let hostingController else { return }
```

jaxdesmarais: Though feels a bit redundant since we assign it just above. Will this ever be nil?

Demo/Application/Features/PaymentButtonViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+                        self?.progressBlock(error.localizedDescription)
+                    }
+                }
+            }
+        }
+
+        if let existingHostingController = hostingController {
+            existingHostingController.willMove(toParent: nil)
+            existingHostingController.view.removeFromSuperview()
+            existingHostingController.removeFromParent()
```
jaxdesmarais, line 66: Do we need a `existingHostingController.didMove` here or will we be good without?

Demo/Application/Features/VenmoPaymentButtonViewController.swift
buzzamus: oh yeah good point!

Demo/Application/Base/Settings/Settings.bundle/Root.plist
buzzamus: @stechiu what are your thoughts?

Demo/Application/Features/PaymentButtonViewController.swift
buzzamus: oh yeah I think you are right. This may have been forgotten to be deleted during a first pass.

Demo/Application/Features/PaymentButtonViewController.swift
buzzamus: i'll investigate

Demo/Application/Base/Settings/Settings.bundle/Root.plist
stechiu: I'm down for that. I branched off of @buzzamus's branch and already made that change

Demo/Application/Features/PaymentButtonViewController.swift
buzzamus: `removeFromParent()` automatically calls `didMove` so it is redundant it looks like - https://developer.apple.com/documentation/uikit/uiviewcontroller/didmove(toparent:)

Demo/Application/Features/PaymentButtonViewController.swift
buzzamus: Technically this still shows up as nil when removing the line, so I guess the guard statement is cleaner than just remaking a new variable like:

```
 let newHostingController = UIHostingController(rootView: venmoButtonView)
  hostingController = newHostingController
  ```
  
  Updating to use the shorthand guard statement though

Comment (jaxdesmarais) [reply Demo/Application/Features/PaymentButtonViewController.swift]: Ah, that makes sense. Thanks for checking.

File: Demo/Application/Features/PaymentButtonViewController.swift (hunk 12 above)
Comment (jaxdesmarais) [reply Demo/Application/Features/PaymentButtonViewController.swift:66]: Oh nice! 🚀

File: Demo/Application/Features/PaymentButtonViewController.swift [hunk 13]
```diff
@@ -0,0 +1,100 @@
...
+        
+        NSLayoutConstraint.activate([
+            colorSegmentedControl.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+            colorSegmentedControl.centerYAnchor.constraint(equalTo: view.centerYAnchor, constant: -80),
+            colorSegmentedControl.widthAnchor.constraint(equalToConstant: 250)
+        ])
+    }
+    
+    @objc private func colorChanged(_ sender: UISegmentedControl) {
+        switch sender.selectedSegmentIndex {
```
jaxdesmarais: Not sure if it's possible to do but we could do something like this maybe:
```swift
selectedColor = switch sender.selectedSegmentIndex {
```
Not even 100% on if that compiles but I _think_ it would.

Demo/Application/Features/PaymentButtonViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+                    self?.progressBlock("Got a nonce 💎!")
+                    self?.completionBlock(nonce)
+                } else if let error {
+                    if error as? BTVenmoError == .canceled {
+                        self?.progressBlock("Canceled 🔰")
+                    } else {
+                        self?.progressBlock(error.localizedDescription)
+                    }
+                }
+            }
```
jaxdesmarais: We could consider extracting this out into it's own private method. Not sure if that'd be cleaner. Take it or leave it, not a blocker.

Demo/Application/Features/PaymentButtonViewController.swift
buzzamus: Nice! Yep that works

Demo/Application/Features/PaymentButtonViewController.swift
buzzamus: Done!

PR #1711, opened 2025-12-05
Add Payment Button Analytics

buzzamus:
### Summary of changes

- Created VenmoButton and PayPalButton analytic events, call each corresponding button's events in view onAppear and in button tap logic.
- added in new initializer to Venmo and PayPal APIClient class types to pass in existing APIClients for passing on existing session information for analytics.
- Renamed VenmoButton's authentication parameter to authorization for parity with all other authentication arguments.

### Checklist

- [x] Added a changelog entry

-

### DIFF
--- a/Sources/BraintreeUIComponents/UIComponentsAnalytics.swift
+++ b/Sources/BraintreeUIComponents/UIComponentsAnalytics.swift
@@ -1,8 +1,14 @@
 import Foundation
 
 enum UIComponentsAnalytics {
+
+    // MARK: - PayPal Button Events
+
     static let payPalButtonPresented = "ui-components:paypal-button:presented"
     static let payPalButtonSelected = "ui-components:paypal-button:selected"
+
+    // MARK: - Venmo Button Events
+
     static let venmoButtonPresented = "ui-components:venmo-button:presented"
     static let venmoButtonSelected = "ui-components:venmo-button:selected"
 }


Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -51,6 +54,7 @@ public struct PayPalButton: View {
         self.color = color
         self.width = width
         self.completion = completion
+        self.apiClient = BTAPIClient(authorization: authorization)
```
buzzamus, line 57: Not sure if this (along with the Venmo one) should actually be able to be created and injected in case merchants need to create one earlier in their checkout flow, or handle it like this?

Demo/Application/Features/PaymentButtonViewController.swift
```diff
@@ -52,7 +52,7 @@ class PaymentButtonViewController: PaymentButtonBaseViewController {
         let venmoRequest = BTVenmoRequest(paymentMethodUsage: .singleUse)
 
         let venmoButtonView = VenmoButton(
-            authentication: authorization,
+            authorization: authorization,
```
jaxdesmarais, line 55: good catch, thanks for updating this

Sources/BraintreeUIComponents/UIComponentsAnalytics.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+enum UIComponentsAnalytics {
+    static let paypalButtonPresented = "ui-components:paypal-button:presented"
+    static let paypalButtonSelected = "ui-components:paypal-button:selected"
```
jaxdesmarais: nit to match what is used in the rest of the SDK:

```suggestion
    static let payPalButtonPresented = "ui-components:paypal-button:presented"
    static let payPalButtonSelected = "ui-components:paypal-button:selected"
```

Sources/BraintreeUIComponents/PayPalButton.swift
agedd: ahh that's a good point, imo i think it's okay to handle it this way

Sources/BraintreeUIComponents/UIComponentsAnalytics.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+enum UIComponentsAnalytics {
+    static let payPalButtonPresented = "ui-components:paypal-button:presented"
+    static let payPalButtonSelected = "ui-components:paypal-button:selected"
+    static let venmoButtonPresented = "ui-components:venmo-button:presented"
```
agedd, line 12: nit: can we add a new line in-between paypal and venmo events and a `MARK` divider i.e:

`// MARK: - Venmo Button Events`

buzzamus: yep! Will add those now

PR #1713, opened 2025-12-09
Persist Session ID across checkout flow

buzzamus:
### Summary of changes

- Creates a Singleton to manage the creation/resetting of the session ID value to pass it across all events that happen in a single checkout experience.

### Checklist

- [x] Added a changelog entry

- @buzzamus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 * BraintreeCore
-  * Add `BTSessionManager` to manage the persisting of Session ID throughout a single checkout session
+  * Add `BTSessionIDManager` to manage the persisting of Session ID throughout a single checkout session
 * BraintreeUIComponents
   * Add BraintreeUIComponents module
   * Add `PayPalButton(authorization:request:color:width:completion:)` to generate a PayPal-branded button for launching PayPal flow
--- a/Sources/BraintreeCore/BTSessionIDManager.swift
+++ b/Sources/BraintreeCore/BTSessionIDManager.swift
@@ -1,16 +1,10 @@
 import Foundation
 
-/// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
-///
 /// Manages a shared session ID across all Braintree payment components to enable end-to-end analytics tracking.
-@_documentation(visibility: private)
-public class BTSessionIDManager {
-
-    // MARK: - Singleton Setup
-    public static let shared = BTSessionIDManager()
-
+class BTSessionIDManager {
+    
+    static let shared = BTSessionIDManager()
     let sessionID: String
-    private let lock = NSLock()
 
     private init() {
         self.sessionID = UUID().uuidString.replacingOccurrences(of: "-", with: "")


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* BraintreeCore
+  * Add `BTSessionManager` To manage the persisting of Session ID throughout a single checkout session
```
stechiu: nit

```suggestion
  * Add `BTSessionManager` to manage the persisting of Session ID throughout a single checkout session
```

stechiu (approved):

Sources/BraintreeCore/BTSessionManager.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+
+/// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+///
+/// Manages a shared session ID across all Braintree payment components to enable end-to-end analytics tracking.
+
+@_documentation(visibility: private)
+public class BTSessionManager {
```
jaxdesmarais: If we don't expect to add anything to this in the future we can rename to:

```suggestion
public class BTSessionIDManager {
```

Sources/BraintreeCore/BTSessionManager.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+
+/// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+///
+/// Manages a shared session ID across all Braintree payment components to enable end-to-end analytics tracking.
+
```
jaxdesmarais:
```suggestion
```

Sources/BraintreeCore/BTSessionManager.swift
```diff
@@ -0,0 +1,33 @@
...
+    // MARK: - Internal Methods
+
+    /// Checks for existing session ID. If a session ID doesn't exist, it creates one to be reused across a single customer session.
+    public func getOrCreateSessionID() -> String {
+        lock.lock()
+        defer { lock.unlock() }
+
+        if let existingSessionID = currentSessionID {
+            return existingSessionID
+        }
```
jaxdesmarais: Is there a reason we can't do this here vs creating a new property:

```suggestion
        if let currentSessionID {
            return currentSessionID
        }
```

Sources/BraintreeCore/BTSessionManager.swift
```diff
@@ -0,0 +1,33 @@
...
+
+        let newSessionID = UUID().uuidString.replacingOccurrences(of: "-", with: "")
+        currentSessionID = newSessionID
+        return newSessionID
```
jaxdesmarais: Do we want to return `currentSessionID` here instead?

UnitTests/BraintreeCoreTests/BTClientMetadata_Tests.swift
```diff
@@ -29,11 +29,26 @@ final class BTClientMetadata_Tests: XCTestCase {
...
+
+    func testSessionID_shouldBeShared_whenNewAPIClientCreated() {
+        let paypalAPIClient = BTAPIClient(authorization: "sandbox_tokenization_key")
+        let paypalSessionID = paypalAPIClient.metadata.sessionID
+
+        let cardAPIClient = BTAPIClient(authorization: "sandbox_tokenization_key")
+        let cardSessionID = cardAPIClient.metadata.sessionID
+
+
+        XCTAssertEqual(paypalSessionID, cardSessionID, "PayPal and Venmo should share the same session ID")
```
jaxdesmarais:
```suggestion
        XCTAssertEqual(paypalSessionID, cardSessionID, "PayPal and Card should share the same session ID")
```

Sources/BraintreeCore/BTSessionManager.swift
jaxdesmarais: Actually this may not even need to be public if it's just used in core!

buzzamus: Updated!

Sources/BraintreeCore/BTSessionManager.swift
buzzamus: Nope! Good catch. I also realized now that we are longer going to ever reset the value this whole class can be simplified, so just committed a changed to set value directly in initializer

Sources/BraintreeCore/BTSessionManager.swift
buzzamus: Same as above comment I actually just simplified this class now

Sources/BraintreeCore/BTSessionManager.swift
buzzamus: true!

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* BraintreeCore
+  * Add `BTSessionManager` to manage the persisting of Session ID throughout a single checkout session
```
agedd: nit: 
```suggestion
  * Add `BTSessionIDManager` to manage the persisting of Session ID throughout a single checkout session
```

Sources/BraintreeCore/BTSessionIDManager.swift
```diff
@@ -0,0 +1,18 @@
...
+///
+/// Manages a shared session ID across all Braintree payment components to enable end-to-end analytics tracking.
+@_documentation(visibility: private)
+public class BTSessionIDManager {
+
+    // MARK: - Singleton Setup
+    public static let shared = BTSessionIDManager()
+
+    let sessionID: String
+    private let lock = NSLock()
```
jaxdesmarais: Looks like we aren't using this!

Sources/BraintreeCore/BTSessionIDManager.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+
+/// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+///
+/// Manages a shared session ID across all Braintree payment components to enable end-to-end analytics tracking.
+@_documentation(visibility: private)
+public class BTSessionIDManager {
+
+    // MARK: - Singleton Setup
+    public static let shared = BTSessionIDManager()
```
jaxdesmarais:
```suggestion
/// Manages a shared session ID across all Braintree payment components to enable end-to-end analytics tracking.
class BTSessionIDManager {

    static let shared = BTSessionIDManager()
```

CHANGELOG.md
buzzamus: Someday I will remember to update the CHANGELOG...but that day is not today 😭 

Updated!

Sources/BraintreeCore/BTSessionIDManager.swift
buzzamus: OOO yep! No longer needed now that we aren' resetting and checking - updated!

Sources/BraintreeCore/BTSessionIDManager.swift
buzzamus: Thanks! Updated

jaxdesmarais (approved):
🚀

PR #1714, opened 2025-12-12
Added loading state to PayPal and Venmo buttons

stechiu:
User cancels or completes payment
https://github.com/user-attachments/assets/ee12f0ef-2761-4247-a807-767c60211894

User hard quits app switch flow
https://github.com/user-attachments/assets/2fb1a821-eb78-40fd-bbaa-b580d27b980e

### Summary of changes

- Added loading states (spinner animation) to PayPal and Venmo payment buttons
- Button is diabled until user cancels, completes payment, or abandons app switch
- Button is re-enabled after payment flow continues, user cancels, or user abandons app switch

### Checklist

- [ ] Added a changelog entry

- @stechiu

### DIFF
--- a/Sources/BraintreeUIComponents/PaymentButtonView.swift
+++ b/Sources/BraintreeUIComponents/PaymentButtonView.swift
@@ -1,7 +1,7 @@
 import SwiftUI
 
 /// Shared payment button view that works with any PaymentButtonColorProtocol
-///  Not to be called directly. Use either BTPayPalButtons or BTVenmoButton instead.
+/// Not to be called directly. Use either BTPayPalButtons or BTVenmoButton instead.
 struct PaymentButtonView<Color: PaymentButtonColorProtocol>: View {
     
     let color: Color
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -1488,8 +1488,8 @@
 				0841353B2ED0CBAD00884FE9 /* PaymentButtonView.swift */,
 				0841353C2ED0CBAD00884FE9 /* UIComponents+Color.swift */,
 				046F007F2ED6905800F08512 /* UIComponents+Bundle.swift */,
-				045241692EC4F80F00C25EA6 /* PrivacyInfo.xcprivacy */,
 				08A83F5C2EE096860076BD28 /* UIComponentsAnalytics.swift */,
+				045241692EC4F80F00C25EA6 /* PrivacyInfo.xcprivacy */,
 			);
 			path = BraintreeUIComponents;
 			sourceTree = "<group>";


Sources/BraintreeUIComponents/VenmoButton.swift
```diff
@@ -58,27 +64,53 @@ public struct VenmoButton: View {
...
+                isLoading = true
+                spinnerRotation = 0
+                invokeVenmoFlow()
+            },
+            isDisabled: isLoading,
+            spinnerImageName: color?.spinnerColor,
+            isLoading: isLoading,
+            spinnerRotation: spinnerRotation
+        )
+        .id(isLoading)
```
jaxdesmarais: Do we need an `id` in the other button or can we remove this?

Sources/BraintreeUIComponents/VenmoButton.swift
```diff
@@ -58,27 +64,53 @@ public struct VenmoButton: View {
...
-            invokeVenmoFlow()
-        }
+            accessibilityHint: "Complete payment using Venmo",
+            action: {
+                apiClient?.sendAnalyticsEvent(UIComponentsAnalytics.venmoButtonSelected)
+                isLoading = true
+                spinnerRotation = 0
+                invokeVenmoFlow()
+            },
+            isDisabled: isLoading,
```
jaxdesmarais: Small nit, can we update the order in the `PaymentButtonView` so action is last? That way we can keep it unnamed which reads a bit cleaner.

Sources/BraintreeUIComponents/VenmoButtonColor.swift
```diff
@@ -59,4 +59,16 @@ public enum VenmoButtonColor: PaymentButtonColorProtocol {
             return Color(hex: "#E9E9E9")
         }
     }
+
+    /// Spinner color when button is tapped
+    var spinnerColor: String? {
+        switch self {
+        case .blue:
+            return "SpinnerWhite"
```
jaxdesmarais: I _think_ this one is supposed to be `SpinnerBlack`? Unless the loading indicator is a different color for Venmo blue vs PayPal?

Sources/BraintreeUIComponents/PaymentButtonView.swift
```diff
@@ -10,12 +10,26 @@ struct PaymentButtonView<Color: PaymentButtonColorProtocol>: View {
...
+            PaymentButtonStyle(
+                color: color,
+                width: width,
+                logoHeight: logoHeight,
+                spinnerImageName: spinnerImageName,
+                isLoading: isLoading,
+                spinnerRotation: spinnerRotation
+            )
+        )
+        .disabled(isDisabled)
```
jaxdesmarais: Could we instead use `isLoading` to track this?

Sources/BraintreeUIComponents/VenmoButtonColor.swift
stechiu: I didn't see a loading state UI for Venmo, only PayPal. For PayPal blue, the spinner is black. I figured since Venmo blue has white logo, the loading state would also be white. I did initially have the spinner as white but upon testing, it's a bit hard to see the spinner because of how dark the blue is.

Who's the UX/UI designer or PM who can confirm?

jaxdesmarais: Sent you the designer name on slack. Lets confirm since they didn't give us a design for this.

Sources/BraintreeUIComponents/VenmoButton.swift
stechiu: Yea looks like we don't _need_ to give the view an ID since it doesn't get used

Sources/BraintreeUIComponents/PaymentButtonView.swift
stechiu: Makes sense, should be good now

Sources/BraintreeUIComponents/VenmoButtonColor.swift
stechiu: Thanks! I'm waiting to hear back from the designer

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -32,6 +32,14 @@ public struct PayPalButton: View {
     /// private BTAPIClient to send analytic events
     private let apiClient: BTAPIClient?
 
+    /// Loading state of button
+    @State private var isLoading: Bool = false
+
+    /// Rotation angle for spinner animation
+    @State private var spinnerRotation: Double = 0
+
+    @Environment(\.scenePhase) private var scenePhase
```
agedd: nit: can we add a small doc string for `scenePhase`

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -88,26 +96,67 @@ public struct PayPalButton: View {
...
+                print("[PayPalButton] App became active via scenePhase, resetting isLoading to false")
+                isLoading = false
+            }
+        }
+        // Additional listener for app foregrounding (more reliable when embedded in UIHostingController)
+        .onReceive(NotificationCenter.default.publisher(for: UIApplication.willEnterForegroundNotification)) { _ in
+            if isLoading {
+                print("[PayPalButton] App will enter foreground via notification, resetting isLoading to false")
+                isLoading = false
+            }
```
agedd, line 123: nice doc-strings explaining all the different states! 🚀

Sources/BraintreeUIComponents/PayPalButton.swift
stechiu: Done!

Sources/BraintreeUIComponents/PaymentButtonView.swift
```diff
@@ -9,13 +9,27 @@ struct PaymentButtonView<Color: PaymentButtonColorProtocol>: View {
     let logoHeight: CGFloat
     let accessibilityLabel: String
     let accessibilityHint: String
+    let isDisabled: Bool
```
jaxdesmarais: We should be able to remove this now

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -88,26 +97,64 @@ public struct PayPalButton: View {
...
+                spinnerRotation = 0
+                withAnimation(Animation.linear(duration: 1).repeatForever(autoreverses: false)) {
+                    spinnerRotation = 360
+                }
+            }
+        }
+        // Listen for app switch abandonment, cancel, or finish via PayPal universal link
+        .onOpenURL { url in
+            // Reset spinner if returning from PayPal via universal link (success or cancel)
+            if url.host?.contains("paypal") == true || url.absoluteString.contains("paypal") {
```
jaxdesmarais: General question, what are we basing this check off of? I am assuming the URL you took a look at has one or both of these in it but wasn't sure.

Sources/BraintreeUIComponents/VenmoButton.swift
```diff
@@ -58,27 +64,51 @@ public struct VenmoButton: View {
...
+            isLoading = true
+            spinnerRotation = 0
             invokeVenmoFlow()
         }
         .onAppear {
             apiClient?.sendAnalyticsEvent(UIComponentsAnalytics.venmoButtonPresented)
+            isLoading = false
+        }
+        // Spinner animation
+        .onChange(of: isLoading) { _, loading in
```
jaxdesmarais: Do we need to add the other states here since Venmo also supports universal links?

Sources/BraintreeUIComponents/VenmoButtonColor.swift
jaxdesmarais: Let me know if you need help getting an answer!

Sources/BraintreeUIComponents/VenmoButton.swift
stechiu: We'll need to keep all the states. I found that `.onReceive` was necessary for the spinner to stop loading if a user app switches, whether it's via a universal link or app. Without that state, the spinner spins into infinity

jaxdesmarais: My question was more why does PayPal have 4 states and venmo only has 2. I would expect that they would both have the same since universal links/app switch is used for both.

Sources/BraintreeUIComponents/PayPalButton.swift
stechiu: I ended up simplifying this to `url.path.contains`. It was originally based off of something I found online

Sources/BraintreeUIComponents/VenmoButton.swift
stechiu: I see, you meant the other states from PayPal button and not within the Venmo file. I originally had those other ones when I was testing the PayPal and Venmo buttons but .onReceive and .onChange will be enough

agedd (approved):
lgtm!

PR #1718, opened 2025-12-17
Convert dataTask() in BTHTTP to an Async Function

agedd:
### Summary of changes

- Replace existing `dataTask(with request: URLRequest, completionHandler: @escaping @Sendable (Data?, URLResponse?, (any Error)?) -> Void) -> URLSessionDataTask` with its async counterpart
- Add async/await unit tests
- **Note**: the original completion handler method still exists in the `BTHTTP` file - will cleanup and remove it in a future PR

### Checklist

- @agedd

### DIFF
--- a/Sources/BraintreeCore/BTHTTP.swift
+++ b/Sources/BraintreeCore/BTHTTP.swift
@@ -290,7 +290,7 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
         response: URLResponse? = nil,
         error: Error? = nil
     ) throws -> (BTJSON, HTTPURLResponse) {
-        if let error = error {
+        if let error {
             throw error
         }
         
@@ -572,10 +572,4 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
         let finalQuery = query.replacingOccurrences(of: queryDiscardHolder, with: "")
         return finalQuery
     }
-}
-
-enum BTHTTPMethod: String {
-    case get = "GET"
-    case post = "POST"
-    case delete = "DELETE"
 }
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -95,6 +95,7 @@
 		62236E1A2B98CFB000CDCC37 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 62236E192B98CFB000CDCC37 /* PrivacyInfo.xcprivacy */; };
 		624454DE2DDE399D00A3383B /* CreateCustomerSessionMutationGraphQLBody.swift in Sources */ = {isa = PBXBuildFile; fileRef = 624454DD2DDE398200A3383B /* CreateCustomerSessionMutationGraphQLBody.swift */; };
 		624B27F72B6AE0C2000AC08A /* BTShopperInsightsError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 624B27F62B6AE0C2000AC08A /* BTShopperInsightsError.swift */; };
+		625719C52F0DBAF6008D019E /* BTHTTPMethod.swift in Sources */ = {isa = PBXBuildFile; fileRef = 625719C42F0DBAE2008D019E /* BTHTTPMethod.swift */; };
 		62970D102B5AF2C000BAB584 /* BTShopperInsightsAnalytics_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 62970D0E2B5AE1E400BAB584 /* BTShopperInsightsAnalytics_Tests.swift */; };
 		6298A1992B91010600E46EDF /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 6298A1982B91010600E46EDF /* PrivacyInfo.xcprivacy */; };
 		62A659A42B98CB23008DFD67 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 62A659A32B98CB23008DFD67 /* PrivacyInfo.xcprivacy */; };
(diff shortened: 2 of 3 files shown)


Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -159,6 +154,50 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
+            var jsonError: Error?
+            handleJSONResponseError(json: json, response: response) { err in
+                jsonError = err
+            }
+            throw jsonError ?? BTHTTPError.clientError([:])
+        }
+        return (json, httpResponse)
+    }
+    
+    func data(for request: URLRequest) async throws -> (Data, URLResponse) {
```
jaxdesmarais: Do we need this? I think we should be able to use the try await on the underlying method directly.

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -159,6 +154,50 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
+            path: path,
+            configuration: configuration,
+            parameters: parameters,
+            headers: headers
+        )
+        let (data, response) = try await session.data(for: request)
+        guard let httpResponse = createHTTPResponse(response: response) else {
+            throw BTHTTPError.httpResponseInvalid
+        }
+        guard let data = data as Data? else {
```
jaxdesmarais: I think this will always pass since the `as Data?` cast will always succeed since data is already Data type from `session.data(for:)`, this check would never fail. If the intent is to verify data exists, the current code doesn't do that since `Data()` is a valid empty data object. Are you able to take a look?

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -159,6 +154,50 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
+            throw BTHTTPError.dataNotFound
+        }
+        if httpResponse.statusCode >= 400 {
+            var error: Error = BTHTTPError.clientError([:])
+            handleHTTPResponseError(response: httpResponse, data: data) { _, err in
```
jaxdesmarais: I think we tend to prefer using the fully named params. Are we able to update this to `error` or something more explicit instead of `err`?

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -159,6 +154,50 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
+                error = err
```
jaxdesmarais: Should we be replacing the empty error in the dictionary or can we instead throw the error directly from this completion vs creating a new object above?

jaxdesmarais: Same question below in the other blocks.

jaxdesmarais (commented):
I think @tdchow had some really good feedback in this PR for this work a while ago if you haven't had a chance to look yet: https://github.com/braintree/braintree_ios/pull/1667

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -77,15 +77,12 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
-            get(path, configuration: configuration, parameters: parameters) { body, response, error in
-                if let error {
-                    continuation.resume(throwing: error)
-                } else {
-                    continuation.resume(returning: (body, response))
-                }
-            }
-        }
+        try await httpRequest(
+            method: "GET",
```
jaxdesmarais: Should we consider extracting the methods into an internal enum?

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -159,6 +154,24 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
+    ) async throws -> (BTJSON, HTTPURLResponse) {
+        let request = try createRequest(
+            method: method,
+            path: path,
+            configuration: configuration,
+            parameters: parameters,
+            headers: headers
+        )
+        let (data, response) = try await session.data(for: request)
+        return try await handleRequestCompletion(data: data, request: request, response: response, error: nil)
```
jaxdesmarais: Since we are throwing errors directly now do we still need to pass `error` into `handleRequestCompletion`?

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -271,6 +284,32 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
+        request: URLRequest?,
+        response: URLResponse?,
+        error: Error?
+    ) async throws -> (BTJSON, HTTPURLResponse) {
+        if let error = error {
+            throw error
+        }
+        guard let response, let httpResponse = createHTTPResponse(response: response) else {
+            throw BTHTTPError.httpResponseInvalid
+        }
```
jaxdesmarais, line 299: Can we add some line breaks in between the statements in this method for readability?

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -271,6 +284,32 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
+        guard let data = data else {
```
jaxdesmarais:
```suggestion
        guard let data else {
```

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -271,6 +284,32 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
         return request
     }
 
+    func handleRequestCompletion(
+        data: Data?,
+        request: URLRequest?,
+        response: URLResponse?,
+        error: Error?
+    ) async throws -> (BTJSON, HTTPURLResponse) {
```
jaxdesmarais: Does this need to be async? Also are we able to remove the old `handleRequestCompletion` or do we need to do that later? If later maybe we can add a TODO?

agedd: ahh yep - added a `TODO` so we can address cleaning up this and other old completion handler methods in a future PR

buzzamus: The task I just picked up is for updating `handleRequestCompletion` so, this should be g2go for this PR

Sources/BraintreeCore/BTHTTP.swift
agedd: done - addressed in 82f6cffe2115f3ec476ad856d9e3b7b83ffa12af

buzzamus (approved):

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -266,11 +279,43 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
...
         return request
     }
 
+    func handleRequestCompletion(
+        data: Data? = nil,
+        request: URLRequest? = nil,
+        response: URLResponse? = nil,
+        error: Error? = nil
+    ) throws -> (BTJSON, HTTPURLResponse) {
+        if let error = error {
```
jaxdesmarais:
```suggestion
        if let error {
```

Sources/BraintreeCore/BTHTTP.swift
```diff
@@ -477,3 +573,9 @@ class BTHTTP: NSObject, URLSessionTaskDelegate {
         return finalQuery
     }
 }
+
+enum BTHTTPMethod: String {
+    case get = "GET"
+    case post = "POST"
+    case delete = "DELETE"
+}
```
jaxdesmarais: Can we move this into it's own file?

PR #1720, opened 2025-12-19
Update PayPalButton For Fallback URL Schema

buzzamus:
### Summary of changes

- Added fallbackURLScheme and universalLink parameters to PaPalButton 
- Updated PayPalButton Demo app implementation to use app switch if PayPal app available
- merged latest version of `main` branch into `ui-components-feature` branch
- Tested out app switch flow to verify it works

### Checklist

- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeUIComponents/PayPalButton.swift
+++ b/Sources/BraintreeUIComponents/PayPalButton.swift
@@ -201,6 +201,6 @@ struct PayPalButton_Previews: PreviewProvider {
     }
 
     // swiftlint:disable:next force_unwrapping
-    static let sampleURL: URL = URL(string: "https://www.example.com/paypal")!
+    static let sampleURL = URL(string: "https://www.example.com/paypal")!
     static func closure(_: BTPayPalAccountNonce?, _: Error?) {}
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
   * Add `BTSessionIDManager` to manage the persisting of Session ID throughout a single checkout session
 * BraintreeUIComponents
   * Add BraintreeUIComponents module
-  * Add `PayPalButton(authorization:request:color:width:completion:)` to generate a PayPal-branded button for launching PayPal flow
+  * Add `PayPalButton(authorization:universalLink:fallbackURLScheme:request:color:width:completion:)` to generate a PayPal-branded button for launching PayPal flow
   * Add `VenmoButton(authorization:universalLink:request:color:width:completion:)` to generate a Venmo-branded button for launching Venmo flow
 * BraintreePayPal
   * Add optional `fallbackURLScheme` parameter to `BTPayPalClient` to support custom URL scheme fallback in PayPal App Switch flow


Demo/Application/Features/PaymentButtonViewController.swift
```diff
@@ -27,10 +27,17 @@ class PaymentButtonViewController: PaymentButtonBaseViewController {
...
+            amount: "10.00",
+            enablePayPalAppSwitch: true,
+            userAuthenticationEmail: nil,
+            userAction: .payNow
+        )
 
         let payPalButtonView = PayPalButton(
             authorization: authorization,
+            // swiftlint:disable:next force_unwrapping
+            universalLink: URL(string: "https://mobile-sdk-demo-site-838cead5d3ab.herokuapp.com/braintree-payments")!,
```
tdchow, line 40: For my learning - why do we need to force unwrap here?

tdchow (approved):

Demo/Application/Features/PaymentButtonViewController.swift
buzzamus: Typically in production code I don't think we would ever do this and would always make URLs optional with some sort of fallback failure...but in this case since we are requiring a URL to make the demo run the only other alternative would be to optionally chain it, but then if we tried to give an alternate value it too would need to then be force unwrapped.

tl;dr - URLs in Swift are fun 🤣

Sources/BraintreeUIComponents/PayPalButton.swift
```diff
@@ -43,12 +49,16 @@ public struct PayPalButton: View {
...
     ///  - authorization: Required. A valid client token or tokenization key.
+    ///  - universalLink: Optional. The URL to use for the PayPal app switch flow. Must be a valid HTTPS URL dedicated to Braintree app switch returns. This URL must be allow-listed in your Braintree Control Panel.
+    ///  - fallbackURLScheme: Optional. A custom URL scheme to use as a fallback if the universal link fails. Pass only the scheme name using alphanumeric characters, hyphens, and periods—without `://` (e.g., `"com.my-app.payments"` not `"com.my-app.payments://"`). This scheme must be registered in your app's Info.plist. You must also contact Braintree to register your URL scheme.
     ///  - request: Required. The PayPal Checkout request.
     ///  - color: Optional. The color of the button. Defaults to `.blue`.
     ///  - width: Optional. The width of the button. Defaults to 300 px.
     ///  - completion: The completion handler to handle client tokenize request success or failure on button press.
     public init(
         authorization: String,
+        universalLink: URL? = nil,
```
jaxdesmarais: What do we think of making `universalLink` required. That way we have everything we need for new merchants to enable app switch via the PayPal button once they pass in enable app switch to the request?

Demo/Application/Features/PaymentButtonViewController.swift
```diff
@@ -27,10 +27,17 @@ class PaymentButtonViewController: PaymentButtonBaseViewController {
     // MARK: - Setup PayPal Payment Button
 
     private func setupPayPalButton() -> UIView {
-        let payPalRequest = BTPayPalCheckoutRequest(amount: "10.00")
+        let payPalRequest = BTPayPalCheckoutRequest(
+            amount: "10.00",
+            enablePayPalAppSwitch: true,
+            userAuthenticationEmail: nil,
+            userAction: .payNow
```
jaxdesmarais, line 34: Do we need these last 2?

buzzamus: Yeah, this was the only way I could actually get the app switch to work vs. using the web version. Once I realized the `PayPalWebCheckoutViewController` app switch flow had them and we had an optional email field I just added the same parameters and got it to run the app switch

buzzamus:
Video of successful app switch:


https://github.com/user-attachments/assets/f14a40c9-4069-499a-b94f-59f102640a3b

Sources/BraintreeUIComponents/PayPalButton.swift
buzzamus: Works for me!

Demo/Application/Features/PaymentButtonViewController.swift
buzzamus: Updated!

Demo/Application/Features/PaymentButtonViewController.swift
jaxdesmarais: ^ we do this in the rest of the demo app too. It's pretty safe to assume it can be force unwrapped since we are setting it right there. 😄 But a bit of a flaw in how URLs are constructed in Swift.

Demo/Application/Features/PaymentButtonViewController.swift
jaxdesmarais: Ah, yeah there are some weird requirements with checkout where app switch only supports `payNow`. We can swap this out to a vault request if we wanted. Since we are setting `userAuthenticationEmail` to nil I wonder why it can't be omitted? That might be a bug if so.

jaxdesmarais (approved):
We likely want to update the CHANGELOG for this but otherwise looks good. Approving ahead of time since that will be a small change.

PR #1726, opened 2026-01-02
Convert BTGraphQLHTTP From Completion Handlers to Async/Await

buzzamus:
### Summary of changes

- switched BTGraphQLHTTP functions using completion handlers to async/await
- Had to create different versions of `post` and `get` functions until the `BTAPIClient` class is converted to async/await since it currently relies on using completion handlers.

### Checklist

- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeCore/BTGraphQLHTTP.swift
+++ b/Sources/BraintreeCore/BTGraphQLHTTP.swift
@@ -3,6 +3,24 @@ import Foundation
 class BTGraphQLHTTP: BTHTTP {
 
     // MARK: - Overrides
+    
+    // TODO: Remove this version of get once BTAPIClient converted to async/await
+    override func get(
+        _ path: String,
+        configuration: BTConfiguration? = nil,
+        parameters: Encodable? = nil,
+        completion: @escaping RequestCompletion
+    ) {
+        callCompletionAsync(with: completion, body: nil, response: nil, error: BTGraphQLHTTPError.unsupportedOperation)
+    }
+    
+    override func get(
+        _ path: String,
+        configuration: BTConfiguration? = nil,
+        parameters: Encodable? = nil
+    ) async throws -> (BTJSON?, HTTPURLResponse?) {
+        throw BTGraphQLHTTPError.unsupportedOperation
+    }
 
     // TODO: Remove this version of post once BTAPIClient converted to async/await
     override func post(
--- a/Braintree.xcodeproj/project.pbxproj
+++ b/Braintree.xcodeproj/project.pbxproj
@@ -25,7 +25,6 @@
 		04AA311E2D0798FC0043ACAB /* BTPageType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 04AA311D2D0798F70043ACAB /* BTPageType.swift */; };
 		04AA31202D07990E0043ACAB /* BTButtonOrder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 04AA311F2D07990A0043ACAB /* BTButtonOrder.swift */; };
 		04B001102D0CF46E00C0060D /* BTExperimentType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 04B0010F2D0CF46900C0060D /* BTExperimentType.swift */; };
-		08CA32AA2F0C5EFF0000CFCB /* BTGraphQLRequestResult.swift in Sources */ = {isa = PBXBuildFile; fileRef = 08CA32A92F0C5EFF0000CFCB /* BTGraphQLRequestResult.swift */; };
 		081F74702EF09793002D53EA /* PayPalButton_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 081F746F2EF09793002D53EA /* PayPalButton_Tests.swift */; };
 		081F74712EF09793002D53EA /* PayPalButton_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 081F746F2EF09793002D53EA /* PayPalButton_Tests.swift */; };
 		081F74732EF099D8002D53EA /* VenmoButton_Tests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 081F74722EF099D8002D53EA /* VenmoButton_Tests.swift */; };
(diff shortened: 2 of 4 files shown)


Sources/BraintreeCore/BTGraphQLHTTP.swift
```diff
@@ -19,121 +18,129 @@ class BTGraphQLHTTP: BTHTTP {
...
     func handleRequestCompletion(
         data: Data?,
         response: URLResponse?,
-        error: Error?,
-        completion: RequestCompletion?
-    ) {
-        guard let completion = completion else { return }
-
+        error: Error?
+        // swiftlint:disable:next large_tuple
```
buzzamus: I don't like this, but wasn't sure how else to handle, so open to suggestions

Sources/BraintreeCore/BTGraphQLHTTP.swift
```diff
@@ -19,121 +18,129 @@ class BTGraphQLHTTP: BTHTTP {
         NSException(name: exceptionName, reason: "GET is unsupported").raise()
     }
 
+    override func get(
```
tdchow, line 17: Not sure if this is an existing pattern - but if GET is not supported, should we not have it defined at all? It's not obvious to me why we have it defined to just throw an error.

Sources/BraintreeCore/BTGraphQLHTTP.swift
```diff
@@ -19,121 +18,129 @@ class BTGraphQLHTTP: BTHTTP {
...
+        }
+    }
+
+    override func post(
+        _ path: String,
+        configuration: BTConfiguration? = nil,
+        parameters: Encodable? = nil,
+        headers: [String: String]? = nil
+    ) async throws -> (BTJSON?, HTTPURLResponse?) {
+        let (body, response, error) = await httpRequest(method: "POST", configuration: configuration, parameters: parameters)
```
tdchow: Could we have the `httpRequest()` function throw an error instead of return an optional error? For where we're still using completion handlers, could we do a try/catch to invoke the completion handler with the error?

Sources/BraintreeCore/BTGraphQLHTTP.swift
tdchow: Maybe an enum with associated values? Success would contain body and httpResponse and Failure would contain httpResponse and error. Probably overkill though. Another option would be to create a class that contains all 3 optional properties.

Sources/BraintreeCore/BTGraphQLHTTP.swift
buzzamus: I think it makes sense to remove it...wasn't sure if there was a specific reason it was doing this or not tbh

tdchow: I don't have the most background knowledge on the iOS repo, so hopefully the second reviewer can provide some context 🙏

agedd: hmm imo i think it's safe to remove it since our graphQL requests are `post` rather than `get` calls but unsure why it was originally added

Sources/BraintreeCore/BTGraphQLHTTP.swift
```diff
@@ -215,3 +222,9 @@ class BTGraphQLHTTP: BTHTTP {
         return errorTree.toDictionary()
     }
 }
+
+// MARK: - UnsupportedOperationError Definition
+struct UnsupportedOperationError: Error, LocalizedError {
```
agedd: nit(optional): thoughts on creating a custom `BTGraphQLHTTPError` enum and calling this in both the completion handler and async/await versions

```
enum BTGraphQLHTTPError: Error {
    case unsupportedOperation

    var localizedDescription: String {
        switch self {
        case .unsupportedOperation:
            return "GET is unsupported"
        }
    }
}
```

Sources/BraintreeCore/BTGraphQLHTTP.swift
agedd: just a suggestion - maybe we could try holding all these properties within a struct:
```
    struct HTTPResult {
        let body: BTJSON?
        let response: HTTPURLResponse?
        let error: Error?
    }
```

Sources/BraintreeCore/BTGraphQLHTTP.swift
buzzamus: That sounds good! Thanks

buzzamus: Actually if we are removing the get function I don't think we need this anymore do we? I just removed UnsupportedOperationError entirely based off of the discussion around get above

Sources/BraintreeCore/BTGraphQLHTTP.swift
jaxdesmarais: This override is used because `BTGraphQLHTTP` subclasses `BTHTTP`. Since BTHTTP` has a get request, we override it here to not allow us to call `get` from the `BTGraphQLHTTP`. Removing this override would allow us to call `BTGraphQLHTTP.get()` which should not be allowed. 

Side note we should update the test `testGETRequests_areUnsupported` as we would expect that test to fail if we remove the `get` override from `BTGraphQLHTTP`. Right now it passes because it doesn't hit the catch block at all.

buzzamus: @jaxdesmarais wait, so add the get override back in correct?

jaxdesmarais: Yes, we should not be able to call get from `BTGraphQLHTTP`. Removing it would allow us to since it subclasses `BTHTTP`. There is likely a better way to architect that long term but having another class both conform to for the shared logic.

Sources/BraintreeCore/BTGraphQLHTTP.swift
buzzamus: Nevermind, added GET method back and created this shared error

Sources/BraintreeCore/BTGraphQLHTTP.swift
buzzamus: done!

Sources/BraintreeCore/BTGraphQLHTTP.swift
buzzamus: Added back and updated test!

Sources/BraintreeCore/BTGraphQLHTTP.swift
buzzamus: Done!

tdchow (approved):

PR #1727, opened 2026-01-07
Funding Source Query Parameter and Instrumentation

NastassiaRodzik:

### Summary of changes

- Adding `fundingSource` to the app switch url.
- Adding `funding_source` as an analytics attribute for tokenize events.

### Checklist

- [x] Added a changelog entry

### Screenshots
Screenshot 2026-01-07 at 1 35 32 PM

### Android Companion PR 
https://github.com/braintree/braintree_android/pull/1511

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -506,18 +506,10 @@ import BraintreeDataCollector
     
     private func getFundingSource(from request: BTPayPalRequest) -> BTPayPalFundingSource {
         if let checkoutRequest = request as? BTPayPalCheckoutRequest {
-            if checkoutRequest.offerCredit {
-                return .credit
-            }
-            if checkoutRequest.offerPayLater {
-                return .payLater
-            }
+            if checkoutRequest.offerCredit { return .credit }
+            if checkoutRequest.offerPayLater { return .payLater }
         }
-        if let vaultRequest = request as? BTPayPalVaultRequest {
-            if vaultRequest.offerCredit {
-                return .credit
-            }
-        }
+        if let vaultRequest = request as? BTPayPalVaultRequest, vaultRequest.offerCredit { return .credit }
         return .payPal
     }
 
--- a/Sources/BraintreePayPal/BTPayPalFundingSource.swift
+++ b/Sources/BraintreePayPal/BTPayPalFundingSource.swift
@@ -1,5 +1,5 @@
 /// Funding sources available when creating a PayPal payment.
-public enum BTPayPalFundingSource: String {
+enum BTPayPalFundingSource: String {
     /// Standard PayPal balance or linked bank account funding.
     case payPal = "paypal"
     /// PayPal Credit revolving line of credit.


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -484,17 +486,30 @@ import BraintreeDataCollector
...
+        if let checkoutRequest = request as? BTPayPalCheckoutRequest {
+            if checkoutRequest.offerCredit {
+                return "credit"
+            }
+            if checkoutRequest.offerPayLater {
+                return "paylater"
+            }
+        }
+        return "paypal"
+    }
```
buzzamus, line 514: Will there potentially be more funding sources in the future? It might make sense to put these in an enum to better keep track of the possible values

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -223,7 +223,8 @@ import UIKit
         isVaultRequest: Bool? = nil,
         linkType: LinkType? = nil,
         pageType: String? = nil,
-        shopperSessionID: String? = nil
+        shopperSessionID: String? = nil,
+        fundingSource: String? = nil
```
jaxdesmarais: This is alphabetized, please update

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -97,7 +98,8 @@ struct FPTIBatchData: Codable {
             pageType: String? = nil,
             requestStartTime: Int? = nil,
             shopperSessionID: String? = nil,
-            startTime: Int? = nil
+            startTime: Int? = nil,
+            fundingSource: String? = nil
```
jaxdesmarais: This is alphabetized, please update

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -405,13 +405,15 @@ import BraintreeDataCollector
     ) {
         self.payPalRequest = request
 
+        let fundingSource: String = getFundingSource(from: request)
```
jaxdesmarais: Should we extract this into a property on the class so we can pass it to all events?

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: ➕ 1 to this comment. It also feels a bit odd to pass a funding source as "paypal" since we are in the paypal module. Is this the default we want to use or can we instead pass nil if it's not credit or pay later?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -706,6 +723,12 @@ import BraintreeDataCollector
     }
 
     private func notifyFailure(with error: Error, completion: @escaping (BTPayPalAccountNonce?, Error?) -> Void) {
+        let fundingSource: String?
+        if let request = payPalRequest {
+            fundingSource = getFundingSource(from: request)
+        } else {
+            fundingSource = nil
+        }
```
jaxdesmarais: This can be eliminated by extracting `fundingSource` into a constant on the class. this also allows us to pass it more easily in every event. Right now we are just passing it into some.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+
+* PayPal
+    * Add `fundingSource` to the app switch url link. 
+    * Add `funding_source` attribute to instrumentation events.
```
```suggestion
* BraintreePayPal
    * Pass `fundingSource` to the app switch url link and to analytics events. 
```

Sources/BraintreePayPal/BTPayPalClient.swift
NastassiaRodzik: @jaxdesmarais "PayPal" is default value

NastassiaRodzik: @buzzamus @jaxdesmarais good suggestion I was also thinking about it. 
Are there any existing conventions or rules in Braintree codebase for naming enums? 
Is `BTPayPalFundingSource` a good name?

Let me know what you think!

jaxdesmarais:
> Hey team @jaxdesmarais @buzzamus Just wanted to call out `BTPayPalVaultRequest` has `offerCredit` property that is currently always `false`, according to the code.
> 
> Are we planning to eventually use it / add **credit** functionality to Vault flow or it should be eliminated?

That statement is incorrect. It defaults to false, but if set to true is set to true and passed upstream. You can use breakpoints to check this.

Sources/BraintreePayPal/BTPayPalClient.swift
buzzamus: I think `BTPayPalFundingSource` works. I think a good example to follow for enum usage is for our analytic events such as `BTCardAnalytics`

buzzamus: One existing convention we have is to put the enum into a separate file though. Could you move the new enum to a new file?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -9,6 +9,12 @@ import BraintreeCore
 import BraintreeDataCollector
 #endif
 
+public enum BTPayPalFundingSource: String {
+    case payPal = "paypal"
+    case credit
+    case payLater = "paylater"
+}
```
buzzamus: I commented below as well, but could this be put in a separate new file? Also, is it deliberate that credit doesn't have a string value?

NastassiaRodzik: Since credit rawValue matches case's name, I didn't explicitly specify it. 
Swift will automatically use the case's name as its string raw value

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -78,6 +78,9 @@ import BraintreeDataCollector
     
     /// Used for analytics purpose to determine if the context type is `BA-TOKEN` or `EC-TOKEN`
     private var contextType: String?
+    
+    /// Used for analytics purpose as well as passed as a parameter to PayPal app
```
agedd: nit: can we make the docstring a bit more clear perhaps something along the lines of 
```suggestion
    /// Used for analytics purpose to determine the funding source of the flow i.e. credit, payLater
```

Sources/BraintreePayPal/BTPayPalFundingSource.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  BTPayPalFundingSource.swift
+//  Braintree
+//
+//  Created by Anastasia Rodzik on 1/8/26.
+//
```
agedd: nit: can we remove these lines and add a simple doc-string to this enum

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -492,6 +503,23 @@ import BraintreeDataCollector
...
+            if checkoutRequest.offerPayLater {
+                return .payLater
+            }
+        }
+        if let vaultRequest = request as? BTPayPalVaultRequest {
+            if vaultRequest.offerCredit {
+                return .credit
+            }
+        }
+        return .payPal
```
jaxdesmarais, line 513: This can be simplified in a way that makes it a bit easier to reason about:
```swift
      if let checkoutRequest = request as? BTPayPalCheckoutRequest {
          if checkoutRequest.offerCredit { return .credit }
          if checkoutRequest.offerPayLater { return .payLater }
      }

      if let vaultRequest = request as? BTPayPalVaultRequest, vaultRequest.offerCredit {
          return .credit
      }

      return .payPal
```

Sources/BraintreePayPal/BTPayPalFundingSource.swift
```diff
@@ -0,0 +1,9 @@
+/// Funding sources available when creating a PayPal payment.
+public enum BTPayPalFundingSource: String {
```
jaxdesmarais: This should not need to be public

PR #1728, opened 2026-01-09
Fix Flakey Analytics Unit Test

buzzamus:
### Summary of changes

- Added `NSLock` to shared access in `MockBackgroundTaskManager`. This was done due to flakey tests around the analytic events. I believe the flakey tests were due to a race condition in the `beginBackgroundTask()` and `endBackgroundTask()` calls. I tested this fix by running tests 200 times to verify no more flakey failures.

One interesting thing to note is the test that was the most flakey (`testSendAnalyticsEventsImmediately_withConcurrentCalls_beginAndEnd_returnSameTaskIDs `) fails almost every time when run by itself. I believe this fix also resolves that issue due to timing differences when running all at once.

This change is only for the mocked version of the background task manager since the production version uses `UIApplication.beginBackgroundTask()`, which is thread safe.

Just a few examples of jobs where these test failed:

 https://github.com/braintree/braintree_ios/actions/runs/20315441931/job/58363751233

https://github.com/braintree/braintree_ios/actions/runs/20767971398/job/59638038574

There were many more, but most of them that I encountered I ended up rerunning the failed jobs, which overwrote the logs.

### Checklist

- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/UnitTests/BraintreeCoreTests/Analytics/MockBackgroundTaskManager.swift
+++ b/UnitTests/BraintreeCoreTests/Analytics/MockBackgroundTaskManager.swift
@@ -1,7 +1,8 @@
+import Foundation
 @testable import BraintreeCore
 
 class MockBackgroundTaskManager: BackgroundTaskManaging {
-    
+
     var didBeginBackgroundTask = false
     var didEndBackgroundTask = false
     var lastTaskName: String?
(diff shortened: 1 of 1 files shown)


UnitTests/BraintreeCoreTests/Analytics/MockBackgroundTaskManager.swift
```diff
@@ -10,11 +11,16 @@ class MockBackgroundTaskManager: BackgroundTaskManaging {
     var endedTaskIDs: Set<UIBackgroundTaskIdentifier> = []
     var begunTaskIDs: Set<UIBackgroundTaskIdentifier> = []
     var taskIDsToReturn: Set<UIBackgroundTaskIdentifier> = []
-    
+
+    private let lock = NSLock()
+
     func beginBackgroundTask(named: String?, expirationHandler handler: (() -> Void)?) -> UIBackgroundTaskIdentifier {
```
tdchow, line 17: For my learning - were we only seeing the flakiness for tests that invoke `beginBackgroundTask()` multiple times within the same test?

tdchow (approved):
Great find!

UnitTests/BraintreeCoreTests/Analytics/MockBackgroundTaskManager.swift
buzzamus: It was in tests with concurrent calls to  `beginBackgroundTask`, so I guess not necessarily multiple calls. This task also made me remember that Sets are not thread safe

jaxdesmarais (approved):

PR #1729, opened 2026-01-12
Add Async Functions to `BTAPIClient`

agedd:
### Summary of changes

- Add async functions to `BTAPIClient`

### Checklist

- @agedd

### DIFF
--- a/UnitTests/BraintreeCoreTests/BTAPIClient_Tests.swift
+++ b/UnitTests/BraintreeCoreTests/BTAPIClient_Tests.swift
@@ -192,7 +192,7 @@ class BTAPIClient_Tests: XCTestCase {
 
         _ = try await apiClient.post("/", parameters: nil, httpType: .graphQLAPI)
 
-        let metaParameters = mockHTTP.lastRequestParameters?["_meta"] as? [String: Any]
+        let metaParameters = mockGraphQLHTTP.lastRequestParameters?["clientSdkMetadata"] as? [String: Any]
         XCTAssertEqual(metaParameters?["integration"] as? String, metadata.integration.stringValue)
         XCTAssertEqual(metaParameters?["source"] as? String, metadata.source.stringValue)
         XCTAssertEqual(metaParameters?["sessionId"] as? String, metadata.sessionID)
@@ -232,7 +232,7 @@ class BTAPIClient_Tests: XCTestCase {
         let postParameters = FakeRequest(testValue: "fake-value")
         _ = try await apiClient.post("/", parameters: postParameters, httpType: .graphQLAPI)
 
-        let metaParameters = mockHTTP.lastRequestParameters?["_meta"] as? [String: Any]
+        let metaParameters = mockGraphQLHTTP.lastRequestParameters?["clientSdkMetadata"] as? [String: Any]
         XCTAssertEqual(metaParameters?["integration"] as? String, metadata.integration.stringValue)
         XCTAssertEqual(metaParameters?["source"] as? String, metadata.source.stringValue)
         XCTAssertEqual(metaParameters?["sessionId"] as? String, metadata.sessionID)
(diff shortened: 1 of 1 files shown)


Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -1,7 +1,9 @@
 import UIKit
 
+// swiftlint:disable:next orphaned_doc_comment
```
jaxdesmarais: I think we only need this because of where the `type_body_length` disable is. Can we move that one up so we can remove this?

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -90,6 +92,16 @@ import UIKit
         }
     }
     
+    public func fetchOrReturnRemoteConfiguration() async throws -> BTConfiguration {
```
jaxdesmarais, line 93: Do we need both this and `fetchConfiguration`? I am trying to understand why we need both. It seems like this is doing more but maybe the other one is incorrect and should be doing this? Not sure if you've looked into that already.

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -132,6 +144,33 @@ import UIKit
...
+            throw BTAPIClientError.invalidAuthorization(authorization.originalValue)
+        }
+        
+        let configuration = try await fetchOrReturnRemoteConfiguration()
+        
+        return try await http(for: httpType)?.get(
+            path,
+            configuration: configuration,
+            parameters: parameters
+        ) ?? (nil, nil)
```
jaxdesmarais: Should we make it so we do not need to do this? Or will that be addressed in a future PR?

Sources/BraintreeCore/BTAPIClient.swift
agedd: ahh yea when i moved it to the top it threw another swift-lint warning related to the orphaned comment - to avoid having to suppress both these warnings, i moved the private helper methods to an extension

Sources/BraintreeCore/BTAPIClient.swift
agedd: ahh yea when i moved it to the top it threw another swift-lint warning related to the orphaned comment - to avoid having to suppress both these warnings, i moved the private helper methods to an extension

Sources/BraintreeCore/BTAPIClient.swift
agedd: ahh yep we don't this, addressed and cleaned it up  in 99c8dc7cb2d27206b4282dfd445f3964630a9791

Sources/BraintreeCore/BTAPIClient.swift
agedd: yea ty for pointing this out - it looks like we don't need them both, removed the old `fetchConfiguration` and updated whenever it's called w/ this async version

Sources/BraintreeCore/BTAPIClient.swift
agedd: ahh nvm, the original warning no longer shows up - updated in 7f302fc4ec1b15b42830c079a2220b561db74913

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -90,8 +90,14 @@ import UIKit
         }
     }
     
-    @MainActor func fetchConfiguration() async throws -> BTConfiguration {
-        try await configurationLoader.getConfig()
+    public func fetchOrReturnRemoteConfiguration() async throws -> BTConfiguration {
```
jaxdesmarais, line 93: Are we able to confirm the test `testCallbacks_useMainDispatchQueue` still passes? I know we duplicated a few tests but it seems like most still use the completion method so want to ensure we don't run into issues later. May be worth converting or manually going through the tests with the async method sooner rather than later unless you've manually tested these.

agedd: ah yep thanks for calling this out ~ the `testCallbacks_useMainDispatchQueue` passes as expected. i've updated the tests to use the async/await version rather than the completion handler method

PR #1731, opened 2026-01-14
Rename funding_source PayPal parameter

NastassiaRodzik:

### Summary of changes

- Rename `funding_source` PayPal Universal link parameter 

### Checklist

- [x] Added a changelog entry

-

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -558,7 +558,7 @@ import BraintreeDataCollector
             URLQueryItem(name: "switch_initiated_time", value: String(Int(round(Date().timeIntervalSince1970 * 1000)))),
             URLQueryItem(name: "flow_type", value: isVaultRequest ? "va" : "ecs"),
             URLQueryItem(name: "merchant", value: merchantID ?? "unknown"),
-            URLQueryItem(name: "funding_source", value: fundingSource?.rawValue ?? BTPayPalFundingSource.payPal.rawValue)
+            URLQueryItem(name: "fundingSource", value: fundingSource?.rawValue ?? BTPayPalFundingSource.payPal.rawValue)
         ]
         
         urlComponents?.queryItems?.append(contentsOf: additionalQueryItems)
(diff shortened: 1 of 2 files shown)


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -558,7 +558,7 @@ import BraintreeDataCollector
             URLQueryItem(name: "switch_initiated_time", value: String(Int(round(Date().timeIntervalSince1970 * 1000)))),
             URLQueryItem(name: "flow_type", value: isVaultRequest ? "va" : "ecs"),
             URLQueryItem(name: "merchant", value: merchantID ?? "unknown"),
-            URLQueryItem(name: "funding_source", value: fundingSource?.rawValue ?? BTPayPalFundingSource.payPal.rawValue)
+            URLQueryItem(name: "fundingSource", value: fundingSource?.rawValue ?? BTPayPalFundingSource.payPal.rawValue)
```
jaxdesmarais, line 561: Why are some snake case and some camel case?

buzzamus (approved):

Sources/BraintreePayPal/BTPayPalClient.swift
anibalb2500: Other integrations are sending this as `fundingSource`. We need to keep consistent with  those integrations so we can track this parameter with one key.

jaxdesmarais: This is the only one that is camel case. If you look at the other properties passed here: `switch_initiated_time` and `flow_type` those are both snake case. Is there a reason we are not instead changing Android code to match the snake case used for every other query parameter? Mixing snake and camel case for these is a pretty bad code smell.

jaxdesmarais: @anibalb2500 please see above

anibalb2500: Hey @jaxdesmarais Sorry for the late response, was on PTO. The dependency is not with Android, but with ModXO. ModXO requires that we append 'fundingSource' (in this exact casing) to the approval url, so that the Pay Later and Credit options are preselected on the RYP page. 

PayPal app will parse the 'fundingSource' and append it to the approval url, as it appears on the app switch url. Other integrations are already sending this parameter as `fundingSource`, so we need to do the same for BT Android and iOS.

PR #1732, opened 2026-01-21
Convert BTAmericanExpressClient to Async/Await

buzzamus:
### Summary of changes

- Updates BTAmericanExpressClient to use the new async/await methods instead of completion handlers.

### Checklist

- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
+++ b/Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
@@ -55,15 +55,11 @@ import BraintreeCore
             let (body, _) = try await apiClient.get("v1/payment_methods/amex_rewards_balance", parameters: parameters)
 
             guard let body else {
-                sendFailureAnalytics(with: BTAmericanExpressError.noRewardsData)
                 throw BTAmericanExpressError.noRewardsData
             }
 
             let rewardsBalance = BTAmericanExpressRewardsBalance(json: body)
             return notifySuccess(with: rewardsBalance)
-        } catch let error as BTAmericanExpressError {
-            // Analytics already sent for BTAmericanExpressError (from guard above)
-            throw error
         } catch {
             sendFailureAnalytics(with: error)
             throw error


UnitTests/BraintreeAmericanExpressTests/BTAmericanExpressClient_Tests.swift
```diff
@@ -37,81 +37,61 @@ class BTAmericanExpressClient_Tests: XCTestCase {
...
-            XCTAssertNil(error)
-            XCTAssertNotNil(rewardsBalance)
-            expectation.fulfill()
-        }
-        waitForExpectations(timeout: 2, handler: nil)
-        
+
+        let rewardsBalance = try await amexClient!.getRewardsBalance(forNonce: "fake-nonce", currencyISOCode: "USD")
+        XCTAssertNotNil(rewardsBalance)
+
```
agedd: nit(optional): can delete new line

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
```diff
@@ -31,28 +31,13 @@ import BraintreeCore
...
-
-            if let error {
-                notifyFailure(with: error, completion: completion)
-                return
-            }
-
-            guard let body else {
-                notifyFailure(with: BTAmericanExpressError.noRewardsData, completion: completion)
-                return
+        Task {
```
agedd, line 34: question: should we consider removing the completion handler version of `getRewardsBalance`

jaxdesmarais: We will want to keep both for now since removing would be a breaking change for merchants. At least for anything public for merchants and not intended for just our use.

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
```diff
@@ -63,32 +48,33 @@ import BraintreeCore
...
+                apiClient.sendAnalyticsEvent(
+                    BTAmericanExpressAnalytics.failed,
+                    errorDescription: error.localizedDescription
+                )
+                throw error
             }
-        }
-    }
-    
-    // MARK: - Analytics Helper Methods
```
jaxdesmarais, line 77: Lets keep the helper methods and maybe convert to use async/awat or throw? For this file it's pretty minor but for larger classes like paypal the helper methods are helpful to maintain consistency.

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
jaxdesmarais, line 112: Should these changes be here or do we need to pull the latest from main?

buzzamus: Nope I need to rebase against main since this was initially pointing at another feature branch. i'll fix and push

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
buzzamus: Updated!

UnitTests/BraintreeAmericanExpressTests/BTAmericanExpressClient_Tests.swift
buzzamus: removed!

Sources/BraintreeCore/Analytics/BTAnalyticsService.swift
buzzamus: Rebased so, only my changes should now show

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
```diff
@@ -31,28 +31,13 @@ import BraintreeCore
...
-
-            guard let body else {
-                notifyFailure(with: BTAmericanExpressError.noRewardsData, completion: completion)
-                return
+        Task {
+            do {
+                let rewardsBalance = try await getRewardsBalance(forNonce: nonce, currencyISOCode: currencyISOCode)
+                completion(rewardsBalance, nil)
+            } catch {
+                completion(nil, error)
```
jaxdesmarais, line 39: Should we be calling notify success and notify failure here?

jaxdesmarais: Ah, I guess in the task when we call the async method it sends the events but want to make sure I understand correctly?

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
```diff
@@ -63,32 +48,38 @@ import BraintreeCore
...
+            let (body, _) = try await apiClient.get("v1/payment_methods/amex_rewards_balance", parameters: parameters)
+
+            guard let body else {
+                try notifyFailure(with: BTAmericanExpressError.noRewardsData)
             }
+
+            let rewardsBalance = BTAmericanExpressRewardsBalance(json: body)
+            return notifySuccess(with: rewardsBalance)
+        } catch let error as BTAmericanExpressError {
+            // Analytics already sent for BTAmericanExpressError (from guard above)
```
jaxdesmarais: Do we need both catch statements here? What about cases where it's deallocated? Guessing that error case isn't relevant now but want to be sure I understand correctly.

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
```diff
@@ -63,32 +48,38 @@ import BraintreeCore
...
-    ) {
+
+    private func notifySuccess(with result: BTAmericanExpressRewardsBalance) -> BTAmericanExpressRewardsBalance {
         apiClient.sendAnalyticsEvent(BTAmericanExpressAnalytics.succeeded)
-        completion(result, nil)
+        return result
     }
 
-    private func notifyFailure(with error: Error, completion: @escaping (BTAmericanExpressRewardsBalance?, Error?) -> Void) {
+    private func notifyFailure(with error: Error) throws -> Never {
```
jaxdesmarais: Are we able to do this:

```suggestion
    private func notifyFailure(with error: Error) throws {
```

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
buzzamus: You are correct: The events are sent in the async method. I was getting duplicate event triggers in certain situations when testing, so this way they are only sent once in `getRewardsBalance`

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
buzzamus: The first catch block is for `BTAmericanExpressError` types only after the notifyFailure method was sent (this sort of goes with your question below with the `Never` usage), so it re-throws the error here to actually be consumed/handled without sending events again.

The second catch block is for the BTAPIClientErrors/BTHTTPErrors. So, it does cover the deallocated and other network errors.

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
buzzamus: If I remove Never it creates a compilation error since there isn't an explicit throw, so this allows it to have an implicit throw here when being used to send analytics and then acts as the throw for line 67

agedd: +1 to removing the `-> Never` if possible

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
agedd: i _think_ it would be safe to remove the `catch let error as BTAmericanExpressError` and handle all errors in the single catch block below on line 66

agedd (approved):
left a couple minor comments, but nothing blocking. approving ahead of time 🚀

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
jaxdesmarais: Gotcha - yeah it seems like in the guard it doesn't know that it throws. What do y'all think about separating out the analytics from throwing? So keep the success helper and update all the error helpers to:
```swift
private func sendFailureAnalytics(with error: Error) {                                                                                                                                                                                                                                        
      apiClient.sendAnalyticsEvent(                                                                                                                                                                                                                                                          
          BTAmericanExpressAnalytics.failed,                                                                                                                                                                                                                                                 
          errorDescription: error.localizedDescription                                                                                                                                                                                                                                       
      )                                                                                                                                                                                                                                                                                      
  }
```
Then at the callsite:
```swift
sendFailureAnalytics(error)                                                                                                                                                                                                                                                            
throw error
```
Would love y'all's input since we would want to use the same pattern across the entire SDK.

buzzamus: That works for me. I can refactor to have analytics just send the call and not throw. I think the reason I did this is it eliminated having to put several different throw keywords in various areas, but I can see this probably ended up just being overkill

buzzamus: Updated!

Sources/BraintreeAmericanExpress/BTAmericanExpressClient.swift
```diff
@@ -63,32 +48,39 @@ import BraintreeCore
...
+            let (body, _) = try await apiClient.get("v1/payment_methods/amex_rewards_balance", parameters: parameters)
+
+            guard let body else {
+                sendFailureAnalytics(with: BTAmericanExpressError.noRewardsData)
+                throw BTAmericanExpressError.noRewardsData
             }
+
+            let rewardsBalance = BTAmericanExpressRewardsBalance(json: body)
+            return notifySuccess(with: rewardsBalance)
+        } catch let error as BTAmericanExpressError {
```
jaxdesmarais: Do we still need this catch with the updates to sending failure events?

buzzamus: yeah, this is there to avoid having the BTAmericanExpressError.noRewardsData error from the guard statement above from throwing the error and sending the analytic event twice

jaxdesmarais: I think if we put the guard outside of the do/catch it would resolve this, right?
```swift
let (body, _): (BTJSON?, HTTPURLResponse?)                                                                            
      do {                                                                                                                  
          (body, _) = try await apiClient.get("v1/payment_methods/amex_rewards_balance", parameters: parameters)            
      } catch {                                                                                                             
          sendFailureAnalytics(with: error)                                                                                 
          throw error                                                                                                       
      }                                                                                                                     
                                                                                                                            
      // Now we're outside the do-catch, so throw will exit directly                                                        
      guard let body else {                                                                                                 
          sendFailureAnalytics(with: BTAmericanExpressError.noRewardsData)                                                  
          throw BTAmericanExpressError.noRewardsData                                                                        
      }                                                                                                                     
                                                                                                                            
      let rewardsBalance = BTAmericanExpressRewardsBalance(json: body)                                                      
      return notifySuccess(with: rewardsBalance)
```

buzzamus: Actually I think I found a way to rewrite this using only 1 catch...I was trying to isolate the scenario with no data early in the guard, but realize that doesn't actually help since we aren't breaking early and it still goes through

jaxdesmarais: ^ very take it or leave it. I am not sure what the best pattern is. This is the first actual module to be converted so will establish the pattern for the rest. Curious to hear opinions!

buzzamus: I was messing around with this:

```
  do {
      let (body, _) = try await apiClient.get("v1/payment_methods/amex_rewards_balance", parameters: parameters)

      guard let body else {
          throw BTAmericanExpressError.noRewardsData
      }

      let rewardsBalance = BTAmericanExpressRewardsBalance(json: body)
      return notifySuccess(with: rewardsBalance)
  } catch {
      sendFailureAnalytics(with: error)
      throw error
  }

buzzamus: Updated to you @jaxdesmarais suggestion

PR #1733, opened 2026-01-21
Update Magnes to 5.6.0 and add Callback

buzzamus:
### Summary of changes

- This updates PPRiskMagnes dependency to version 5.6.0 as well as updates the internal `collectAndSubmit` call to include the new completion handler argument

### Checklist

- [x] Added a changelog entry

- @buzzamus


Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -50,7 +50,6 @@
 		809E86A62AD00AF4004998B0 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 809E86A52AD00AF4004998B0 /* AppDelegate.swift */; };
 		809E86A82AD01FE7004998B0 /* SceneDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 809E86A72AD01FE7004998B0 /* SceneDelegate.swift */; };
 		80D36DEE2A7967F20035380E /* VenmoViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 80D36DED2A7967F20035380E /* VenmoViewController.swift */; };
-		9C36BD2926B3071B00F0A559 /* PPRiskMagnes.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9C36BD2826B3071A00F0A559 /* PPRiskMagnes.xcframework */; };
```
jaxdesmarais, line 53: Was this a duplicate or unintentionally removed?

Frameworks/XCFrameworks/PPRiskMagnes.xcframework/_CodeSignature/CodeDirectory
jaxdesmarais, line 1: It looks like there is still a code signature directory. Can we try removing and re-adding the entire `PPRiskMagnes.xcframework` again? The diff looks a bit off.

Sources/BraintreeDataCollector/BTDataCollector.swift
```diff
@@ -134,6 +137,35 @@ import BraintreeCore
 
         return result?.getPayPalClientMetaDataId() ?? ""
     }
+
+    func collectClientMetadataID(with configuration: BTConfiguration) -> String {
```
jaxdesmarais: I think we can use/update the current method vs a new `collectClientMetadataID`/`submitToMagnes`. We want this new logic to be used across the entire class/all methods.

Demo/Demo.xcodeproj/project.pbxproj
buzzamus: I don't recall doing anything that would have changed the Demo project file directly

buzzamus: There, reverted back...not sure what happened

Sources/BraintreeDataCollector/BTDataCollector.swift
```diff
@@ -43,21 +43,27 @@ import BraintreeCore
...
                 return
             }
-            
-            completion(deviceData, nil)
+
+            // Only invoke completion after Magnes collectAndSubmit internal callback finishes
+            _ = try? MagnesSDK.shared().collectAndSubmit(
+                withPayPalClientMetadataId: clientMetadataID,
+                withAdditionalData: [:]
+            ) { _, _ in
```
jaxdesmarais: Do we want to use the completion here to check the status values and only return once we get a success status? Same with the one below when a custom CMID is passed. Then we should be able to add breakpoints to check we are getting a success as well as return a failure or timeout error if it falls into either of those. We will want to do the same in the `generateClientMetadataID` method to wait for the callback.

Sources/BraintreeDataCollector/BTDataCollector.swift
```diff
@@ -43,21 +43,27 @@ import BraintreeCore
...
-            
+
             guard let deviceData = String(data: jsonData, encoding: .utf8) else {
                 completion(nil, BTDataCollectorError.encodingFailure)
                 return
             }
-            
-            completion(deviceData, nil)
+
+            // Only invoke completion after Magnes collectAndSubmit internal callback finishes
```
agedd: super clean docs 🚀

agedd:
i believe we may also need to update the `Cartfile` to point to the Magnes `5.6.0` version

Sources/BraintreeDataCollector/BTDataCollector.swift
buzzamus: Updated to check status in completion here

Sources/BraintreeDataCollector/BTDataCollector.swift
```diff
@@ -127,14 +142,13 @@ import BraintreeCore
             magnesSource: .BRAINTREE
         )
 
-        let result = try? MagnesSDK.shared().collectAndSubmit(
+        let result = try? MagnesSDK.shared().collect(
```
jaxdesmarais: Are we able to use the callback one here? I think the main usecase for this is for merchants that pass in a custom pairing ID via the `clientMetadataID` method.

buzzamus: In this one we can ignore the new callback's response though I take it? Since it isn't handling the response at all and just wants the clientMetadataID?

buzzamus: So, like this?:
```
        let result = try? MagnesSDK.shared().collectAndSubmit(
            withPayPalClientMetadataId: clientMetadataID ?? "",
            withAdditionalData: data ?? [:]
        ) { _, _ in }
 ```

Comment (jaxdesmarais) [reply Sources/BraintreeDataCollector/BTDataCollector.swift]: I don't think we want to ignore the callback success status for either. For both we want to wait on the callback to return the result. If it's not possible for this one, lets sync up on slack on some options.

Comment (buzzamus) [reply Sources/BraintreeDataCollector/BTDataCollector.swift]: Yeah I think I found a way and working on it now, but it was just a much bigger change than initially thought since I have to create a mock and update unit tests as well as the changes needed in the modules

File: Sources/BraintreeDataCollector/BTDataCollectorError.swift [hunk 6]
```diff
@@ -30,6 +36,12 @@ public enum BTDataCollectorError: Int, Error, CustomNSError, LocalizedError, Equ
 
         case .encodingFailure:
             return "The device data could not be encoded."
+
+        case .magnesSubmitError:
+            return "Magnes SDK failed to submit device data."
+
+        case .magnesSubmitTimeout:
+            return "Magnes SDK timed out while submitting device data."
```
jaxdesmarais: Small suggestion since the name of this SDK is mostly obfuscated from merchants:
```suggestion
        case .callbackSubmitError:
            return "Failed to submit device data."

        case .callbackSubmitTimeout:
            return "Timed out while submitting device data."
```

buzzamus: Updated!

Sources/BraintreeDataCollector/BTDataCollector.swift
buzzamus: I updated to instead deprecate `clientMetadataID` and have `collectAndSubmit` either accept a value, or generate one based off our discussion

Sources/BraintreeDataCollector/BTDataCollector.swift
```diff
@@ -66,11 +96,12 @@ import BraintreeCore
     ///  We recommend that you call this method as early as possible, e.g. at app launch. If that's too early,
     ///  calling it when the customer initiates checkout is also fine.
     ///  Use the return value on your server, e.g. with `Transaction.sale`.
+    /// - Parameter riskCorrelationID: Optional risk correlation ID to associate with this collection. If not provided, one will be generated.
```
agedd: small nit: 

```suggestion
    /// - Parameter riskCorrelationID: Optional. A risk correlation ID to associate with this collection. If not provided, one will be generated.
```

agedd (approved):
lgtm! thanks for making this change 🚀

Sources/BraintreeDataCollector/BTDataCollector.swift
```diff
@@ -27,6 +27,10 @@ import BraintreeCore
     /// - Parameter pairingID: A pairing ID to associate with this clientMetadataID must be 10-32 chars long or null
     /// - Returns: A client metadata ID to send as a header
     /// - Note: This returns a raw client metadata ID, which is not the correct format for device data when creating a transaction. Instead, it is recommended to use `collectDeviceData`.
+    @available(
```
jaxdesmarais, line 31: To the top above the comment here can you add a next major version note? We typically do them like `NEXT_MAJOR_VERSION: remove this method`

Sources/BraintreeDataCollector/BTDataCollector.swift
```diff
@@ -36,28 +40,54 @@ import BraintreeCore
...
             }
-            
-            let clientMetadataID: String = self.generateClientMetadataID(with: configuration)
+
+            let clientMetadataID: String
+            if let riskCorrelationID {
+                clientMetadataID = riskCorrelationID
+            } else {
+                clientMetadataID = self.generateClientMetadataID(with: configuration)
+            }
```
jaxdesmarais: You should be able to do this instead:

```suggestion
            let clientMetadataID = riskCorrelationID ?? self.generateClientMetadataID(with: configuration)
```

Sources/BraintreeDataCollector/BTDataCollector.swift
jaxdesmarais: Nice! Now that we updated this we can add a `NEXT_MAJOR_VERSION` comment to this message to remove it in the future.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## unreleased
 
+* BraintreeDataCollector
+  * Update PPRiskMagnes to 5.6.0 with internal completion handler to ensure device data is not returned until Magnes SDK completes data submission
+  * Deprecate `BTDataCollector.clientMetadataID`
```
```suggestion
  * Deprecate `BTDataCollector.clientMetadataID` - use `BTDataCollector(riskCorrelationID:completion:)` instead
```

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -474,31 +477,42 @@ import BraintreeDataCollector
...
-                case .payPalApp(let url):
-                    self.didPayPalServerAttemptAppSwitch = true
-                    guard (self.isVaultRequest ? approvalURL.baToken : approvalURL.ecToken) != nil else {
-                        self.notifyFailure(
-                            with: self.isVaultRequest ? BTPayPalError.missingBAToken : BTPayPalError.missingECToken,
-                            completion: completion
-                        )
+                let dataCollector = self.dataCollector ?? BTDataCollector(authorization: self.apiClient.authorization.originalValue)
+                dataCollector.collectDeviceData(
+                    riskCorrelationID: self.payPalRequest?.riskCorrelationID
```
jaxdesmarais: Small nit, based on the logic above it looks like if `self.payPalRequest?.riskCorrelationID` is nil we instead want to use the `self.contextID` as the riskCorrelationID.

buzzamus: Updated!

UnitTests/BraintreePayPalTests/MockBTDataCollector.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+@testable import BraintreeDataCollector
+
+class MockBTDataCollector: BTDataCollector {
```
jaxdesmarais: Take it or leave it but since we have 3 of these we could move them to the BraintreeTestShared module and use it across all 3 if we wanted.

buzzamus: Yep this is a better idea. Done!

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
```diff
@@ -68,28 +71,38 @@ import BraintreeDataCollector
...
 
-            let dataCollector = BTDataCollector(authorization: self.apiClient.authorization.originalValue)
-            request.correlationID = dataCollector.clientMetadataID(nil)
-
             guard let configuration else {
                 self.notifyFailure(with: BTLocalPaymentError.fetchConfigurationFailed, completion: completion)
                 return
             }
 
-            if !configuration.isLocalPaymentEnabled {
```
jaxdesmarais, line 79: Small nit but we may be okay to still check both of these before the datacollector runs.

buzzamus: Updated to move checks before the data collector call

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: Looks like this change maybe got reverted?

buzzamus: ah it did, but I didn't realize this was the same line we discussed. I removed it, because when it reverts to the contextID it is the shorter format when I was testing. I can move it back if that is the desired flow, but removing it means a new one will be generated in the longer string value format

buzzamus: Added this back in

jaxdesmarais (approved):

PR #1734, opened 2026-01-22
Add Async/Await To BTVenmoClient

agedd:
### Summary of changes

- Converted the `vault()` and `tokenize()` functions to async/await variants
-  Updated unit tests and added new ones to test async/await functionality

### Checklist

- @agedd

### DIFF
--- a/Sources/BraintreeVenmo/BTVenmoClient.swift
+++ b/Sources/BraintreeVenmo/BTVenmoClient.swift
@@ -122,7 +122,16 @@ import BraintreeCore
             merchantProfileID: merchantProfileID
         )
 
-        let (body, _) = try await apiClient.post("", parameters: graphQLParameters, httpType: .graphQLAPI)
+        let (body, _): (BTJSON?, URLResponse?)
+        do {
+            (body, _) = try await apiClient.post("", parameters: graphQLParameters, httpType: .graphQLAPI)
+        } catch {
+            let nsError = error as NSError
+            let jsonResponse = nsError.userInfo[BTCoreConstants.jsonResponseBodyKey] as? BTJSON
+            let errorMessage = jsonResponse?["error"]["message"].asString()
+            let defaultMessage = "Failed to fetch a Venmo paymentContextID while constructing the requestURL."
+            throw BTVenmoError.invalidRedirectURL(errorMessage ?? defaultMessage)
+        }
 
         guard let body else {
             throw BTVenmoError.invalidBodyReturned


Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -175,17 +177,70 @@ import BraintreeCore
...
     public func tokenize(_ request: BTVenmoRequest) async throws -> BTVenmoAccountNonce {
-        try await withCheckedThrowingContinuation { continuation in
-            tokenize(request) { nonce, error in
-                if let error {
-                    continuation.resume(throwing: error)
-                } else if let nonce {
-                    continuation.resume(returning: nonce)
-                }
+        apiClient.sendAnalyticsEvent(BTVenmoAnalytics.tokenizeStarted, isVaultRequest: shouldVault)
+
```
buzzamus: Are these new analytic events, or how were they firing previously?

buzzamus: Nevermind, I see. I realized the completion handler version is still here.

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -315,32 +329,23 @@ import BraintreeCore
...
+        let (body, _) = try await apiClient.post("v1/payment_methods/venmo_accounts", parameters: parameters)
+        
+        guard let body else {
+            throw BTVenmoError.invalidBodyReturned
+        }
+        
+        let venmoAccountJSON: BTJSON = body["venmoAccounts"][0]
+        
+        if let venmoJSONError = venmoAccountJSON.asError() {
+            throw venmoJSONError
```
buzzamus, line 373: Do we need to send the failure analytic event here?

agedd: hmm yea i think this error should bubble up and get caught on line 263 where we fire the failure analytics event 👀

jaxdesmarais (commented):
It looks like there are still a few methods in this class that we would want to convert along with the GraphQL API calls. Will those be a separate PR or should they be included here?

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -310,37 +249,99 @@ import BraintreeCore
...
-        apiClient.post("v1/payment_methods/venmo_accounts", parameters: parameters) { body, _, error in
-            if let error {
-                self.notifyFailure(with: error, completion: self.appSwitchCompletion)
-                return
-            }
+    
+    // MARK: - Private Helpers
+    
+    private func handlePaymentContextSuccess(_ returnURL: BTVenmoAppSwitchReturnURL) {
+        Task { [weak self] in
```
jaxdesmarais: I think this and the `handleDirectSuccess` method creates new Task blocks from synchronous `handleOpen`, which is called from `handleReturnURL`. The `appSwitchCompletion` is then stored as an instance variable, but the async `Task` may complete after the `startVenmoFlow` continuation has already been fulfilled, causing issues. I think we will instead want to make sure these both and `handleOpen` happens on the main thread.

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -310,37 +249,99 @@ import BraintreeCore
...
+            let venmoAccountNonce = BTVenmoAccountNonce.venmoAccount(with: json)
+            _ = self.notifySuccess(with: venmoAccountNonce)
+        }
+    }
+    
+    private func handleVaultingIfNeeded(for venmoAccountNonce: BTVenmoAccountNonce) async {
+        if shouldVault && apiClient.authorization.type == .clientToken {
+            do {
+                let vaultedNonce = try await vault(venmoAccountNonce.nonce)
+                _ = notifySuccess(with: vaultedNonce)
```
jaxdesmarais: Are we able to just do this here and below:

```suggestion
                return notifySuccess(with: vaultedNonce)
```

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -359,38 +360,40 @@ import BraintreeCore
...
+    private func notifySuccess(with result: BTVenmoAccountNonce) -> BTVenmoAccountNonce {
         apiClient.sendAnalyticsEvent(
             BTVenmoAnalytics.tokenizeSucceeded,
             contextID: contextID,
             isVaultRequest: shouldVault,
             linkType: linkType
         )
-        completion(result, nil)
+        appSwitchCompletion(result, nil)
+        appSwitchCompletion = { _, _ in }
```
jaxdesmarais: I am not sure we need these, I think instead we can do this at the callsite for the completion methods:
```swift
let venmoAccoutNonce = try await _async_method_here_
appSwitchCompletion(venmoAccoutNonce, nil)
```

UnitTests/BraintreeVenmoTests/BTVenmoClient_Tests.swift
```diff
@@ -83,6 +85,13 @@ class BTVenmoClient_Tests: XCTestCase {
         venmoClient.bundle = FakeBundle()
 
         venmoClient.tokenize(venmoRequest) { _, _ in }
+        
+        // Add delay for async execution
+        let expectation = expectation(description: "Wait for async execution")
```
jaxdesmarais: I think we need to add all of these right now because of the race condition with `handleOpen`/`handlePaymentContextSuccess`/`handleDirectSuccess` - if we fix those we should be able to remove all of the delays here to wait for the actual async work.

Sources/BraintreeVenmo/BTVenmoClient.swift
agedd: yep ty for the suggestion, addressed in ae2b0365d3fd8587e6ee42ddea8c896599c9e821

Sources/BraintreeVenmo/BTVenmoClient.swift
agedd: ah great catch, thanks for letting me know! i've addressed this in edc56db14a80b51f9364a539a70df638e492d468

UnitTests/BraintreeVenmoTests/BTVenmoClient_Tests.swift
agedd: removed and cleaned up all the polling delays in edc56db14a80b51f9364a539a70df638e492d468

UnitTests/BraintreeVenmoTests/BTVenmoClient_Tests.swift
```diff
@@ -312,10 +328,9 @@ class BTVenmoClient_Tests: XCTestCase {
         let expectation = expectation(description: "Callback invoked")
         venmoClient.tokenize(venmoRequest) { venmoAccount, error in
             XCTAssertNil(venmoAccount)
-            guard let error = error as NSError? else {return}
-            XCTAssertEqual(error.domain, "com.braintreepayments.BTVenmoErrorDomain")
-            XCTAssertEqual(error.code, BTVenmoError.invalidRedirectURL("").errorCode)
-            XCTAssertEqual(error.localizedDescription, "Failed to fetch a Venmo paymentContextID while constructing the requestURL.")
+            guard let error = error as NSError? else { return }
+            XCTAssertEqual(error.domain, "Venmo Error")
+            XCTAssertEqual(error.code, 100)
```
jaxdesmarais: Is this the correct error or should this be:
```swift
XCTAssertEqual(error.domain, "com.braintreepayments.BTVenmoErrorDomain")
XCTAssertEqual(error.code, BTVenmoError.invalidRedirectURL("").errorCode)
XCTAssertEqual(error.localizedDescription, "Failed to fetch a Venmo paymentContextID while constructing the requestURL.")
```

PR #1737, opened 2026-01-28
Add Async/Await To `BTSEPADirectDebitClient`

agedd:
### Summary of changes

- Convert the internal completion handlers in `BTSEPADirectDebitClient` to use the async function variant.

### Checklist

- @agedd

### DIFF
--- a/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
+++ b/Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
@@ -107,8 +107,7 @@ import BraintreeCore
     func createMandate(request: BTSEPADirectDebitRequest) async throws -> CreateMandateResult {
         apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.createMandateChallengeRequired)
         do {
-            let createMandateResult = try await sepaDirectDebitAPI.createMandate(sepaDirectDebitRequest: request)
-            return createMandateResult
+            return try await sepaDirectDebitAPI.createMandate(sepaDirectDebitRequest: request)
         } catch {
             throw error
         }
--- a/UnitTests/BraintreeSEPADirectDebitTests/BTSEPADirectDebitClient_Tests.swift
+++ b/UnitTests/BraintreeSEPADirectDebitTests/BTSEPADirectDebitClient_Tests.swift
@@ -220,6 +220,60 @@ class BTSEPADirectDebitClient_Tests: XCTestCase {
     }
     
     @MainActor
+    func testTokenizeWithPresentationContext_callsTokenizeWithAlreadyApprovedMandate_returnsSuccess_andSendsAnalytics() async {
+        let mockWebAuthenticationSession = MockWebAuthenticationSession()
+        let mockSepaDirectDebitAPI = SEPADirectDebitAPI(apiClient: mockAPIClient)
+
+        let mockCreateMandateResult = BTJSON(
+            value: [
+                "message": [
+                    "body": [
+                        "sepaDebitAccount": [
+                            "approvalUrl": "null",
+                            "last4": "1234",
+                            "merchantOrPartnerCustomerId": "a-customer-id",
+                            "bankReferenceToken": "a-bank-reference-token",
+                            "mandateType": "ONE_OFF"
+                        ]
+                    ]
+                ]
+            ]
+        )
+
+        let sepaDirectDebitClient = BTSEPADirectDebitClient(authorization: authorization,
+            webAuthenticationSession: mockWebAuthenticationSession,
+            sepaDirectDebitAPI: mockSepaDirectDebitAPI
+        )
+        
+        sepaDirectDebitClient.apiClient = mockAPIClient
+
+        let mockTokenizeResponse = BTJSON(
+            value: [
+                "nonce": "a-fake-payment-method-nonce",
+                "details": [
+                    "ibanLastChars": "1234",
+                    "merchantOrPartnerCustomerId": "a-customer-id",
+                    "mandateType": "RECURRENT"
+                ]
+            ] as [String: Any]
+        )
+
+        mockAPIClient.cannedResponseBody = mockCreateMandateResult
+        mockAPIClient.cannedResponseBody = mockTokenizeResponse
+        
+        do {
+            let nonce = try await sepaDirectDebitClient.tokenize(sepaDirectDebitRequest)
+            XCTAssertEqual(nonce.nonce, "a-fake-payment-method-nonce")
+            XCTAssertEqual(nonce.ibanLastFour, "1234")
+            XCTAssertEqual(nonce.customerID, "a-customer-id")
+            XCTAssertEqual(nonce.mandateType?.description, "RECURRENT")
+            XCTAssertTrue(self.mockAPIClient.postedAnalyticsEvents.contains(BTSEPADirectAnalytics.tokenizeSucceeded))
+        } catch let error as NSError {
+            XCTFail("This request should be successful.")
+        }
+    }
+    
+    @MainActor
     func testTokenize_callsTokenizeWithAlreadyApprovedMandate_returnsSuccess_andSendsAnalytics() async {
         let mockWebAuthenticationSession = MockWebAuthenticationSession()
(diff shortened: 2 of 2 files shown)


Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -93,14 +93,33 @@ import BraintreeCore
...
+            // if the SEPADirectDebitAPI.tokenize API calls returns a "null" URL, the URL has already been approved.
+            if createMandateResult.approvalURL == CreateMandateResult.mandateAlreadyApprovedURLString {
+                apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.createMandateSucceeded)
+                return try await tokenize(createMandateResult: createMandateResult)
+            } else if let url = URL(string: createMandateResult.approvalURL) {
+                let success = try await startAuthenticationSession(url: url, context: self)
+                if success {
+                    return try await tokenize(createMandateResult: createMandateResult)
+                } else {
+                    throw BTSEPADirectDebitError.webFlowCanceled
```
buzzamus: take it or leave it suggestion: This section is a little hard to follow at first just due to the varying levels of if statements. I don't really have a better suggestion to make it more readable, but just a thought. Otherwise LGTM

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -93,14 +93,33 @@ import BraintreeCore
...
                 }
+            } else {
+                apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.createMandateFailed)
```
jaxdesmarais, line 79: Do we want a similar help method to PR #1732 for handling errors?

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -93,14 +93,33 @@ import BraintreeCore
...
+                throw BTSEPADirectDebitError.approvalURLInvalid
             }
+        } catch {
+            apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.createMandateFailed)
+            apiClient.sendAnalyticsEvent(
```
jaxdesmarais: Should we be sending both of these here?

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -118,6 +137,26 @@ import BraintreeCore
         }
     }
     
+    /// Calls `SEPADirectDebitAPI.createMandate` to create the mandate and returns the `approvalURL` in the `CreateMandateResult`
+    /// that is used to display the mandate to the user during the web flow.
+    /// - Parameter request: a `BTSEPADirectDebitRequest`
+    /// - Returns: A `CreateMandateResult` containing the approval URL
+    /// - Throws: An `Error` describing the failure
+    func createMandate(request: BTSEPADirectDebitRequest) async throws -> CreateMandateResult {
```
jaxdesmarais, line 107: We should be able to delete the competition ones and call the async/await method in task blocks. For both this and the new method below.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
agedd: ahh yep agreed it's a bit confusing with the nested ifs, tried to clean it up a bit in ee3395d942c19ca3056c849895930c44784a99fa

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
agedd: ah yep good catch, cleaned up the notify failure method to only fire the event similar to the PR #1732

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -42,98 +42,91 @@ import BraintreeCore
...
+            // If the approval URL is invalid, fail early
+            guard createMandateResult.approvalURL != CreateMandateResult.mandateAlreadyApprovedURLString,
+                let url = URL(string: createMandateResult.approvalURL) else {
+                // Mandate already approved, skip authentication
+                if createMandateResult.approvalURL == CreateMandateResult.mandateAlreadyApprovedURLString {
+                    apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.createMandateSucceeded)
+                    return try await tokenize(createMandateResult: createMandateResult)
                 }
+                // Invalid approval URL
+                apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.createMandateFailed)
```
jaxdesmarais, line 79: I think we want to update this to a send failure helper method for sending analytics with the correct failue and error events. I think the amex PR and some other PRs use this pattern.

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -42,98 +42,91 @@ import BraintreeCore
...
+            // Authenticate and tokenize
+            let success = try await startAuthenticationSession(url: url, context: self)
+            guard success else {
+                throw BTSEPADirectDebitError.webFlowCanceled
             }
+            
+            apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.createMandateSucceeded)
+            return try await tokenize(createMandateResult: createMandateResult)
+        } catch {
+            apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.createMandateFailed)
```
jaxdesmarais, line 94: If you do a find for `sendAnalyticsEvent` we should be able to replace anything outside of the started or 2 helper methods with `notifyFailure(with: error_here)`

agedd: ah yep i checked and made sure to use `notifyFailure `wherever appropriate. i noticed we've been using `sendAnalyticsEvent `for the additional SEPA analytics events, so i kept that pattern consistent. 

my thinking was this creates a clean separation between conversion and non-conversion events. thoughts?

jaxdesmarais: Gotcha - didn't realize that was the pattern we were following before. Thanks for explaining!

Sources/BraintreeSEPADirectDebit/BTSEPADirectDebitClient.swift
```diff
@@ -42,98 +42,89 @@ import BraintreeCore
...
+    /// - Parameter request: a `BTSEPADirectDebitRequest`
+    /// - Returns: A `CreateMandateResult` containing the approval URL
+    /// - Throws: An `Error` describing the failure
+    func createMandate(request: BTSEPADirectDebitRequest) async throws -> CreateMandateResult {
         apiClient.sendAnalyticsEvent(BTSEPADirectAnalytics.createMandateChallengeRequired)
-        sepaDirectDebitAPI.createMandate(sepaDirectDebitRequest: request) { result, error in
-            completion(result, error)
+        do {
+            let createMandateResult = try await sepaDirectDebitAPI.createMandate(sepaDirectDebitRequest: request)
+            return createMandateResult
```
jaxdesmarais: Take it or leave it, but not sure if we can do this here:

```suggestion
            return try await sepaDirectDebitAPI.createMandate(sepaDirectDebitRequest: request)
```

UnitTests/BraintreeSEPADirectDebitTests/BTSEPADirectDebitClient_Tests.swift
```diff
@@ -208,49 +218,9 @@ class BTSEPADirectDebitClient_Tests: XCTestCase {
             }
         }
     }
-
-    func testTokenizeWithPresentationContext_handleWebAuthenticationSessionSuccessURLInvalid_returnsError_andSendsAnalytics() {
```
jaxdesmarais, line 212: I think this test is useful, can we add it back?

PR #1738, opened 2026-01-29
Convert BTCardClient to Async/Await

buzzamus:
### Summary of changes

- Updated BTCardClient to use Async/Await instead of completion handlers by creating new Tokenize method.

### Checklist

- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeCard/BTCardClient.swift
+++ b/Sources/BraintreeCard/BTCardClient.swift
@@ -73,12 +73,12 @@ import BraintreeCore
                     let cardNonce = BTCardNonce(graphQLJSON: cardJSON)
                     return notifySuccess(with: cardNonce)
                 } catch {
-                    let nsError = error as NSError
-                    let response: HTTPURLResponse? = nsError.userInfo[BTCoreConstants.urlResponseKey] as? HTTPURLResponse
+                    let error = error as NSError
+                    let response: HTTPURLResponse? = error.userInfo[BTCoreConstants.urlResponseKey] as? HTTPURLResponse
                     var callbackError: Error = error
 
                     if response?.statusCode == 422 {
-                        callbackError = constructCallbackError(with: nsError.userInfo, error: nsError) ?? error
+                        callbackError = constructCallbackError(with: error.userInfo, error: error) ?? error
                     }
 
                     notifyFailure(with: callbackError)


Sources/BraintreeCard/BTCardClient.swift
```diff
@@ -32,97 +32,90 @@ import BraintreeCore
...
-            }
+        do {
+            let configuration = try await apiClient.fetchOrReturnRemoteConfiguration()
 
-            if self.isGraphQLEnabled(for: configuration) {
+            if isGraphQLEnabled(for: configuration) {
                 if card.authenticationInsightRequested && card.merchantAccountID == nil {
-                    self.notifyFailure(with: BTCardError.integration, completion: completion)
-                    return
+                    throw BTCardError.integration
```
jaxdesmarais, line 58: Do we need to send analytics here?

Sources/BraintreeCard/BTCardClient.swift
```diff
@@ -32,97 +32,90 @@ import BraintreeCore
...
-                        return
+                        notifyFailure(with: cardJSONError)
+                        throw cardJSONError
                     }
 
                     let cardNonce = BTCardNonce(graphQLJSON: cardJSON)
+                    notifySuccess()
+                    return cardNonce
+                } catch {
+                    let nsError = error as NSError
```
jaxdesmarais: Can we do this instead here:

```suggestion
                    let error = error as NSError
```

Sources/BraintreeCard/BTCardClient.swift
```diff
@@ -182,19 +175,14 @@ import BraintreeCore
...
-    
+
     // MARK: - Analytics Helper Methods
-    
-    private func notifySuccess(
-        with result: BTCardNonce,
-        completion: @escaping (BTCardNonce?, Error?) -> Void
-    ) {
+
+    private func notifySuccess() {
```
jaxdesmarais: Can we have this return a `BTCardNonce` and do analytics and the result together?

Sources/BraintreeCard/BTCardClient.swift
buzzamus: Yep sorry missed this. Updated

Sources/BraintreeCard/BTCardClient.swift
buzzamus: updated!

Sources/BraintreeCard/BTCardClient.swift
buzzamus: updated to return here as well

Sources/BraintreeCard/BTCardClient.swift
```diff
@@ -32,97 +32,89 @@ import BraintreeCore
...
-                        self.notifyFailure(with: cardJSONError, completion: completion)
-                        return
+                        notifyFailure(with: cardJSONError)
+                        throw cardJSONError
                     }
 
                     let cardNonce = BTCardNonce(graphQLJSON: cardJSON)
+                    return notifySuccess(with: cardNonce)
+                } catch {
+                    let nsError = error as NSError
```
jaxdesmarais: Not blocking, but looks like we changed below to this but not this one:

```suggestion
                    let error = error as NSError
```

PR #1739, opened 2026-01-30
Convert BTApplePayClient to Async/Await

buzzamus:
### Summary of changes

- Convert BTApplePayClient to Async/Await and remove underlying completion handlers.

### Checklist

- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeApplePay/BTApplePayClient.swift
+++ b/Sources/BraintreeApplePay/BTApplePayClient.swift
@@ -60,8 +60,7 @@ import BraintreeCore
             paymentRequest.merchantIdentifier = configuration.applePayMerchantIdentifier ?? ""
             paymentRequest.supportedNetworks = configuration.applePaySupportedNetworks ?? []
 
-            notifyPaymentRequestSuccess()
-            return paymentRequest
+            return notifyPaymentRequestSuccess(with: paymentRequest)
         } catch {
             sendPaymentRequestFailureAnalytics(with: error)
             throw error
@@ -134,8 +133,7 @@ import BraintreeCore
                 throw BTApplePayError.failedToCreateNonce
             }
 
-            notifyTokenizeSuccess()
-            return applePayNonce
+            return notifyTokenizeSuccess(with: applePayNonce)
         } catch {
             sendTokenizeFailureAnalytics(with: error)
             throw error
@@ -144,16 +142,18 @@ import BraintreeCore
     
     // MARK: - Analytics Helper Methods
 
-    private func notifyTokenizeSuccess() {
+    private func notifyTokenizeSuccess(with result: BTApplePayCardNonce) -> BTApplePayCardNonce {
         apiClient.sendAnalyticsEvent(BTApplePayAnalytics.tokenizeSucceeded)
+        return result
     }
 
     private func sendTokenizeFailureAnalytics(with error: Error) {
         apiClient.sendAnalyticsEvent(BTApplePayAnalytics.tokenizeFailed, errorDescription: error.localizedDescription)
     }
 
-    private func notifyPaymentRequestSuccess() {
+    private func notifyPaymentRequestSuccess(with result: PKPaymentRequest) -> PKPaymentRequest {
         apiClient.sendAnalyticsEvent(BTApplePayAnalytics.paymentRequestSucceeded)
+        return result
     }
 
     private func sendPaymentRequestFailureAnalytics(with error: Error) {


Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -138,48 +111,52 @@ import BraintreeCore
...
+                sendTokenizeFailureAnalytics(with: BTApplePayError.failedToCreateNonce)
+                throw BTApplePayError.failedToCreateNonce
             }
+
+            notifyTokenizeSuccess()
+            return applePayNonce
+        } catch {
+            sendTokenizeFailureAnalytics(with: error)
+            throw error
         }
```
buzzamus, line 140: Tried to follow the same pattern of throwing catching once that we discussed in the Amex ticket

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -29,17 +29,29 @@ import BraintreeCore
     /// - Parameter completion: A completion block that returns the payment request or an error.
     @objc(makePaymentRequest:)
     public func makePaymentRequest(completion: @escaping (PKPaymentRequest?, Error?) -> Void) {
+        Task { @MainActor in
```
stechiu, line 32: Do we need @ MainActor here?

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -138,48 +111,52 @@ import BraintreeCore
...
-        try await withCheckedThrowingContinuation { continuation in
-            tokenize(payment) { applePayNonce, error in
-                if let error {
-                    continuation.resume(throwing: error)
-                } else if let applePayNonce {
-                    continuation.resume(returning: applePayNonce)
-                }
+        apiClient.sendAnalyticsEvent(BTApplePayAnalytics.tokenizeStarted)
+
+        do {
```
stechiu, line 115: Should this be wrapped in `Task` block?

Sources/BraintreeApplePay/BTApplePayClient.swift
buzzamus: Yeah, it has to be added, because otherwise it causes crashes. Since the results of this task block will be used to update the UI it avoids crashes or hanging result handling

Sources/BraintreeApplePay/BTApplePayClient.swift
buzzamus: The actual call to this method is already wrapped in a Task so, it is just indirectly

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -138,48 +111,52 @@ import BraintreeCore
...
+            let configuration = try await apiClient.fetchOrReturnRemoteConfiguration()
+
+            guard configuration.isApplePayEnabled else {
+                sendTokenizeFailureAnalytics(with: BTApplePayError.unsupported)
+                throw BTApplePayError.unsupported
+            }
+
+            let parameters = BTApplePaymentTokensRequest(token: payment.token)
+            let (body, _) = try await apiClient.post("v1/payment_methods/apple_payment_tokens", parameters: parameters)
```
jaxdesmarais, line 124: Not for this PR at all but I wonder if we can get rid of the unnamed param and just return the body at some point. Not sure if it's used anywhere, but could be a fun open dev thing to look into.

Sources/BraintreeApplePay/BTApplePayClient.swift
```diff
@@ -138,48 +111,52 @@ import BraintreeCore
...
     }
     
     // MARK: - Analytics Helper Methods
-    
-    private func notifySuccess(
-        with result: BTApplePayCardNonce,
-        completion: @escaping (BTApplePayCardNonce?, Error?) -> Void
-    ) {
+
+    private func notifyTokenizeSuccess() {
```
jaxdesmarais: I had thought for the amex work we were able to do both for the success case but maybe I am misremembering completely.

buzzamus: ah nope you are right. I'll update

buzzamus: Updated! Although initially I thought you meant return the result and send analytic like you mentioned in the other PR, but now I am wondering if you mean combine the two type of success/failure events? This is certainly doable by making the arguments/return generics, but this class is a bit different than the other client flows since this has both a tokenize and payment request call in it vs. just a tokenize call for other payment methods.

PR #1741, opened 2026-02-03
Add Async/Await To `BTLocalPayments`

stechiu:
### Summary of changes

- Convert the internal completion handlers in BTLocalPayments to async/await
- Tested on demo app and confirmed that everything works as expected

### Checklist

- @stechiu

### DIFF
--- a/Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
+++ b/Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
@@ -223,11 +223,13 @@ import BraintreeDataCollector
                 }
             } sessionDidCancel: { [weak self] in
                 guard let self else { return }
+                let canceledError = BTLocalPaymentError.canceled(self.request?.paymentType ?? "")
                 if !webSessionReturned {
                     apiClient.sendAnalyticsEvent(BTLocalPaymentAnalytics.browserLoginAlertCanceled)
                 }
                 apiClient.sendAnalyticsEvent(BTLocalPaymentAnalytics.paymentCanceled)
-                continuation.resume(throwing: BTLocalPaymentError.canceled(self.request?.paymentType ?? ""))
+                sendFailureAnalytics(with: canceledError)
+                continuation.resume(throwing: canceledError)
             }
         }
     }
--- a/UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
+++ b/UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
@@ -251,9 +251,7 @@ class BTLocalPaymentClient_UnitTests: XCTestCase {
         client.apiClient = mockAPIClient
         client.webAuthenticationSession = makeCancelingWebSession()
 
-        Task {
-            _ = try? await client.start(localPaymentRequest)
-        }
+        _ = try? await client.start(localPaymentRequest) 
 
         XCTAssertEqual(mockAPIClient.postedContextID, "123aaa-123-543-777")
     }


Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
```diff
@@ -115,74 +115,73 @@ import BraintreeDataCollector
...
-                NSLog(
-                    """
-                    %@ Payment cannot be processed: the redirectUrl or paymentToken is nil. Contact Braintree support if the error persists.
-                    """,
-                    BTLogLevelDescription.string(for: .critical)
-                )
-                self.notifyFailure(with: BTLocalPaymentError.appSwitchFailed, completion: self.merchantCompletion)
+            } catch {
+                notifyFailure(with: error, completion: merchantCompletion)
                 return
```
buzzamus: Instead of referencing the merchant completion and using it throughout, the pattern we have been doing is just keep it isolated to the public method that starts everything and then here the response would be returned to bubble up to the public method to use the completion for. 

So, for example:

```
  public func start(_ request: BTLocalPaymentRequest, completion: @escaping (BTLocalPaymentResult?, Error?) -> Void) {
      Task { @MainActor in
          do {
              let result = try await start(request)
              completion(result, nil)
          } catch {
              completion(nil, error)
          }
      }
  }
```

And then for the configuration request in the public start there is also now an async/await version in the APIClient, so it can be made with:
```let configuration = try await apiClient.fetchOrReturnRemoteConfiguration()```

File: Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift [hunk 2]
```diff
@@ -241,10 +242,7 @@ import BraintreeDataCollector
 
     // MARK: - Analytics Helper Methods
 
-    private func notifySuccess(
-        with result: BTLocalPaymentResult,
-        completion: @escaping (BTLocalPaymentResult?, Error?) -> Void
-    ) {
+    private func notifySuccess(with result: BTLocalPaymentResult, completion: @escaping (BTLocalPaymentResult?, Error?) -> Void) {
         apiClient.sendAnalyticsEvent(BTLocalPaymentAnalytics.paymentSucceeded, contextID: contextID)
         completion(result, nil)
```
buzzamus: for the notifySuccess and failure we can also get rid of the completion handler and just send the event. For the success event we have actually been using it for the return of the result. Example: https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreeCard/BTCardClient.swift#L180-L183

Then, we can just use ``` return notifySuccess(with: tokenizedLocalPayment)```

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
stechiu: Thanks I'll make the update

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
```diff
@@ -115,74 +115,73 @@ import BraintreeDataCollector
         webSessionReturned = true
     }
 
-    func handleOpen(_ url: URL) {
-        // canceled case
```
jaxdesmarais, line 119: Can we add this docstring back?

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
```diff
@@ -115,74 +115,73 @@ import BraintreeDataCollector
...
+    func handleOpen(_ url: URL) async {
```
jaxdesmarais: This can either use a task block or be marked async but we do not need both. We can either remove async here and use the task block or leave async and use do/catch with await below.

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
```diff
@@ -115,74 +115,73 @@ import BraintreeDataCollector
...
-            }
-
-            if
-                let paymentID = body?["paymentResource"]["paymentToken"].asString(),
-                let approvalURLString = body?["paymentResource"]["redirectUrl"].asString(),
-                let url = URL(string: approvalURLString) {
-
-                if !paymentID.isEmpty {
-                    self.contextID = paymentID
+        Task {
```
jaxdesmarais: This currently creates an unstructured `Task` from a synchronous context. This is fire-and-forget, so errors might not propagate correctly. Since this is private I _think_ we can make it async and use do/catch outside of a task. If it must stay synchronous, we can add `@MainActor` to the Task. Would need to test to see which is best.

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
```diff
@@ -213,7 +212,9 @@ import BraintreeDataCollector
             }
 
             if let url {
-                handleOpen(url)
+                Task { [weak self] in
```
jaxdesmarais: Are there instances where this will be deallocated unexpectedly that we should be concerned with?

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
jaxdesmarais: ➕ 1 we can get rid of the completion here and in the failure by following the previously established patterns @buzzamus linked.

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
```diff
@@ -76,8 +74,14 @@ class BTLocalPaymentClient_UnitTests: XCTestCase {
     func testStartPayment_postsAllCreationParameters() {
         mockAPIClient.cannedConfigurationResponseBody = BTJSON(value: [ "paypalEnabled": true ])
 
+        let mockWebAuthenticationSession = MockWebAuthenticationSession()
+        mockWebAuthenticationSession.cannedSessionDidDisplay = true
+        mockWebAuthenticationSession.cannedResponseURL = nil
+        mockWebAuthenticationSession.cannedErrorResponse = nil
```
jaxdesmarais: Why is this now needed?

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
```diff
@@ -116,7 +120,13 @@ class BTLocalPaymentClient_UnitTests: XCTestCase {
         
         client.start(paymentRequest) { _, _ in }
 
-        XCTAssertEqual(mockAPIClient.lastPOSTParameters!["merchant_account_id"] as? String, "customer-nl-merchant-account")
+        let expectation = expectation(description: "Start payment completes")
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
jaxdesmarais: I think this will lead to some flakiness on CI. Is there another way we can approach this?

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
```diff
@@ -233,8 +243,14 @@ class BTLocalPaymentClient_UnitTests: XCTestCase {
         let client = BTLocalPaymentClient(authorization: tempClientToken)
         client.apiClient = mockAPIClient
         client.webAuthenticationSession = mockWebAuthenticationSession
-        client.start(localPaymentRequest) { _, _ in }
+        
+        let expectation = expectation(description: "Start payment completes")
```
jaxdesmarais: Are you able to share why this is needed now?

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
```diff
@@ -316,16 +344,27 @@ class BTLocalPaymentClient_UnitTests: XCTestCase {
             ]
         )
 
-        client.start(localPaymentRequest) { _, _ in }
+        let expectation = self.expectation(description: "Tokenization completes")
+        client.start(localPaymentRequest) { _, _ in
+            expectation.fulfill()
+        }
+
+        try? await Task.sleep(nanoseconds: 100_000_000)
```
jaxdesmarais: This is a timing-based test that can be flaky. Tests might pass/fail randomly depending on system load. We should exceptions here instead of a sleep.

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
```diff
@@ -340,17 +379,26 @@ class BTLocalPaymentClient_UnitTests: XCTestCase {
...
         )
 
-        client.start(localPaymentRequest) { _, _ in }
+        let expectation = self.expectation(description: "Tokenization completes")
+        client.start(localPaymentRequest) { _, _ in
+            expectation.fulfill()
+        }
 
-        client.handleOpen(URL(string: "com.braintreepayments.demo.payments://x-callback-url/braintree/local-payment/success")!)
+        try? await Task.sleep(nanoseconds: 100_000_000)
```
jaxdesmarais: Same here to update test vs using sleep

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
```diff
@@ -380,14 +430,24 @@ class BTLocalPaymentClient_UnitTests: XCTestCase {
...
+            }
             XCTAssertEqual(error.domain, BTLocalPaymentError.errorDomain)
-            XCTAssertEqual(error.code, BTLocalPaymentError.canceled("flow-type").errorCode)
+            XCTAssertEqual(error.code, BTLocalPaymentError.canceled("ideal").errorCode)
+            expectation.fulfill()
         }
 
-        client.handleOpen(URL(string: "com.braintreepayments.demo.payments://x-callback-url/braintree/local-payment/cancel?paymentId=PAY-79C90584AX7152104LNY4OCY")!)
+        // Wait for start to complete, then trigger the cancel
+        try? await Task.sleep(nanoseconds: 100_000_000) // 0.1 seconds
```
jaxdesmarais: Same here to remove sleep

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
```diff
@@ -430,16 +491,48 @@ class BTLocalPaymentClient_UnitTests: XCTestCase {
...
+
+        var startCompleted = false
+        client.start(paymentRequest) { _, _ in
+            if startCompleted {
+                tokenizeExpectation.fulfill()
+            } else {
+                startCompleted = true
+                startExpectation.fulfill()
+            }
+        }
```
jaxdesmarais: Are you able to share more details on this change? I am not sure I understand why it was needed? This likely points to an issue in the code where we are not structuring things correctly.

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
```diff
@@ -430,16 +491,48 @@ class BTLocalPaymentClient_UnitTests: XCTestCase {
...
-        client.handleOpen(
+        await client.handleOpen(
             URL(string: "com.braintreepayments.demo.payments://x-callback-url/braintree/local-payment/success")!
         )
 
+        // Wait for tokenization to complete
+        await fulfillment(of: [tokenizeExpectation], timeout: 2)
+
+        // Wait for POST to completes
+        try? await Task.sleep(nanoseconds: 50_000_000)
```
jaxdesmarais: We will want to remove this sleep as well.

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
stechiu: I didn't reply yesterday but yes, I'm trying out removal of completion in success and failure. Started working on it yesterday and I'm halfway through fixing unit test failures. I had followed this pattern before but ran into many issues and determined that keeping them as is until there was team discussion was better

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
stechiu: I'll try alternatives. The test was changed to `DispatchQueue.main.asyncAfter(deadline: .now()` after looking at other unit tests in our code base and how similar test cases have been handled

jaxdesmarais: It looks like that code is only used in 3 cases unrelated to async/await work where we need a delay to simulate certain mocked behavior: https://github.com/search?q=repo%3Abraintree%2Fbraintree_ios%20DispatchQueue.main.asyncAfter(deadline%3A%20.now()%20&type=code

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
stechiu: I'll keep `async` then since `handleOpen` is used elsewhere in the file asynchronously and in unit tests

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
stechiu: I see. That may be why I have unit test errors related to how async vs sync is getting fired. I'll remove Task and try with async  so that errors don't only fire in the Test. Will update here if I find a better solution

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
stechiu: It's for in the cases where `self` in `self?.handleOpen(url)` deallocates before Task completes, like if local payments is canceled.

I've updated to the below so the `self` guard statement is moved into the task block and `self` isn't set as an optional, although the guard statement could be left at the beginning of this function
```
            if let url {
                Task { [weak self] in
                    guard let self else {
                        NSLog("%@ BTLocalPaymentClient has been deallocated.", BTLogLevelDescription.string(for: .critical))
                        return
                    }
                    await self.handleOpen(url)
                }
```

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
stechiu: This was added in due to consistent failures with the original `waitForExistence` timeout failures. This block was handling how `BTLocalPaymentClient` was originally setup. I've made updates to `BTLocalPaymentClient` based on your and Brent's suggestion so we  don't need this anymore

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
stechiu: There was a failure in one of the assertions where some of the post parameters couldn't be found so I added this block to simulate a browser switch. Since updating `BTLocalPaymentClient` after addressing PR comments, I tested and this isn't needed anymore

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
stechiu: Thanks for pointing out that this one isn't ideal for async/await! There are different ways to approach this so I'll update more robust alternative

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
stechiu: I made changes to this test and removed this block of code on Tuesday but didn't push the update here since it was a WIP. It was here was so tests wait for async payment flow to complete before asserting

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
```diff
@@ -115,146 +109,146 @@ import BraintreeDataCollector
...
         }
 
-        guard let url else {
-            notifyFailure(with: BTLocalPaymentError.missingRedirectURL, completion: merchantCompletion)
-            return
+        if !paymentID.isEmpty {
+            self.contextID = paymentID
         }
-
-        webSessionReturned = false
```
jaxdesmarais, line 202: Do we need to reset this still at some point?

jaxdesmarais: Oh ignore me I see it below, just needed to keep scrolling. 😄

UnitTests/BraintreeLocalPaymentTests/BTLocalPaymentClient_UnitTests.swift
```diff
@@ -21,64 +43,64 @@ class BTLocalPaymentClient_UnitTests: XCTestCase {
...
-            expectation.fulfill()
+        do {
+            _ = try await client.start(localPaymentRequest)
+            XCTFail("This request should throw an error.")
+        } catch let error as NSError {
+            XCTAssertEqual(error.domain, BTAPIClientError.errorDomain)
+            XCTAssertEqual(error.code, BTAPIClientError.configurationUnavailable.errorCode)
         }
-
-        waitForExpectations(timeout: 2)
```
jaxdesmarais, line 38: Take it or leave it, we can maybe add the following to this test, `testStartPayment_returnsErrorWhenLocalPaymentsNotEnabled` and `testStartPayment_returnsErrorWhenLocalPaymentDelegateIsNil` to ensure we are sending the failure events as expected:
```swift
XCTAssertTrue(mockAPIClient.postedAnalyticsEvents.contains(BTLocalPaymentAnalytics.paymentFailed))
// or whatever event we are supposed to be sending
```

jaxdesmarais: Looks like there is another failing test too in the local payments client tests

Sources/BraintreeLocalPayment/BTLocalPaymentClient.swift
```diff
@@ -57,204 +54,198 @@ import BraintreeDataCollector
...
+    
     // MARK: - Internal Methods
 
     func applicationDidBecomeActive(notification: Notification) {
         webSessionReturned = true
     }
 
-    func handleOpen(_ url: URL) {
+    func handleOpen(_ url: URL) async throws -> BTLocalPaymentResult {
         // canceled case
```
buzzamus, line 108: This is listed as an internal method and I only ever see it called here, but just wanted to make sure this method would never be specifically called anywhere where this would be considered a breaking change is it? I don't know enough yet about local payment methods, but assuming it is fine since it is really an internal method.

agedd: ah yep great question — yep, it wouldn't be a breaking change and i think it is only ever called in `launchWebSession`

PR #1743, opened 2026-02-06
BTPayPalMessagingview - Update to async/await

stechiu:
### Summary of changes

- Convert the internal completion handlers in BTPayPalMessagingView to use the async/await
- Tested PayPal Messaging in Demo app and works as expected

### Checklist

- @stechiu

### DIFF
--- a/Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
+++ b/Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
@@ -44,7 +44,7 @@ public class BTPayPalMessagingView: UIView {
         
         apiClient.sendAnalyticsEvent(BTPayPalMessagingAnalytics.started)
 
-        Task {
+        Task { @MainActor in
             do {
                 let configuration = try await apiClient.fetchOrReturnRemoteConfiguration()
                 
--- a/Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
+++ b/Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
@@ -22,7 +22,7 @@ public class BTPayPalMessagingView: UIView {
     ///  Initializes a `BTPayPalMessagingView`.
     /// - Parameter authorization: A valid client token or tokenization key used to authorize API calls.
     public init(authorization: String) {
-        self.apiClient = BTAPIClient(authorization: authorization)
+        apiClient = BTAPIClient(authorization: authorization)
 
         super.init(frame: .zero)
     }
@@ -49,7 +49,7 @@ public class BTPayPalMessagingView: UIView {
                 let configuration = try await apiClient.fetchOrReturnRemoteConfiguration()
                 
                 guard let clientID = configuration.json?["paypal"]["clientId"].asString() else {
-                    self.notifyFailure(with: BTPayPalMessagingError.payPalClientIDNotFound)
+                    notifyFailure(with: BTPayPalMessagingError.payPalClientIDNotFound)
                     return
                 }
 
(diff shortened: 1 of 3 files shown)


buzzamus (approved):
😎

Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
```diff
@@ -43,42 +43,39 @@ public class BTPayPalMessagingView: UIView {
...
-            guard let configuration else {
-                self.notifyFailure(with: BTPayPalMessagingError.fetchConfigurationFailed)
-                return
-            }
+        Task {
+            do {
+                let configuration = try await apiClient.fetchOrReturnRemoteConfiguration()
+                
+                guard let clientID = configuration.json?["paypal"]["clientId"].asString() else {
+                    self.notifyFailure(with: BTPayPalMessagingError.payPalClientIDNotFound)
```
jaxdesmarais: Small cleanup: we can remove `self` from all of these, same with the one below. Since we are no longer in the completion block of  `fetchOrReturnRemoteConfiguration` we no longer need it. 😎

stechiu: Thanks, I'll update and check if there are other `self`'s that can be removed!

agedd (approved):
lgtm! 🚀

Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
```diff
@@ -43,42 +43,39 @@ public class BTPayPalMessagingView: UIView {
...
-            if let error {
-                self.notifyFailure(with: error)
-                return
-            }
 
-            guard let configuration else {
-                self.notifyFailure(with: BTPayPalMessagingError.fetchConfigurationFailed)
-                return
-            }
+        Task {
```
jaxdesmarais: Are we able to check if this needs the following:

```suggestion
        Task { @MainActor in
```
Since `start` updates the UI I would assume we may want some of this task to happen on the main thread.

UnitTests/BraintreePayPalMessagingTests/BTPayPalMessagingView_Tests.swift
```diff
@@ -9,32 +9,45 @@ final class BTPayPalMessagingView_Tests: XCTestCase {
...
+        mockDelegate.didReceiveErrorExpectation = expectation
+
         payPalMessageView.start()
 
-        XCTAssertEqual(mockDelegate.error as? BTPayPalMessagingError, BTPayPalMessagingError.fetchConfigurationFailed)
-        XCTAssertEqual((mockDelegate.error as? BTPayPalMessagingError)?.errorCode, 0)
-        XCTAssertEqual((mockDelegate.error as? BTPayPalMessagingError)?.errorDescription, "Failed to fetch Braintree configuration.")
+        await fulfillment(of: [expectation], timeout: 2)
+
+        XCTAssertNotNil(mockDelegate.error)
```
jaxdesmarais: We want to assert the actual error messages here - right now we are just testing the error is not nil not the specific errors which reduces the test specificity.

stechiu: Got it, I accidentally asserted `not nil` on the errors and have test added for asserting the errers

Sources/BraintreePayPalMessaging/BTPayPalMessagingView.swift
stechiu: I looked into it and added `@MainActor` to Task line so it would guarantee all code runs on the main thread and since `start()` calls different functions that update the UI

PR #1744, opened 2026-02-06
Update BTDataCollector To Use Async/Await

buzzamus:
### Summary of changes

- Convert BTDataCollector to use Async/Await calls and remove completion handler closures.

### Checklist

- [ ] Added a changelog entry

- @buzzamus


agedd (approved):
🚀

PR #1745, opened 2026-02-06
[OPEN DEV] Update `AmexViewController` to Use `async/await` Methods

jaxdesmarais:
### Summary of changes

- Now that we are converting internal methods to all use async/await I wanted to spend an open dev converting 1 of the demo app features to do the same - figured we could slowly convert these as we had some time on open devs or otherwise

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/Demo/Application/Features/AmexViewController.swift
+++ b/Demo/Application/Features/AmexViewController.swift
@@ -27,18 +27,24 @@ class AmexViewController: PaymentButtonBaseViewController {
     }
 
     @objc func tappedValidCard() {
-        getRewards(for: "371260714673002")
+        Task {
+            await getRewards(for: "371260714673002")
+        }
     }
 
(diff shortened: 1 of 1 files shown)


Demo/Application/Features/AmexViewController.swift
```diff
@@ -27,18 +27,24 @@ class AmexViewController: PaymentButtonBaseViewController {
     }
 
     @objc func tappedValidCard() {
-        getRewards(for: "371260714673002")
+        Task {
```
jaxdesmarais, line 30: I thought it was cleanest to add the `Task` at the boundary where you go from synchronous (@objc) to async so wrapped `getRewards` in a task block in each of these. But if folks feel differently, let me know!

stechiu:
LGTM. Question - Is this what we'll recommend merchants to do and update our dev docs with async/await?

PR #1747, opened 2026-02-10
Revert "BTPayPalMessagingview - Update to async/await"

stechiu:
Reverts braintree/braintree_ios#1743


jaxdesmarais:
Merging in as their are failures that seem to be related to this change

PR #1748, opened 2026-02-11
Convert BTPayPalClient to Async

buzzamus:
### Summary of changes

- Converted BTPayPalClient to async/await

This conversion is a little different than the previous patterns we have established. Due to the app switch and browser flows, completion handlers are a bit more prevalent and so, are passed around a bit more than other modules due to different scenarios calling them.

### Checklist

- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreePayPal/BTPayPalClient.swift
+++ b/Sources/BraintreePayPal/BTPayPalClient.swift
@@ -400,7 +400,6 @@ import BraintreeDataCollector
         guard let returnURL = BTPayPalReturnURL(.payPalApp(url: url)) else {
             notifyFailure(with: BTPayPalError.invalidURL("App Switch return URL cannot be nil"))
             appSwitchCompletion(nil, BTPayPalError.invalidURL("App Switch return URL cannot be nil"))
-            BTPayPalClient.payPalClient = nil
             return
         }
 
@@ -409,7 +408,6 @@ import BraintreeDataCollector
             guard let payPalRequest else {
                 notifyFailure(with: BTPayPalError.missingPayPalRequest)
                 appSwitchCompletion(nil, BTPayPalError.missingPayPalRequest)
-                BTPayPalClient.payPalClient = nil
                 return
             }
 
@@ -420,12 +418,10 @@ import BraintreeDataCollector
                 } catch {
                     appSwitchCompletion(nil, error)
                 }
-                BTPayPalClient.payPalClient = nil
             }
         case .unknownPath:
             notifyFailure(with: BTPayPalError.appSwitchReturnURLPathInvalid)
             appSwitchCompletion(nil, BTPayPalError.appSwitchReturnURLPathInvalid)
-            BTPayPalClient.payPalClient = nil
         }
     }
 
@@ -824,7 +820,9 @@ extension BTPayPalClient: BTAppContextSwitchClient {
     /// :nodoc:
     @_documentation(visibility: private)
     @objc public static func handleReturnURL(_ url: URL) {
-        payPalClient?.handleReturnURL(url)
+        guard let client = payPalClient else { return }
+        payPalClient = nil
+        client.handleReturnURL(url)
     }
 
     /// :nodoc:


Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -378,7 +366,8 @@ import BraintreeDataCollector
             appSwitchCompletion = completion
         } else {
             hasOpenedURL = false
-            notifyFailure(with: BTPayPalError.appSwitchFailed, completion: completion)
+            notifyFailure(with: BTPayPalError.appSwitchFailed)
+            completion(nil, BTPayPalError.appSwitchFailed)
```
stechiu, line 389: Do we need this completion block? since we've removed the completion part of notifyFailure

buzzamus: Yeah, in this case this is using the completion the merchant passes in and giving them the error still for handling

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -424,83 +424,81 @@ import BraintreeDataCollector
...
+            universalLink: self.universalLink,
+            fallbackURLScheme: self.fallbackURLScheme
+        )
+
+        do {
+            let (body, _) = try await self.apiClient.post(request.hermesPath, parameters: parameters)
+
+            guard let body, let approvalURL = BTPayPalApprovalURLParser(body: body) else {
+                notifyFailure(with: BTPayPalError.invalidURL("Missing approval URL in gateway response."))
+                throw BTPayPalError.invalidURL("Missing approval URL in gateway response.")
```
jaxdesmarais, line 476: Are we able to check that this won't send duplicate failure events with the catch below?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -424,83 +424,81 @@ import BraintreeDataCollector
...
-                return
-            }
-            
-            self.isConfigFromCache = configuration.isFromCache
+        let configuration: BTConfiguration
+        do {
+            configuration = try await apiClient.fetchOrReturnRemoteConfiguration()
+        } catch {
+            notifyFailure(with: BTPayPalError.fetchConfigurationFailed)
+            throw BTPayPalError.fetchConfigurationFailed
```
jaxdesmarais: I think here we may want to do this to persist the config error:
```suggestion
            notifyFailure(with: error)
            throw error
```

Sources/BraintreePayPal/BTPayPalClient.swift
buzzamus: Updated! It was firing twice, so removed the notify here and let the catch handle.

Sources/BraintreePayPal/BTPayPalClient.swift
buzzamus: Updated!

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -252,18 +249,16 @@ import BraintreeDataCollector
                 url: url,
                 didPayPalServerAttemptAppSwitch: didPayPalServerAttemptAppSwitch ?? false
             ) else {
-            notifyFailure(with: BTPayPalError.invalidURLAction, completion: completion)
-            return
+            throw BTPayPalError.invalidURLAction
```
agedd, line 253: q: do we need to use `notifyFailure` here to fire these events 👀

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -814,7 +822,6 @@ extension BTPayPalClient: BTAppContextSwitchClient {
     @_documentation(visibility: private)
     @objc public static func handleReturnURL(_ url: URL) {
         payPalClient?.handleReturnURL(url)
-        BTPayPalClient.payPalClient = nil
```
agedd, line 817: curious why this was removed

buzzamus: Good question! This is to make sure app switch plays nice with the new async/await flow since it was initially running synchronously, which led to it being deallocated before the switch finished. This line is now moved into the Task after the await completes.

Sources/BraintreePayPal/BTPayPalClient.swift
buzzamus: ope you are correct. This won't be handled in the catch, so will add back in

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -410,29 +397,40 @@ import BraintreeDataCollector
         hasOpenedURL = false
 
         guard let returnURL = BTPayPalReturnURL(.payPalApp(url: url)) else {
-            notifyFailure(with: BTPayPalError.invalidURL("App Switch return URL cannot be nil"), completion: appSwitchCompletion)
+            notifyFailure(with: BTPayPalError.invalidURL("App Switch return URL cannot be nil"))
+            appSwitchCompletion(nil, BTPayPalError.invalidURL("App Switch return URL cannot be nil"))
+            BTPayPalClient.payPalClient = nil
```
jaxdesmarais: Do we want to consider putting this in some sort of deinit instead?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -274,23 +271,17 @@ import BraintreeDataCollector
...
-                return
-            }
+        let (body, _) = try await apiClient.post("/v1/payment_methods/paypal_accounts", parameters: encodableParams)
 
-            self.notifySuccess(with: tokenizedAccount, completion: completion)
+        guard
+            let payPalAccount = body?["paypalAccounts"].asArray()?.first,
+            let tokenizedAccount = BTPayPalAccountNonce(json: payPalAccount)
+        else {
+            throw BTPayPalError.failedToCreateNonce
```
jaxdesmarais, line 282: Do we need to call `notifyFailure` here?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -446,82 +444,82 @@ import BraintreeDataCollector
...
+        }
+
+        self.isConfigFromCache = configuration.isFromCache
+
+        guard configuration.isPayPalEnabled else {
+            notifyFailure(with: BTPayPalError.disabled)
+            throw BTPayPalError.disabled
+        }
+
+        self.payPalRequest = request
```
jaxdesmarais: Are we setting this twice? Also on line 434?

buzzamus: oh strange looks like there have historically been 2 lines setting it. I'll remove one.

Sources/BraintreePayPal/BTPayPalClient.swift
buzzamus: hm...i'm not sure...it looks like BTVenmoClient nils the client at the top of the static handleReturnURL method. So, yeah having it in multiple places here is not a good approach for sure....Thoughts?

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: Oh fun! Haha. Wonder why we would have done that or if we just didn't catch the duplicate at some point.

Sources/BraintreePayPal/BTPayPalClient.swift
jaxdesmarais: Gotcha - if this makes the most since that's totally fine.

Sources/BraintreePayPal/BTPayPalClient.swift
buzzamus: updated

Sources/BraintreePayPal/BTPayPalClient.swift
buzzamus: updated!

Sources/BraintreePayPal/BTPayPalClient.swift
buzzamus: You have far more context than me! I was just nervous of a deinit in any scenario where a merchant might purposely have a reference still to the client and any sort of weirdness happening, but if that isn't a problem a deinit might make sense for both classes actually

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -162,15 +164,12 @@ import BraintreeDataCollector
...
-                    continuation.resume(throwing: error)
-                } else if let nonce {
-                    continuation.resume(returning: nonce)
-                }
-            }
-        }
+        isVaultRequest = true
+        contextType = "BA-TOKEN"
+        fundingSource = getFundingSource(from: request)
+        shouldRequestBillingAgreement = false
```
agedd, line 170: nit: it looks like we set this value from the request object i.e. `request.requestBillingAgreement` in the past, just wanted to check if we should be doing the same or false is okay here

buzzamus: Good question! 

Looks like this was changed recently when the inner sourcing team changed things up for analytics - https://github.com/braintree/braintree_ios/pull/1749

buzzamus: But yeah, I just used the existing values as they were and moved them with the code changes, so the logic is the same as main currently is

PR #1749, opened 2026-02-11
Add Billing With Purchase analytics tags

NastassiaRodzik:

### Summary of changes

- Adding **is_purchase**, **is_billing_agreement**, and **billing_plan_type** to FPTIBatchData so that it is sent with analytic events.
This is for observability over the Billing With Purchase feature, so we can properly distinguish these flows.

Screenshot 2026-02-11 at 12 23 03 PM

Android [Companion PR ](https://github.com/braintree/braintree_android/pull/1529)

### Checklist

- [x] Added a changelog entry

- nrodzik

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 ## unreleased
 * BraintreePayPal
     * Pass `is_billing_agreement` and `billing_plan_type` to analytics events in order to track the billing with purchase flows.
-    
+* BraintreeCore
+  * Make `BTPostalAddress` properties publicly accessible (fixes #1751)
+
 ## 7.4.0 (2026-02-12)
 * BraintreeVenmo
     * Expose the `riskCorrelationID` property on `BTVenmoRequest` 
--- a/Sources/BraintreeCore/BTPostalAddress.swift
+++ b/Sources/BraintreeCore/BTPostalAddress.swift
@@ -3,17 +3,31 @@ import Foundation
 ///  Generic postal address
 @objcMembers public class BTPostalAddress: NSObject, Encodable {
 
-    // MARK: - Internal Properties
-    
+    // MARK: - Public Properties
+
     // Property names follow the `Braintree_Address` convention as documented at:
     // https://developer.paypal.com/braintree/docs/reference/request/address/create
-    let recipientName: String?
-    let streetAddress: String?
-    let extendedAddress: String?
-    let locality: String?
-    let countryCodeAlpha2: String?
-    let postalCode: String?
-    let region: String?
+
+    /// Recipient name for shipping address
+    public let recipientName: String?
+
+    /// Line 1 of the Address (e.g. number, street, etc)
+    public let streetAddress: String?
+
+    /// Line 2 of the Address (e.g. suite, apt #, etc.)
+    public let extendedAddress: String?
+
+    /// City name
+    public let locality: String?
+
+    /// 2 letter country code
+    public let countryCodeAlpha2: String?
+
+    /// Zip code or equivalent
+    public let postalCode: String?
+
+    /// Either a two-letter state code (for the US), or an ISO-3166-2 country subdivision code of up to three letters
+    public let region: String?
     
     // MARK: - Initializer
     


Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -50,9 +52,13 @@ struct FPTIBatchData: Codable {
         let endTime: Int?
         let errorDescription: String?
         let eventName: String
+        /// Whether or not billing agreement will be created - customer opted to save PayPal for future purchases and a vaulted billing agreement was created with the charge.
+        let isBillingAgreement: Bool?
```
buzzamus: We already have `isVaultRequest` as a parameter value. Is it possible to leverage that instead of adding a new similar/duplicate parameter?

NastassiaRodzik: **isBillingAgreement** is a flag which means that billing agreement was requested by the user when making a purchase, meaning ` isVaultRequest = false `

buzzamus (approved):

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -50,9 +52,13 @@ struct FPTIBatchData: Codable {
         let endTime: Int?
         let errorDescription: String?
         let eventName: String
+        /// Whether or not billing agreement will be created - customer opted to save PayPal for future purchases and a vaulted billing agreement was created with the charge.
```
saperi22: Can we explicitly call out that this is a param that's set for opting in to BA with one time checkout?

saperi22: Is it only a BA with purchase if both `isBillingAgreement` and `isPurchase` are true?
What does it mean if `isBillingAgreement` is true but `isPurchase` is false? I'd like to understand which flow would lead to this combination; do we not expect this to happen?

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 * BraintreeDataCollector
   * Update PPRiskMagnes to 5.6.0 
   * Add `collectDeviceDataOnSuccess(riskCorrelationID: completion:)` method that ensures Magnes SDK completes data submission prior to returning device data.
+* BraintreePayPal
+    * Pass `is_billing_agreement`, `is_purchase`, and `billing_plan_type` to analytics events.
```
agedd: nit: can we be a bit more explicit and add that these tags were added to track the billing with purchase flows.

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -1515,6 +1527,22 @@ class BTPayPalClient_Tests: XCTestCase {
         XCTAssertFalse(mockAPIClient.postedIsVaultRequest)
         XCTAssertEqual(mockAPIClient.postedContextType, "EC-TOKEN")
         XCTAssertEqual(mockAPIClient.postedFundingSource, "paylater")
+        XCTAssertFalse(mockAPIClient.postedIsBillingAgreement)
+        XCTAssertTrue(mockAPIClient.postedIsPurchase)
+        XCTAssertNil(mockAPIClient.postedBillingPlanType)
+    }
+    
+    func testTokenize_whenCheckoutRequestBillingAgreementTrue_setsBillingWithPurchaseAnalytics() async {
```
agedd, line 1530: nit: do we need to also add a similar test for `BTPayPalVaultRequest` 👀

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
NastassiaRodzik: `Is it only a BA with purchase if both isBillingAgreement and isPurchase are true?`
- Yes

`What does it mean if isBillingAgreement is true but isPurchase is false?`
- User requested a billing agreement (i.e. save paypal for future purchases) but checkout request didn't have an amount

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
NastassiaRodzik: @agedd it's already there, line 1471 and 1485: 

```
func testTokenize_whenVaultRequest_setsVaultAnalyticsTags()
func testTokenize_whenVaultCreditRequest_setsVaultAnalyticsTags()
```

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -33,6 +33,8 @@ struct FPTIBatchData: Codable {
         let applicationState: String?
 
         let appSwitchURL: String?
+        /// Recurring billing plan type, or charge pattern.
+        let billingPlanType: String?
```
jaxdesmarais: Please name these after the properties in the SDK they represent. The coding keys can map them to the underlying type.

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
jaxdesmarais: Please name this after the property it represents

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
```diff
@@ -50,9 +52,13 @@ struct FPTIBatchData: Codable {
         let endTime: Int?
         let errorDescription: String?
         let eventName: String
+        /// Whether or not billing agreement will be created - customer opted to save PayPal for future purchases and a vaulted billing agreement was created with the charge.
+        let isBillingAgreement: Bool?
         /// True if the `BTConfiguration` was retrieved from local cache after `tokenize()` call.
         /// False if the `BTConfiguration` was fetched remotely after `tokenize()` call.
         let isConfigFromCache: Bool?
+        /// Whether or not there's a purchase as part of the transaction. True if amount > 0
+        let isPurchase: Bool?
```
jaxdesmarais: This name does not make sense - lets think of something that more accurately represents what this is for

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 * BraintreeDataCollector
   * Update PPRiskMagnes to 5.6.0 
   * Add `collectDeviceDataOnSuccess(riskCorrelationID: completion:)` method that ensures Magnes SDK completes data submission prior to returning device data.
+* BraintreePayPal
+    * Pass `is_billing_agreement`, `is_purchase`, and `billing_plan_type` to analytics events in order to track the billing with purchase flows.
```
jaxdesmarais: Pull latest from main - this is not under the correct section

jaxdesmarais (changes requested):

Sources/BraintreeCore/Analytics/FPTIBatchData.swift
NastassiaRodzik: @jaxdesmarais what do you think about `isPurchaseFlow` name?

UnitTests/BraintreeTestShared/MockAPIClient.swift
```diff
@@ -14,8 +14,10 @@ public class MockAPIClient: BTAPIClient {
     public var postedAnalyticsEvents: [String] = []
     public var postedApplicationState: String? = nil
     public var postedAppSwitchURL: [String: String?] = [:]
+    public var postedBillingPlanType: String? = nil
```
jaxdesmarais: Update to match the property name pattern used:
```suggestion
    public var postedRecurringBillingPlanType: String? = nil
```

UnitTests/BraintreeTestShared/MockAPIClient.swift
```diff
@@ -14,8 +14,10 @@ public class MockAPIClient: BTAPIClient {
...
     public var postedButtonOrder: String? = nil
     public var postedButtonType: String? = nil
+    public var postedIsBillingAgreement = false
```
jaxdesmarais:
```suggestion
    public var postedShouldRequestBillingAgreement = false
```

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
...
```
jaxdesmarais: This is still under the wrong section - please pull the latest from `main` and update with the accurate tags we are passing

jaxdesmarais (approved):

PR #1765, opened 2026-03-16
Convert Completion Handlers in BTHTTP to Async Functions

agedd:
### Summary of changes

-  Remove all the completion-based versions of the HTTP methods.

### Checklist

- @agedd

### DIFF
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -99,45 +99,7 @@ import UIKit
             throw error
         }
     }
-    
-    /// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
-    ///
-    /// Perfom an HTTP GET on a URL composed of the configured from environment and the given path.
-    /// - Parameters:
-    ///   - path: The endpoint URI path.
-    ///   - parameters: Optional set of query parameters to be encoded with the request.
-    ///   - httpType: The underlying `BTAPIClientHTTPService` of the HTTP request. Defaults to `.gateway`.
-    ///   - completion:  A block object to be executed when the request finishes.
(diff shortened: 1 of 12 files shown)


Sources/BraintreeCore/Configuration/ConfigurationLoader.swift
```diff
@@ -60,7 +60,7 @@ class ConfigurationLoader {
             do {
                 let (body, response) = try await http.get(configPath, parameters: BTConfigurationRequest())
 
-                if response?.statusCode != 200 || body == nil {
+                if response.statusCode != 200 {
                     throw BTAPIClientError.configurationUnavailable
```
buzzamus, line 64: Is the nil body check no longer needed?

agedd: ahh yep that's correct - in the completion handler version, the body was `BTJSON?` however in the async/await version, we return non-optional `(BTJSON, HTTPURLResponse)` and failure would surface as a thrown error so by this point i think body should be valid.

buzzamus (approved):
😎

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -1,6 +1,5 @@
 import UIKit
 
-// swiftlint:disable type_body_length file_length
```
jaxdesmarais, line 3: 😍

jaxdesmarais (commented):
I think we can also remove:
- `BTHTTP.dispatchQueue`
- `BTHTTP.typealias RequestCompletion`
- `BTAPIClient.RequestCompletion`

We could also consider removing `fetchOrReturnRemoteConfiguration` with completion. Not sure if we want to do that in this PR or spin up a new ticket.

jaxdesmarais (approved):
Assuming all flows manually are still working as expected we should be good to go!

PR #1769, opened 2026-03-30
Create BTCard extension for allowing card data to be optional

buzzamus:
### Summary of changes

- extended BTCard to allow card data values to be optional. This will allow a merchant integration using the Card Fields to pass in all optional data without the card data (which will be included with the Card Fields).
### Checklist

- [] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeUIComponents/CardFields/Shared/BTCard+CardFields.swift
+++ b/Sources/BraintreeUIComponents/CardFields/Shared/BTCard+CardFields.swift
@@ -1,0 +1,51 @@
+#if canImport(BraintreeCard)
+import BraintreeCard
+#endif
+
+extension BTCard {
+
+    /// Creates a `BTCard` with only optional metadata fields for use with `BTCardFields`.
+    /// Card number, expiration month, expiration year, and CVV are managed internally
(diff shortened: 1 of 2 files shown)


agedd (approved):
lgtm!! 🚀

jaxdesmarais (approved):

Sources/BraintreeUIComponents/CardFields/Shared/BTCard+CardFields.swift
```diff
@@ -0,0 +1,51 @@
+#if canImport(BraintreeCard)
+import BraintreeCard
+#endif
+
+extension BTCard {
+
+    /// Creates a `BTCard` with only optional metadata fields for use with `BTCardFields`.
+    /// Card number, expiration month, expiration year, and CVV are managed internally
+    /// by the `BTCardFields` form and should not be set by the merchant.
+    public convenience init(
```
jaxdesmarais, line 10: Since it's just used for this module are we able to make this internal vs public?

buzzamus: yep! I'll make sure that gets done on a future PR

PR #1771, opened 2026-04-02
Add card number field UI

buzzamus:
### Summary of changes

- This PR adds the UI for the card number fields. 
Note: This PR is for the UI only. The state changes (such as spacing between the digits and rules around card types) as well as validation checks/UI changes will be handled in later PRs.

Screenshot 2026-04-02 at 10 13 18 AM

### Checklist

- [] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldView.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldView.swift
@@ -1,12 +1,46 @@
 import SwiftUI
 
-struct CardNumberFieldView: View {
+struct CardNumberFieldView<ViewModel: CardFieldsViewModelProtocol>: View {
+    
+    @ObservedObject var viewModel: ViewModel
+    var onAutoAdvance: (() -> Void)?
+    @FocusState private var isFocused: Bool
+
     
     var body: some View {
-        Text(/*@START_MENU_TOKEN@*/"Hello, World!"/*@END_MENU_TOKEN@*/)
+        CardFieldsContainerView(
+            validationState: viewModel.validationState,
+            isFocused: isFocused
+        ) {
+            CardBrandView()
+
+            VStack(alignment: .leading, spacing: 2) {
+                Text("Card number")
+                    .font(.system(size: 12))
+                    .foregroundColor(Color(.secondaryLabel))
+
+                TextField("", text: Binding(
+                    get: { viewModel.value },
+                    set: { viewModel.updateValue($0) }
+                ))
+                .keyboardType(.numberPad)
+                .focused($isFocused)
+                .font(.system(size: 16))
+                .foregroundColor(Color(.label))
+            }
+
+            Spacer()
+        }
+        // TODO: update to new onChange when minimum target moved to iOS 17.0
+        .onChange(of: isFocused) { focused in
+            viewModel.isFocused = focused
+        }
+        .onChange(of: viewModel.shouldAutoAdvance) { shouldAdvance in
+            if shouldAdvance { onAutoAdvance?() }
+        }
     }
 }
 
 #Preview {
-    CardNumberFieldView()
+    CardNumberFieldView(viewModel: CardNumberFieldViewModel())
 }
--- a/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
@@ -2,5 +2,15 @@ import Foundation
 
 class CardNumberFieldViewModel: CardFieldsViewModelProtocol {
     
-    var state: String = ""
+    @Published private(set) var value: String = ""
+    @Published private(set) var validationState: ValidationResult = .valid
+    @Published var isFocused: Bool = false
+
+    // TODO: Implement auto-advance logic w/ state change
+    var shouldAutoAdvance: Bool { false }
+
+    // TODO: Implement validation and formatting
+    func updateValue(_ newValue: String) {
+        value = newValue
+    }
 }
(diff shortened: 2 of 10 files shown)


Sources/BraintreeUIComponents/CardFields/Shared/CardFieldsContainerView.swift
```diff
@@ -0,0 +1,38 @@
+import SwiftUI
+
+struct CardFieldsContainerView<Content: View>: View {
+
```
buzzamus, line 4: I don't know if ContainerView is the right name for this, but I couldn't think of anything better....essentially this will act like a protocol for the views to store all of the shared attributes...although instead of a protocol it uses ViewBuilder.

agedd: `ContainerView` makes sense to me!

agedd (approved):
this is looking great! 😍

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldView.swift
```diff
@@ -1,12 +1,46 @@
...
+                .foregroundColor(Color(.label))
+            }
+
+            Spacer()
+        }
+        // TODO: update to new onChange when minimum target moved to iOS 17.0
+        .onChange(of: isFocused) { focused in
+            viewModel.isFocused = focused
+        }
+        .onChange(of: viewModel.shouldAutoAdvance) { shouldAdvance in
```
jaxdesmarais, line 38: Are we resetting this anywhere? Thinking of cases when the user re-focuses the field after auto-advancing, it won't re-trigger. Maybe this should be a future TODO or is already on your list though.

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
```diff
@@ -2,5 +2,15 @@ import Foundation
 
 class CardNumberFieldViewModel: CardFieldsViewModelProtocol {
```
jaxdesmarais, line 4: Should this class be a `@MainActor` or are we okay here?

Sources/BraintreeUIComponents/CardFields/Shared/CardFieldsViewModelProtocol.swift
```diff
@@ -1,5 +1,11 @@
 import Foundation
 
-protocol CardFieldsViewModelProtocol {
-    var state: String { get }
+protocol CardFieldsViewModelProtocol: ObservableObject {
+    
+    var value: String { get }
+    var validationState: ValidationResult { get }
+    var isFocused: Bool { get set }
```
jaxdesmarais, line 7: Should this be a `private(set)`? Not sure if we are creating 2 sources of truth with this and `isFocused` in the view model and view?

Sources/BraintreeUIComponents/CardFields/Shared/ValidationResult.swift
```diff
@@ -0,0 +1,5 @@
+enum ValidationResult: Equatable {
+    case valid
+    case invalid(String)
+    case validating
```
jaxdesmarais, line 4: Doesn't seem like this case us used yet. I assume implementing `validating` will be in a future PR?

jaxdesmarais (commented):
Just a few general questions

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldView.swift
buzzamus: I can add a TODO! The focusing/auto-advancing functionality will be completed finalized in a later task to handle the state change logic

Sources/BraintreeUIComponents/CardFields/Shared/ValidationResult.swift
buzzamus: Correct this will be part of the state change task to come

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
buzzamus: Ah yep! This will be needed when everything is hooked up to handle state changes. I will update now

Sources/BraintreeUIComponents/CardFields/Shared/CardFieldsViewModelProtocol.swift
buzzamus: hm...it should either be `private(set)` or really just several methods to react to the view's `isFocused`. I kind of got ahead of myself by trying to add some things I thought we would need between all the fields with the shared protocol, but it might be better to remove it and make it more clear to just react to the state changes sent from the view...thoughts?

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
buzzamus: Updated

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldView.swift
jaxdesmarais: Cool - as long as it's on your mind/in a future ticket totally fine. Mainly just making sure we had it somewhere. 😄

Sources/BraintreeUIComponents/CardFields/Shared/CardFieldsViewModelProtocol.swift
jaxdesmarais: Up to you. If we will use/update eventually I am okay leaving as is for now. Whatever makes the most sense.

buzzamus: I say leave it for now since this may change and right now I would only be guessing what pattern makes the most sense for when I actually dive into the state handling task. I kind of want to try both choices mentioned above and see what works, but that won't come until the state management ticket is worked on

jaxdesmarais (approved):
Thanks for addressing my comments and the TODOs.

PR #1773, opened 2026-04-06
Create cvv field

buzzamus:
### Summary of changes

- Adds UI for CVV field. State change handling and validations will come later. Also, you will notice that the CVV field runs the width of the device. This is because the CVV field view itself does not control its own width and instead the completed CardField component will set the height of each individual input field.

Screenshot 2026-04-06 at 1 29 17 PM

Screenshot 2026-04-06 at 1 31 33 PM

### Checklist

- [] Added a changelog entry
- []  Tested and confirmed payment flows affected by this change are functioning as expected

- @buzzamus

### DIFF
--- a/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
@@ -1,12 +1,97 @@
 import SwiftUI
 
-struct CVVFieldView: View {
-    
+struct CVVField<ViewModel: CVVFieldViewModel>: View {
+
+    // MARK: - Internal Properties
+
+    @ObservedObject var viewModel: ViewModel
+    var onAutoAdvance: (() -> Void)?
+
+    // MARK: - Private Properties
+
+    @FocusState private var isFocused: Bool
+    @State private var showCVVHint: Bool = false
+
+    // MARK: - View
+
     var body: some View {
-        Text(/*@START_MENU_TOKEN@*/"Hello, World!"/*@END_MENU_TOKEN@*/)
+        CardFieldsContainerView(
+            validationState: viewModel.validationState,
+            isFocused: isFocused
+        ) {
+            VStack(alignment: .leading, spacing: 2) {
+                Text("CVV")
+                    .font(.system(size: 12))
+                    .foregroundColor(Color(.secondaryLabel))
+
+                ZStack(alignment: .leading) {
+                    TextField("", text: Binding(
+                        get: { viewModel.rawValue },
+                        set: { viewModel.updateValue($0) }
+                    ))
+                    .keyboardType(.numberPad)
+                    .focused($isFocused)
+                    .foregroundColor(.clear)
+                    .tint(Color(.label))
+                    .font(.system(size: 16))
+
+                    if viewModel.characters.isEmpty {
+                        Text("•••")
+                            .font(.system(size: 16))
+                            .foregroundColor(Color(.placeholderText))
+                    } else {
+                        HStack(spacing: 2) {
+                            ForEach(viewModel.characters) { character in
+                                Text(character.isMasked ? "•" : String(character.value))
+                                    .font(.system(size: 16))
+                                    .foregroundColor(Color(.label))
+                                    .animation(.easeInOut(duration: 0.3), value: character.isMasked)
+                            }
+                        }
+                    }
+                }
+                .frame(height: 24)
+            }
+
+            Spacer()
+
+            // Help icon
+            Button {
+                showCVVHint = true
+            } label: {
+                Image(systemName: "questionmark.circle")
+                    .foregroundColor(Color(.secondaryLabel))
+                    .font(.system(size: 16))
+            }
+            .popover(isPresented: $showCVVHint, arrowEdge: .top) {
+                VStack(alignment: .leading, spacing: 8) {
+                    Text("What is a CVV?")
+                        .font(.system(size: 16, weight: .semibold))
+                        .foregroundColor(Color(.label))
+
+                    Text("The CVV is the 3 or 4-digit number on the back of your card")
+                        .font(.system(size: 14))
+                        .foregroundColor(Color(.secondaryLabel))
+                        .fixedSize(horizontal: false, vertical: true)
+                }
+                .padding(16)
+                .presentationCompactAdaptation(.popover)
+            }
+        }
+        .onChange(of: isFocused) { _, focused in
+            viewModel.isFocused = focused
+        }
+        .onChange(of: viewModel.shouldAutoAdvance) { _, shouldAdvance in
+            if shouldAdvance { onAutoAdvance?() }
+        }
+        .contentShape(Rectangle())
+        .onTapGesture {
+            isFocused = true
+        }
     }
 }
 
 #Preview {
-    CVVFieldView()
+    CVVField(viewModel: CVVFieldViewModel())
+        .padding()
 }
--- a/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
@@ -1,6 +1,6 @@
 import SwiftUI
 
-struct CVVField<ViewModel: CVVFieldViewModel>: View {
+struct CVVFieldView<ViewModel: CVVFieldViewModel>: View {
 
     // MARK: - Internal Properties
 
@@ -92,6 +92,6 @@ struct CVVField<ViewModel: CVVFieldViewModel>: View {
 }
 
 #Preview {
-    CVVField(viewModel: CVVFieldViewModel())
+    CVVFieldView(viewModel: CVVFieldViewModel())
         .padding()
 }
--- a/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
@@ -63,9 +63,9 @@ struct CVVField<ViewModel: CVVFieldViewModel>: View {
                     .foregroundColor(Color(.secondaryLabel))
                     .font(.system(size: 16))
             }
-            .popover(isPresented: $showCVVHint, arrowEdge: .top) {
+            .popover(isPresented: $showCVVHint, arrowEdge: .bottom) {
                 VStack(alignment: .leading, spacing: 8) {
-                    Text("What is a CVV?")
+                    Text("CVV")
                         .font(.system(size: 16, weight: .semibold))
                         .foregroundColor(Color(.label))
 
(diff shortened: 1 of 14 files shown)


Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
```diff
@@ -1,12 +1,97 @@
 import SwiftUI
 
-struct CVVFieldView: View {
-    
+struct CVVField<ViewModel: CVVFieldViewModel>: View {
```
agedd: nit:
```suggestion
struct CVVFieldView<ViewModel: CVVFieldViewModel>: View {
```

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
```diff
@@ -1,12 +1,97 @@
...
+
+    // MARK: - Internal Properties
+
+    @ObservedObject var viewModel: ViewModel
+    var onAutoAdvance: (() -> Void)?
+
+    // MARK: - Private Properties
+
+    @FocusState private var isFocused: Bool
```
agedd, line 12: do we need to set a default value of `false` here

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
```diff
@@ -1,4 +1,50 @@
-import Foundation
+import SwiftUI
 
-// TODO: have conform to CardFieldsViewModelProtocol
-class CVVFieldViewModel {}
+class CVVFieldViewModel: CardFieldsViewModelProtocol {
+
+    // MARK: - CardFieldViewModelProtocol
```
agedd, line 6: nit: 
```suggestion
    // MARK: - CardFieldViewModelProtocol Conformance
```

Sources/BraintreeUIComponents/CardFields/Shared/BTCard+CardFields.swift
```diff
@@ -0,0 +1,51 @@
+#if canImport(BraintreeCard)
+import BraintreeCard
+#endif
+
+extension BTCard {
+
+    /// Creates a `BTCard` with only optional metadata fields for use with `BTCardFields`.
+    /// Card number, expiration month, expiration year, and CVV are managed internally
+    /// by the `BTCardFields` form and should not be set by the merchant.
+    public convenience init(
```
agedd, line 10: nit: did we want to make this initializer internal rather than public

agedd (approved):
looking great to me ~ just added a few minor cleanup comments

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
buzzamus: SwiftUI implicitly sets a value of `false` for booleans when a value isn't set for `@FocusState`

Sources/BraintreeUIComponents/CardFields/Shared/BTCard+CardFields.swift
buzzamus: This has to be public so that developers can pass in any non-card data they wish to include with the form generated card values (such as cardholder name or address info)

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
buzzamus: good catch!

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: Updated

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
```diff
@@ -1,4 +1,50 @@
...
+        let newCount = digits.count
+
+        if newCount > oldCount {
+            // fade transition animation for masking new characters
+            for i in oldCount..<newCount {
+                let index = digits.index(digits.startIndex, offsetBy: i)
+                characters.append(CVVCharacter(value: digits[index]))
+
+                let charIndex = i
+                DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) { [weak self] in
```
jaxdesmarais: Will this cause issues if the number is quickly added/deleted before the timer fires? Wondering if we will run into issues where the character no longer exists at the index and we attempt to mask it.

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
```diff
@@ -1,4 +1,50 @@
...
+    // TODO: Implement auto-advance logic with state changes
+    var shouldAutoAdvance: Bool { false }
+
+    // MARK: - Internal Properties
+
+    /// Individual characters with their masking state — drives the custom display
+    @Published private(set) var characters: [CVVCharacter] = []
+
+    /// Raw digits only — no masking, used for tokenization
+    private(set) var rawValue: String = ""
```
jaxdesmarais: Does this one also need to be `@Published` since it drives some behavior?

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
```diff
@@ -1,12 +1,97 @@
...
+                }
+                .padding(16)
+                .presentationCompactAdaptation(.popover)
+            }
+        }
+        .onChange(of: isFocused) { _, focused in
+            viewModel.isFocused = focused
+        }
+        .onChange(of: viewModel.shouldAutoAdvance) { _, shouldAdvance in
+            if shouldAdvance { onAutoAdvance?() }
```
jaxdesmarais, line 89: Will this be implemented in the future PR?

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
```diff
@@ -1,12 +1,97 @@
...
+
+            // Help icon
+            Button {
+                showCVVHint = true
+            } label: {
+                Image(systemName: "questionmark.circle")
+                    .foregroundColor(Color(.secondaryLabel))
+                    .font(.system(size: 16))
+            }
+            .popover(isPresented: $showCVVHint, arrowEdge: .bottom) {
```
jaxdesmarais, line 68: Do we want an accessibility label on the popover?

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: ah good point. i'll use the character's ID (UUID) for the check instead to account for this

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: Yep! i'll add that

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
buzzamus: Yeah this will be part of the state handling task

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
buzzamus: Added accessibility label for both popover as well as the help icon button

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: done!

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: Updated to instead use the UUID and added unit tests for the behavior

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
```diff
@@ -1,4 +1,52 @@
...
+                DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) { [weak self] in
+                    guard let self, let characterIndex = self.characters.firstIndex(where: { $0.id == characterID }) else { return }
+                    withAnimation(.easeInOut(duration: 0.3)) {
+                        self.characters[characterIndex].isMasked = true
+                    }
+                }
+            }
+        } else if newCount < oldCount {
+            characters = Array(characters.prefix(newCount))
+        }
```
jaxdesmarais, line 62: Do we also need to handle updating if someone types "1234" then pastes "5678" over it so the `characters` updates? I think in this case nether branch would run, right?

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
```diff
@@ -1,4 +1,52 @@
...
+
+        if newCount > oldCount {
+            for i in oldCount..<newCount {
+                let index = digits.index(digits.startIndex, offsetBy: i)
+                let newCharacter = CVVCharacter(value: digits[index])
+                characters.append(newCharacter)
+
+                // Capture the stable ID rather than the index
+                let characterID = newCharacter.id
+                DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) { [weak self] in
```
jaxdesmarais: Take it or leave it, we could consider doing this in a Task instead since we are moving toward structured concurrency.

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
jaxdesmarais: We should be able to test this too with something like this if we want to check:
```swift
  func testUpdateValue_sameLength_differentDigits_updatesCharacters() {
      viewModel.updateValue("123")
      viewModel.updateValue("456")
      XCTAssertEqual(viewModel.rawValue, "456")
      XCTAssertEqual(viewModel.characters.map { $0.value }, ["4", "5", "6"])
  }
```

buzzamus: Yep you are right! That test fails, so i'll add an else block to capture handling this

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: Updated to use a Task

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: Added an else statement to handle this behavior and got the new unit test to pass

jaxdesmarais (approved):
Thanks for addressing my comments! This looks great. 🚀

PR #1774, opened 2026-04-07
 Create Expiration Date Field Form

agedd:
### Summary of changes

- Add expiration date field form based off of figma design

### Checklist

- @agedd

### DIFF
--- a/Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
+++ b/Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
@@ -2,8 +2,13 @@ import SwiftUI
 
 struct ExpirationDateFieldView<ViewModel: CardFieldsViewModelProtocol>: View {
     
+    // MARK: - Internal Properties
+    
     @ObservedObject var viewModel: ViewModel
     var onAutoAdvance: (() -> Void)?
+    
+    // MARK: - Private Properties
+    
     @FocusState private var isFocused: Bool
     
     var body: some View {
@@ -31,12 +36,10 @@ struct ExpirationDateFieldView<ViewModel: CardFieldsViewModelProtocol>: View {
             
             Spacer()
         }
-        .frame(width: 179, height: 64)
-        // update `onChange(of:perform:)` signature once min deployment target is iOS 17
-        .onChange(of: isFocused) { focused in
+        .onChange(of: isFocused) { _, focused in
             viewModel.isFocused = focused
         }
-        .onChange(of: viewModel.shouldAutoAdvance) { shouldAutoAdance in
+        .onChange(of: viewModel.shouldAutoAdvance) { _, shouldAutoAdance in
             if shouldAutoAdance { onAutoAdvance?() }
         }
     }
--- a/Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
+++ b/Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
@@ -24,9 +24,9 @@ struct ExpirationDateFieldView<ViewModel: CardFieldsViewModelProtocol>: View {
                 TextField(
                     "",
                     text: Binding(
-                    get: { viewModel.value },
-                    set: { viewModel.updateValue($0) }
-                  )
+                        get: { viewModel.value },
+                        set: { viewModel.updateValue($0) }
+                    )
                 )
                 .keyboardType(.numberPad)
                 .focused($isFocused)
--- a/Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
+++ b/Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
@@ -1,12 +1,47 @@
 import SwiftUI
 
-struct ExpirationDateFieldView: View {
+struct ExpirationDateFieldView<ViewModel: CardFieldsViewModelProtocol>: View {
+    
+    @ObservedObject var viewModel: ViewModel
+    var onAutoAdvance: (() -> Void)?
+    @FocusState private var isFocused: Bool
     
     var body: some View {
-        Text(/*@START_MENU_TOKEN@*/"Hello, World!"/*@END_MENU_TOKEN@*/)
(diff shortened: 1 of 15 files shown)


buzzamus (approved):
😎

jaxdesmarais (commented):
It looks like there are maybe some merge conflicts and additional changes/diffs that need to be addressed here first before reviewing. Do you want to address those first?

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
```diff
@@ -1,12 +1,50 @@
...
+                .font(.body)
+                .foregroundColor(Color(.label))
+            }
+            
+            Spacer()
+        }
+        .onChange(of: isFocused) { _, focused in
+            viewModel.isFocused = focused
+        }
+        .onChange(of: viewModel.shouldAutoAdvance) { _, shouldAutoAdance in
```
jaxdesmarais: Typo here:

```suggestion
        .onChange(of: viewModel.shouldAutoAdvance) { _, shouldAutoAdvance in
```

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
```diff
@@ -1,12 +1,50 @@
...
     
     var body: some View {
-        Text(/*@START_MENU_TOKEN@*/"Hello, World!"/*@END_MENU_TOKEN@*/)
+        CardFieldsContainerView(
+            validationState: viewModel.validationState,
+            isFocused: isFocused
+        ) {
+            VStack(alignment: .leading, spacing: 2) {
+                Text("Expiration (MM/YY)")
+                    .font(.caption)
```
jaxdesmarais: The fonts used here are dynamic both here with `.caption` and with `.body` below. I think in other PRs it's a fixed font. I think dynamic will be better unless visually it greatly changes the experience. Not sure if we want to align this PR to the others and then fix them all or merge this one in and open a small PR to update the other ones? Curious on y'all's thoughts @buzzamus and @agedd.

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
```diff
@@ -1,12 +1,50 @@
...
+    @FocusState private var isFocused: Bool
     
     var body: some View {
-        Text(/*@START_MENU_TOKEN@*/"Hello, World!"/*@END_MENU_TOKEN@*/)
+        CardFieldsContainerView(
+            validationState: viewModel.validationState,
+            isFocused: isFocused
+        ) {
+            VStack(alignment: .leading, spacing: 2) {
+                Text("Expiration (MM/YY)")
```
jaxdesmarais, line 20: I think this formatting hint should be a placeholder not the label. Do we need merchants/customers to care though? Or can we get the expiry in the correct format internally regardless if someone types `mm/YY` vs `MM/YYYY`.

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldViewModel.swift
```diff
@@ -1,7 +1,16 @@
 import Foundation
 
-// TODO: have conform to CardFieldsViewModelProtocol
-class ExpirationDateFieldViewModel {
+class ExpirationDateFieldViewModel: CardFieldsViewModelProtocol {
```
jaxdesmarais, line 4: iirc this needs to be a `@MainActor` class since it uses published vars.

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldViewModel.swift
```diff
@@ -1,7 +1,16 @@
...
     
-    var state: String = ""
+    @Published private(set) var value: String = ""
+    @Published private(set) var validationState: ValidationResult = .valid
+    @Published var isFocused: Bool = false
+    
+    // TODO: Update auto-advance logic
+    var shouldAutoAdvance: Bool { false }
+    
+    // TODO: Implement expiration fiel validation and formatting checks
```
jaxdesmarais:
```suggestion
    // TODO: Implement expiration field validation and formatting checks
```

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
buzzamus: I am fine with dynamic. I spent probably too much time trying to go through fonts to see what was the match in the design, but in the end I don't think it really made much of a difference at all. I think if this is the way we go then a small PR to fix the others after this is merged makes sense.

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
buzzamus: I initially had in mind to have support for both MM/YYYY and MM/YY, but the designs only called out MM/YY, so I decided to stick strictly to that when creating the tasks. If we do want to support MM/YYYY then that is fine, but the validations will need to be updated to allow for both and parse the right format and also add some additional checks. I personally feel like maybe this could be a fast follow though?

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
agedd: yea i agree! i think i may have missed the text label font size 12 setting in the figma design, but imo it doesn't make much of a difference from a ui perspective. on board to create a smaller PR to adjust fonts to dynamic for the other elements

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
agedd: agreed, i feel like the client-side validations we add later on can address the different formats the merchant/user may type in this field

jaxdesmarais (approved):
Thanks for addressing my comments! This PR looks great.

PR #1776, opened 2026-04-09
Create Card Field Validator

buzzamus:
### Summary of changes

- Adds in validation checks for card number field

### Checklist

- [] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
@@ -11,7 +11,8 @@ struct CardNumberFieldValidator: CardFieldsValidatorProtocol {
             return .invalid("Card number is required")
         }
 
-        guard value.filter({ !$0.isNumber && $0 != " " }).isEmpty else {
+        let invalidCharacters = value.filter { !$0.isNumber && $0 != " " }
+        guard invalidCharacters.isEmpty else {
             return .invalid("Card number is invalid")
         }
 
--- a/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
@@ -18,7 +18,7 @@ struct CardNumberFieldValidator: CardFieldsValidatorProtocol {
         let brand = detectBrand(from: digits)
 
         guard digits.count >= brand.minLength else {
-            return .valid
+            return .validating
         }
 
         guard digits.count <= brand.maxLength else {
@@ -41,27 +41,19 @@ struct CardNumberFieldValidator: CardFieldsValidatorProtocol {
     func detectBrand(from digits: String) -> CardBrand {
         // Pass 1: strict prefixes
         for brand in CardBrand.allCases {
-            for pattern in brand.prefixPatterns where matches(pattern: pattern, input: digits) {
+            for pattern in brand.prefixPatterns where digits.prefixMatch(of: pattern) != nil {
                 return brand
             }
         }
 
         // Pass 2: relaxed prefixes
         for brand in CardBrand.allCases {
-            for pattern in brand.relaxedPrefixPatterns where matches(pattern: pattern, input: digits) {
+            for pattern in brand.relaxedPrefixPatterns where digits.prefixMatch(of: pattern) != nil {
                 return brand
             }
         }
 
         return .unknown
-    }
-
-    // MARK: - Private
-
-    private func matches(pattern: String, input: String) -> Bool {
-        guard let regex = try? NSRegularExpression(pattern: pattern) else { return false }
-        let range = NSRange(input.startIndex..., in: input)
-        return regex.firstMatch(in: input, range: range) != nil
     }
 
     /// Luhn algorithm - Processes digits right to left, doubling every second digit.
--- a/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
@@ -1,8 +1,94 @@
 import Foundation
 
-final class CardNumberFieldValidator: CardFieldsValidatorProtocol {
-    
-    func isValid(_ text: String) -> Bool {
-        return false
+struct CardNumberValidator: CardFieldValidatorProtocol {
+
+    // MARK: - CardFieldValidatorProtocol
+
+    func validate(_ value: String) -> ValidationResult {
+        let digits = value.filter { $0.isNumber }
+
+        guard !digits.isEmpty else {
+            return .invalid("Card number is required")
+        }
+
+        guard value.filter({ !$0.isNumber && $0 != " " }).isEmpty else {
+            return .invalid("Card number must contain numbers only")
+        }
+
+        let brand = detectBrand(from: digits)
+
+        guard digits.count >= brand.minLength else {
+            // Still typing — not invalid until we know they've finished
+            return .valid
+        }
+
+        guard digits.count <= brand.maxLength else {
+            return .invalid("Card number is too long")
+        }
+
+        guard brand.validLengths.contains(digits.count) else {
+            return .invalid("Invalid card number length")
+        }
+
+        guard isLuhnValid(digits) else {
+            return .invalid("Invalid card number")
+        }
+
+        return .valid
+    }
+
+    // MARK: - Brand Detection
+
+    /// Two-pass detection mirroring the drop-in:
+    /// 1. Check strict prefixes across all brands
+    /// 2. If no match, check relaxed prefixes (Maestro only)
+    func detectBrand(from digits: String) -> CardBrand {
+        // Pass 1: strict prefixes
+        for brand in CardBrand.allCases {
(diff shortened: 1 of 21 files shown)


Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
```diff
@@ -1,8 +1,85 @@
...
+            }
+        }
+
+        return .unknown
+    }
+
+    // MARK: - Private
+
+    private func matches(pattern: String, input: String) -> Bool {
+        guard let regex = try? NSRegularExpression(pattern: pattern) else { return false }
```
jaxdesmarais: iirc `NSRegularExpression` is recompiled every keystroke. Could we compile once and cache potentially or use `Regex` that handles caching automatically? What do you think?

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
```diff
@@ -1,8 +1,85 @@
...
+        }
+
+        guard value.filter({ !$0.isNumber && $0 != " " }).isEmpty else {
+            return .invalid("Card number is invalid")
+        }
+
+        let brand = detectBrand(from: digits)
+
+        guard digits.count >= brand.minLength else {
+            return .valid
```
jaxdesmarais: Should we use `.validating` here instead?

UnitTests/BraintreeUIComponentsTests/CardNumberFieldValidatorTests.swift
```diff
@@ -0,0 +1,147 @@
+import XCTest
+@testable import BraintreeUIComponents
+
+final class CardNumberFieldValidatorTests: XCTestCase {
```
jaxdesmarais, line 4: These tests look great overall! Will we add tests for spacing at a later point? Same with making sure all spaces would not return as valid?

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
buzzamus: Ah did not know that! I'll update to use swift regex style

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
buzzamus: ah yep good catch!

UnitTests/BraintreeUIComponentsTests/CardNumberFieldValidatorTests.swift
buzzamus: Yeah, I initially had spacing checks, but removed them am holding off for now since the spacing that is added in visually every 4 digits will come in the state handling ticket, so this may be updated a bit to take care of that once the whole state handling cycle is done

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
```diff
@@ -1,8 +1,77 @@
...
+    // MARK: - CardFieldValidatorProtocol Conformance
+
+    func validate(_ value: String) -> ValidationResult {
+        let digits = value.filter { $0.isNumber }
+
+        guard !digits.isEmpty else {
+            return .invalid("Card number is required")
+        }
+
+        guard value.filter({ !$0.isNumber && $0 != " " }).isEmpty else {
```
agedd: nit(take it or leave it): can we extract the value filter into a variable above for clearer readibility

Sources/BraintreeUIComponents/CardFields/Shared/CardBrand.swift
```diff
@@ -0,0 +1,92 @@
...
+    case dinersClub
+    case unionPay
+    case maestro
+    case hiper
+    case hipercard
+    case unknown
+
+    // MARK: - Prefix Patterns
+
+    /// Strict prefix patterns checked first across all brands before relaxed patterns.
```
agedd, line 18: amazing doc strings! 🚀

Sources/BraintreeUIComponents/CardFields/Shared/CardBrand.swift
```diff
@@ -0,0 +1,92 @@
...
+            return [12, 13, 14, 15, 16, 17, 18, 19]
+        case .unionPay:
+            return [16, 17, 18, 19]
+        default:
+            return [16]
+        }
+    }
+
+    var minLength: Int {
+        validLengths.min() ?? 16
```
agedd, line 80: qq: curious as to why 16 is the default value here

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldValidator.swift
buzzamus: Done!

Sources/BraintreeUIComponents/CardFields/Shared/CardBrand.swift
buzzamus: This was largely following what the Drop-in did tbh. I am guessing just due to 16 being the most common length and that we can't actually do proper length validation checks if it is an unknown card type (which hypothetically shouldn't happen)

UnitTests/BraintreeUIComponentsTests/CardNumberFieldValidatorTests.swift
```diff
@@ -0,0 +1,147 @@
...
+
+    // MARK: - Non-Numeric Characters
+
+    func testValidate_nonNumericCharacters_returnsInvalid() {
+        XCTAssertEqual(validator.validate("4111abcd11111111"), .invalid("Card number is invalid"))
+    }
+
+    func testValidate_spacesAllowed_doesNotReturnNonNumericError() {
+        let result = validator.validate("4111 1111 1111 1111")
+        XCTAssertNotEqual(result, .invalid("Card number must contain numbers only"))
```
jaxdesmarais, line 32: I am not sure I fully understand the intent of this test. We are asserting that it is anything other than invalid and this message?

UnitTests/BraintreeUIComponentsTests/CardNumberFieldValidatorTests.swift
```diff
@@ -0,0 +1,147 @@
...
+
+    func testDetectBrand_dinersClub36_returnsDinersClub() {
+        XCTAssertEqual(validator.detectBrand(from: "36111111111111"), .dinersClub)
+    }
+
+    func testDetectBrand_dinersClub300_returnsDinersClub() {
+        XCTAssertEqual(validator.detectBrand(from: "30011111111111"), .dinersClub)
+    }
+
+    func testDetectBrand_maestroRelaxed_returnsMaestro() {
```
jaxdesmarais: It looks like this is testing the strict not relaxed maestro?

Sources/BraintreeUIComponents/CardFields/Shared/CardBrand.swift
```diff
@@ -0,0 +1,92 @@
...
+    case hiper
+    case hipercard
+    case unknown
+
+    // MARK: - Prefix Patterns
+
+    /// Strict prefix patterns checked first across all brands before relaxed patterns.
+    /// Order of cases in CaseIterable matters — Discover must come before UnionPay
+    /// to correctly handle the 622126–622925 co-branded range.
+    var prefixPatterns: [Regex<Substring>] {
```
jaxdesmarais, line 21: For all of these, I cannot remember if we need the `^` or if it can be omitted. Would you mind testing? I thought `prefixMatch(of:)` anchored to the start.

UnitTests/BraintreeUIComponentsTests/CardNumberFieldValidatorTests.swift
buzzamus: Ah whoops I will update this...this was the exact test I was working on when I stopped and had to think through if I wanted the validator to care if there were spaces or not and I think I ended up not completely finishing my desire in the test.

UnitTests/BraintreeUIComponentsTests/CardNumberFieldValidatorTests.swift
buzzamus: Ope yep i'll update and then also add a real relaxed as well as the strict

Sources/BraintreeUIComponents/CardFields/Shared/CardBrand.swift
buzzamus: Just checked and it is not needed, removed

UnitTests/BraintreeUIComponentsTests/CardNumberFieldValidatorTests.swift
buzzamus: Updated

UnitTests/BraintreeUIComponentsTests/CardNumberFieldValidatorTests.swift
buzzamus: Updated

jaxdesmarais (approved):

agedd (approved):
🚀

PR #1781, opened 2026-04-20
Add Pre-Commit Hook

agedd:

### Summary of changes

-  standardize our commit format by adopting commit prefixes 
- `.pre-commit-config.yaml` declares the commit-msg hook
- `scripts/commit-msg` is the validation script

**note**: once the PR is merged, every contributor must run the following once from the repo root after cloning:

```
mkdir -p .git/hooks
cp scripts/commit-msg .git/hooks/commit-msg
chmod +x .git/hooks/commit-msg
```
#### install (via `pre-commit` framework)
If you have `pre-commit` installed (or install it via `pip3 install pre-commit`):
```
pre-commit install --hook-type commit-msg
```

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

- @agedd


scripts/commit-msg
```diff
@@ -0,0 +1,46 @@
...
+COMMIT_MSG_FILE="$1"
+COMMIT_MSG=$(cat "$COMMIT_MSG_FILE")
+
+# Skip merge commits
+if echo "$COMMIT_MSG" | grep -qE "^Merge "; then
+  exit 0
+fi
+
+# Allowed type prefixes
+TYPES="fix|feat|docs|style|refactor|test"
```
jaxdesmarais: Do we want to add "PR feedback" or something similar? I see us use this regularly and is often helpful to see what comment was fixed in what commit. What do you think?

scripts/commit-msg
```diff
@@ -0,0 +1,46 @@
...
+  echo "   Examples:"
+  echo "     feat: add paypal vaulting support"
+  echo "     fix(venmo): resolve null pointer on auth"
+  echo "     docs: update README installation steps"
+  echo "     style: reformat card module files"
+  echo "     refactor(core): simplify token parsing logic"
+  echo "     test: add unit tests for 3DS flow"
+  echo ""
+  echo "   Allowed types: fix | feat | docs | style | refactor | test"
+  echo "   Rule: description must start lowercase after the colon"
```
jaxdesmarais: Do we want to be slightly flexible on the casing? I think the format is the most important part vs all lowercase, but not sure how easy that is to enforce.

scripts/commit-msg
jaxdesmarais: I would say "feature" should also be okay vs just feat?

buzzamus (approved):
I like the additional of the optional scope to make it quick and easy to see if it is just a specific module.

jaxdesmarais (approved):
This is going to make our commit hygiene much better overall. Thanks for making it a bit more flexible to work a bit better for us.

PR #1785, opened 2026-04-28
Increase Integration Test Coverage

agedd:
### Summary of changes
- Increase Integration Test Coverage by adding coverage for `BTLocalPaymentClient`, `BTSEPADirectDebitClient` and `BTAmexExpressClient` methods i.e `tokenize` (mostly before we launch a session)
- Combined Test Coverage for UI Tests and Integration Tests is now ~ 75% which meets the target for the SDLC framework

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

- @agedd

### DIFF
--- a/IntegrationTests/BraintreeAmexExpress_IntegrationTests.swift
+++ b/IntegrationTests/BraintreeAmexExpress_IntegrationTests.swift
@@ -1,9 +1,25 @@
+import Foundation
 import XCTest
+@testable import BraintreeCore
 @testable import BraintreeAmericanExpress
 @testable import BraintreeCard
-@testable import BraintreeCore
 
-class BraintreeAmexExpress_IntegrationTests: XCTestCase {
+class BTAmericanExpressClient_IntegrationTests: XCTestCase {
+    
+    // MARK: - Properties
+    
+    var americanExpressClient: BTAmericanExpressClient!
+    var cardClient: BTCardClient!
+    
+    // MARK: - Setup
+    
+    override func setUp() {
+        super.setUp()
+        americanExpressClient = BTAmericanExpressClient(authorization: BTIntegrationTestsConstants.sandboxClientToken)
+        cardClient = BTCardClient(authorization: BTIntegrationTestsConstants.sandboxClientToken)
+    }
+    
+    // MARK: - getRewardsBalance
     
     func testGetRewardsBalance_returnsResult() async {
         let cardClient = BTCardClient(authorization: BTIntegrationTestsConstants.sandboxClientTokenVersion3)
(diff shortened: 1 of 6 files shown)


IntegrationTests/BraintreeAmexExpress_IntegrationTests.swift
```diff
@@ -1,9 +1,25 @@
...
+    override func setUp() {
+        super.setUp()
+        americanExpressClient = BTAmericanExpressClient(authorization: BTIntegrationTestsConstants.sandboxClientToken)
+        cardClient = BTCardClient(authorization: BTIntegrationTestsConstants.sandboxClientToken)
+    }
+    
+    // MARK: - getRewardsBalance
     
     func testGetRewardsBalance_returnsResult() async {
         let cardClient = BTCardClient(authorization: BTIntegrationTestsConstants.sandboxClientTokenVersion3)
```
buzzamus: Just curious (this may be a lapse in my knowledge for V3 client tokens): is the difference in client token here noteworthy and needed for this specific test?

buzzamus: This is from an existing test, so not directly related to your changes, but since we have a cardClient being created in the setup it caught my attention

agedd: ah great question afaik i don't think there is a difference and it's not needed for this test

buzzamus (approved):
😎

IntegrationTests/BraintreeAmexExpress_IntegrationTests.swift
```diff
@@ -1,31 +1,90 @@
...
+    
+    override func setUp() {
+        super.setUp()
+        americanExpressClient = BTAmericanExpressClient(authorization: BTIntegrationTestsConstants.sandboxClientToken)
+        cardClient = BTCardClient(authorization: BTIntegrationTestsConstants.sandboxClientToken)
+    }
+    
+    // MARK: - getRewardsBalance
     
-    func testGetRewardsBalance_returnsResult() async {
```
jaxdesmarais, line 8: It looks like we removed the success case from our tests but maybe the diff is reading weird. If we did remove this, we for sure want a integration tests for the happy path.

IntegrationTests/BraintreeAmexExpress_IntegrationTests.swift
```diff
@@ -1,31 +1,90 @@
...
+                return
+            }
+            
+            visaNonce = tokenizedCard.nonce
+            cardExpectation.fulfill()
+        }
+        
+        waitForExpectations(timeout: 5)
+        
+        guard let nonce = visaNonce else {
```
jaxdesmarais:
```suggestion
        guard let visaNonce else {
```

IntegrationTests/BraintreeAmexExpress_IntegrationTests.swift
agedd: ahh yep good catch ~ addressed this and added a happy path test in 3fb2f3c394d9509e611982663ca1151216fd23a3

saralvasquez (approved):
The test cases make sense to me! I'm not 100% on the semantics and such given the iOS of it all but the logic makes sense

jaxdesmarais (approved):

PR #1789, opened 2026-05-01
[Open Dev] Add Code Coverage Check Workflow

buzzamus:
### Summary of changes
- Added code coverage percentage check as part of CI process within unit test workflow after tests have run:

Screenshot 2026-05-01 at 3 54 15 PM

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used Claude to help troubleshoot issues with getting code coverage check to work locally and overcome an Xcode version number issue.

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,34 @@ jobs:
       - name: Install CocoaPod dependencies
         run: pod install
       - name: Run Unit Tests
-        run: set -o pipefail && xcodebuild -workspace 'Braintree.xcworkspace' -sdk 'iphonesimulator' -configuration 'Debug' -scheme 'UnitTests' -destination 'name=iPhone 16 Pro,OS=18.5,platform=iOS Simulator' -derivedDataPath DerivedData test | ./Pods/xcbeautify/xcbeautify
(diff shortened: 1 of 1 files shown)


jaxdesmarais (approved):
This is really cool! I wonder if there is a way to automatically output it as a comment in the PR in the future?

buzzamus:
> This is really cool! I wonder if there is a way to automatically output it as a comment in the PR in the future?

That was originally my intention and would love to get there...may explore this further on a future open dev day.

saralvasquez (approved):
This looks good to me. You mentioned that adding this to the test workflow was the only reasonable way to get the coverage data. Is that because we need some kind of data collected in the test step? It would be nice to be able to get coverage data without needing to run all the test which is the only reason I ask

buzzamus:
> This looks good to me. You mentioned that adding this to the test workflow was the only reasonable way to get the coverage data. Is that because we need some kind of data collected in the test step? It would be nice to be able to get coverage data without needing to run all the test which is the only reason I ask

Yeah, using the Xcode coverarage feature unfortunately requires running all tests in order to get the percentage. I know there are other tools out there that I hope to explore at some point, but this was the tool used to actually measure our current percentage, so just started with this. I agree it would be great to get the coverage without needing to run all the tests, so hopefully I can find something else long term

PR #1793, opened 2026-05-12
Add Create Payment Context Events

agedd:
### Summary of changes
- Add events when create payment context is started/succeeded/failed

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**

writing and formatting unit test cases for the new analytic events

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

- @agedd

### DIFF
--- a/Sources/BraintreeVenmo/BTVenmoClient.swift
+++ b/Sources/BraintreeVenmo/BTVenmoClient.swift
@@ -114,7 +114,7 @@ import BraintreeCore
         let metadata = apiClient.metadata
         metadata.source = .venmoApp
 
-        apiClient.sendAnalyticsEvent(BTVenmoAnalytics.createPaymentContextStarted)
+        apiClient.sendAnalyticsEvent(BTVenmoAnalytics.createPaymentContextStarted, isVaultRequest: shouldVault)
         
         let graphQLParameters = VenmoCreatePaymentContextGraphQLBody(
             request: request,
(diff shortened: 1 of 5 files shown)


buzzamus (approved):
😎 
Looks good to me!

This is not related to your changes at all, but reviewing the unit tests made me realize that there are multiple unit tests with a bunch of repeatable code....I feel like this might be a good candidate for in the future to see if the repeated code could be put into a helper function or something...may not be possible just thinking out loud

Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -123,6 +125,7 @@ import BraintreeCore
         do {
             (body, _) = try await apiClient.post("", parameters: graphQLParameters, httpType: .graphQLAPI)
         } catch {
+            apiClient.sendAnalyticsEvent(BTVenmoAnalytics.createPaymentContextFailed, isVaultRequest: shouldVault)
```
jaxdesmarais: Do we want to also send `errorDescription` and `contextID` for the success and failure calls? Or anything else from the the other start/succeed/failed events?

jaxdesmarais (commented):
I think we added a changelog entry on the android PR if we want to match that here

Sources/BraintreeVenmo/BTVenmoClient.swift
agedd: ahh yep good callout - addressed in 1258ff889fe0b1f0868bb19fce34a3b0fc17f5a3

jaxdesmarais (commented):
I think just missing changelog to match Android

jaxdesmarais (approved):

PR #1794, opened 2026-05-12
Handle Card Number Field State

buzzamus:
### Summary of changes
- Injects Validator into Card Fields View Model and sets the focus and valid state for field to be used in the CardFields component

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used to review code and ensured it followed pattern already highlighted by CVV state handling PR

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
@@ -2,25 +2,39 @@ import Foundation
 
 @MainActor
 class CardNumberFieldViewModel: ObservableObject {
-    
+
     @Published private(set) var value: String = ""
     @Published private(set) var validationState: ValidationResult = .valid
     @Published private(set) var cardBrand: CardBrand = .unknown
-    @Published var isFocused: Bool = false
+    @Published var isFocused: Bool = false {
+        didSet {
+            // Show validation errors only after the user leaves the field
+            if !isFocused {
+                validationState = validator.validate(value)
+            }
+        }
+    }
 
-    private let validator = CardNumberFieldValidator()
+    var shouldAutoAdvance: Bool { validationState == .valid && !value.isEmpty }
+    var maxLength: Int { cardBrand.maxLength }
 
-    // TODO: Implement auto-advance logic w/ state change
-    var shouldAutoAdvance: Bool { false }
+    // MARK: - Private Properties
 
-    // TODO: Implement validation and formatting w/ state change
+    private let validator: CardNumberFieldValidator
+
+    init(validator: CardNumberFieldValidator = CardNumberFieldValidator()) {
+        self.validator = validator
+    }
+
+    // MARK: - Internal Methods
+
     func updateValue(_ newValue: String) {
         value = newValue
-        
-        // TODO: call updateCardBrand based off of value changes from validator (handled w/ state change ticket)
-    }
-    
-    private func updateCardBrand(with brand: CardBrand) {
-        self.cardBrand = brand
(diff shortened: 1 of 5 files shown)


Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
```diff
@@ -2,25 +2,52 @@ import Foundation
...
 class CardNumberFieldViewModel: ObservableObject {
-    
+
     @Published private(set) var value: String = ""
     @Published private(set) var validationState: ValidationResult = .valid
     @Published private(set) var cardBrand: CardBrand = .unknown
-    @Published var isFocused: Bool = false
+    @Published var isFocused: Bool = false {
+        didSet {
+            // Show validation errors only after the user leaves the field
```
agedd, line 11: quick question: do we want to show the validation error while the card number field is in focused state 👀

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
```diff
@@ -2,25 +2,52 @@ import Foundation
...
+    }
+
+    // MARK: - Internal Methods
 
-    // TODO: Implement auto-advance logic w/ state change
-    var shouldAutoAdvance: Bool { false }
+    func formatted(digits: String) -> String {
+        var result = ""
+        var index = digits.startIndex
+        for (i, groupSize) in cardBrand.digitGroups.enumerated() {
```
jaxdesmarais: Can we rename `i` here to what it represents?

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
```diff
@@ -2,25 +2,52 @@ import Foundation
...
-        
-        // TODO: call updateCardBrand based off of value changes from validator (handled w/ state change ticket)
-    }
-    
-    private func updateCardBrand(with brand: CardBrand) {
-        self.cardBrand = brand
+        cardBrand = validator.detectBrand(from: newValue)
+
+        let result = validator.validate(newValue)
+        if result == .valid {
```
jaxdesmarais: could we combine these into a single if let?

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
buzzamus: The only icky part about that is then having to decipher when someone is still typing and something isn't currently valid yet and to ignore vs. just waiting until after the user is off the field. Briefly it was discussed on having some sort of delay, but that could get sticky as well.

I think we could do an in depth look into this after the initial release of the card fields though for sure

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
buzzamus: Done!

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldViewModel.swift
buzzamus: I used `if case` since it involves an enum value

jaxdesmarais (approved):

agedd (approved):
😎

PR #1796, opened 2026-05-13
Add Query Payment Context Events

agedd:
### Summary of changes
- Add events when query payment context is called

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

- @agedd

### DIFF
--- a/Sources/BraintreeVenmo/BTVenmoClient.swift
+++ b/Sources/BraintreeVenmo/BTVenmoClient.swift
@@ -302,12 +302,6 @@ import BraintreeCore
                     let error = BTVenmoError.invalidBodyReturned
                     notifyFailure(with: error)
                     appSwitchCompletion(nil, error)
-                    apiClient.sendAnalyticsEvent(
-                        BTVenmoAnalytics.queryPaymentContextFailed,
-                        contextID: contextID,
-                        errorDescription: error.localizedDescription,
-                        isVaultRequest: shouldVault
-                    )
                     return
                 }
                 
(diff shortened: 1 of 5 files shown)


Sources/BraintreeVenmo/BTVenmoClient.swift
```diff
@@ -290,16 +290,33 @@ import BraintreeCore
...
                     return
                 }
-
+                
+                apiClient.sendAnalyticsEvent(
+                    BTVenmoAnalytics.queryPaymentContextSucceeded,
+                    contextID: contextID,
+                    isVaultRequest: shouldVault
+                )
+
```
buzzamus, line 313: This is mostly a question due to me not quite knowing what is considered to be a successful query payment context, but should this event be sent after the handleVaultingIfNeeded call is done, or since it has separate events we consider it a completely separate part of our analytics trigger flow....does my question make sense?

agedd: yeah that makes sense ~ great question!  the query payment context events are specifically scoped to just the GraphQL fetch, so `queryPaymentContextSucceeded` fires as soon as we get a valid response back — before vaulting kicks off.  i believe vaulting has its own events ( i.e `tokenizeSucceeded`), so keeping them separate makes it easier to pinpoint exactly where something went wrong in the flow.

jaxdesmarais (approved):
Thanks for adding these events and additional details!

buzzamus (approved):
😎

PR #1797, opened 2026-05-13
Make cc number optional

buzzamus:
### Summary of changes
- Make BTCard number optional to be in line with Android and Web SDK behavior. This allows some certain use cases where merchants wish to tokenize information other than just CVV (but without card number).

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Fix my bad commit history since I was on the wrong branch. Also, to double check and make sure I found all the areas that I needed to change

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

- @buzzamus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   * Add `payPalPayLater` case to `BTButtonType` enum
 * BraintreeVenmo
   * Add analytics tracking for `createPaymentContext` GraphQL call
+* BraintreeCard
+  * Make `number` field optional in `BTCard`
 
 ## 7.5.0 (2026-02-25)
 * BraintreePayPal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   * Add analytics tracking for `createPaymentContext` GraphQL call
   * Add analytics tracking for `queryPaymentContext` GraphQL call
 * BraintreeCard
-  * Make `number` field optional in `BTCard`
+  * Add `BTCard(expirationMonth:expirationYear:cvv:)` to allow expiration date and cvv only tokenization
 
 ## 7.5.0 (2026-02-25)
 * BraintreePayPal
--- a/Sources/BraintreeCard/BTCard.swift
+++ b/Sources/BraintreeCard/BTCard.swift
@@ -31,41 +31,12 @@ import BraintreeCore
     let authenticationInsightRequested: Bool
     let merchantAccountID: String?
 
-    // MARK: - Initializer
-    
-    /// Creates a Card
-    /// - Parameters:
-    ///   - number: Optional: The card number. If omitted, only the provided fields (e.g. expiration date and CVV) will be tokenized.
-    ///   - expirationMonth: Required: The expiration month as a one or two-digit number on the Gregorian calendar.
-    ///   - expirationYear: Required: The expiration year as a two or four-digit number on the Gregorian calendar.
-    ///   - cvv: Required: The card verification code (like CVV or CID).
-    ///   - postalCode: Optional: The postal code associated with the card's billing address.
-    ///   - cardholderName: Optional: The cardholder's name.
-    ///   - firstName: Optional: First name on the card.
-    ///   - lastName: Optional: Last name on the card.
-    ///   - company: Optional: Company name associated with the card.
-    ///   - streetAddress: Optional: The street address associated with the card's billing address.
-    ///   - extendedAddress: Optional: The extended address associated with the card's billing address.
-    ///   - locality: Optional: The city associated with the card's billing address.
-    ///   - region: Optional: Either a two-letter state code (for the US), or an ISO-3166-2 country subdivision code of up to three letters.
-    ///   - countryName: Optional: The country name associated with the card's billing address.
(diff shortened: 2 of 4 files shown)


Sources/BraintreeCard/BTCard.swift
```diff
@@ -35,7 +35,7 @@ import BraintreeCore
     
     /// Creates a Card
     /// - Parameters:
-    ///   - number: Required: The card number.
+    ///   - number: Optional: The card number. If omitted, only the provided fields (e.g. expiration date and CVV) will be tokenized.
```
agedd: nit: should we maybe consider making a third init for this change and make `number` optional there? that way we can leave the original public init untouched - wdyt?

buzzamus: Ah yeah I think that makes sense!

buzzamus: Updated!

agedd (approved):
😎

Sources/BraintreeCard/BTCard.swift
```diff
@@ -86,36 +132,49 @@ import BraintreeCore
...
+
     /// Creates a new instance of `BTCard` with only a CVV value,
     /// setting default values for all other parameters.
     /// This initializer should only be used if you wish to create a
     /// CVV-only payment method nonce to verify a card already stored in your Vault.
     /// - Parameters:
     ///   - cvv: The card verification code (like CVV or CID).
     public convenience init(cvv: String) {
-        self.init(number: "", expirationMonth: "", expirationYear: "", cvv: cvv)
+        self.init(number: nil, expirationMonth: "", expirationYear: "", cvv: cvv)
```
jaxdesmarais, line 177: Does switching this from empty to nil blow anything up in the API?

Sources/BraintreeCard/BTCard.swift
```diff
@@ -64,7 +110,7 @@ import BraintreeCore
     ///     - Note: Use this flag with caution. By enabling client-side validation, certain tokenize card requests may result in adding the card to the vault. These semantics are not currently documented.
     ///   - authenticationInsightRequested: Optional: If authentication insight is requested. If this property is set to `true`, a `merchantAccountID` must be provided. Defaults to `false`.
     ///   - merchantAccountID: Optional: The merchant account ID.
-    public init(
+    public convenience init(
         number: String,
```
jaxdesmarais, line 114: I can't quite wrap my head around how number could be required here but optional? Am I missing something obvious or can you eli5?

Sources/BraintreeCard/BTCard.swift
buzzamus: We don't have a CVV only integration test or a verification flow in the demo (we probably need both tbh), BUT I can successfully tokenize and get a nonce back for both cvv only and cvv and expiration date in our credit card form demo. I get an error when submitting for payment since it is a txn.sale call, but the tokenize call is successful. I can reach out the the cards or gateway team though if you think maybe the payment method update or some other call may be different?

Sources/BraintreeCard/BTCard.swift
buzzamus: This is definitely the quirky part about all the convenience initializers. Here are all the scenarios:

credit card number is included: Since  `BTCard(number:) `will be called and the compiler will enforce number as non-optional for that init flow. The corresponding convenience init is called and casts it to a non-nil optional String?

credit card number not included: Since it will be one of the initializers that doesn't have number as a parameter and won't accept one the corresponding convenience init will have a value of `number: nil`. when this is encoded to JSON the value will be omitted since it is a nil value.

So, tl;dr - in the scenarios where number can't be passed nil will always be passed through the corresponding internal inits. When number is passed it will always have a value so, it won't hit the flow where it can be nil.

I hope that made sense, because I typed this forever trying to figure out the right words 🤣

Sources/BraintreeCard/BTCard.swift
buzzamus: I added an integration test, but it is more just for future assurance than anything

jaxdesmarais: Nope that makes sense, thanks for the explanation. :)

Sources/BraintreeCard/BTCard.swift
jaxdesmarais: Gotcha gotcha, I think that makes sense, but yeah the nuances of convenience inits always takes me a bit to wrap my head around.

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 * BraintreeVenmo
   * Add analytics tracking for `createPaymentContext` GraphQL call
   * Add analytics tracking for `queryPaymentContext` GraphQL call
+* BraintreeCard
+  * Make `number` field optional in `BTCard`
```
jaxdesmarais: Take or leave, but may be a bit more accurate:

```suggestion
  * Add `BTCard(expirationMonth:expirationYear:cvv:)` to allow expiration date and cvv only tokenization
```

jaxdesmarais (approved):

PR #1800, opened 2026-05-19
Create Card Fields Component

buzzamus:
### Summary of changes
- This PR creates the shared CardFields component that houses all the fields and hooks up the state handling to all 3.

Screenshot 2026-05-19 at 3 48 07 PM
Screenshot 2026-05-19 at 3 47 41 PM
Screenshot 2026-05-19 at 3 47 34 PM
Screenshot 2026-05-19 at 3 47 21 PM

### AI Usage

**Which AI Agent Was Used?**
- [] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Helped to fix bugs in behavior of the validation triggers

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus 

----

### DIFF
--- a/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
@@ -71,7 +71,7 @@ struct CVVFieldView: View {
             .accessibilityLabel("CVV help")
             .accessibilityHint("Tap for information about where to find your CVV")
             .popover(isPresented: $showCVVHint, arrowEdge: .bottom) {
-                VStack(alignment: .leading, spacing: 8) {
+                VStack(alignment: .leading, spacing: 12) {
                     Text("CVV")
                         .font(.system(size: 16, weight: .semibold))
                         .foregroundColor(Color(.label))
@@ -80,9 +80,8 @@ struct CVVFieldView: View {
                         .font(.system(size: 14))
                         .foregroundColor(Color(.secondaryLabel))
                         .fixedSize(horizontal: false, vertical: true)
-                    // width fixed 261, height Hug 104
-                    // border radius 12
                 }
+                .frame(width: 261)
                 .padding(16)
                 .accessibilityElement(children: .combine)
                 .accessibilityLabel("CVV help information")
--- a/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
@@ -80,6 +80,8 @@ struct CVVFieldView: View {
                         .font(.system(size: 14))
                         .foregroundColor(Color(.secondaryLabel))
                         .fixedSize(horizontal: false, vertical: true)
+                    // width fixed 261, height Hug 104
+                    // border radius 12
                 }
                 .padding(16)
                 .accessibilityElement(children: .combine)
--- a/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
@@ -95,6 +95,9 @@ struct CVVFieldView: View {
         .onChange(of: viewModel.shouldAutoAdvance) { _, shouldAdvance in
             if shouldAdvance { onAutoAdvance?() }
         }
+        .onChange(of: viewModel.isFocused) { _, focused in
+            if focused { isFocused = true }
+        }
         .contentShape(Rectangle())
         .onTapGesture {
             isFocused = true
--- a/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
@@ -9,7 +9,8 @@ class CVVFieldViewModel: ObservableObject {
         didSet {
             // Show validation errors only after the user leaves the field
             if !isFocused {
-                validationState = validator.validate(rawValue)
+                let result = validator.validate(rawValue)
+                validationState = result == .validating ? .invalid("CVV is invalid") : result
             }
         }
     }
@@ -36,14 +37,20 @@ class CVVFieldViewModel: ObservableObject {
 
     // MARK: - Internal Methods
 
+    func updateExpectedLength(_ length: Int?) {
+        validator.expectedLength = length
+    }
+
     func updateValue(_ newValue: String) {
-        let digits = String(newValue.filter { $0.isNumber }.prefix(4))
+        let digits = String(newValue.filter { $0.isNumber }.prefix(maxLength))
         rawValue = digits
         value = digits
 
         let result = validator.validate(digits)
         if result == .valid {
             validationState = .valid
+        } else if validationState == .valid {
+            validationState = .validating
         }
 
         let oldCount = characters.count
--- a/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
@@ -9,13 +9,13 @@ class CVVFieldViewModel: ObservableObject {
         didSet {
             // Show validation errors only after the user leaves the field
             if !isFocused {
-                let result = validator.validate(rawValue)
+                let result = validator.validate(value)
                 validationState = result == .validating ? .invalid("CVV is invalid") : result
             }
         }
     }
 
-    var shouldAutoAdvance: Bool { validationState == .valid && !rawValue.isEmpty }
+    var shouldAutoAdvance: Bool { validationState == .valid && !value.isEmpty }
 
     var maxLength: Int { validator.expectedLength ?? 4 }
 
@@ -23,9 +23,6 @@ class CVVFieldViewModel: ObservableObject {
 
     /// Individual characters with their masking state — drives the custom display
     @Published private(set) var characters: [CVVCharacter] = []
-
-    /// Raw digits only — no masking, used for tokenization
-    @Published private(set) var rawValue: String = ""
 
     // MARK: - Private Properties
 
@@ -43,14 +40,18 @@ class CVVFieldViewModel: ObservableObject {
 
     func updateValue(_ newValue: String) {
         let digits = String(newValue.filter { $0.isNumber }.prefix(maxLength))
-        rawValue = digits
         value = digits
 
         let result = validator.validate(digits)
-        if result == .valid {
+        switch result {
+        case .valid:
             validationState = .valid
-        } else if validationState == .valid {
-            validationState = .validating
+        case .invalid:
+            break // preserve existing state — don't regress from .invalid
+        case .validating:
+            if case .invalid(_) = validationState { } else {
+                validationState = .validating
+            }
         }
 
         let oldCount = characters.count
--- a/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
@@ -49,7 +49,7 @@ class CVVFieldViewModel: ObservableObject {
         case .invalid:
             break // preserve existing state — don't regress from .invalid
         case .validating:
-            if case .invalid(_) = validationState { } else {
+            if case .invalid = validationState { } else {
                 validationState = .validating
             }
         }
--- a/Sources/BraintreeUIComponents/CardFields/CardFields.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CardFields.swift
@@ -22,9 +22,12 @@ struct CardFields: View {
             )
 
             HStack(spacing: 12) {
-                ExpirationDateFieldView(viewModel: viewModel.expirationDateViewModel) {
-                    viewModel.cvvViewModel.isFocused = true
-                }
+                ExpirationDateFieldView(
+                    viewModel: viewModel.expirationDateViewModel,
+                    onAutoAdvance: {
+                        viewModel.cvvViewModel.isFocused = true
+                    }
+                )
 
                 CVVFieldView(viewModel: viewModel.cvvViewModel)
             }
--- a/Sources/BraintreeUIComponents/CardFields/CardFields.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CardFields.swift
@@ -22,12 +22,9 @@ struct CardFields: View {
             )
 
             HStack(spacing: 12) {
-                ExpirationDateFieldView(
-                    viewModel: viewModel.expirationDateViewModel,
-                    onAutoAdvance: {
-                        viewModel.cvvViewModel.isFocused = true
-                    }
-                )
+                ExpirationDateFieldView(viewModel: viewModel.expirationDateViewModel) {
+                    viewModel.cvvViewModel.isFocused = true
+                }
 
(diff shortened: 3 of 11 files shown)


Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
```diff
@@ -1,6 +1,11 @@
...
-    var state: String = ""
+@MainActor
+final class CardFieldsViewModel: ObservableObject {
+
+    // MARK: - Internal Properties
+
+    let cardNumberViewModel = CardNumberFieldViewModel()
+    let expirationDateViewModel = ExpirationDateFieldViewModel()
+    let cvvViewModel = CVVFieldViewModel()
 }
```
buzzamus, line 30: I'm not sure I totally like this route of having 3 separate view models just called in this, but 2 things made me lean this way:

1. It seemed like a lot of work and redundant code to just wrap each actual view model's call into a separate function
2. This view model will have more logic added to it when the tokenize call and other remaining pieces are added in.

I'm still opening to suggestions though

Sources/BraintreeUIComponents/CardFields/CardFields.swift
```diff
@@ -0,0 +1,38 @@
...
+                },
+                onBrandChanged: { brand in
+                    let length: Int? = brand == .unknown ? nil : brand.cvvLength
+                    viewModel.cvvViewModel.updateExpectedLength(length)
+                }
+            )
+
+            HStack(spacing: 12) {
+                ExpirationDateFieldView(viewModel: viewModel.expirationDateViewModel) {
+                    viewModel.cvvViewModel.isFocused = true
```
agedd, line 29: q: should this be within the `onAutoAdvance` closure 👀 like on line 15

Sources/BraintreeUIComponents/CardFields/CardFields.swift
```diff
@@ -0,0 +1,38 @@
...
+                }
+
```
agedd, line 31: nit(optional): can delete empty line

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
```diff
@@ -1,6 +1,11 @@
 import Foundation
 
-final class CardFieldsViewModel {
-    
-    var state: String = ""
+@MainActor
+final class CardFieldsViewModel: ObservableObject {
```
agedd, line 5: nit: maybe we can add some doc strings to this class

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
```diff
@@ -36,14 +37,20 @@ class CVVFieldViewModel: ObservableObject {
...
-        let digits = String(newValue.filter { $0.isNumber }.prefix(4))
+        let digits = String(newValue.filter { $0.isNumber }.prefix(maxLength))
         rawValue = digits
         value = digits
 
         let result = validator.validate(digits)
         if result == .valid {
             validationState = .valid
+        } else if validationState == .valid {
+            validationState = .validating
```
agedd: looks like the `else if` may be no-op ->  if `validationState == .valid` we're setting it to `.valid` again, same as the first branch 👀

Sources/BraintreeUIComponents/CardFields/CardFields.swift
buzzamus: ope yep mixed up syntax type. Will fix thanks!

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: It is setting it to `validating` in line 53, but I think this just points to how confusing this is to read....I had to do this to cover a scenario where a user has a valid CVV, but then deletes a digit so it goes to back to `.validating` to avoid UI issues...I wonder if this makes it easier to read:

```
  if result == .valid {
      validationState = .valid
  } else if validationState != .invalid {
      validationState = .validating
  }
  ```
  
  Thoughts? That way the switching in terms by getting `invalid `in there makes it stand out more

File: Sources/BraintreeUIComponents/CardFields/CardFields.swift (hunk 3 above)
Comment (buzzamus) [reply Sources/BraintreeUIComponents/CardFields/CardFields.swift:31]: good eye!

File: Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift (hunk 4 above)
Comment (buzzamus) [reply Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift:5]: Ah yep I should probably start on those 🤣 . I think I just forgot since this still needs the tokenize call and the button added, but best to start adding some now I suppose. 

I'll update with the parts that make sense now

File: Sources/BraintreeUIComponents/CardFields/CardFields.swift (hunk 2 above)
Comment (buzzamus) [reply Sources/BraintreeUIComponents/CardFields/CardFields.swift:29]: Actually since this is the last argument our linter doesn't like using this syntax for the trailing closure ugh

File: Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift (hunk 5 above)
Comment (agedd) [reply Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift]: ah yep, this reads much clearer to me! we could also go with a switch if that's easier — good either way: 
```
switch result {
case .valid:
    validationState = .valid
case .invalid:
    break // preserve existing state — don't regress from .invalid
case .validating:
    if validationState != .invalid {
        validationState = .validating
    }
}
```

File: Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift [hunk 6]
```diff
@@ -9,7 +9,8 @@ class CVVFieldViewModel: ObservableObject {
         didSet {
             // Show validation errors only after the user leaves the field
             if !isFocused {
-                validationState = validator.validate(rawValue)
+                let result = validator.validate(rawValue)
```
agedd: nit(optional): can we rename `rawValue` to `value` here to match the other two view models

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
agedd: ah makes sense — each field has enough of its own logic to justify the separation, and the parent VM as a coordinator is a clean pattern. once tokenization lands it'll have more work to do anyway. i'm good with this.

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: oh yeah I think a switch makes it easier to read. I'll add that!

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
buzzamus: Actually I just realized this class is going to be internal and not really touched by the merchant directly, but I'll start adding some to the public class (which is just CardFields)

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: Updated! This was actually redundant since I already have value in there, but separated them out and really didn' t need both

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
buzzamus: Put doc strings in CardFields

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: It looks a little wonky, but I had to do this:

```
        case .validating:
            if case .invalid = validationState { } else {
                validationState = .validating
            }
        }
```
Due to .invalid having a string argument

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
```diff
@@ -81,6 +81,7 @@ struct CVVFieldView: View {
                         .foregroundColor(Color(.secondaryLabel))
                         .fixedSize(horizontal: false, vertical: true)
                 }
+                .frame(width: 261)
```
jaxdesmarais, line 84: Will this fix width case issues on different sized devices? We could also tackle it across the board in a follow up PR but I assume we will want all of these to be somewhat flexible based on device sizes.

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
```diff
@@ -36,14 +34,24 @@ class CVVFieldViewModel: ObservableObject {
...
+        let digits = String(newValue.filter { $0.isNumber }.prefix(maxLength))
         value = digits
 
         let result = validator.validate(digits)
-        if result == .valid {
+        switch result {
+        case .valid:
             validationState = .valid
+        case .invalid:
+            break // preserve existing state — don't regress from .invalid
```
jaxdesmarais, line 50: Does this mean it stays invalid until you're done typing?

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
```diff
@@ -47,6 +47,9 @@ struct ExpirationDateFieldView: View {
         .onChange(of: viewModel.shouldAutoAdvance) { _, shouldAutoAdvance in
             if shouldAutoAdvance { onAutoAdvance?() }
         }
+        .onChange(of: viewModel.isFocused) { _, focused in
+            if focused { isFocused = true }
```
jaxdesmarais, line 51: When does this changed back to false?

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldView.swift
```diff
@@ -4,6 +4,7 @@ struct CardNumberFieldView: View {
 
     @ObservedObject var viewModel: CardNumberFieldViewModel
     var onAutoAdvance: (() -> Void)?
+    var onBrandChanged: ((CardBrand) -> Void)?
```
jaxdesmarais: Does this need to be optional or do we always have a brand?

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
jaxdesmarais: Should these have a Published/Observable object wrapper?  If any child VM publishes changes, the parent won't automatically trigger view updates unless the child VMs are themselves ObservableObject and the view observes them directly (which it does via @StateObject)? Or is it fine?

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
buzzamus: ah yeah good point. I was putting in the width values from the Figma design, but I think these do need to have a min/max or some sort of value that allows them to expand to some degree

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
buzzamus: Yes exactly. I think @agedd and I were discussing this (and I touched base with Android too) and I thought this was the best approach vs. potentially too much noise with it saying their cvv (or any field really) was invalid while they were still typing

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldView.swift
buzzamus: We always have a CardBrand since the default image is used for default values (so, in this case CardBrand value of `.unknown`)

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
buzzamus: SwiftUI actually sets isFocused automatically when a user focuses on another element

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
buzzamus: No, the CardFields delegates each individual field's behavior to their own separate viewmodel and re-render any changes independently. Although thinking out loud I think I need to add a published event for when the form is valid so when we go to tokenize we check to make sure all 3 are. I'll add something in for that. Other than the `isFormValid` check the view models were intended to handle their own field by design, but you making me talk through it made me realize I missed a piece

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
jaxdesmarais: Yeah, I think we could take a few approaches, we could either let a merchant pass a main width in but set a min (I think we did this for buttons). Depending on the width we would then adjust the cvv + expiry field based on what is set. Or we could set it to the device width minus padding as a default. I'll defer to you on what we think is best. If it needs to be a follow up, that's fine too.

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldViewModel.swift
jaxdesmarais: That makes sense to me! As long as we are aligned across platforms, I think that's the big thing.

Sources/BraintreeUIComponents/CardFields/ExpirationDateField/ExpirationDateFieldView.swift
jaxdesmarais: Oh dope!

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldView.swift
jaxdesmarais: Nice - I think my question was around the closure as a whole so we don't need to unwrap it. Not a deal breaker, was just curious if we could do `((CardBrand) -> Void)` instead.

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
buzzamus: Updated to use Combine to check for all 3 fields to be valid. I think when we drop support for iOS 16 a cleaner solution would be to use @ObservedObject, but this is the next best thing IMO

Sources/BraintreeUIComponents/CardFields/CVVField/CVVFieldView.swift
buzzamus: I'll create a card for this

Sources/BraintreeUIComponents/CardFields/CardNumberField/CardNumberFieldView.swift
buzzamus: Good call! I think at one point in my head I was going to allow a nil value, but not anymore. Updated to drop the Optional version

jaxdesmarais (approved):

agedd (approved):
🚀

PR #1802, opened 2026-05-20
Run Data Collector Setup on Main Thread

buzzamus:
### Summary of changes
-  During the setup process, Magnes uses UIPasteboard, which must run on the main thread. This can cause race conditions that lead to intermittent crashes. This fix puts the setup call and the `generateClientMetadataID` function (which calls setup) on the Main thread.

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Assistance in discovering issue around UIPasteboard requiring main thread as well as assistance in how to write a test for this.

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

- @buzzamus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+
+* BraintreeDataCollector
+  * Fix intermittent crash in `collectDeviceData` caused by an internal SDK component accessing `UIPasteboard` off the main thread
+
 ## 7.6.0 (2026-05-18)
 * Fix inconsistency in minimum deployment target, which is now consistently iOS 16 (fixes #1757)
 * BraintreeShopperInsights
--- a/Sources/BraintreeDataCollector/BTDataCollector.swift
+++ b/Sources/BraintreeDataCollector/BTDataCollector.swift
@@ -62,7 +62,10 @@ import BraintreeCore
     public func collectDeviceData() async throws -> String {
         let configuration = try await fetchConfiguration()
 
-        let clientMetadataID: String = generateClientMetadataID(with: configuration)
+        // MagnesSDK.setUp accesses UIPasteboard internally, which requires the main thread.
+        let clientMetadataID: String = await MainActor.run {
+            generateClientMetadataID(with: configuration)
+        }
         let dataDictionary: [String: String] = ["correlation_id": clientMetadataID]
 
         guard let jsonData = try? JSONSerialization.data(withJSONObject: dataDictionary) else {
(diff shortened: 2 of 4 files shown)


agedd (approved):
ty for fixing this! 🚀

Sources/BraintreeDataCollector/BTDataCollector.swift
```diff
@@ -62,7 +62,10 @@ import BraintreeCore
     public func collectDeviceData() async throws -> String {
         let configuration = try await fetchConfiguration()
 
-        let clientMetadataID: String = generateClientMetadataID(with: configuration)
+        // MagnesSDK.setUp accesses UIPasteboard internally, which requires the main thread.
```
agedd: nit(take it or leave it): 
```suggestion
        // MagnesSDK.shared().setUp accesses UIPasteboard internally, which requires the main thread.
```

Sources/BraintreeDataCollector/BTDataCollector.swift
```diff
@@ -101,13 +104,16 @@ import BraintreeCore
...
                 let magnesEnvironment = self.getMagnesEnvironment(from: self.config)
 
-                try? MagnesSDK.shared().setUp(
-                    setEnviroment: magnesEnvironment,
-                    setOptionalAppGuid: self.deviceIdentifier(),
-                    disableRemoteConfiguration: false,
-                    disableBeacon: false,
-                    magnesSource: .BRAINTREE
-                )
+                // MagnesSDK.setUp accesses UIPasteboard internally, which requires the main thread.
```
agedd: nit(optional): same as above

buzzamus: Good call thanks!

Sources/BraintreeDataCollector/BTDataCollector.swift
```diff
@@ -62,7 +62,10 @@ import BraintreeCore
     public func collectDeviceData() async throws -> String {
         let configuration = try await fetchConfiguration()
 
-        let clientMetadataID: String = generateClientMetadataID(with: configuration)
+        // MagnesSDK.shared().setUp accesses UIPasteboard internally, which requires the main thread.
```
jaxdesmarais, line 65: not a blocker, cannot remember if we do `//` or `///` here and below?

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+
+* BraintreeDataCollector
+  * Fix intermittent crash in `collectDeviceData` caused by an internal SDK component accessing `UIPasteboard` off the main thread
```
jaxdesmarais, line 6: We can add the `(fixes #XXXX)` here too

jaxdesmarais (approved):

CHANGELOG.md
buzzamus: This isn't actually related to the issue brought up after all

jaxdesmarais: Oh dope, then ignore. :)

Sources/BraintreeDataCollector/BTDataCollector.swift
buzzamus: I think it looks like /// is for public API doc comments whereas // is for internal comments on code

PR #1804, opened 2026-05-22
Put PayPal MetadataID Call On Main Thread

buzzamus:
### Summary of changes
- This is the same bug as the Data Collector and Local Payments fixes where due to a Magnes use of UIPasteboard the call needs to be on the main thread

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used to find the location of the call to put on main thread since this is the same issue that came up in another place

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

- @buzzamus

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * BraintreeDataCollector
   * Fix intermittent crash in `collectDeviceData` caused by an internal SDK component accessing `UIPasteboard` off the main thread
+* BraintreePayPal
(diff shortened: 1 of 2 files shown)


agedd (approved):
ty for fixing this! 🚀

noguier (approved):
Approving, since the change is similar to #1802 which was reviewed thoroughly.

PR #1807, opened 2026-05-27
Fix ClientMetadataID Being Nil For PayPal Checkout With BillingAgreement Flow

buzzamus:
### Summary of changes
- When a merchant uses the checkout flow with PayPal and has `shouldRequestBillingAgreement` set to `true` the PP checkout token is returned for the CMID, but once the flow is completed a BA token is returned since a billing agreement has been requested. This caused the new BA token to override the existing CMID, which could no longer find the new token in the existing dictionary.  This fix resolves that issue by setting to the EC token for this specific scenario.

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
For creating a quick testing scneario in the Demo app by adding a BA request toggle.

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

- @buzzamus

### DIFF
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -3,7 +3,7 @@ import UIKit
 import BraintreePayPal
 import BraintreeCore
 
-// swiftlint:disable type_body_length
+// swiftlint:disable type_body_length file_length
 class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
 
     lazy var payPalClient = BTPayPalClient(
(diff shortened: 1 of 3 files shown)


Demo/Application/Features/PayPalWebCheckoutViewController.swift
```diff
@@ -78,6 +78,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
     let rbaDataToggle = Toggle(title: "Recurring Billing (RBA) Data")
     let contactInformationToggle = Toggle(title: "Add Contact Information")
     let amountBreakdownToggle = Toggle(title: "Amount Breakdown")
+    let requestBillingAgreementToggle = Toggle(title: "Request Billing Agreement")
```
buzzamus, line 82: I included this with the fix as we didn't have an easy way to test a checkout flow with requestBillingAgreement enabled, so figured this would be good to keep

agedd (commented):
ty for fixing this! can we add a unit test to test this scenario (approving ahead of time) 🚀

jaxdesmarais (approved):
Thanks for adding docstrings so future us does not forget this fix. 🚀

buzzamus:
> ty for fixing this! can we add a unit test to test this scenario (approving ahead of time) 🚀

@agedd I actually started to, but both `shouldRequestBillingAgreement` and `extractToken` are private, so there isn't really a super good way to test this easily the way it currently is

agedd (approved):

PR #1814, opened 2026-06-03
Add CardFields to Demo

buzzamus:
### Summary of changes
- Adds CardFields to Demo app
- fixes issue where pay button was enabled before anything was entered into fields
- fixes issue where UnionPay had to pass Luhn check before it could be submitted (not all UnionPay cards pass luhn algorithm)

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Set up basic structure in Demo app for adding CardFields to options, create unit tests

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus


Demo/Application/Base/ContainmentViewController.swift
```diff
@@ -229,6 +229,8 @@ class ContainmentViewController: UIViewController {
             return ApplePayViewController(authorization: authorization)
         case "CardTokenizationViewController":
             return CardTokenizationViewController(authorization: authorization)
+        case "CardFieldsViewController":
+            return CardFieldsViewController(authorization: authorization)
```
jaxdesmarais: Curious what folks think about changing the Payment Buttons demo to just UI Components and having the buttons + card fields all displayed together vs 2 separate views?

Demo/Application/Features/CardFieldsViewController.swift
```diff
@@ -0,0 +1,71 @@
+import SwiftUI
+import UIKit
+import BraintreeCard
+import BraintreeUIComponents
+
+class CardFieldsViewController: BaseViewController {
```
jaxdesmarais: I think if we subclass `PaymentButtonBaseViewController` we will have access to authorization as a property automatically and can remove the 2 inits.

Demo/Application/Base/ContainmentViewController.swift
buzzamus: I think that makes sense to me

buzzamus: Would we treat this as a single form then and remove the button color toggles, or each element have its own section?

jaxdesmarais: Up to you what looks nicer. We can also change how the button toggles exist to something that makes more sense (not sure what that would be but open to ideas). I think my thought is make it look more like the figma where buttons + card fields exist together as they would if someone is replacing the drop in. But understand what your saying with where to then put the color options. 🤔

agedd: +1 i'm on board with that - we can create a new integration in the demo app something titled i.e. `Card Fields` which would include the card fields component at the top followed by either or both (depending on what we support) standard blue PP and Venmo buttons that would trigger the tokenize flow.

buzzamus: Ah sorry I didn't see this until now and did something different:

Simulator Screenshot - iPhone 17 - 2026-06-03 at 15 49 14

jaxdesmarais: This looks great! My only very minor nit is to maybe move the buttons to the top or below the pay button. Since pay only applies to card fields and the other flows will launch without that being enabled it keeps things grouped together nicely. But if you feel strongly about this layout, not a deal breaker for me.

Demo/Application/Features/CardFieldsViewController.swift
jaxdesmarais: ^ looks like we may have missed this one in the new VC

Demo/Application/Base/ContainmentViewController.swift
buzzamus: nope no strong opinions either way here! I'll move them below the pay button since I feel like most payment forms I see have the alternate payment methods below the card forms

Demo/Application/Features/CardFieldsViewController.swift
buzzamus: ope!

buzzamus: Updated!

jaxdesmarais (approved):

agedd (approved):
this looks so nice! 😎

PR #1816, opened 2026-06-04
fix: use PAT for release workflow

saralvasquez:
### Summary of changes
This changes the release workflow to use a personal access token for the git commands at the end of the release workflow. This should resolve the issue where the version change push to main fails due to the ruleset requirements we have in place. The PAT is coming from a secure new github account/user that only the SDK team has access to. This user has been included in the bypass list for the ruleset and I've added the token to the repo secrets

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [ ] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
no AI was used

- [ ] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

### Authors
- @saralvasquez


buzzamus (approved):
This is great thank you! So, the only issue was with needing a token or was there another change you had to make in the workflow as well?

agedd (approved):
😎

saralvasquez:
> This is great thank you! So, the only issue was with needing a token or was there another change you had to make in the workflow as well?

So we needed to link the git commands to a token that could be added to the bypass list. Before this it was using a token generated by github actions itself and for some reason we couldn't add that entity to the bypass list which is why nothing we did got around the ruleset. Along with this PR I also added the token to the repo secrets and added the user that generated the token to the bypass list

saralvasquez:
Do we have any idea what's up with that failing test?

PR #1817, opened 2026-06-08
Card fields analytics

buzzamus:
### Summary of changes
- Added 2 new analytic events for tracking CardFields when they render on screen and when the tokenize request is started. Verified the events triggered and carry same session ID throughout 

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [ ] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
N/A: I coded this like our ancient forefathers did: 

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus


Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
```diff
@@ -45,9 +48,15 @@ final class CardFieldsViewModel: ObservableObject {
...
     // MARK: - Internal Methods
 
+    func sendAnalyticsEvent(_ event: String) {
+        apiClient.sendAnalyticsEvent(event)
+    }
+
     func tokenize() {
         guard isFormValid else { return }
 
+        sendAnalyticsEvent(UIComponentsAnalytics.cardFieldsSelected)
```
agedd: nit: could we call`apiClient.sendAnalyticsEvent()` directly here instead of going through the helper 👀

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
```diff
@@ -15,6 +16,7 @@ final class CardFieldsViewModel: ObservableObject {
 
     // MARK: - Private Properties
 
+    var apiClient: BTAPIClient
```
agedd: does this need to be private 👀

UnitTests/BraintreeUIComponentsTests/CardFieldsViewModelTests.swift
```diff
@@ -0,0 +1,43 @@
...
+import BraintreeCard
+import BraintreeCore
+@testable import BraintreeTestShared
+@testable import BraintreeUIComponents
+
+@MainActor
+final class CardFieldsViewModelTests: XCTestCase {
+
+    private var mockAPIClient = MockAPIClient(authorization: "development_testing_tokenization_key_sandbox")
+    private var viewModel = CardFieldsViewModel(
```
agedd: nit: it looks like we're initializing the same on line 19 
```suggestion
    private var viewModel = CardFieldsViewModel
```

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
buzzamus: ah yep! Good call thanks

buzzamus: Actually sorry it is exposed for testing like the current pattern for the other clients. I moved and added docstring to follow existing pattern

UnitTests/BraintreeUIComponentsTests/CardFieldsViewModelTests.swift
buzzamus: ope yep fixed!

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
```diff
@@ -45,9 +50,15 @@ final class CardFieldsViewModel: ObservableObject {
 
     // MARK: - Internal Methods
 
+    func sendAnalyticsEvent(_ event: String) {
+        apiClient.sendAnalyticsEvent(event)
+    }
```
jaxdesmarais: nit: can we add a docstring to this? It took me a second to wrap my head around what was going on. Alternatively we could consider constructing the `apiClient` in `CardFields` and passing it as a param to the view model? That way we aren't constructing it twice which on fast connections could mean 2 config calls before it's in the cache. Then we can send both events in the same place. What do you think?

buzzamus: yep passing the client in from CardFields works! I briefly thought about going down that road early on, but I think it makes much more sense than having this weird phantom 2nd apiClient.

buzzamus: Updated

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
buzzamus: Your comment and Jax's below sort of intersected, so ended up changing this up

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
buzzamus: Updated again to remove the apiClient directly from the viewModel. I had to create a `submitAction` to house the analytics and tokenize call to use it in the onvalid closure check

jaxdesmarais (approved):

agedd (approved):
🚀

Issue #1818, opened 2026-06-10
Main-thread crash (objc_retain) in BTAPIClient.setupHTTPCredentials via init-fired fetchOrReturnRemoteConfiguration (v7)


BhosaleNikhil:
### General information

- SDK version: **7.5.0** (verified the crashing code path is **unchanged on `main` / 7.7.0**)
- Integration: Swift Package Manager
- Affected clients: `BTThreeDSecureClient`, `BTDataCollector`, `BTApplePayClient` (each created with `authorization: String`)
- Environment: production, real devices

### Describe the bug

Since upgrading from 6.32.0 to 7.x we see a recurring main-thread crash. It is an `objc_retain` on a freed/torn pointer, occurring in the `@MainActor` continuation of `fetchOrReturnRemoteConfiguration(_:)` that runs `setupHTTPCredentials` after the config load resumes.

We receive two crash groups with **byte-identical** stack traces:

```
Crashed: com.apple.main-thread
0  libobjc.A.dylib   objc_retain_x0 + 16
1  libobjc.A.dylib   objc_retain + 16
2  app               BTAPIClient.setupHTTPCredentials(_:) (BTHTTP.swift)
3  app               closure #1 in BTAPIClient.fetchOrReturnRemoteConfiguration(_:) (BTAPIClient.swift:86)
4  libswift_Concurrency  swift::runJobInEstablishedExecutorContext(swift::Job*) + 472
5  libswift_Concurrency  swift_job_runImpl(...) + 156
6  libdispatch       _dispatch_main_queue_drain.cold.6 + 612
...
15 app               main (AppDelegate.swift)
```

### Analysis

`BTAPIClient.init` fires `fetchOrReturnRemoteConfiguration { _, _ in }`, which spawns `Task { @MainActor in let config = try await configurationLoader.getConfig(); setupHTTPCredentials(config); ... }`. `getConfig()` is isolated to the app-wide `@ConfigurationActor`, while `setupHTTPCredentials` runs on `@MainActor`, and `deinit` (both `BTAPIClient` and `ConfigurationLoader`, calling `session.finishTasksAndInvalidate()`) runs on an arbitrary thread.

[...]

PR #1821, opened 2026-06-11
Fix Thread Crash Issue In APIClient

buzzamus:
### Summary of changes
- This is a fix for issue #1818
- `setupHTTPCredentials()` function marked with @MainActor to ensure all function runs happen on same thread since this is used by both `graphQLHTTP` and `payPalHTTP`
- `fetchOrReturnRemoteConfiguration()` function. marked with @MainActor before calling `setupHTTPCredentials()` to ensure entire flow is on same thread
- `post()` calls `fetchOrReturnRemoteConfiguration()` instead of `setupHTTPCredentials` directly to use the now thread-safe path
- cleaned up `deinit` and added `payPalHTTP?.session.finishTasksAndInvalidate()` to ensure URLSession is removed

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used to verify initial fix of `setupHTTPCredentials()` and `fetchOrReturnRemoteConfiguration()` were correct as well as review for any additional fixes. Claude pointed out last 2 bullet points ( verified with multiple agents)

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [x] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeCore/BTAPIClient.swift
+++ b/Sources/BraintreeCore/BTAPIClient.swift
@@ -23,7 +23,7 @@ import UIKit
     weak var analyticsService: AnalyticsSendable? = BTAnalyticsService.shared
 
     // MARK: - Initializers
-   
+
     /// :nodoc: This method is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
     /// Initialize a new API client.
     /// - Parameter authorization: Your tokenization key or client token.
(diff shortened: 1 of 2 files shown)


Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -50,13 +50,9 @@ import UIKit
     // MARK: - Deinit
 
     deinit {
-        if http != nil && http?.session != nil {
```
jaxdesmarais, line 53: Thanks for cleaning this redundancy up!

jaxdesmarais (approved):

agedd (approved):
ty for fixing this @buzzamus! 🚀

Sources/BraintreeCore/BTAPIClient.swift
```diff
@@ -241,14 +232,17 @@ import UIKit
     
     // MARK: Private Helpers
     
+    // @MainActor serializes writes to graphQLHTTP and payPalHTTP, preventing a data race
+    // where concurrent callers on different executors could each observe nil and race to set them.
+    @MainActor
```
agedd, line 237: nice docstrings

PR #1822, opened 2026-06-17
Fix CVV validation error state

buzzamus:
### Summary of changes
- Fixes an issue where if CVV is entered and card number changes to a card type where CVV length changes, the validation error state wouldn't trigger unless the field was specifically focused again.
- Fixes an issue where typing in CVV would be laggy in certain situations.

https://github.com/user-attachments/assets/38c4a1c3-39f0-4d71-878b-660ccd958121

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Track down cause of laggy CVV typing, write unit tests

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
+++ b/Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
@@ -13,6 +13,13 @@ final class CardFieldsViewModel: ObservableObject {
 
     @Published private(set) var isFormValid: Bool = false
 
+    // Debounced so rapid delete+retype coalesces — never delivers an intermediate invalid state.
+    lazy var formValidityPublisher: AnyPublisher<Bool, Never> = $isFormValid
+        .dropFirst()
+        .removeDuplicates()
+        .debounce(for: .milliseconds(50), scheduler: DispatchQueue.main)
+        .eraseToAnyPublisher()
+
     // MARK: - Private Properties
 
     private let cardClient: BTCardClient
@@ -40,6 +47,7 @@ final class CardFieldsViewModel: ObservableObject {
 
         Publishers.CombineLatest3(cardValid, expValid, cvvValid)
             .map { $0 && $1 && $2 }
+            .removeDuplicates()
             .assign(to: &$isFormValid)
     }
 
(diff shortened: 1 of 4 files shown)


Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
```diff
@@ -40,6 +47,7 @@ final class CardFieldsViewModel: ObservableObject {
 
         Publishers.CombineLatest3(cardValid, expValid, cvvValid)
             .map { $0 && $1 && $2 }
+            .removeDuplicates()
```
agedd, line 50: curious why we added this check here 👀

agedd (commented):
would you be able to attach a screen recording of this validation check/fix in effect

UnitTests/BraintreeUIComponentsTests/CVVFieldViewModelTests.swift
```diff
@@ -75,6 +75,39 @@ import XCTest
          XCTAssertEqual(viewModel.characters.map { $0.value }, ["4", "5", "6"])
      }
 
+
+     // MARK: - updateExpectedLength
```
agedd, line 79: nice unit tests! 🚀

Sources/BraintreeUIComponents/CardFields/CardFieldsViewModel.swift
buzzamus: This ignores duplicates of the same validity, so for example if the form is valid, more typing occurs, and the form is still valid it does not have to re-render the view, which avoids stuttering issues for the CVV masking

buzzamus:
> would you be able to attach a screen recording of this validation check/fix in effect

Updated description with video of the fix!

jaxdesmarais (approved):
Thanks for fixing this bug and adding tests for it as well!

agedd (approved):
ty for fixing this! 🚀

PR #1827, opened 2026-07-10
OPEN DEV: convert SEPA Direct Debit demo to SwiftUI view

buzzamus:
### Summary of changes
- Converted SEPA Direct Debit Demo to use a SwiftUI view as a sort of POC for slowly migrating all other views to SwiftUI. 

After diving into this view it has become clear this won't be a 1 part lift and shift, but instead will need to first have the SwiftUI views created and called in the view controller. A 2nd part will also need to be done where we eventually remove a lot of the shared logic housed in the Base view controller. 

https://github.com/user-attachments/assets/c763bab4-fe99-4826-b128-b3aaa536107f

### AI Usage

**Which AI Agent Was Used?**
- [x] Copilot 
- [ ] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used for discovering what was causing the Constraints errors and develop a fix

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/Demo/Application/Features/SEPADirectDebitView.swift
+++ b/Demo/Application/Features/SEPADirectDebitView.swift
@@ -1,11 +1,3 @@
-//
-//  SEPADirectDebitView.swift
-//  Demo
-//
-//  Created by Brent Busby on 7/10/26.
-//  Copyright © 2026 braintree. All rights reserved.
-//
-
 import SwiftUI
 import AuthenticationServices
 import BraintreeCore
@@ -30,7 +22,11 @@ struct SEPADirectDebitView: View {
 
     var body: some View {
         VStack {
-            Button(action: requestTokenization) {
+            Button {
+                Task {
+                    await requestTokenization()
+                }
+            } label: {
                 Text("SEPA Direct Debit")
                     .frame(maxWidth: .infinity)
             }
@@ -41,7 +37,7 @@ struct SEPADirectDebitView: View {
         }
     }
 
-    private func requestTokenization() {
+    private func requestTokenization() async {
         onProgress("Tapped SEPA Direct Debit")
 
         let billingAddress = BTPostalAddress(
@@ -62,15 +58,14 @@ struct SEPADirectDebitView: View {
             merchantAccountID: "EUR-sepa-direct-debit"
         )
 
-        sepaDirectDebitClient.tokenize(sepaDirectDebitRequest) { sepaDirectDebitNonce, error in
-            if let sepaDirectDebitNonce {
-                onComplete(sepaDirectDebitNonce)
-            } else if let error {
-                if error as? BTSEPADirectDebitError == .webFlowCanceled {
-                    onProgress("Canceled")
-                } else {
-                    onProgress(error.localizedDescription)
-                }
+        do {
+            let sepaDirectDebitNonce = try await sepaDirectDebitClient.tokenize(sepaDirectDebitRequest)
+            onComplete(sepaDirectDebitNonce)
+        } catch {
+            if error as? BTSEPADirectDebitError == .webFlowCanceled {
+                onProgress("Canceled")
+            } else {
+                onProgress(error.localizedDescription)
             }
         }
     }
--- a/Demo/Application/Features/SEPADirectDebitView.swift
+++ b/Demo/Application/Features/SEPADirectDebitView.swift
@@ -1,0 +1,85 @@
+//
+//  SEPADirectDebitView.swift
+//  Demo
+//
+//  Created by Brent Busby on 7/10/26.
+//  Copyright © 2026 braintree. All rights reserved.
+//
+
+import SwiftUI
+import AuthenticationServices
+import BraintreeCore
+import BraintreeSEPADirectDebit
+
+struct SEPADirectDebitView: View {
+
+    let sepaDirectDebitClient: BTSEPADirectDebitClient
+    let onProgress: (String?) -> Void
+    let onComplete: (BTPaymentMethodNonce?) -> Void
+
+    init(
+        authorization: String,
+        onProgress: @escaping (String?) -> Void = { _ in },
+        onComplete: @escaping (BTPaymentMethodNonce?) -> Void = { _ in }
+    ) {
+        self.sepaDirectDebitClient = BTSEPADirectDebitClient(authorization: authorization)
+        self.onProgress = onProgress
+        self.onComplete = onComplete
+    }
+
+
+    var body: some View {
+        VStack {
+            Button(action: requestTokenization) {
+                Text("SEPA Direct Debit")
+                    .frame(maxWidth: .infinity)
+            }
+            .buttonStyle(.borderedProminent)
+            .buttonBorderShape(.capsule)
+            .controlSize(.large)
+            .padding(.horizontal)
+        }
+    }
+
+    private func requestTokenization() {
+        onProgress("Tapped SEPA Direct Debit")
+
+        let billingAddress = BTPostalAddress(
+            streetAddress: "Kantstraße 70",
+            extendedAddress: "#170",
+            locality: "Freistaat Sachsen",
+            countryCodeAlpha2: "FR",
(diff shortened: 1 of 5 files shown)


buzzamus:
This PR can also be used as a conversation about anything related to SwiftUI when we decide to do this. Feel free to discuss any thoughts on patterns or any other opinions!

Demo/Application/Features/SEPADirectDebitView.swift
```diff
@@ -0,0 +1,85 @@
+//
+//  SEPADirectDebitView.swift
+//  Demo
+//
+//  Created by Brent Busby on 7/10/26.
+//  Copyright © 2026 braintree. All rights reserved.
+//
+
```
jaxdesmarais: nit: we generally remove these

Demo/Application/Features/SEPADirectDebitView.swift
```diff
@@ -0,0 +1,85 @@
...
+        let sepaDirectDebitRequest = BTSEPADirectDebitRequest(
+            accountHolderName: "John Doe",
+            iban: BTSEPADirectDebitTestHelper.generateValidSandboxIBAN(),
+            customerID: generateRandomCustomerID(),
+            billingAddress: billingAddress,
+            mandateType: .oneOff,
+            merchantAccountID: "EUR-sepa-direct-debit"
+        )
+
+        sepaDirectDebitClient.tokenize(sepaDirectDebitRequest) { sepaDirectDebitNonce, error in
```
jaxdesmarais: As part of this should we also consider moving these to async/await?

Demo/Application/Features/SEPADirectDebitView.swift
buzzamus: whoops! I'll remove

Demo/Application/Features/SEPADirectDebitView.swift
buzzamus: Yep! I actually meant to after copy/pasting all the existing logic to get it working, but forgot. Good call!

buzzamus: updated!

Demo/Application/Features/SEPADirectDebitView.swift
buzzamus: Removed!

Demo/Application/Features/SEPADirectDebitView.swift
```diff
@@ -0,0 +1,80 @@
...
+        authorization: String,
+        onProgress: @escaping (String?) -> Void = { _ in },
+        onComplete: @escaping (BTPaymentMethodNonce?) -> Void = { _ in }
+    ) {
+        self.sepaDirectDebitClient = BTSEPADirectDebitClient(authorization: authorization)
+        self.onProgress = onProgress
+        self.onComplete = onComplete
+    }
+
+
```
jaxdesmarais, line 22: nit to remove double space
```suggestion
```

jaxdesmarais (approved):

Demo/Application/Features/SEPADirectDebitViewController.swift
```diff
@@ -1,59 +1,40 @@
...
-            iban: BTSEPADirectDebitTestHelper.generateValidSandboxIBAN(),
-            customerID: generateRandomCustomerID(),
-            billingAddress: billingAddress,
-            mandateType: .oneOff,
-            merchantAccountID: "EUR-sepa-direct-debit"
-        )
+        let hostingController = UIHostingController(rootView: demoView)
+        addChild(hostingController)
+        hostingController.view.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(hostingController.view)
```
agedd: nit: maybe we can extract this logic into a shared helper function i.e. `UIViewController` extension like `embed(_ view: some UIView)` since we'll be repeating this for every future view controller conversion

Demo/Application/Features/SEPADirectDebitViewController.swift
```diff
@@ -1,59 +1,40 @@
...
-            }
-        }
+        NSLayoutConstraint.activate([
+            hostingController.view.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
+            hostingController.view.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor),
+            hostingController.view.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor),
+            hostingController.view.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor)
+        ])
+
+        hostingController.didMove(toParent: self)
```
agedd: nit: can also pull this into the shared helper function as well

Demo/Application/Features/SEPADirectDebitViewController.swift
```diff
@@ -1,59 +1,40 @@
...
     }
 
-    private func generateRandomCustomerID() -> String {
-        String(UUID().uuidString.replacingOccurrences(of: "-", with: "").prefix(20))
+    // This is to supress Constraint warnings when the payment button is not overriden. The actual Payment Button is within the SwiftUI view
```
agedd: small nit: 
```suggestion
    // This is to suppress Constraint warnings when the payment button is not overriden. The actual Payment Button is within the SwiftUI view
```

agedd: might be worth adding a  `// TODO: ` here so this doesn't get lost once several views are created using this placeholder

Demo/Application/Features/SEPADirectDebitView.swift
```diff
@@ -0,0 +1,80 @@
...
+import BraintreeSEPADirectDebit
+
+struct SEPADirectDebitView: View {
+
+    let sepaDirectDebitClient: BTSEPADirectDebitClient
+    let onProgress: (String?) -> Void
+    let onComplete: (BTPaymentMethodNonce?) -> Void
+
+    init(
+        authorization: String,
```
agedd: nit: looks like the view builds its own dependency instead of receiving it. thoughts on modifying this to take the `BTSEPADirectDebitClient` as an injected dependency that way it'd be easier to unit test the view in isolation? 

```
init(
        client: BTSEPADirectDebitClient,
        onProgress: @escaping (String?) -> Void = { _ in },
        onComplete: @escaping (BTPaymentMethodNonce?) -> Void = { _ in }
    ) {
        self.sepaDirectDebitClient = client
        self.onProgress = onProgress
        self.onComplete = onComplete
    }
```

Demo/Application/Features/SEPADirectDebitViewController.swift
buzzamus: yep! Good idea

Demo/Application/Features/SEPADirectDebitViewController.swift
buzzamus: yep i'll do that!

Demo/Application/Features/SEPADirectDebitViewController.swift
buzzamus: Updated

Demo/Application/Features/SEPADirectDebitViewController.swift
buzzamus: Updated

Demo/Application/Features/SEPADirectDebitViewController.swift
buzzamus: Updated

Demo/Application/Features/SEPADirectDebitView.swift
buzzamus: Updated

agedd (approved):
this looks great! ty for doing this 🚀

PR #1830, opened 2026-07-20
Add PA M3 Base Interface and Types

agedd:
### Summary of changes
- Add basic types and interface related to Payment Actions M3

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
AI was used to cross-reference against the integration guide and the ADRs published by the other teams

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

- @agedd


jaxdesmarais:
I think this should be going into a feature branch, right?

Sources/BraintreeCore/PaymentActions/BTPaymentActionSelectedPaymentMethod.swift
```diff
@@ -0,0 +1,28 @@
...
+@_documentation(visibility: private)
+public enum BTPaymentActionSelectedPaymentMethodDetails {
+    
+    // swiftlint:disable:next enum_case_associated_values_count
+    case creditCard(
+        last4: String,
+        bin: String,
+        expirationMonth: String,
+        expirationYear: String,
+        cardholderName: String,
```
buzzamus: Is cardholder name required for payment actions?

Sources/BraintreeCore/PaymentActions/BTPaymentActionStatus.swift
```diff
@@ -0,0 +1,28 @@
...
+    case readyForConfirmation
+    case succeeded
+    case requiresCapture
+    case canceled
+    case expired
+    case requiresCustomerAction
+    case processing
+    case unknown
+    
+    public init(rawValue: String) {
```
jaxdesmarais: Is there a reason we aren't making this a string enum? The raw value logic reads a bit different than how we do this in the rest of the SDK. Additionally I think we will need to conform to `Int` and have a variable with the raw value like we do elsewhere for Obj-C support. What do you think?

Sources/BraintreeCore/PaymentActions/BTPaymentActionsGraphQLBody.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+/// Marker protocol refining `BTGraphQLEncodablebody` for Payment Actions mutations.
```
jaxdesmarais:
```suggestion
/// Marker protocol refining `BTGraphQLEncodableBody` for Payment Actions mutations.
```

Sources/BraintreeCore/PaymentActions/BTPaymentActionSelectedPaymentMethod.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+/// Reflects the `selectedPaymentMethod` field on a `PaymentAction`.
+@_documentation(visibility: private)
+public struct BTPaymentActionSelectedPaymentMethod {
```
jaxdesmarais: Are the types here correct for Obj-C interoperability?

Sources/BraintreeCore/PaymentActions/BTPaymentActionsAnalytics.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+// TODO: These keys are currently placeholder values that must be changed when we get the official values.
+enum BTPaymentActionsAnalytics {
```
jaxdesmarais: I think we will likely need the analytics to live in the feature modules so we know what feature is triggering the event?

Sources/BraintreeCore/PaymentActions/BTNextAction.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+/// Represents the different shapes `nextAction` can take when returned from a Payment Action GraphQL mutation
+/// relating to a 3DS flow.
+@_documentation(visibility: private)
+enum BTNextAction {
```
jaxdesmarais: Should we omit 3DS work for now and save that for when it's ready?

Sources/BraintreeCore/PaymentActions/BTPaymentActionSelectedPaymentMethod.swift
agedd: ah, I believe so — based on the `SetPaymentActionPaymentMethod` mutation ADR. that said, it's subject to change, i've added TODO in c13703f2808ed51a70d097ba3493eadb12983de0 to mark this

Sources/BraintreeCore/PaymentActions/BTPaymentActionSelectedPaymentMethod.swift
```diff
@@ -0,0 +1,28 @@
...
+    public let details: BTPaymentActionSelectedPaymentMethodDetails
+}
+
+/// The different shapes `selectedPaymentMethod.details` can take. `.creditCard` is the only
+/// case exercised in M3; other payment methods add cases as they're scoped.
+@_documentation(visibility: private)
+public enum BTPaymentActionSelectedPaymentMethodDetails {
+    
+    // swiftlint:disable:next enum_case_associated_values_count
+    case creditCard(
```
mzlangreder: This is because we're using different casing here so it can be encoded to GraphQL?

Sources/BraintreeCore/PaymentActions/BTNextAction.swift
agedd: ah yep good call, addressed this in c13703f2808ed51a70d097ba3493eadb12983de0

Sources/BraintreeCard/BTCardPaymentActionsAnalytics.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+// TODO: These keys are currently placeholder values that must be changed when we get the official values.
+enum BTCardPaymentActionsAnalytics {
```
mzlangreder: Are these going to be specific to Card?

Sources/BraintreeCard/BTCardPaymentActionResult.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+public struct BTCardPaymentActionResult {
```
mzlangreder: Are you planning on building out the DTO's for payment actions in a future PR? This is the only result type I see on this PR so far.

When we build up our `PaymentActionsService`, what type is it going to map the returned GraphQL response to?

Sources/BraintreeCard/CardSetPaymentMethodGraphQLBody.swift
```diff
@@ -0,0 +1,104 @@
...
+        // swiftlint:disable indentation_width
+        """
+        mutation SetPaymentActionPaymentMethod($input: SetPaymentActionPaymentMethodInput!) {
+          setPaymentActionPaymentMethod(input: $input) {
+            paymentAction {
+              id
+              status
+              nextAction {
+                type
+                songbirdUrl
```
mzlangreder: There was a bit of [discussion](https://github.com/braintree/braintree_android/pull/1654#discussion_r3624863157) on whether we should keep parts of the query we wont end up using. I ended up removing `songbirdUrl` since we wont make use of it due to how we work with Cardinal.

Sources/BraintreeCard/CardSetPaymentMethodGraphQLBody.swift
```diff
@@ -0,0 +1,104 @@
...
+    
+    static func setPaymentActionPaymentMethodMutation() -> String {
+        // swiftlint:disable indentation_width
+        """
+        mutation SetPaymentActionPaymentMethod($input: SetPaymentActionPaymentMethodInput!) {
+          setPaymentActionPaymentMethod(input: $input) {
+            paymentAction {
+              id
+              status
+              nextAction {
```
mzlangreder: With each payment method owning the query, we should probably limit the entire selection set to only what Credit Card is going to need. At least for M3 we wont need to have nextAction returned or consumed by the Result type you have defined so far.  Should it be dropped here? What about `selectedPaymentMethod`?

Sources/BraintreeCard/CardSetPaymentMethodGraphQLBody.swift
mzlangreder, line 1: All other files/types are prefixed wit `BT`, should this be too?

Sources/BraintreeCard/BTCardPaymentActionsAnalytics.swift
agedd: yep typically we house analytics per module, so i've moved these into the `BTCard` module and as per @jaxdesmarais's suggestion. we'll still need to align with web on the actual analytic string constants (have a `TODO` to indicate this)

Sources/BraintreeCard/BTCardPaymentActionResult.swift
agedd: ah good catch and call - i've added the `BTPaymentActionResult` now

Sources/BraintreeCard/CardSetPaymentMethodGraphQLBody.swift
agedd: checked against the precedent in our SDK and it seems we use un-prefixed structs i.e. `CreditCardGraphQLBody` same as this one, so decided to leave this as-is.

Sources/BraintreeCard/CardSetPaymentMethodGraphQLBody.swift
agedd: agreed, will trim the selection set to just id and status for m3, it matches what the result type actually consumes rght now and `selectedPaymentMethod` is still an open api-review question, so imo can also be taken out as of now.

Sources/BraintreeCard/CardSetPaymentMethodGraphQLBody.swift
agedd: agreed, removed `songbirdURL` and m4 specific types in the query selection set and result types whenever appropriate

Sources/BraintreeCard/BTCardPaymentActionsAnalytics.swift
mzlangreder: Ok, I'll make the same change over in my PR.

mzlangreder (approved):

Sources/BraintreeCard/BTCardPaymentActionResult.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+@objcMembers public class BTCardPaymentActionResult: NSObject {
```
jaxdesmarais, line 10: we may want a todo for docstrings here

Sources/BraintreeCard/BTCardPaymentActionResult.swift
```diff
@@ -0,0 +1,16 @@
...
+    
+    public let id: String
+    public let status: BTPaymentActionStatus
+    
+    public init(id: String, status: BTPaymentActionStatus) {
```
jaxdesmarais: I am not sure the init will need to be public. Similar to a nonce I assume we will get a BTJSON object back from the API that we then parse and set the properties with, correct?

Sources/BraintreeCard/BTCardPaymentActionsAnalytics.swift
jaxdesmarais: I think we can house these in the existing analytics enum with a docstring for payment actions? The we can do this to match our naming conventions:
```swift
    static let paymentActionsSetPaymentMethodStarted = "card:payment-actions:set-payment-method:started"
```
Not even sure we need `readyForConfirmation` at least not yet since that is not a conversion related event. Curious to hear y'all's thoughts though.

Sources/BraintreeCore/PaymentActions/BTPaymentActionResult.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+struct BTPaymentActionResult {
```
jaxdesmarais, line 5: Do we still need this one?

Sources/BraintreeCore/PaymentActions/BTPaymentActionsGraphQLBody.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+/// Marker protocol refining `BTGraphQLEncodableBody` for Payment Actions mutations.
+/// Each payment method's request body (i.e. `CardSetPaymentMethodGraphQLBody`) conforms to this.
+public protocol BTPaymentActionsGraphQLBody: BTGraphQLEncodableBody { }
```
jaxdesmarais: Do we expect the body to require a different structure or can we just use `BTGraphQLEncodableBody` in `CardSetPaymentMethodGraphQLBody`? Since both are BT GraphQL I expect the structure will be the same?

Sources/BraintreeCore/PaymentActions/BTPaymentActionStatus.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+/// The lifecycle status of a Payment Action. 
+@objc public enum BTPaymentActionStatus: Int {
```
jaxdesmarais, line 4: Can we add a todo for docstrings for this?

Sources/BraintreeCore/PaymentActions/BTPaymentActionStatus.swift
```diff
@@ -0,0 +1,28 @@
...
+        case "READY_FOR_CONFIRMATION": return .readyForConfirmation
+        case "SUCCEEDED": return .succeeded
+        case "REQUIRES_CAPTURE": return .requiresCapture
+        case "CANCELED": return .canceled
+        case "EXPIRED": return .expired
+        case "REQUIRES_CUSTOMER_ACTION": return .requiresCustomerAction
+        case "PROCESSING": return .processing
+        default: return .unknown
+        }
+    }
```
jaxdesmarais, line 44: Can we use the structure we use for the rest of our enum cases or remove this if we don't know how we will use it yet?

Sources/BraintreeCard/BTCardPaymentActionsAnalytics.swift
mzlangreder: I think I find the explicit grouping appealing.

Sources/BraintreeCore/PaymentActions/BTPaymentActionStatus.swift
```diff
@@ -0,0 +1,28 @@
...
+    case requiresPaymentMethod
+    case readyForConfirmation
+    case succeeded
+    case requiresCapture
+    case canceled
+    case expired
+    case requiresCustomerAction
+    case processing
+    case unknown
+
```
mzlangreder, line 20: Should we drop the cases here that don't belong to M3 as well? I think that would make sense given our decision to only build what's explicitly required.

Sources/BraintreeCore/PaymentActions/BTPaymentActionSelectedPaymentMethod.swift
agedd: removed these types for now as they're not relevant to m3

Sources/BraintreeCore/PaymentActions/BTPaymentActionResult.swift
agedd: yep, this is the internal result type that each payment-method-specific client will use to construct its own public-facing result type.

Sources/BraintreeCore/PaymentActions/BTPaymentActionStatus.swift
agedd: added doc-strings directly for each case

Sources/BraintreeCore/PaymentActions/BTPaymentActionStatus.swift
agedd: yep mirrored this similar to how we organize and model `BTSEPADirectDebitMandateType`

Sources/BraintreeCore/PaymentActions/BTPaymentActionStatus.swift
agedd: ah yes, addressed this in 77ec964805381229a2d037ea7bff109c8787c8b3

Sources/BraintreeCore/PaymentActions/BTPaymentActionsGraphQLBody.swift
agedd: ah yes - structurally, they're identical but leaned on proceeding with the marker protocol to constrain the `setPaymentMethod` generic to only accept payment-action specific bodies

Sources/BraintreeCard/BTCardPaymentActionsAnalytics.swift
jaxdesmarais: 🤔 my only concern is that it goes against the rest of the SDK where we use something like `MARK` to indicate the start of 1 set of groupings. I worry that creating a new set of files for a new flow now means there are multiple sources of truth for where analytics live for a given module. I am okay deviating from the pattern just want to think through the implications for other teams/future us.

Sources/BraintreeCore/PaymentActions/BTPaymentActionResult.swift
jaxdesmarais: It seems like we aren't using it anywhere, should it be similar to `BTPaymentMethodNonce` where all nonce types conform to it and we extend it from there while adding payment method specifics on the payment method specific result? If so we should implement that here and implement the inits in a future PR.

Sources/BraintreeCore/PaymentActions/BTPaymentActionsGraphQLBody.swift
jaxdesmarais: I think if we aren't using it, we can omit for now and add when we get to that implementation. What do you think?

Sources/BraintreeCore/PaymentActions/BTPaymentActionStatus.swift
```diff
@@ -0,0 +1,45 @@
...
+    /// The Payment Action has completed successfully. The underlying transaction has been submitted for settlement (or settled) and any selected payment details have been cleared.
+    case succeeded
+    
+    /// The payment method has been authorized but funds have not yet been captured.
+    case requiresCapture
+    
+    /// The status returned by the server could not be mapped to a known case.
+    case unknown
+    
+    var descprition: String {
```
jaxdesmarais: Little typo:
```suggestion
    var description: String {
```

Sources/BraintreeCore/PaymentActions/BTPaymentActionsGraphQLBody.swift
agedd: sure that works!

Sources/BraintreeCore/PaymentActions/BTPaymentActionResult.swift
agedd: ah yes this type won't be until we build out the `setPaymentMethod` in a future PR - this way the client only exposes what it actually needs rather than being locked to the internal shape.

Sources/BraintreeCard/BTCardPaymentActionsAnalytics.swift
agedd: ahh i see that's a great callout, agreed - it makes more sense to house this under the existing `BTCardAnalytics` enum, addressed in 4f34f7d6d92f8dccc4a6b6de08821a93895fed15

Sources/BraintreeCore/PaymentActions/BTPaymentActionResult.swift
jaxdesmarais: Ah - cool cool, maybe a todo will help or if it's documented somewhere in jira that's cool too.

jaxdesmarais (approved):

PR #1832, opened 2026-07-21
Feature/ paypal campaign create order

Juan219:
### Summary of changes
- PR 1 of 2 for PayPal Campaign IDs.
- Base branch: `braintree:paypal-campaigns-feature`.
- Dependency: none.
- Adds shared `BTPayPalCampaign` model in `BraintreeCore`.
- Adds public `campaigns` support to `BTPayPalCheckoutRequest`.
- Sends valid campaign IDs in PayPal Checkout create order requests as `paypal_campaigns`.
- Omits `paypal_campaigns` when the list is empty or contains only empty IDs.
- Preserves campaign ID order when serializing the request.
- Adds unit coverage for empty, single, multiple, and mixed campaign ID inputs.
- Adds Demo App support for entering comma-separated campaign IDs in the PayPal Checkout flow.
- Adds scrollable Demo layout support so the PayPal Checkout form content remains accessible.
- Test steps:
  - Run `xcodebuild test -project Braintree.xcodeproj -scheme BraintreePayPal -destination 'platform=iOS Simulator,name=iPhone 16 Pro Max,OS=18.5' -only-testing:BraintreePayPalTests/BTPayPalCheckoutRequest_Tests -only-testing:BraintreePayPalTests/BTPayPalClient_Tests`.
  - Build/run the Demo App and verify PayPal Checkout requests include `paypal_campaigns` when campaign IDs are entered.

Video Evidence:

https://github.com/user-attachments/assets/0464aabd-2137-4717-b1c0-9bd52df88032
  
  CAL Logs for Create Order
  
Screenshot 2026-07-22 at 12 32 46 PM

  
```
  [BT SDK Request] POST https://api.sandbox.braintreegateway.com:443/merchants/dcpspy2brwdjr3qn/client_api/v1/paypal_hermes/create_payment_resource
[BT SDK Request Headers] ["Accept": "application/json", "User-Agent": "Braintree/iOS/7.8.0", "Content-Type": "application/json; charset=utf-8", "Accept-Language": "en-US"]
[BT SDK Request Summary] correlation_id=<not present> paypal_campaigns=Can-123,cam-ii99
[BT SDK Request Body] {
  "_meta" : {
    "integration" : "custom",
    "platform" : "iOS",
    "sessionId" : "47A886FDA98F4A259F568C29014819EF",
    "source" : "unknown",
    "version" : "7.8.0"
  },
  "amount" : "5.00",
  "authorization_fingerprint" : "eyJraWQiOiIyMDE4MDQyNjE2LXNhbmRib3giLCJpc3MiOiJodHRwczovL2FwaS5zYW5kYm94LmJyYWludHJlZWdhdGV3YXkuY29tIiwiYWxnIjoiRVMyNTYifQ.eyJleHAiOjE3ODQ4MjY4MzUsImp0aSI6IjNmYTgyMThhLWNhNDktNDc2OC1iMzRiLWE0OWVhNjg0MmZjYiIsInN1YiI6ImRjcHNweTJicndkanIzcW4iLCJpc3MiOiJodHRwczovL2FwaS5zYW5kYm94LmJyYWludHJlZWdhdGV3YXkuY29tIiwibWVyY2hhbnQiOnsicHVibGljX2lkIjoiZGNwc3B5MmJyd2RqcjNxbiIsInZlcmlmeV9jYXJkX2J5X2RlZmF1bHQiOnRydWUsInZlcmlmeV93YWxsZXRfYnlfZGVmYXVsdCI6dHJ1ZX0sInJpZ2h0cyI6WyJtYW5hZ2VfdmF1bHQiXSwic2NvcGUiOlsiQnJhaW50cmVlOlZhdWx0IiwiQnJhaW50cmVlOkNsaWVudFNESyIsIkJyYWludHJlZTpBWE8iLCJCcmFpbnRyZWU6UGF5bWVudHNSZWFkeSJdLCJvcHRpb25zIjp7ImN1c3RvbWVyX2lkIjoiODI2OTdDODYtM0IyRS00MEE5LUE0QjgtQTRGOERFNjM5NURFIiwicGF5cGFsX2NsaWVudF9pZCI6IkFiNFRqckhJcjZqMWhJVWNUYVpaQmJmTGVBWjlxX1FBUmpQYUVVNXZ6TTFQUmJfZEUzR04zWk9LSExvcl9lSWdaRWszV08zSzFXVDE5VFg2In19.TaCzpJ9owSHvUpWqAE9Y7V_TYRv6cs8RSlq7RNzNlRk-nC1oIHvmDzcZ-vlJWNtcrG-rGXL0GZEVanIGiBrKIw?customer_id=",
  "cancel_url" : "sdk.ios.braintree:\/\/onetouch\/v1\/cancel",
  "contact_preference" : "UPDATE_CONTACT_INFO",
  "currency_iso_code" : "USD",
  "experience_profile" : {
    "address_override" : false,
    "brand_name" : "Acme Widgets, Ltd. (Sandbox)",
    "no_shipping" : true
  },
  "intent" : "authorize",
  "line_items" : [
    {
      "image_url" : "https:\/\/www.example.com\/example.jpg",
      "kind" : "debit",
      "name" : "item one 1234567",
      "quantity" : "1",
      "unit_amount" : "5.00",
      "upc_code" : "123456789",
      "upc_type" : "UPC-A"
    }
  ],
  "offer_pay_later" : false,
  "offer_paypal_credit" : false,
  "paypal_campaigns" : [
    {
      "id" : "Can-123"
    },
    {
      "id" : "cam-ii99"
    }
  ],
  "return_url" : "sdk.ios.braintree:\/\/onetouch\/v1\/success"
}

```

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [ ] Claude 
- [x] Other: Codex

**How was AI used?**
Codex was used to inspect the SDK request flow, implement campaign ID support for PayPal Checkout create order requests, add focused unit tests, update the Demo App, and help split the changes into reviewable PRs.

- [ ] less than 30% 
- [ ] 30 - 60% 
- [x] 60 - 100% 

### Checklist
- [x] Added a changelog entry

- @Juan219

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* BraintreePayPal
+  * Add `payPalCampaigns` to `BTPayPalCheckoutRequest` and pass them as `paypal_campaigns` in PayPal checkout requests.
 * BraintreeUIComponents
   * Fix minimum target version to iOS 16.0 to match all other modules
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 * BraintreePayPal
-  * Add `campaigns` to `BTPayPalCheckoutRequest` and pass them in PayPal checkout requests so eligible campaign context can be persisted during order creation and surfaced in the PayPal checkout experience.
+  * Add `campaigns` to `BTPayPalCheckoutRequest` so eligible campaign context can be persisted during order creation and surfaced in the PayPal checkout experience.
 * BraintreeUIComponents
   * Fix minimum target version to iOS 16.0 to match all other modules
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 * BraintreePayPal
-  * Add `campaigns` to `BTPayPalCheckoutRequest` and pass them as `paypal_campaigns` in PayPal checkout requests so eligible campaign context can be persisted during order creation and surfaced in the PayPal checkout experience.
+  * Add `campaigns` to `BTPayPalCheckoutRequest` and pass them in PayPal checkout requests so eligible campaign context can be persisted during order creation and surfaced in the PayPal checkout experience.
 * BraintreeUIComponents
   * Fix minimum target version to iOS 16.0 to match all other modules
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 * BraintreePayPal
-  * Add `payPalCampaigns` to `BTPayPalCheckoutRequest` and pass them as `paypal_campaigns` in PayPal checkout requests.
+  * Add `campaigns` to `BTPayPalCheckoutRequest` and pass them as `paypal_campaigns` in PayPal checkout requests.
 * BraintreeUIComponents
   * Fix minimum target version to iOS 16.0 to match all other modules
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## unreleased
 * BraintreePayPal
-  * Add `campaigns` to `BTPayPalCheckoutRequest` and pass them as `paypal_campaigns` in PayPal checkout requests.
+  * Add `campaigns` to `BTPayPalCheckoutRequest` and pass them as `paypal_campaigns` in PayPal checkout requests so eligible campaign context can be persisted during order creation and surfaced in the PayPal checkout experience.
 * BraintreeUIComponents
   * Fix minimum target version to iOS 16.0 to match all other modules
 
--- a/Demo/Application/Base/PaymentButtonBaseViewController.swift
+++ b/Demo/Application/Base/PaymentButtonBaseViewController.swift
@@ -6,7 +6,10 @@ class PaymentButtonBaseViewController: BaseViewController {
     let authorization: String
 
     var heightConstraint: CGFloat?
+    var usesIntrinsicContentHeight = false
 
+    private let scrollView = UIScrollView()
+    private let contentView = UIView()
     private var paymentButton = UIView()
 
     override init(authorization: String) {
@@ -25,14 +28,61 @@ class PaymentButtonBaseViewController: BaseViewController {
         view.backgroundColor = .systemBackground
 
         paymentButton = createPaymentButton()
-        view.addSubview(paymentButton)
+        paymentButton.translatesAutoresizingMaskIntoConstraints = false
 
-        NSLayoutConstraint.activate([
-            paymentButton.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 20),
-            paymentButton.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -20),
-            paymentButton.centerYAnchor.constraint(equalTo: view.centerYAnchor),
-            paymentButton.heightAnchor.constraint(equalToConstant: heightConstraint ?? 100)
-        ])
+        scrollView.keyboardDismissMode = .interactive
+        scrollView.translatesAutoresizingMaskIntoConstraints = false
+        contentView.translatesAutoresizingMaskIntoConstraints = false
+
+        view.addSubview(scrollView)
+        scrollView.addSubview(contentView)
+        contentView.addSubview(paymentButton)
+
+        let contentViewEqualHeightConstraint = contentView.heightAnchor.constraint(
+            equalTo: scrollView.frameLayoutGuide.heightAnchor
+        )
+        contentViewEqualHeightConstraint.priority = .defaultLow
+
+        let paymentButtonTopConstraint = paymentButton.topAnchor.constraint(
+            equalTo: contentView.topAnchor,
+            constant: 20
+        )
+        paymentButtonTopConstraint.priority = .defaultHigh
+
+        let paymentButtonBottomConstraint = paymentButton.bottomAnchor.constraint(
+            equalTo: contentView.bottomAnchor,
+            constant: -20
+        )
+        paymentButtonBottomConstraint.priority = .defaultHigh
+
+        var constraints = [
+            scrollView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            scrollView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            scrollView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
+            scrollView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor),
+
+            contentView.leadingAnchor.constraint(equalTo: scrollView.contentLayoutGuide.leadingAnchor),
+            contentView.trailingAnchor.constraint(equalTo: scrollView.contentLayoutGuide.trailingAnchor),
+            contentView.topAnchor.constraint(equalTo: scrollView.contentLayoutGuide.topAnchor),
+            contentView.bottomAnchor.constraint(equalTo: scrollView.contentLayoutGuide.bottomAnchor),
+            contentView.widthAnchor.constraint(equalTo: scrollView.frameLayoutGuide.widthAnchor),
+            contentView.heightAnchor.constraint(greaterThanOrEqualTo: scrollView.frameLayoutGuide.heightAnchor),
+            contentViewEqualHeightConstraint,
+
+            paymentButton.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: 20),
+            paymentButton.trailingAnchor.constraint(equalTo: contentView.trailingAnchor, constant: -20),
+            paymentButton.topAnchor.constraint(greaterThanOrEqualTo: contentView.topAnchor, constant: 20),
+            paymentButton.bottomAnchor.constraint(lessThanOrEqualTo: contentView.bottomAnchor, constant: -20),
+            paymentButtonTopConstraint,
+            paymentButtonBottomConstraint,
+            paymentButton.centerYAnchor.constraint(equalTo: contentView.centerYAnchor)
+        ]
+
+        if !usesIntrinsicContentHeight {
+            constraints.append(paymentButton.heightAnchor.constraint(equalToConstant: heightConstraint ?? 100))
+        }
+
+        NSLayoutConstraint.activate(constraints)
     }
 
     /// A factory method that subclasses must implement to return a payment button view.
--- a/Demo/Application/Base/PaymentButtonBaseViewController.swift
+++ b/Demo/Application/Base/PaymentButtonBaseViewController.swift
@@ -43,21 +43,23 @@ class PaymentButtonBaseViewController: BaseViewController {
         )
         contentViewEqualHeightConstraint.priority = .defaultLow
 
+        let contentInset: CGFloat = usesIntrinsicContentHeight ? 10 : 20
+
         let paymentButtonTopConstraint = paymentButton.topAnchor.constraint(
             equalTo: contentView.topAnchor,
-            constant: 20
+            constant: contentInset
         )
         paymentButtonTopConstraint.priority = .defaultHigh
 
         let paymentButtonBottomConstraint = paymentButton.bottomAnchor.constraint(
             equalTo: contentView.bottomAnchor,
-            constant: -20
+            constant: -contentInset
         )
         paymentButtonBottomConstraint.priority = .defaultHigh
 
         var constraints = [
-            scrollView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
-            scrollView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            scrollView.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor),
+            scrollView.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor),
             scrollView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
             scrollView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor),
 
@@ -71,15 +73,15 @@ class PaymentButtonBaseViewController: BaseViewController {
 
             paymentButton.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: 20),
             paymentButton.trailingAnchor.constraint(equalTo: contentView.trailingAnchor, constant: -20),
-            paymentButton.topAnchor.constraint(greaterThanOrEqualTo: contentView.topAnchor, constant: 20),
-            paymentButton.bottomAnchor.constraint(lessThanOrEqualTo: contentView.bottomAnchor, constant: -20),
+            paymentButton.topAnchor.constraint(greaterThanOrEqualTo: contentView.topAnchor, constant: contentInset),
+            paymentButton.bottomAnchor.constraint(lessThanOrEqualTo: contentView.bottomAnchor, constant: -contentInset),
             paymentButtonTopConstraint,
-            paymentButtonBottomConstraint,
-            paymentButton.centerYAnchor.constraint(equalTo: contentView.centerYAnchor)
+            paymentButtonBottomConstraint
         ]
 
         if !usesIntrinsicContentHeight {
             constraints.append(paymentButton.heightAnchor.constraint(equalToConstant: heightConstraint ?? 100))
+            constraints.append(paymentButton.centerYAnchor.constraint(equalTo: contentView.centerYAnchor))
         }
 
         NSLayoutConstraint.activate(constraints)
--- a/Demo/Application/Base/PaymentButtonBaseViewController.swift
+++ b/Demo/Application/Base/PaymentButtonBaseViewController.swift
@@ -34,9 +34,9 @@ class PaymentButtonBaseViewController: BaseViewController {
         scrollView.translatesAutoresizingMaskIntoConstraints = false
         contentView.translatesAutoresizingMaskIntoConstraints = false
 
-        view.addSubview(scrollView)
-        scrollView.addSubview(contentView)
         contentView.addSubview(paymentButton)
+        scrollView.addSubview(contentView)
+        view.addSubview(scrollView)
 
         let contentViewEqualHeightConstraint = contentView.heightAnchor.constraint(
             equalTo: scrollView.frameLayoutGuide.heightAnchor
--- a/Demo/Application/Features/PayPalWebCheckoutViewController.swift
+++ b/Demo/Application/Features/PayPalWebCheckoutViewController.swift
@@ -74,6 +74,39 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         stackView.distribution = .fillEqually
         return stackView
     }()
+
+    lazy var payPalCampaignsLabel: UILabel = {
+        let label = UILabel()
+        label.text = "PayPal Campaigns:"
+        return label
+    }()
+
+    lazy var payPalCampaignsTextField: UITextField = {
+        let textField = UITextField()
+        textField.placeholder = "campaign-123-id"
+        textField.textAlignment = .right
+        textField.backgroundColor = .systemBackground
+        return textField
+    }()
+
+    lazy var payPalCampaignsStackView: UIStackView = {
+        let stackView = UIStackView(arrangedSubviews: [payPalCampaignsLabel, payPalCampaignsTextField])
+        stackView.distribution = .fillEqually
+        return stackView
+    }()
+
+    lazy var customerInfoStackView: UIStackView = {
+        let stackView = UIStackView(arrangedSubviews: [
+            emailStackView,
+            countryCodeStackView,
+            nationalNumberStackView,
+            payPalCampaignsStackView
+        ])
+        stackView.axis = .vertical
+        stackView.distribution = .fill
+        stackView.spacing = 8
+        return stackView
+    }()
     
     let rbaDataToggle = Toggle(title: "Recurring Billing (RBA) Data")
     let contactInformationToggle = Toggle(title: "Add Contact Information")
@@ -81,7 +114,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
     let requestBillingAgreementToggle = Toggle(title: "Request Billing Agreement")
 
     override func viewDidLoad() {
-        super.heightConstraint = 500
+        super.usesIntrinsicContentHeight = true
         super.viewDidLoad()
     }
 
@@ -129,9 +162,7 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
         vaultStackView.spacing = 12
 
         let stackView = UIStackView(arrangedSubviews: [
-            emailStackView,
-            countryCodeStackView,
-            nationalNumberStackView,
+            customerInfoStackView,
             oneTimeCheckoutStackView,
             vaultStackView
         ])
@@ -187,7 +218,8 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
             userPhoneNumber: BTPayPalPhoneNumber(
                 countryCode: countryCodeTextField.text ?? "",
                 nationalNumber: nationalNumberTextField.text ?? ""
-            )
+            ),
+            campaigns: payPalCampaigns
         )
 
         if amountBreakdownToggle.isOn {
@@ -231,7 +263,8 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
                 merchantAccountID: "quantumleapsandboxtesting-1",
                 recurringBillingDetails: recurringBillingDetails,
                 recurringBillingPlanType: .subscription,
-                requestBillingAgreement: true
+                requestBillingAgreement: true,
+                campaigns: payPalCampaigns
             )
         }
 
@@ -315,7 +348,8 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
             amount: amount,
             enablePayPalAppSwitch: true,
             userAuthenticationEmail: emailTextField.text,
-            userAction: .payNow
+            userAction: .payNow,
+            campaigns: payPalCampaigns
         )
 
         payPalClient.tokenize(request) { nonce, error in
@@ -339,7 +373,8 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
             enablePayPalAppSwitch: true,
             userAuthenticationEmail: emailTextField.text,
             userAction: .payNow,
-            offerCredit: true
+            offerCredit: true,
+            campaigns: payPalCampaigns
         )
 
         payPalClient.tokenize(request) { nonce, error in
@@ -385,7 +420,8 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
             enablePayPalAppSwitch: true,
             userAuthenticationEmail: emailTextField.text,
             userAction: .payNow,
-            offerPayLater: true
+            offerPayLater: true,
+            campaigns: payPalCampaigns
         )
 
         payPalClient.tokenize(request) { nonce, error in
@@ -398,5 +434,13 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
 
             self.completionBlock(nonce)
         }
+    }
+
+    private var payPalCampaigns: [BTPayPalCampaign] {
+        payPalCampaignsTextField.text?
+            .split(separator: ",")
+            .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
+            .filter { !$0.isEmpty }
+            .map { BTPayPalCampaign(id: $0) } ?? []
     }
 }
--- a/Sources/BraintreeCore/BTPayPalCampaign.swift
+++ b/Sources/BraintreeCore/BTPayPalCampaign.swift
@@ -1,0 +1,17 @@
+import Foundation
+
+/// A PayPal campaign applied to a checkout transaction.
+@objcMembers public class BTPayPalCampaign: NSObject, Encodable {
+
+    // MARK: - Public Properties
+
+    public let id: String
+
+    // MARK: - Public Initializer
+
+    /// Initializes a PayPal campaign.
+    /// - Parameter id: The PayPal campaign ID.
+    public init(id: String) {
+        self.id = id
+    }
+}
--- a/Sources/BraintreeCore/BTPayPalCampaign.swift
+++ b/Sources/BraintreeCore/BTPayPalCampaign.swift
@@ -10,7 +10,7 @@ import Foundation
     // MARK: - Public Initializer
 
     /// Initializes a PayPal campaign.
-    /// - Parameter id: The PayPal campaign ID.
+    /// - Parameter id: Required. The PayPal campaign ID.
     public init(id: String) {
         self.id = id
     }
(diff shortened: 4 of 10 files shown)


joaaraujo-paypal (approved):

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -59,6 +59,7 @@ import BraintreeCore
     var lineItems: [BTPayPalLineItem]?
     var localeCode: BTPayPalLocaleCode?
     var merchantAccountID: String?
+    var campaigns: [BTPayPalCampaign]
```
vkspune: This is optional parameter merchant may or may not send it as part of request

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -113,6 +114,51 @@ import BraintreeCore
...
+        enablePayPalAppSwitch: Bool,
+        userAuthenticationEmail: String? = nil,
+        intent: BTPayPalRequestIntent = .authorize,
+        userAction: BTPayPalRequestUserAction = .none,
+        offerPayLater: Bool = false,
+        currencyCode: String? = nil,
+        requestBillingAgreement: Bool = false,
+        contactPreference: BTContactPreference = .none,
+        offerCredit: Bool = false,
+        campaigns: [BTPayPalCampaign]
```
vkspune: this should be optional parameter

vkspune (changes requested):

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
Juan219: Agreed, this is optional for merchants. The initializer defaults `campaigns` to an empty array, so existing integrations do not need to pass it. When the array is empty, or when all campaign IDs are empty/blank, the SDK omits `paypal_campaigns` from the request body.

I kept the stored property as a non-optional array to avoid separate `nil` vs empty states, since both mean “do not send campaigns” for the wire request.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
Juan219: This is optional for merchants because they can continue using the existing App Switch initializer that does not include `campaigns`. This overload is only for merchants who want to pass campaigns, so the parameter is required here. The main Checkout initializer defaults `campaigns` to an empty array, and the request body omits `paypal_campaigns` when no valid campaign IDs are provided.

Juan219: The intent is that campaigns are optional for merchants to send, but I'd prefer not to model the parameter as a nullable array.

For Swift callers, the main Checkout initializer already defaults `campaigns` to an empty array, so merchants are not required to pass it when they do not have campaigns. For App Switch, we keep the existing initializer without `campaigns` because this API is exposed to Objective-C, and Swift default parameter values do not create a shorter Objective-C selector. That means Objective-C merchants who do not send campaigns can continue using the existing no-campaign initializer, while merchants who do send campaigns can opt into this overload.

Making `campaigns` nullable would add a second "no campaigns" state (`nil` and `[]`) even though both would have the same wire behavior: the SDK omits `paypal_campaigns` when there are no valid campaign IDs. That would add API ambiguity, extra normalization/tests, and more documentation surface without improving the backend contract or Objective-C ergonomics.

vkspune (approved):

agedd:
hi @Juan219 - please move the pr to ready for review status so we can start taking a look.

Demo/Application/Base/PaymentButtonBaseViewController.swift
```diff
@@ -25,14 +28,61 @@ class PaymentButtonBaseViewController: BaseViewController {
...
-        ])
+        paymentButton.translatesAutoresizingMaskIntoConstraints = false
+
+        scrollView.keyboardDismissMode = .interactive
+        scrollView.translatesAutoresizingMaskIntoConstraints = false
+        contentView.translatesAutoresizingMaskIntoConstraints = false
+
+        view.addSubview(scrollView)
+        scrollView.addSubview(contentView)
+        contentView.addSubview(paymentButton)
```
agedd: nit: can we reverse the order of this starting with `contentView.addSubview()` to `view.addSubview()`

buzzamus:
Hey @Juan219 is this ready for review, or still meant to be a draft?

buzzamus:
> Hey @Juan219 is this ready for review, or still meant to be a draft?

Ope sorry saw @agedd already commented

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
buzzamus: Is there a reason this needs to be in a new initializer altogether, though instead of just adding `campaigns: [BTPayPalCampaign] = []` to the existing init? That would eliminate the need to have to maintain 2 similar inits with just one argument difference.

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
```diff
@@ -63,6 +64,13 @@ struct PayPalCheckoutPOSTBody: Encodable {
         self.offerPayLater = payPalRequest.offerPayLater
         self.offerCredit = payPalRequest.offerCredit
         self.amountBreakdown = payPalRequest.amountBreakdown
+
+        let validPayPalCampaigns = payPalRequest.campaigns.filter {
+            !$0.id.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+        }
```
buzzamus: Could we add a unit test where the id is something like " abc "? I think this just checks the id without whitespace, but never actually transforms the value does it? I think filter will just return the original object here.

Sources/BraintreePayPal/Models/BTPayPalCampaign.swift
```diff
@@ -0,0 +1,17 @@
...
+    public let id: String
+
+    // MARK: - Public Initializer
+
+    /// Initializes a PayPal campaign.
+    /// - Parameter id: The PayPal campaign ID.
+    public init(id: String) {
+        self.id = id
+    }
+}
```
buzzamus, line 17: Since this is a PayPal model could this be moded to `BraintreePayPal/Models/`?

Juan219: Good question. I kept `BTPayPalCampaign` in `BraintreeCore` because it is shared by both `BraintreePayPal` and `BraintreeShopperInsights`.

`BraintreeShopperInsights` currently depends only on `BraintreeCore`. If we move this model into `BraintreePayPal/Models`, Shopper Insights would either need to depend on `BraintreePayPal` or define a duplicate campaign model. That would add unnecessary coupling for a shared request model.

There is a second dependent PR that I’ll open after this one is merged. That PR wires the same campaign model into the Shopper Insights APIs so campaign IDs can be sent on create/update customer session and generate customer recommendations requests. Keeping the model in Core allows that follow-up PR to reuse the same public type without changing Shopper Insights’ dependency graph.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
Juan219: Yes, the reason is Objective-C compatibility for the App Switch initializer.

For Swift callers, I agree that `campaigns: [BTPayPalCampaign] = []` is the cleaner shape, and the main `BTPayPalCheckoutRequest` initializer already uses that approach.

The separate initializer is only for the App Switch convenience initializer because this API is exposed to Objective-C. Swift default arguments are not available from Objective-C, so adding `campaigns` to the existing initializer would change the generated Objective-C selector and force Objective-C merchants to pass campaigns even when they do not use them.

Keeping the existing initializer preserves the current Objective-C selector, while the new overload adds an opt-in path for merchants who need to send campaigns. It does mean maintaining two similar App Switch initializers, but it avoids making this optional campaign field a source-breaking change for Objective-C consumers.

Demo/Application/Base/PaymentButtonBaseViewController.swift
Juan219: Sure, I will make the change.

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
Juan219: Yes, exactly. I updated this to keep filter only, so the SDK uses trimming only to decide whether an ID is blank, but it does not transform the merchant-provided value.
I also added/updated the unit test with an ID like " campaign-123-id " to confirm the original value is preserved in the request body. So " abc " would still be sent as " abc ", while "" or "   " are omitted.

Sources/BraintreeCore/BTPayPalCampaign.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+/// A PayPal campaign applied to a checkout transaction.
+@objcMembers public class BTPayPalCampaign: NSObject, Encodable {
+
+    // MARK: - Public Properties
+
+    public let id: String
```
jaxdesmarais: This should be internal. The property should only be exposed on the init.

Sources/BraintreeCore/BTPayPalCampaign.swift
```diff
@@ -0,0 +1,17 @@
...
+
+    // MARK: - Public Initializer
+
+    /// Initializes a PayPal campaign.
+    /// - Parameter id: The PayPal campaign ID.
```
jaxdesmarais:
```suggestion
    /// - Parameter id: Required. The PayPal campaign ID.
```

Sources/BraintreePayPal/Models/BTPayPalCampaign.swift
jaxdesmarais: I am not aware of this being accepted in the `BraintreeShopperInsights` module and it is not defined in the LLD for this feature.

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
```diff
@@ -26,6 +26,7 @@ struct PayPalCheckoutPOSTBody: Encodable {
     private var merchantAccountID: String?
     private var osType: String?
     private var osVersion: String?
+    private var payPalCampaigns: [BTPayPalCampaign]?
```
jaxdesmarais, line 29:
```suggestion
    private var campaigns: [BTPayPalCampaign]?
```

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
```diff
@@ -63,6 +64,14 @@ struct PayPalCheckoutPOSTBody: Encodable {
         self.offerPayLater = payPalRequest.offerPayLater
         self.offerCredit = payPalRequest.offerCredit
         self.amountBreakdown = payPalRequest.amountBreakdown
+
+        let validPayPalCampaigns = payPalRequest.campaigns.compactMap {
+            let trimmedID = $0.id.trimmingCharacters(in: .whitespacesAndNewlines)
```
jaxdesmarais: Does the API not do validation for whitespace/newlines in IDs? This logic is potentially mutating the ID sent to the SDK, which we should not do.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -113,6 +114,51 @@ import BraintreeCore
...
+    ///   - currencyCode: Optional: A three-character ISO-4217 ISO currency code to use for the transaction. Defaults to merchant currency code if not set.
+    ///   See https://developer.paypal.com/docs/api/reference/currency-codes/ for a list of supported currency codes.
+    ///   - requestBillingAgreement: Optional: If set to `true`, this enables the Checkout with Vault flow, where the customer will be prompted to consent to a billing agreement
+    ///   during checkout. Defaults to `false`.
+    ///   - contactPreference: Optional: Preference for the contact information section within the payment flow. Defaults to `.none` if not set.
+    ///   - offerCredit: Optional: Offers PayPal Credit if the customer qualifies. Defaults to `false`.
+    ///   - campaigns: Optional: PayPal campaigns applied to the transaction.
+    /// - Warning: This initializer should be used for merchants using the PayPal App Switch flow. This feature is currently in beta and may change or be removed in future releases.
+    /// - Note: The PayPal App Switch flow currently only supports the production environment.
+    public convenience init(
```
jaxdesmarais, line 92: Why do we need a whole new init for this? This should be removed and added to the end of the existing init with a default value/nil only.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* BraintreePayPal
+  * Add `campaigns` to `BTPayPalCheckoutRequest` and pass them as `paypal_campaigns` in PayPal checkout requests.
```
```suggestion
  * Add `campaigns` to `BTPayPalCheckoutRequest`
```

jaxdesmarais (changes requested):

Demo/Application/Base/PaymentButtonBaseViewController.swift
```diff
@@ -25,14 +28,61 @@ class PaymentButtonBaseViewController: BaseViewController {
...
+
+        let paymentButtonBottomConstraint = paymentButton.bottomAnchor.constraint(
+            equalTo: contentView.bottomAnchor,
+            constant: -20
+        )
+        paymentButtonBottomConstraint.priority = .defaultHigh
+
+        var constraints = [
+            scrollView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            scrollView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
```
agedd:
```suggestion
            scrollView.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor),
            scrollView.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor),
```

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
```diff
@@ -63,6 +64,14 @@ struct PayPalCheckoutPOSTBody: Encodable {
         self.offerPayLater = payPalRequest.offerPayLater
         self.offerCredit = payPalRequest.offerCredit
         self.amountBreakdown = payPalRequest.amountBreakdown
+
+        let validPayPalCampaigns = payPalRequest.campaigns.compactMap {
```
agedd: small nit: 
```suggestion
        let validCampaigns = payPalRequest.campaigns.compactMap {
```

CHANGELOG.md
agedd: nit: can we also bake in why we are adding campaigns to this request into the changelog entry

agedd:
are you able to pull in the latest from main?

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
Juan219: I only added because of this comment: https://github.com/braintree/braintree_ios/pull/1832#discussion_r3640881479, it was a map before, but since you own the code I'm not sure what is the standard practice.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
Juan219: Already address this comment here: https://github.com/braintree/braintree_ios/pull/1832#discussion_r3660118921

Let me know if this is good or I should make the proposal.

Sources/BraintreePayPal/Models/BTPayPalCampaign.swift
Juan219: Thanks for calling this out. Shopper Insights support is part of the broader documented scope, but it is being split into a dependent follow-up PR to keep this first PR focused on PayPal Checkout / create order.

The feature documentation calls out campaign ID support for shopper session create/update/fetch recommendations, payment-ready, and transaction request/create order flows. The second PR will wire this same campaign model into `BraintreeShopperInsights` for those shopper session/recommendations APIs.

That is why I placed `BTPayPalCampaign` in `BraintreeCore`: both `BraintreePayPal` in this PR and `BraintreeShopperInsights` in the follow-up PR need to reference the same public campaign model without making Shopper Insights depend on the PayPal module.

If we prefer to keep this PR strictly scoped to PayPal Checkout only, I can move the model to `BraintreePayPal/Models` now and promote it to Core in the follow-up PR, but that would create extra file movement/API churn between the two dependent PRs.

Sources/BraintreeCore/BTPayPalCampaign.swift
Juan219: This depends on the result from another comment, if BTPayPalCampaign will continue living on `BraintreeCore` or will be moved to `BraintreePayPal/Models`

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
Juan219: I’d prefer to keep this as `payPalCampaigns` because it mirrors the API field this value represents. Even though `CodingKeys` controls the final wire key, keeping the private property aligned with the backend concept makes the encoding intent clearer and avoids having two names for the same thing in this model.

For merchant-facing API, we expose the cleaner `campaigns` parameter on `BTPayPalCheckoutRequest`. At the POST body layer, this property is no longer merchant-facing; it represents the serialized PayPal campaigns field, so `payPalCampaigns` feels appropriate here.

Sources/BraintreePayPal/Models/BTPayPalCampaign.swift
jaxdesmarais: I agree with keeping things scoped properly to reduce PR size. But we need to think through this as the LLD we approved did not include adding campaign IDs to shopper insights. We also need to think about future scope where say we want Venmo campaigns or some other payment method type. I think calling something PayPal and scoping it to the core SDK is a pretty large anti-pattern, as Brent called out. I am open on suggestions though on how we want to scale this to other feature sets in the future, but agree with moving `BTPayPalCampaign` into the PayPal module.

Demo/Application/Base/PaymentButtonBaseViewController.swift
Juan219: For context, this view was updated so the Demo App can display the new campaign input without clipping content on smaller screens. The main fix was adding the scroll view behavior so all fields/buttons remain reachable.

I’m okay with anchoring the scroll view horizontally to the safe area as a small consistency improvement, but it is not required for the campaign ID flow itself. If we prefer to keep this PR focused on the minimum Demo App change needed to test campaigns, I can leave the existing horizontal anchors unchanged.

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
jaxdesmarais: That makes sense - thanks for the reasoning.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
jaxdesmarais: We can default to nil and avoid the duplications. Same as `lineItems`

Juan219: That makes sense from the Swift side, and I agree it would remove the duplicated App Switch convenience initializer.

The reason I added a separate initializer was to preserve the existing Objective-C selector. Since this class is exposed to Objective-C through `@objcMembers`, Swift default arguments are not exported to Objective-C as additional shorter overloads.

For example, if we only keep one Swift initializer like:

```swift
public convenience init(
    amount: String,
    enablePayPalAppSwitch: Bool,
    ...
    offerCredit: Bool = false,
    campaigns: [BTPayPalCampaign]? = nil
)
Swift callers can omit campaigns because of the default value:
BTPayPalCheckoutRequest(
    amount: "1.00",
    enablePayPalAppSwitch: true
)
But Objective-C callers would need to use the full selector that includes campaigns, even when they do not have campaigns to send:
[[BTPayPalCheckoutRequest alloc] initWithAmount:@"1.00"
                           enablePayPalAppSwitch:YES
                        userAuthenticationEmail:nil
                                         intent:BTPayPalRequestIntentAuthorize
                                     userAction:BTPayPalRequestUserActionNone
                                 offerPayLater:NO
                                   currencyCode:nil
                        requestBillingAgreement:NO
                              contactPreference:BTContactPreferenceNone
                                    offerCredit:NO
                                      campaigns:nil];
Keeping the existing initializer plus a campaign-specific overload avoids that Objective-C source change, while still giving campaign merchants an opt-in path.
That said, if the team is comfortable with the Objective-C selector change for this beta App Switch initializer, I can simplify this by removing the new initializer and adding campaigns: [BTPayPalCampaign]? = nil to the end of the existing one, similar to lineItems.

Comment (jaxdesmarais) [reply Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift:92]: We are okay with the latter and understand it means changes for Obj-C callers.

File: Sources/BraintreePayPal/Models/BTPayPalCampaign.swift (hunk 5 above)
Comment (Juan219) [reply Sources/BraintreePayPal/Models/BTPayPalCampaign.swift:17]: Yes, agreed. I moved BTPayPalCampaign into BraintreePayPal/Models/.

Since Shopper Insights now uses its own BTShopperInsightsCampaign model, keeping BTPayPalCampaign scoped to the PayPal module makes the ownership clearer and avoids exposing a PayPal Checkout-specific model from BraintreeCore.

File: Sources/BraintreeCore/BTPayPalCampaign.swift (hunk 6 above)
Comment (Juan219) [reply Sources/BraintreeCore/BTPayPalCampaign.swift]: Agreed. I updated id to be internal and kept it exposed only through the public initializer.
Now that BTPayPalCampaign lives only in the BraintreePayPal module, the stored property does not need to be public. Merchants can still create campaigns with BTPayPalCampaign(id:), while the SDK can encode the ID internally.

File: Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift (hunk 10 above)
Comment (Juan219) [reply Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift:92]: Sure, I have updated the code to keep one single init.

File: Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift (hunk 2 above)
Comment (Juan219) [reply Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift]: No need to create a new initializer, keeping one based on this comment thread: https://github.com/braintree/braintree_ios/pull/1832/files/f8f333b2408317f7d8fc5ad66c43ecca6a1cafea#diff-49e322ceda0295e096bd87fcbac95670577c9eb9ce74916c97faba40eee54195

**Comment by Juan219** [comment]
> are you able to pull in the latest from main?

@agedd  I can't update the feature branch.

File: Demo/Application/Features/PayPalWebCheckoutViewController.swift [hunk 14]
```diff
@@ -74,14 +74,47 @@ class PayPalWebCheckoutViewController: PaymentButtonBaseViewController {
...
+    }()
     
     let rbaDataToggle = Toggle(title: "Recurring Billing (RBA) Data")
     let contactInformationToggle = Toggle(title: "Add Contact Information")
     let amountBreakdownToggle = Toggle(title: "Amount Breakdown")
     let requestBillingAgreementToggle = Toggle(title: "Request Billing Agreement")
 
     override func viewDidLoad() {
-        super.heightConstraint = 500
+        super.usesIntrinsicContentHeight = true
```
jaxdesmarais, line 117: Just want to double check we have made sure this doesn't break other views in the demo app correct? Same with the changes to the base view controller?

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -175,7 +180,8 @@ import BraintreeCore
         shopperSessionID: String? = nil,
         userAuthenticationEmail: String? = nil,
         userPhoneNumber: BTPayPalPhoneNumber? = nil,
-        offerCredit: Bool = false
+        offerCredit: Bool = false,
+        campaigns: [BTPayPalCampaign] = []
```
jaxdesmarais: This can also default to nil correct?

UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift
```diff
@@ -152,6 +161,99 @@ class BTPayPalCheckoutRequest_Tests: XCTestCase {
...
+
+        guard let parameters = try? request.encodedPostBodyWith(configuration: configuration).toDictionary() else {
+            XCTFail()
+            return
+        }
+
+        XCTAssertNil(parameters["paypal_campaigns"])
+    }
+
+    func testParametersWithConfiguration_whenSomePayPalCampaignIDsAreEmpty_returnsOnlyValidPayPalCampaignsInOrder() {
```
jaxdesmarais: We should likely add a docstring to the property that it cannot include whitespace. This type of validation an stripping of characters on the client vs handling this as invalid in the API still feels a bit odd to me. Are we doing the same on Android? cc: @santugowda

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* BraintreePayPal
+  * Add `campaigns` to `BTPayPalCheckoutRequest` and pass them in PayPal checkout requests so eligible campaign context can be persisted during order creation and surfaced in the PayPal checkout experience.
```
jaxdesmarais: nit: the second part feels a bit redundant

```suggestion
  * Add `campaigns` to `BTPayPalCheckoutRequest` so eligible campaign context can be persisted during order creation and surfaced in the PayPal checkout experience.
```

UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift
santugowda: On Android we're not stripping or validating whitespace : the id is passed through in both modules (PayPalCampaign → paypal_campaigns, ShopperInsightsCampaign → paypalCampaigns), sent as `{ "id": "<value>" }` with no trim/replace/isBlank. 
So we don't mirror the iOS client-side stripping.  Please let me know if we need to do this on Android.

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
Juan219: Updated.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
Juan219: Yes, good call. I can make this campaigns: [BTPayPalCampaign]? = nil and normalize it internally with campaigns ?? []. That keeps the public initializer aligned with the fact that campaigns are optional, allows Objective-C callers to pass nil, and avoids carrying optional handling through the rest of the request/encoding logic.

Demo/Application/Features/PayPalWebCheckoutViewController.swift
Juan219: I re-tested the Demo App integrations manually to make sure the shared PaymentButtonBaseViewController layout change does not introduce unintended visual regressions.
For validation, I captured before/after screenshots using braintree/paypal-campaigns-feature as the baseline and this PR branch as the after state. The only expected visual difference is in the PayPal Checkout screen, where the campaign field is added and the content uses the updated top-aligned scroll layout. The other Demo integrations were checked to confirm their UI remains visually unchanged.

UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift
Juan219: Good point. The current iOS logic is not stripping or mutating campaign IDs that are sent to the API. The trim is only used to identify empty / whitespace-only IDs so we do not send a no-op paypal_campaigns array. For example, " abc " is preserved and sent as " abc ", and there is test coverage for that case.
Based on the Android response, Android also passes the ID through without trim/replace/isBlank validation. To keep this aligned, I would avoid documenting that campaign IDs “cannot include whitespace.” Instead, I can clarify the initializer parameter docstring to say that empty or whitespace-only IDs are omitted from the request, while non-empty IDs are sent unchanged.

CHANGELOG.md
Juan219: Updated.

UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift
jaxdesmarais: > Instead, I can clarify the initializer parameter docstring to say that empty or whitespace-only IDs are omitted from the request, while non-empty IDs are sent unchanged.

Can you and @santugowda instead confirm with the API what should be passed and where this should happen? Say I am a merchant and I incorrectly pass in:
- Campaign 1: "      "
- Campaign 2: "valid"
- Campaign 3: "also valid"

The question here is:
1. Should the SDK pass in 3 campaigns which the API then returns an error such as "Campaign 1 is invalid"
2. Should the SDK remove campaign 1 completely and return only 2 - what does this do from a merchant point of view if Campaign 2 becomes Campaign 1 because the SDK removed Campaign 1 for being invalid?

My gut of how APIs and SDKs work says we should be doing option 1 here, the SDK is a passthrough not a validation point, but we you all to do some confirmation so we can get these PRs across.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -198,6 +204,7 @@ import BraintreeCore
         self.lineItems = lineItems
         self.localeCode = localeCode
         self.merchantAccountID = merchantAccountID
+        self.campaigns = campaigns ?? []
```
jaxdesmarais: Do we need to default to `??` we don't seem to do that with lineItems so can do the same here and above to match that implimentation.

UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift
jaxdesmarais: @Juan219 this is the only open comment which impacts this code in both PRs:
```swift
        let validCampaigns = payPalRequest.campaigns.filter {
            !$0.id.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
        }
```
If we don't know the answer, lets let the API handle validation and just check `.isEmpty` to align with Android.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* BraintreePayPal
+  * Add `campaigns` to `BTPayPalCheckoutRequest` so eligible campaign context can be persisted during order creation and surfaced in the PayPal checkout experience.
 * BraintreeUIComponents
```
jaxdesmarais, line 8: You'll need to pull in the latest. This was already released so should be under a new unreleased section.

jaxdesmarais:
@Juan219 want to make sure the comments aren't getting lost here so we can merge this in

UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift
Juan219: Updated. Same as Android.

Sources/BraintreePayPal/Models/PayPalCheckoutPOSTBody.swift
```diff
@@ -63,6 +64,15 @@ struct PayPalCheckoutPOSTBody: Encodable {
...
         self.amountBreakdown = payPalRequest.amountBreakdown
+
+        if let campaigns = payPalRequest.campaigns {
+            let validCampaigns = campaigns.filter {
+                !$0.id.isEmpty
+            }
+            if !validCampaigns.isEmpty {
+                self.payPalCampaigns = validCampaigns
+            }
+        }
```
jaxdesmarais: We can match the line items logic exactly:

```suggestion
        if let campaigns = payPalRequest.campaigns, !campaigns.isEmpty {
            self.campaigns = campaigns
        }
```

Juan219: Updated

jaxdesmarais (approved):

buzzamus (approved):

PR #1838, opened 2026-07-29
Create Payment Actions Service

agedd:
### Summary of changes
- Create `BTPaymentActionsService` 
- Add `BTPaymentActionsError`
- Add related unit tests 

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used Claude to help map out and write the unit tests

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

- @agedd


Sources/BraintreeCore/PaymentActions/BTPaymentActionStatus.swift
```diff
@@ -24,6 +30,8 @@ import Foundation
         case .readyForConfirmation: return "READY_FOR_CONFIRMATION"
         case .succeeded: return "SUCCEEDED"
         case .requiresCapture: return "REQUIRES_CAPTURE"
+        case .canceled: return "CANCELED"
+        case .expired: return "EXPIRED"
         case .unknown: return "UNKNOWN"
         }
```
buzzamus, line 36: Do these new values need to be in the `status` function too?

Sources/BraintreeCore/PaymentActions/BTPaymentActionsError.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+
+public enum BTPaymentActionsError: Error, CustomNSError, LocalizedError, Equatable {
+    
+    /// 0. Unknown error
+    case unknown
+    
+    /// 1. Customer provided invalid input
+    case customerInputInvalid([String: Any])
```
jaxdesmarais: Do we know the shape of the error we will get back from GraphQL for this method?

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
```diff
@@ -0,0 +1,52 @@
...
+    // MARK: - Initializer
+    
+    init(apiClient: BTAPIClient) {
+        self.apiClient = apiClient
+    }
+    
+    // MARK: - Internal Methods
+    
+    func setPaymentMethod<Body: BTGraphQLEncodableBody>(_ body: Body) async throws -> BTPaymentActionResult {
+
```
jaxdesmarais:
```suggestion
```

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
```diff
@@ -0,0 +1,52 @@
...
+    
+    func setPaymentMethod<Body: BTGraphQLEncodableBody>(_ body: Body) async throws -> BTPaymentActionResult {
+        
+        let responseBody: BTJSON?
+        
+        do {
+            (responseBody, _) = try await apiClient.post("", parameters: body, httpType: .graphQLAPI)
+        } catch let error as NSError {
+            if error.domain == BTCoreConstants.httpErrorDomain,
+            error as? BTHTTPError == .clientError([:]),
```
jaxdesmarais: I think we need to follow the GraphQL pattern for this instead. We can take a look at the other requests that hit GraphQL for the shape but iirc this is all very HTTP error leaning currently.

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
```diff
@@ -0,0 +1,52 @@
...
+                throw BTPaymentActionsError.customerInputInvalid(userInfo)
+            }
+            throw error
+        }
+        
+        let setPaymentActionPaymentMethodJSON: BTJSON = responseBody?["data"]["setPaymentActionPaymentMethod"] ?? BTJSON()
+        
+        if let jsonError = setPaymentActionPaymentMethodJSON.asError() {
+            throw jsonError
+        }
```
jaxdesmarais: Same here, this is also very REST API specific it seems.

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
```diff
@@ -0,0 +1,52 @@
...
+        
+        let paymentActionJSON: BTJSON = setPaymentActionPaymentMethodJSON["paymentAction"]
+        let paymentActionID = paymentActionJSON["id"].asString() ?? ""
+        let status = BTPaymentActionStatus.status(from: paymentActionJSON["status"].asString() ?? "")
```
jaxdesmarais: If we don't get a status back is that an issue? Should any of these be a guard vs let?

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
```diff
@@ -0,0 +1,57 @@
...
+    }
+    
+    // MARK: - Internal Methods
+    
+    /// Submits a payment method to the `setPaymentActionPaymentMethod` GraphQL mutation.
+    /// - Parameter body: the GraphQL request body to submit, encoding the payment method details and the fields
+    /// to return on the `paymentAction` selection set.
+    /// - Returns: a `BTPaymentActionResult` containing the Payment Action `id` and `status`.
+    /// - Throws: `BTPaymentActionsError.customerInputInvalid` if the server returns a 422 with an invalid-input message, or the underlying error otherwise.
+    func setPaymentMethod<Body: BTGraphQLEncodableBody>(_ body: Body) async throws -> BTPaymentActionResult {
```
jaxdesmarais: It looks like we took a slightly different approach in the Android PR. I don't have a strong feeling on which one to align with but will likely be easier for us to reason about if we do the same across both. Curious to hear yours and @mzlangreder thoughts!

Sources/BraintreeCore/PaymentActions/BTPaymentActionsError.swift
agedd: ah yep, good callout. i've actually removed this error type for now since we don't know the exact shape just yet.

Sources/BraintreeCore/PaymentActions/BTPaymentActionStatus.swift
agedd: good catch! added these

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
agedd: ahh yep, removed this in 7f462f4603de5c77b23694555d612ce9f354e67b

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
agedd: ah good callout - dug into `BTGraphQLHTTP` and it looks like it already inspects the top-level `errors` array and throws before this method ever sees a response body so `asError()` was unnecessary so removed it

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
agedd: ah yes, updated so now `id` and `status` are both guard-fail if the key is missing and this aligns with android as well

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
agedd: confirmed with the Android PR, missing `id`/`status` key -> failure, unrecognized but present status -> still success with `unknown`.

buzzamus (approved):
Looks good!

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
```diff
@@ -0,0 +1,47 @@
...
+        
+        // TODO: no error type currently exists for this service -- using NSError as a placeholder until the team decides how errors should be surfaced here.
+        guard let paymentActionID = paymentActionJSON["id"].asString(), !paymentActionID.isEmpty else {
+            throw NSError(
+                domain: "com.braintreepayments.BTPaymentActionsErrorDomain",
+                code: 0,
+                userInfo: [NSLocalizedDescriptionKey: "Payment Action response is missing an id."]
+            )
+        }
+        guard let statusString = paymentActionJSON["status"].asString(), !statusString.isEmpty else {
```
jaxdesmarais: It looks like Android sets the status as `.unknown` if we do not get one back where iOS throws an error. What is the expected flow?

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
```diff
@@ -0,0 +1,47 @@
...
+        
+        let paymentActionJSON: BTJSON = responseBody?["data"]["setPaymentActionPaymentMethod"]["paymentAction"] ?? BTJSON()
+        
+        // TODO: no error type currently exists for this service -- using NSError as a placeholder until the team decides how errors should be surfaced here.
+        guard let paymentActionID = paymentActionJSON["id"].asString(), !paymentActionID.isEmpty else {
+            throw NSError(
+                domain: "com.braintreepayments.BTPaymentActionsErrorDomain",
+                code: 0,
+                userInfo: [NSLocalizedDescriptionKey: "Payment Action response is missing an id."]
+            )
```
jaxdesmarais: I think we can address this in this PR with an error type and message for this case.

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
```diff
@@ -0,0 +1,47 @@
...
+    
+    /// Submits a payment method to the `setPaymentActionPaymentMethod` GraphQL mutation.
+    /// - Parameter body: the GraphQL request body to submit, encoding the payment method
+    ///   details and the fields to return on the `paymentAction` selection set.
+    /// - Returns: a `BTPaymentActionResult` containing the Payment Action `id` and `status`.
+    /// - Throws: the underlying error from the network layer or GraphQL response.
+    func setPaymentActionPaymentMethod<Body: BTGraphQLEncodableBody>(
+        _ body: Body
+    ) async throws -> BTPaymentActionResult {
+        let (responseBody, _) = try await apiClient.post("", parameters: body, httpType: .graphQLAPI)
```
jaxdesmarais: We should be able to wrap this in a `do/catch` and throw an error? We may need a todo for the exact error shape but can implement the `do/catch` now.

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
agedd: hmm i think we throw `.unknown` when we don't recognize the raw wire value (out of the currently defined status types) - adjusted this to throw an `missingStatus` error in the case we get back nil for this and can confirm later in a future PR.

Sources/BraintreeCore/PaymentActions/BTPaymentActionsService.swift
agedd: ah yep, addressed this in de42105e5ad91481356949ec84f1bdd05273b1bf

Sources/BraintreeCore/PaymentActions/BTPaymentActionError.swift
```diff
@@ -0,0 +1,43 @@
...
+    public var errorUserInfo: [String: Any] {
+        switch self {
+        case .missingID:
+            return [NSLocalizedDescriptionKey: "Payment Action response is missing an id."]
+        case .missingStatus:
+            return [NSLocalizedDescriptionKey: "Payment Action response is missing a status."]
+        case .decodingFailure:
+            return [NSLocalizedDescriptionKey: "Failed to decode Payment Action response."]
+        }
+    }
```
jaxdesmarais: Can we use `errorDescription` here instead?

agedd: yep, resolved in 126804d2b246eee3c5552fd9ff4956ba1b1b8ff6

Braintree.xcodeproj/project.pbxproj
```diff
@@ -41,7 +41,6 @@
 		087831392F7C1666003BB460 /* ValidationResult.swift in Sources */ = {isa = PBXBuildFile; fileRef = 087831382F7C1666003BB460 /* ValidationResult.swift */; };
 		0878313B2F7C2839003BB460 /* CardFieldsContainerView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0878313A2F7C2839003BB460 /* CardFieldsContainerView.swift */; };
 		0878313D2F83FBDA003BB460 /* CVVCharacter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0878313C2F83FBDA003BB460 /* CVVCharacter.swift */; };
-		2786AB6520BA4E79A5EFB08B /* CVVHintCard.swift in Sources */ = {isa = PBXBuildFile; fileRef = 731BD3E21BC44E928B80A155 /* CVVHintCard.swift */; };
```
jaxdesmarais, line 44: We should figure out why these diffs keep popping up for the unrelated files at some point. :)

jaxdesmarais (approved):

PR #1842, opened 2026-08-07
[Open Dev] Use fresh Client Token for PayPal CI Jobs

buzzamus:
### Summary of changes
- This PR Fixes CI test flakiness with the following changes:
      - uses a fresh client token instead of our static one to test and see if CI successes increase at all for the PayPal integration tests
      - Increases timeout from 5 to 10 seconds due to rare slower response times from sandbox on card transactions
      - checks for PP button to appear before clicking instead of just attempting to click (this causes instant failure rather than making wait times longer, therefore cutting down time on tests that will result in failure regardless)

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Analyzed what tests were failing most and create a fresh client token for those tests

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus


IntegrationTests/Helpers/BTIntegrationTestsConstants.swift
```diff
@@ -6,4 +6,23 @@ class BTIntegrationTestsConstants {
...
     static let sandboxClientTokenVersion3 = "eyJ2ZXJzaW9uIjozLCJhdXRob3JpemF0aW9uRmluZ2VycHJpbnQiOiIxYzM5N2E5OGZmZGRkNDQwM2VjNzEzYWRjZTI3NTNiMzJlODc2MzBiY2YyN2M3NmM2OWVmZjlkMTE5MjljOTVkfGNyZWF0ZWRfYXQ9MjAxNy0wNC0wNVQwNjowNzowOC44MTUwOTkzMjUrMDAwMFx1MDAyNm1lcmNoYW50X2lkPWRjcHNweTJicndkanIzcW5cdTAwMjZwdWJsaWNfa2V5PTl3d3J6cWszdnIzdDRuYzgiLCJjb25maWdVcmwiOiJodHRwczovL2FwaS5zYW5kYm94LmJyYWludHJlZWdhdGV3YXkuY29tOjQ0My9tZXJjaGFudHMvZGNwc3B5MmJyd2RqcjNxbi9jbGllbnRfYXBpL3YxL2NvbmZpZ3VyYXRpb24ifQ=="
+
+    /// Fetches a fresh, short-lived client token from the live sandbox demo merchant server.
+    /// Client tokens expire after 24 hours, so tests that specifically exercise client-token
+    /// auth must request one at run time rather than relying on a hardcoded fixture.
+    static func fetchClientToken() async throws -> String {
+        guard let url = URL(string: "https://braintree-demo-merchant-63b7a2204f6e.herokuapp.com/client_token") else {
```
mzlangreder: Does this demo URL ever change?

Is it weird to have a function in a file for constants?

buzzamus: the URL doesn't change and yeah I see your point. I threw it in there for testing to see if this worked at all and then forgot to circle back, so i'll throw it somewhere else where it makes sense.

buzzamus: Updated to move into the Helpers

mzlangreder (approved):

jaxdesmarais (approved):
🚀 thanks for fixing this to resolve the flaky CI job.

PR #1843, opened 2026-08-07
Add `BTPaymentActionsClient` and Payment Actions Module

agedd:
### Summary of changes

- Add a new `BraintreePaymentActions` module that will house everything PA-related

- Add  public `BTPaymentActionsClient` with a generic `BTPaymentActionRequest `protocol plus `AnyEncodable` and graphQL request body so payment-method-specific payloads can be encoded and submitted via GraphQL.

- Add `SetPaymentActionPaymentMethodGraphQLBody`  and helper `AnyEncodable` wrapper struct to allow encoding a payment-method specific payload with the client needing to manage its type.

- Add generic `BTPaymentActionRequest` that each payment-method request will conform to

- Adds a card-specific request - `BTCardPaymentActionRequest` - and shared result/status/error types; includes `PrivacyInfo.xcprivacy` for the new module.

- Updates packaging/build: Package.swift (new SPM library/target), Podspec subspec, and large Xcodeproj edits to add the BraintreePaymentActions framework + test target; removes older card-specific GraphQL/result files.

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
- Used for initial research on how to structure the files and add helper methods for the generic request type handled by the client.

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

- @agedd


Braintree.xcodeproj/project.pbxproj
```diff
@@ -144,13 +144,20 @@
...
+		622C948E302A2CC7005001C2 /* PrivacyInfo.xcprivacy in Resources */ = {isa = PBXBuildFile; fileRef = 622C948C302A2CC7005001C2 /* PrivacyInfo.xcprivacy */; };
+		622C9497302A4EB8005001C2 /* BTPaymentActionRequest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 626E89503026748000AE6692 /* BTPaymentActionRequest.swift */; };
+		622C9498302A4EB8005001C2 /* BTPaymentActionError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 626E89473026698A00AE6692 /* BTPaymentActionError.swift */; };
+		622C9499302A4EB8005001C2 /* BTPaymentActionResult.swift in Sources */ = {isa = PBXBuildFile; fileRef = 626E89483026698A00AE6692 /* BTPaymentActionResult.swift */; };
+		622C949A302A4EB8005001C2 /* BTPaymentActionsClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 626E89493026698A00AE6692 /* BTPaymentActionsClient.swift */; };
+		622C949B302A4EB8005001C2 /* BTPaymentActionStatus.swift in Sources */ = {isa = PBXBuildFile; fileRef = 626E894A3026698A00AE6692 /* BTPaymentActionStatus.swift */; };
+		622C949C302A4EC8005001C2 /* BTPaymentActionsClientTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 622C9490302A4C40005001C2 /* BTPaymentActionsClientTests.swift */; };
```
buzzamus: I see the other PR for tests is open, but the Tests file is in this PR and not that one, so may need to switch up this stuff in your PRs, or just be cognizant of the merging order since I am guessing locally these are not a problem

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
```diff
@@ -1,15 +1,31 @@
...
+    public init(authorization: String) {
+        self.apiClient = BTAPIClient(authorization: authorization)
+    }
+    
+    // MARK: - Public Methods
+    
+    // TODO: - Add Obj-C interop in a follow-up PR.
+    @nonobjc public func submitForPaymentAction(_ request: any BTPaymentActionRequest) async throws -> BTPaymentActionStatus {
+        let body = SetPaymentActionPaymentMethodGraphQLBody(request: request)
+        let result = try await setPaymentActionPaymentMethod(body)
```
buzzamus: will this have a catch to handle errors later, or will the  code that calls this function handle the error?

Sources/BraintreePaymentActions/Models/BTCardPaymentActionRequest.swift
```diff
@@ -0,0 +1,139 @@
+import Foundation
+
+/// The card details to submit as the payment method for a Payment Action.
+@objcMembers public class BTCardPaymentActionRequest: NSObject, BTPaymentActionRequest {
```
jaxdesmarais: I thought for requests in v7 we were keeping the constants internal and only exposing the init?

Sources/BraintreePaymentActions/Models/BTCardPaymentActionRequest.swift
```diff
@@ -0,0 +1,139 @@
...
+
+    /// The postal or ZIP code of the cardholder's billing address.
+    public let postalCode: String?
+
+    /// The 2-letter ISO 3166-1 country code of the cardholder's billing address.
+    public let countryCodeAlpha2: String?
+
+    // MARK: - Initializer
+
+    public init(
```
jaxdesmarais, line 36: Docstrings should go here instead vs on the properties per the v7 pattern.

Sources/BraintreePaymentActions/Models/BTCardPaymentActionRequest.swift
```diff
@@ -0,0 +1,139 @@
...
+        self.countryCodeAlpha2 = countryCodeAlpha2
+    }
+}
+
+// MARK: - BTPaymentActionRequest
+
+extension BTCardPaymentActionRequest {
+
+    public func paymentMethodParameters() -> any Encodable {
+        PaymentMethodPayload(card: .init(request: self))
```
jaxdesmarais: I find that we generally don't use the `.init` pattern in our codebase and instead initialize the actual type. Can we update this to the same?

Sources/BraintreePaymentActions/Models/BTCardPaymentActionRequest.swift
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    struct PaymentMethodPayload: Encodable {
```
jaxdesmarais: It looks like other modules have these separated out differently vs within the same file as the request. Can we match the existing pattern used?

Sources/BraintreePaymentActions/Models/BTCardPaymentActionRequest.swift
```diff
@@ -0,0 +1,139 @@
...
+        self.locality = locality
+        self.region = region
+        self.postalCode = postalCode
+        self.countryCodeAlpha2 = countryCodeAlpha2
+    }
+}
+
+// MARK: - BTPaymentActionRequest
+
+extension BTCardPaymentActionRequest {
```
jaxdesmarais: This is pretty small - the entire method is 3 lines, what's the advantage to putting it as an extension vs within the request class? Also I believe this method can be internal correct?

Sources/BraintreePaymentActions/BTPaymentActionRequest.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+/// A payment method-specific request to submit to a Payment Action via `BTPaymentActionsClient`.
+public protocol BTPaymentActionRequest {
```
jaxdesmarais: This should be able to be internal vs public, correct?

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
agedd: ah yep good catch, add this in c25e39dac445bc013c7c3c9618cf2bb5850ae48a

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
```diff
@@ -1,15 +1,31 @@
...
+    /// Creates a Payment Actions Client.
+    /// - Parameter authorization: A valid client token or tokenization key used to authorize API calls.
+    public init(authorization: String) {
+        self.apiClient = BTAPIClient(authorization: authorization)
+    }
+    
+    // MARK: - Public Methods
+    
+    // TODO: - Add Obj-C interop in a follow-up PR.
+    @nonobjc public func submitForPaymentAction(_ request: any BTPaymentActionRequest) async throws -> BTPaymentActionStatus {
```
jaxdesmarais: Do we return just the status or the result? I also think we can handle the obj-c interop now, same with it missing from the init. iirc `AnyObject` protocols bridge to Obj-C.

Braintree.xcodeproj/project.pbxproj
agedd: ah yep good callout ~ removed the unit test reference here so it's cleaner

jaxdesmarais (changes requested):

Sources/BraintreePaymentActions/Models/BTCardPaymentActionRequest.swift
agedd: ah yes, addressed in 01042513e42669cf55ebc40ef4bef5aaa995fcd3

Sources/BraintreePaymentActions/Models/BTCardPaymentActionRequest.swift
agedd: yep, updated in 01042513e42669cf55ebc40ef4bef5aaa995fcd3

Sources/BraintreePaymentActions/Models/BTCardPaymentActionRequest.swift
agedd: ah yep, removed the extension and placed within the request class itself

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
agedd: ah i believe we agreed to returning the `status` fully for now, but this is subject to change. also added obj-c interop now in 78251492cf300479a05398130311b986a79b1d81

Sources/BraintreePaymentActions/SetPaymentActionPaymentMethodGraphQLBody.swift
```diff
@@ -0,0 +1,49 @@
...
+        
+        let input: Input
+        
+        init(request: any BTPaymentActionRequest) {
+            self.input = Input(request: request)
+        }
+        
+        struct Input: Encodable {
+            
+            let paymentActionId: String
```
jaxdesmarais: we will need a custom coding key but to match the rest of the SDK pattern:
```suggestion
            let paymentActionID: String
```

Sources/BraintreePaymentActions/BTPaymentActionRequest.swift
```diff
@@ -0,0 +1,12 @@
+import Foundation
+
+/// A payment method-specific request to submit to a Payment Action via `BTPaymentActionsClient`.
+protocol BTPaymentActionRequest {
```
buzzamus: Super small nit: Looks like our other protocols end the name with `Protocol`...but not sure if that is an actual style rule we have or if it is really just an unnecessary extra word

buzzamus (approved):
This looks good! Just one tiny tiny nit/question/convo starter that is a take it or leave it

Sources/BraintreePaymentActions/BTPaymentActionRequest.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+/// A payment method-specific request to submit to a Payment Action via `BTPaymentActionsClient`.
+/// This base class is public only because `BTPaymentActionsClient.submitForPaymentAction` and
+/// public subclasses (e.g. `BTCardPaymentActionRequest`) both require it to be.
+public class BTPaymentActionRequest: NSObject {
+    
+    /// The ID of the Payment Action to submit this payment method for.
+    public let paymentActionID: String
```
jaxdesmarais: Same here as other public class, I think we can make the property internal and the init public? We should also add `objcMembers` to the public types.

Sources/BraintreePaymentActions/BTPaymentActionResult.swift
```diff
@@ -1,8 +1,8 @@
 import Foundation
 
-/// The internal, shared result of a Payment Actions `setPaymentMethod` GraphQL call. Each payment-method-specific
+/// The result of a Payment Actions `setPaymentActionPaymentMethod` GraphQL call. Each payment-method-specific
 /// client maps this into its own public result type, surfacing fields only relevant to that payment method.
-struct BTPaymentActionResult {
+public struct BTPaymentActionResult {
```
jaxdesmarais: Should this be a class with NSObject/objcMembers?

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
```diff
@@ -0,0 +1,100 @@
...
+    // MARK: - Internal Properties
+    
+    /// Exposed for testing to get the instance of BTAPIClient
+    var apiClient: BTAPIClient
+    
+    // MARK: - Initializer
+    
+    /// Creates a Payment Actions Client.
+    /// - Parameter authorization: A valid client token or tokenization key used to authorize API calls.
+    public init(authorization: String) {
```
jaxdesmarais, line 19: We will need to expose this for objc

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
```diff
@@ -0,0 +1,100 @@
...
+    
+    // MARK: - Public Methods
+    
+    /// Submits a payment method to a Payment Action.
+    /// - Parameters:
+    ///    - request: The payment method details to submit — e.g. a `BTCardPaymentActionRequest`.
+    ///    - completion: A completion block that is invoked when the submission has completed. If it succeeds,
+    ///    `status` will contain the resulting `BTPaymentActionStatus` and `error` will be `nil`; if it fails,
+    ///    `error` will describe the failure.
+    @objc(submitForPaymentAction:completion:)
```
jaxdesmarais:
```suggestion
    @objc(submitForPaymentActionWithRequest:completion:)
```

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
```diff
@@ -0,0 +1,100 @@
...
+        try await submit(request)
+    }
+    
+    // MARK: - Internal Methods
+    
+    /// Submits a payment method to the `setPaymentActionPaymentMethod` GraphQL mutation.
+    /// - Parameter request: any payment-method-specific request subclassing `BTPaymentActionRequest`.
+    /// - Returns: the resulting `BTPaymentActionStatus`.
+    /// - Throws: the underlying error from the network layer or GraphQL response.
+    func submit(_ request: BTPaymentActionRequest) async throws -> BTPaymentActionStatus {
```
jaxdesmarais: It looks like on Android we return the entire `PaymentActionResult` not just the status. Can we align iOS with Android?

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
```diff
@@ -0,0 +1,100 @@
...
+    
+    /// Submits a payment method to the `setPaymentActionPaymentMethod` GraphQL mutation.
+    /// - Parameter body: the GraphQL request body to submit, encoding the payment method
+    ///   details and the fields to return on the `paymentAction` selection set.
+    /// - Returns: a `BTPaymentActionResult` containing the Payment Action `id` and `status`.
+    /// - Throws: the underlying error from the network layer or GraphQL response.
+    func setPaymentActionPaymentMethod<Body: BTGraphQLEncodableBody>(
+        _ body: Body
+    ) async throws -> BTPaymentActionResult {
+
```
jaxdesmarais:
```suggestion
```

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
```diff
@@ -0,0 +1,100 @@
...
+            
+            let paymentActionJSON: BTJSON = body?["data"]["setPaymentActionPaymentMethod"]["paymentAction"] ?? BTJSON()
+            
+            // TODO: Verify and ensure these error types are correct in a later PR.
+            guard let paymentActionID = paymentActionJSON["id"].asString(), !paymentActionID.isEmpty else {
+                throw BTPaymentActionError.missingID
+            }
+            guard let statusString = paymentActionJSON["status"].asString(), !statusString.isEmpty else {
+                throw BTPaymentActionError.missingStatus
+            }
```
jaxdesmarais, line 85: I think a couple line breaks here could help with readability:

```suggestion
            }

            guard let statusString = paymentActionJSON["status"].asString(), !statusString.isEmpty else {
                throw BTPaymentActionError.missingStatus
            }

```

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
agedd: yep updated this, in 7ffd691637def8a19e65c13c26dded2e5d86ea26

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
```diff
@@ -0,0 +1,124 @@
...
+        try await submit(request)
+    }
+    
+    // MARK: - Internal Methods
+    
+    /// Submits a payment method to the `setPaymentActionPaymentMethod` GraphQL mutation.
+    /// - Parameter request: any payment-method-specific request subclassing `BTPaymentActionRequest`.
+    /// - Returns: the resulting `BTPaymentActionStatus`.
+    /// - Throws: the underlying error from the network layer or GraphQL response.
+    func submit(_ request: BTPaymentActionRequest) async throws -> BTPaymentActionResult {
```
mzlangreder: Are we using submit from other functions? Could this method body be folded in to the `submitForPaymentAction` above?

Sources/BraintreePaymentActions/BTPaymentActionResult.swift
```diff
@@ -0,0 +1,25 @@
...
+/// The result of submitting a payment method to a Payment Action, mapped from the Payment Action's
+/// lifecycle status into the next action the merchant should take.
+@objcMembers public class BTPaymentActionResult: NSObject {
+    
+    // MARK: - Internal Properties
+    
+    let type: BTPaymentActionResultType
+    let id: String
+    let serverAction: BTServerAction
+
```
mzlangreder, line 12: Should these be public?

mzlangreder (approved):
Looking good, just a couple questions.

Sources/BraintreePaymentActions/BTPaymentActionResult.swift
agedd: ah yep, keeping them internal to match the current v7 style where the properties are internal and injected into a public init - scanning through the codebase, i've identified other places where the old pattern is used so will create tickets for those types.

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
agedd: ah good catch, i've folded this into `submitForPaymentAction` above

PR #1844, opened 2026-08-10
Source edit_billing_agreement_jwt from the client token for PayPal Edit FI

saurabh-P9525:
Source the `edit_billing_agreement_jwt` for the PayPal Edit FI flow from the **client token** and thread it into `create_payment_resource`. The JWT is never merchant-supplied, and the flow is **scoped to the `BraintreePayPalSavedPaymentMethod` module** rather than exposed on the public API.

- The JWT (`paymentMethodIDJWT`) is extracted from the client token and exposed through a dedicated `ClientTokenAuthorizationProviding` protocol, so it stays off the general `ClientAuthorization`.
- `editBillingAgreement` is **not** merchant-facing. It is an internal flag on `BTPayPalCheckoutRequest`, enabled only by the Edit FI module through an `@_spi` extension — a merchant not using that module cannot reach a property that would do nothing for them.
- `edit_billing_agreement_jwt` is only sent when the module opted in **and** the client token carries the JWT — otherwise the key is omitted (not null), so existing flows send byte-identical bodies.

### Summary of changes
- **`ClientTokenAuthorizationProviding`** (new): declares `paymentMethodIDJWT: String? { get }`. Only `BTClientToken` conforms, so `TokenizationKey`, `InvalidAuthorization`, and test mocks are untouched and `ClientAuthorization` keeps no client-token-specific field.
- **`BTClientToken`**: extracts the JWT from the client-token JSON via `json["paymentMethodIdJwt"].asString()`, matching how `configURL` is read.
- **`BTPayPalCheckoutRequest`**: `editBillingAgreement` is an internal `Bool` and is **not** a parameter on either public initializer. It is set only via `@_spi(BraintreePayPalSavedPaymentMethod) enableEditBillingAgreement()`, declared `@nonobjc` so it stays out of the generated Objective-C header (the class is `@objcMembers`). Adds a checkout-only `encodedPostBodyWith(...)` overload taking `paymentMethodIDJWT`, alongside the one satisfying `BTPayPalRequest`.
- **`BTPayPalClient`**: reads the JWT via `authorization as? ClientTokenAuthorizationProviding` and passes it to the checkout overload; vault requests use the protocol signature unchanged.
- **`BTPayPalRequest` protocol + `BTPayPalVaultRequest`**: unchanged — the JWT is checkout-only, so it stays off the shared protocol and vault never sees it.
- **`PayPalCheckoutPOSTBody`**: adds `edit_billing_agreement_jwt` only when `editBillingAgreement == true` && `paymentMethodIDJWT != nil`.
- **Tests**: cover opt-in + JWT present (sent), opt-in false (omitted), and opt-in + JWT nil (omitted), plus `BTClientToken` parsing with and without the claim.

### Notes for reviewers
- **`@_spi` has no precedent in this repo.** It is the only mechanism that is cross-module *and* hidden from merchants: `internal` is unreachable from `BraintreePayPalSavedPaymentMethod`, and `public` would expose the knob to everyone. Flagging it explicitly for a codeowners call.
- **`enableEditBillingAgreement()` has no caller in this PR.** Its only consumer is `BraintreePayPalSavedPaymentMethod` (#1850), which lands on top of this branch. The POST-body tests here exercise the internal flag directly via `@testable`.
- Verified against the generated `BraintreePayPal-Swift.h`: `BTPayPalCheckoutRequest` is still exported to Objective-C, while `editBillingAgreement` / `enableEditBillingAgreement` are absent.

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used Claude (guided by the repo's `CLAUDE.md`) to extract the JWT from the client token via `ClientAuthorization`/`BTClientToken`, thread it through the PayPal request/client/POST body with the opt-in gate, update the protocol + vault conformance, and write the unit tests.

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @saurabh-P9525

----
### Inner Source Process

### Inner Source Checklist
- [x] Provide steps to test the flows changed, if applicable in the summary
- [ ] Demo video of the functionality, if applicable
- [ ] All upstream dependencies are merged in and this PR can be released at any time; PRs should not be opened until this is true


Sources/BraintreeCore/Authorization/BTClientToken.swift
```diff
@@ -38,6 +41,7 @@ class BTClientToken: ClientAuthorization {
         self.bearer = authorizationFingerprint
         self.configURL = configURL
         self.originalValue = clientToken
+        self.paymentMethodIDJWT = json["paymentMethodIDJWT"].asString()
     }
```
rainaarya: This should be `paymentMethodIdJwt` as per decoded JSON

saurabh-P9525: addressed

UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift
```diff
@@ -244,6 +244,47 @@ class BTPayPalCheckoutRequest_Tests: XCTestCase {
         XCTAssertEqual(parameters["cancel_url"] as? String, "sdk.ios.braintree://onetouch/v1/cancel")
     }
 
+    func testParametersWithConfiguration_whenEditBillingAgreementTrueAndJWTPresent_includesEditBillingAgreementJWT() {
+        let request = BTPayPalCheckoutRequest(amount: "1", editBillingAgreement: true)
```
rainaarya: Right now the tests pass the JWT straight into `encodedPostBodyWith`, (without touching BTClientToken), so tests don't verify that the JWT is read from client token, or that it reaches the POST body through BTPayPalClient.

Can u also add tests for these? like a BTClientToken test that a client token containing paymentMethodIdJwt populates the property, and one without it returns nil.

and a BTPayPalClient test (using mockAPIClient) that with editBillingAgreement: true and a client token carrying the JWT, edit_billing_agreement_jwt actually shows up in the create_payment_resource body .

rainaarya:
Let's also keep the `CHANGELOG` updated as we are adding a new param (`editBillingAgreement`) bool

UnitTests/BraintreePayPalTests/BTPayPalCheckoutRequest_Tests.swift
saurabh-P9525: addressed

rainaarya (approved):

binoybt (approved):
LGTM - Binoy

Sources/BraintreeCore/Authorization/ClientAuthorization.swift
```diff
@@ -17,6 +17,18 @@ public protocol ClientAuthorization {
...
+    /// PayPal Edit FI flow to reference an existing billing agreement. `nil` when not present
+    /// (e.g. tokenization keys, or client tokens that don't include it).
+    var paymentMethodIDJWT: String? { get }
+}
+
+/// :nodoc: Default so existing conformers are unaffected; only client tokens provide a value.
+@_documentation(visibility: private)
+extension ClientAuthorization {
+
+    public var paymentMethodIDJWT: String? { nil }
```
jaxdesmarais: Why is this needed and what are the alternatives?

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -91,6 +92,7 @@ import BraintreeCore
         amount: String,
         enablePayPalAppSwitch: Bool,
         userAuthenticationEmail: String? = nil,
+        editBillingAgreement: Bool = false,
```
jaxdesmarais: We are missing documentation for this parameter. It's also unclear why it's necessary. Merchants will opt into this flow by using the new module and cannot use this feature outside of the UI based flow.

Sources/BraintreeCore/Authorization/ClientAuthorization.swift
saurabh-P9525: `ClientAuthorization` is adopted by `BTClientToken`, `TokenizationKey`, and `InvalidAuthorization`. Since only client tokens contain a JWT, the protocol requirement defaults to `nil` through an extension, and `BTClientToken` overrides it with the actual value. The alternative would be to implement the property on every conforming type and return `nil` where it doesn’t apply.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
saurabh-P9525: Even though using the module implies an **edit-FI intent**, the module itself has no way to set that intent on the merchant’s behalf. It needs to be explicitly specified on the request object provided by the merchant.

Sources/BraintreeCore/Authorization/ClientAuthorization.swift
jaxdesmarais: This has code smell though due to scoping. Why not have a new protocol like `ClientTokenAuthorizationProviding`?

saurabh-P9525: Sounds good.  I’ll move it to `ClientTokenAuthorizationProviding`. The main tradeoff is that we replace a static property read with a runtime `as?` cast at the call site, and introduce a new public protocol plus mock conformance in tests for a single field and caller. That said, it does keep the responsibility more cleanly scoped.

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
jaxdesmarais: Removed internal documentation as we are in public GitHub and cannot per security. We should instead extend the class as a merchant not using this module could access this property and it would have no functionality, which would be a poor experience.

Sources/BraintreeCore/Authorization/BTClientToken.swift
```diff
@@ -38,6 +41,7 @@ class BTClientToken: ClientAuthorization {
         self.bearer = authorizationFingerprint
         self.configURL = configURL
         self.originalValue = clientToken
+        self.paymentMethodIDJWT = json["paymentMethodIdJwt"].asString()
```
buzzamus, line 44: I just want to double check that the JSON key is actually `paymentMethodIdJwt` since this is different than the variable name. This seems pretty open to a typing error if we aren't careful

saurabh-P9525: yes that correct json key to extract paymentMethodIdJwt from client token

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
saurabh-P9525: Updated in `a623651c` — please take a look. `editBillingAgreement` is no longer merchant-facing: removed from both public inits, property stays internal, and it's set only through an SPI-gated extension that the Edit FI module calls on the merchant's behalf.

Sources/BraintreeCore/Authorization/BTClientToken.swift
```diff
@@ -38,6 +48,7 @@ class BTClientToken: ClientAuthorization {
         self.bearer = authorizationFingerprint
         self.configURL = configURL
         self.originalValue = clientToken
+        self.paymentMethodIDJWT = json[JSONKeys.paymentMethodIDJWT].asString()
```
agedd: i think we can refactor this to something like json["paymentMethodIdJwt"].asString() the same way we read and extract the `configURL` to be aligned with the style we use in our SDK

Sources/BraintreeCore/Authorization/BTClientToken.swift
```diff
@@ -1,7 +1,14 @@
 import Foundation
 
 /// An authorization string used to initialize the Braintree SDK
-class BTClientToken: ClientAuthorization {
+class BTClientToken: ClientAuthorization, ClientTokenAuthorizationProviding {
+
+    // MARK: - Constants
+
+    private enum JSONKeys {
```
agedd: nit: can delete this and read directly from the json object

Sources/BraintreePayPal/BTPayPalRequest.swift
```diff
@@ -30,7 +30,8 @@ protocol BTPayPalRequest {
         configuration: BTConfiguration,
         isPayPalAppInstalled: Bool,
         universalLink: URL?,
-        fallbackURLScheme: String?
+        fallbackURLScheme: String?,
+        paymentMethodIDJWT: String?
```
agedd: question: do we need to pass in `paymentMethodIDJWT` into the respective POST body

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
```diff
@@ -121,7 +121,8 @@ import BraintreeCore
         configuration: BTConfiguration,
         isPayPalAppInstalled: Bool = false,
         universalLink: URL? = nil,
-        fallbackURLScheme: String? = nil
+        fallbackURLScheme: String? = nil,
+        paymentMethodIDJWT: String? = nil
```
agedd: same question as above

Sources/BraintreeCore/Authorization/BTClientToken.swift
saurabh-P9525: Done , now reads inline, matching how configURL is parsed above:
`self.paymentMethodIDJWT = json["paymentMethodIdJwt"].asString()`

Sources/BraintreeCore/Authorization/BTClientToken.swift
saurabh-P9525: Removed the JSONKeys enum and read directly from the JSON object.

Sources/BraintreePayPal/BTPayPalRequest.swift
saurabh-P9525: it's only needed for checkout, since `edit_billing_agreement_jwt` goes in the `create_payment_resource` body. Removed it from the protocol; `BTPayPalCheckoutRequest` now has a checkout-only overload that takes it, and `BTPayPalClient` passes it there.

Sources/BraintreePayPal/BTPayPalVaultRequest.swift
saurabh-P9525: Removed — vault never used it.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -462,12 +462,24 @@ import BraintreeDataCollector
...
 
-        let parameters = request.encodedPostBodyWith(
-            configuration: configuration,
-            isPayPalAppInstalled: self.application.isPayPalAppInstalled(),
-            universalLink: self.universalLink,
-            fallbackURLScheme: self.fallbackURLScheme
-        )
+        let parameters: Encodable
+
+        if let checkoutRequest = request as? BTPayPalCheckoutRequest {
```
agedd, line 467: design nit: instead of type-casting on the request type here, can we rather expand the `BTPayPalRequest` protocol to always take in `paymentMethodIDJWT` and let each conforming type decide what to do with it i.e.: 

```suggestion
 // BTPayPalRequest.swift
protocol BTPayPalRequest {
    ...
    func encodedPostBodyWith(
        configuration: BTConfiguration,
        isPayPalAppInstalled: Bool,
        universalLink: URL?,
        fallbackURLScheme: String?,
        paymentMethodIDJWT: String?
    ) -> Encodable
}
```

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPal
+  * Send `edit_billing_agreement_jwt` on `create_payment_resource` for the PayPal Edit FI flow, sourced from the client token
```
agedd: nit: 
```suggestion
  * Send the `edit_billing_agreement_jwt` sourced from the client token for the PayPal Edit FI beta flow"
```

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -90,6 +90,37 @@ class BTPayPalClient_Tests: XCTestCase {
...
+        }
+
+        wait(for: [expectation], timeout: 1)
+
+        XCTAssertEqual("v1/paypal_hermes/create_payment_resource", mockAPIClientWithJWT.lastPOSTPath)
+        guard let lastPostParameters = mockAPIClientWithJWT.lastPOSTParameters else { XCTFail(); return }
+
+        XCTAssertEqual(lastPostParameters["edit_billing_agreement_jwt"] as? String, "edit-fi-jwt")
+    }
+
```
agedd, line 123: nit: could we add a sibling test here - i.e. same setup but with `MockAPIClient(authorization: BTTokenizationKey(...))` instead of a client token

Sources/BraintreePayPal/BTPayPalClient.swift
saurabh-P9525: Happy to make this change. I just wanted to confirm the direction, since it would bring the signature back closer to the earlier implementation.

Following the earlier comment on [[a623651](https://github.com/braintree/braintree_ios/pull/1844#discussion_r3807136905)], `paymentMethodIDJWT` was removed from `BTPayPalRequest` based on the question around whether it needed to be passed there. That change is also what introduced the type cast at the call site.

Adding it back would simplify the call site and restore a uniform signature, which I’m happy to do if that’s the preferred direction. and also`BTPayPalVaultRequest` will again have to introduce `paymentMethodIDJWT: String? = nil` params as it confirms to BTPayPalRequest protocol

CHANGELOG.md
saurabh-P9525: Applied, thanks.

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
saurabh-P9525: Added a test to verify `edit_billing_agreement_jwt` is omitted when using a tokenization key, since `BTTokenizationKey` doesn’t conform to `ClientTokenAuthorizationProviding`.

Sources/BraintreePayPal/BTPayPalClient.swift
agedd: good question, thinking about it a bit more, let's keep as-is with the type cast here. the vault request never needing this field is actually a useful signal. imo the cast doesn't read as cleanly, but it's clear about which type requires the JWT so let's keep this as-is. thanks for confirming

agedd (approved):

Sources/BraintreePayPal/BTPayPalCheckoutRequest.swift
```diff
@@ -216,13 +217,44 @@ import BraintreeCore
...
+extension BTPayPalCheckoutRequest {
+
+    /// :nodoc: Not part of the public API. `@nonobjc` keeps this out of the generated Objective-C header.
+    @_documentation(visibility: private)
+    @_spi(BraintreePayPalSavedPaymentMethod)
+    @nonobjc
+    public func enableEditBillingAgreement() {
+        editBillingAgreement = true
+    }
+}
```
buzzamus, line 260: Where is this function actually called at? I noticed the notes mention it is in a later PR, but I don't see it there either. If it is going to be there I think that is fine, but 1) wanted to double check and 2) in order to avoid the confusion going forward I think we should probably just put code like that in the PR using it to avoid having to check multiple PRs

saurabh-P9525: You’re right that the call site isn’t visible yet, the changes will be in follow up PR (UI + integration PR) stacked on top of #1850. #1850 is scoped to the API and data models, while the integration PR is where `BraintreePayPalSavedPaymentMethod` builds the `BTPayPalCheckoutRequest` and calls `BTPayPalClient`, making it the only consumer of this SPI.

We split the PR stack along module boundaries to keep each change focused.

buzzamus (approved):

mzlangreder (approved):

PR #1850, opened 2026-08-13
Add View/Edit FI and credit messaging data layer to BraintreePayPalSavedPaymentMethod

rainaarya:
### Summary of changes

Adds the data layer for the View/Edit FI component in `BraintreePayPalSavedPaymentMethod` (funding instrument fetch and the credit messaging fetch).

**Funding instrument fetch**
- `BTPayPalSavedPaymentMethodClient.fetchPaymentMethod(fundingInstrumentType:orderID:merchantAccountID:)` calls the BT GQL `paypalFundingInstrumentDetails` GraphQL query over the existing `.graphQLAPI` rail.
- `STICKY_FI` reads the payment method ID JWT from `apiClient.authorization.paymentMethodIDJWT`; `FI_FROM_APPROVED_CHECKOUT` requires an `orderID`. Exactly one identity field is sent, as the API requires.
- `BTPayPalSavedPaymentMethodSummary` / `BTPayPalSavedPaymentMethod` / `BTPayPalPayer` model the instrument and display-only payer responses.

**Credit messaging fetch**
- `BTPayPalSavedPaymentMethodClient.fetchCreditPresentmentMessages(amount:currencyCode:)` POSTs to `/v2/credit/fetch-presentment-messages` over the existing `.payPalAPI` rail.
- `BTPayPalCreditMessagingResult` returns the main, disclaimer and action content blocks in the order PayPal returns them, plus the impression beacon.

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot
- [x] Claude
- [ ] Other (Type Name Here)

**How was AI used?**
Code generation, unit tests, and refactoring. The backend contracts were verified against the Atmosphere schema and the Offer Presentment API dev docs. AI drafted the client, models and tests.

- [ ] less than 30%
- [ ] 30 - 60%
- [x] 60 - 100%

### Checklist
- [ ] Added a changelog entry

### Authors
- @rainaarya

----
### Inner Source Process

### Inner Source Checklist
- [ ] Provide steps to test the flows changed, if applicable in the summary
- [ ] Demo video of the functionality, if applicable — N/A, no UI in this PR
- [ ] All upstream dependencies are merged in and this PR can be released at any time


Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,117 @@
...
+
+        let (body, _) = try await apiClient.post("", parameters: parameters, httpType: .graphQLAPI)
+
+        guard let body else {
+            throw BTPayPalSavedPaymentMethodError.emptyBodyReturned
+        }
+
+        guard let summary = BTPayPalSavedPaymentMethodSummary(json: body["data"]["paypalFundingInstrumentDetails"]) else {
+            throw BTPayPalSavedPaymentMethodError.failedToParseSummary
+        }
```
saurabh-P9525, line 85: If `paypalFundingInstrumentDetails` comes back `null` (or an object with empty `payer` + empty `paymentMethods`) for a buyer with nothing to display, this throws `failedToParseSummary`. That's a valid "no FI" state the UI renders as brand-only/hidden — it shouldn't be an error. Can we confirm the empty-result shape and return an empty `BTPayPalSavedPaymentMethodSummary` for it instead of throwing? As written, "`nothing to show`" is indistinguishable from a genuine parse failure.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodSummary.swift
```diff
@@ -0,0 +1,31 @@
...
+
+    /// The funding instruments PayPal can charge. The first entry is the one that will be charged.
+    let paymentMethods: [BTPayPalSavedPaymentMethod]
+
+    // MARK: - Initializer
+
+    init?(json: BTJSON) {
+        guard json.isObject else {
+            return nil
+        }
```
saurabh-P9525, line 25: init? returning nil for a non-object is what makes the client throw on the empty case. Consider an explicit .empty (payer nil, paymentMethods []) so "no FI" is representable without failing.

Sources/BraintreePayPalSavedPaymentMethod/Models/BTPayPalCreditMessagingResult.swift
```diff
@@ -0,0 +1,50 @@
...
+
+        guard preferredMessage.isObject else {
+            return nil
+        }
+
+        let content = preferredMessage["content"]
+
+        self.messageID = preferredMessage["id"].asString()
+        self.messageType = preferredMessage["type"].asString()
+        self.mainItems = content["main_items"].asArray()?.map(BTPayPalCreditMessageItem.init) ?? []
```
saurabh-P9525: init? succeeds as long as preferred_message is an object, even with empty main_items. Consider requiring mainItems non-empty ?

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: The empty case already works, `{"payer": null, "paymentMethods": []}` is an object, so it parses to an empty summary and doesn't throw error. 

You can check the `testInit_whenPayloadIsEmpty_returnsEmptySummary` test for reference.
Error will fire only on a malformed response.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodSummary.swift
rainaarya: Similar to above, `.empty` is already the normal path

Sources/BraintreePayPalSavedPaymentMethod/Models/BTPayPalCreditMessagingResult.swift
rainaarya: updated

saurabh-P9525 (approved):

Sources/BraintreePayPalSavedPaymentMethod/Models/BTPayPalCreditMessagingResult.swift
```diff
@@ -0,0 +1,51 @@
...
+
+    /// Parses the preferred message out of a `v2/credit/fetch-presentment-messages` response.
+    /// - Returns: `nil` when PayPal has no message to show for this buyer.
+    init?(json: BTJSON) {
+        let preferredMessage = json["messages"][0]["preferred_message"]
+        let content = preferredMessage["content"]
+        let mainItems = content["main_items"].asArray()?.map(BTPayPalCreditMessageItem.init) ?? []
+
+        // Reporting success with no message text would fire the impression beacon for a message the buyer never saw.
+        guard !mainItems.isEmpty else {
```
saurabh-P9525: This guard checks the item count, not whether the message has any copy ... so it doesn't quite deliver what the commit title promises. Since BTPayPalCreditMessageItem.init is non-failable, every array entry survives : 
```
guard mainItems.contains(where: { $0.text?.isEmpty == false }) else {
    return nil
}
```

Sources/BraintreePayPalSavedPaymentMethod/Models/BTPayPalCreditMessageItem.swift
```diff
@@ -0,0 +1,44 @@
...
+
+    /// A non-unique identifier for the block, for example `"periodic_payment_count"` or `"paypal_logo"`.
+    let name: String?
+
+    /// Whether `clickURL` may be loaded in an embedded web view rather than an external browser.
+    let isEmbeddable: Bool
+
+    // MARK: - Initializer
+
+    init(json: BTJSON) {
```
saurabh-P9525: Nit: this is the only model in the module with a non-failable init ... `BTPayPalSavedPaymentMethod`, `BTPayPalPayer`, and `BTPayPalSavedPaymentMethodSummary` are all init? with guard json.isObject. For consistency can we do same here ?

Sources/BraintreePayPalSavedPaymentMethod/Models/BTPayPalCreditMessagingResult.swift
rainaarya: addressed

Sources/BraintreePayPalSavedPaymentMethod/Models/BTPayPalCreditMessageItem.swift
rainaarya: addressed, made it init? and changed `.map` to `.compactMap`

saurabh-P9525 (approved):

binoybt (approved):
Lgtm

jaxdesmarais (changes requested):
@rainaarya please break this PR down into smaller chunks as 400 should be the limit for PRs to provide accurate reviews. Additionally nothing in this PR is Obj-C compatible or public for merchants to use. This PR is also missing docstrings and not following existing patterns in several places. Moving this back to draft in the meantime, please reopen when ready.

rainaarya:
> @rainaarya please break this PR down into smaller chunks as 400 should be the limit for PRs to provide accurate reviews. Additionally nothing in this PR is Obj-C compatible or public for merchants to use. This PR is also missing docstrings and not following existing patterns in several places. Moving this back to draft in the meantime, please reopen when ready.

Hi @jaxdesmarais,

We are following Ehud's guidelines to limit the no. of open PRs for this deliverable and we aligned earlier that it will increase overall LOC per PR. Happy to chat more on Slack if you have specific questions. 

This PR is focused on the data models and API calls which are internal to the module. The next PR which will cover UI layer will expose the functionality for the merchant (public) and will invoke the APIs introduced in this PR.

All types and stored properties have  `///` doc-strings. I have further added docstrings in  `BTPayPalCreditMessageItemType`/`BTPayPalSavedPaymentMethodType`  enum cases and the `init?(json:)` parsers. Let us know if there are any additional places to add in.

We have tried to follow the existing codebase patterns. Let us know any specific places that you think may require alignment with code patterns. We have modelled this based on the existing modules, (`BTGraphQLEncodableBody` + nested Variables, POST-body naming, the standard error/enum conformances, `Models/` layout, etc.).

UnitTests/BraintreePayPalSavedPaymentMethodTests/BTPayPalSavedPaymentMethodClient_Tests.swift
```diff
@@ -0,0 +1,203 @@
...
+    override func tearDown() {
+        sut = nil
+        mockAPIClient = nil
+        super.tearDown()
+    }
+
+    // MARK: - Request
+
+    func testFetchPaymentMethod_whenStickyFI_postsQueryWithPaymentMethodIDJWT() async throws {
+        _ = try? await sut.fetchPaymentMethod(fundingInstrumentType: .stickyFI, merchantAccountID: "fake-merchant-account-id")
```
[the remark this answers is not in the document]
rainaarya: It is intentional, with no canned response the parse step throws after the POST, and this test only asserts request shape.
 If no request went out, `XCTAssertEqual(lastPOSTPath, "")` and `XCTUnwrap(lastPOSTParameters)` below both fail, so it can't pass silently.

UnitTests/BraintreePayPalSavedPaymentMethodTests/BTPayPalCreditMessaging_Tests.swift
```diff
@@ -0,0 +1,281 @@
...
+    override func tearDown() {
+        sut = nil
+        mockAPIClient = nil
+        super.tearDown()
+    }
+
+    // MARK: - Request
+
+    func testFetchCreditPresentmentMessages_postsTreatmentARequestToThePayPalAPI() async throws {
+        _ = try? await sut.fetchCreditPresentmentMessages(amount: "55.00", currencyCode: "USD")
```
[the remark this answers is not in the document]
rainaarya: Same as above, the throw would happen after the POST, and the request assertions below still fail if nothing was sent.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,117 @@
...
+
+    /// Exposed for testing to get the instance of BTAPIClient
+    var apiClient: BTAPIClient
+
+    // MARK: - Initializer
+
+    /// Creates a `BTPayPalSavedPaymentMethodClient`
+    /// - Parameter authorization: A client token generated with the buyer's payment method ID
+    init(authorization: String) {
+        self.apiClient = BTAPIClient(authorization: authorization)
```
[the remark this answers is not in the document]
rainaarya: `BTAPIClient.init(authorization:)` is non-failable, so this compiles. It is similar to how other modules initialize it.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodSummary.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
```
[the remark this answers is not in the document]
rainaarya: This is the repo-wide pattern already (e.g. BTPayPalAccountNonce). Importing unconditionally may break the CocoaPods builds.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodSummary.swift
```diff
@@ -0,0 +1,30 @@
...
+    let payer: BTPayPalPayer?
+
+    /// The funding instruments PayPal can charge. The first entry is the one that will be charged.
+    let paymentMethods: [BTPayPalSavedPaymentMethod]
+
+    // MARK: - Initializer
+
+    /// Parses the `paypalFundingInstrumentDetails` field of a `PaypalFundingInstrumentDetails` response.
+    /// - Returns: `nil` when the field is not an object.
+    init?(json: BTJSON) {
```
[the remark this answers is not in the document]
rainaarya: Same as above, it is a repo pattern.

Sources/BraintreePayPalSavedPaymentMethod/Models/PayPalCreditMessagingPOSTBody.swift
```diff
@@ -0,0 +1,55 @@
+import Foundation
+
+// swiftlint:disable nesting
+/// The POST body for `v2/credit/fetch-presentment-messages`
```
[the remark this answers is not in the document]
rainaarya: there's already a matching `// swiftlint:enable` nesting after the type, and the file contains only that one type.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodError.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+
+enum BTPayPalSavedPaymentMethodError: Int, Error, CustomNSError, LocalizedError, Equatable {
```
jaxdesmarais: This does not follow the existing error pattern and is not exposed to merchants at all so they cannot access our error. Errors and their types must all be public.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,117 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+/// Fetches what to display for a buyer's vaulted PayPal payment method: the funding instrument PayPal will charge, and the
+/// Pay Later message that accompanies it.
+final class BTPayPalSavedPaymentMethodClient {
```
jaxdesmarais, line 9: Who constructs this? It doesn't need to be public?

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,117 @@
...
+    ///   - orderID: The approved checkout order ID. Required for `fiFromApprovedCheckout` and ignored otherwise.
+    ///   - merchantAccountID: The merchant account the funding instrument is fetched against.
+    /// - Returns: A `BTPayPalSavedPaymentMethodSummary` describing what to display for the buyer
+    /// - Throws: A `BTPayPalSavedPaymentMethodError` if the request cannot be built or the response cannot be parsed
+    func fetchPaymentMethod(
+        fundingInstrumentType: BTPayPalFundingInstrumentFetchType,
+        orderID: String? = nil,
+        merchantAccountID: String? = nil
+    ) async throws -> BTPayPalSavedPaymentMethodSummary {
+        guard apiClient.authorization.type == .clientToken else {
```
jaxdesmarais, line 45: Let's add the client token requirement to the docstrings

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        let parameters = PayPalFundingInstrumentDetailsGraphQLBody(
+            fundingInstrumentType: fundingInstrumentType,
+            paymentMethodIDJWT: paymentMethodIDJWT,
+            orderID: resolvedOrderID,
+            merchantAccountID: merchantAccountID
+        )
+
+        let (body, _) = try await apiClient.post("", parameters: parameters, httpType: .graphQLAPI)
```
jaxdesmarais, line 77: We are not handling errors returned here? Should this be a do/catch? Is there a case where we need to check the json with `asError` like we need to do with some other GrpahQL APIs?

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,117 @@
...
+    ///   - amount: The order amount the message is calculated from, for example `"55.00"`.
+    ///   - currencyCode: The ISO-4217 currency code for `amount`, for example `"USD"`.
+    /// - Returns: A `BTPayPalCreditMessagingResult` describing the message to render
+    /// - Throws: A `BTPayPalSavedPaymentMethodError` if the request cannot be built or PayPal returns no message
+    /// - Note: The message is additive. Callers are expected to hide the row when this throws rather than fail checkout.
+    func fetchCreditPresentmentMessages(
+        amount: String,
+        currencyCode: String
+    ) async throws -> BTPayPalCreditMessagingResult {
+        guard apiClient.authorization.type == .clientToken else {
```
jaxdesmarais, line 105: Lets also add this requirement to the docstring

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,117 @@
...
+            throw BTPayPalSavedPaymentMethodError.invalidAuthorization
+        }
+
+        let parameters = PayPalCreditMessagingPOSTBody(amount: amount, currencyCode: currencyCode)
+
+        let (body, _) = try await apiClient.post(
```
jaxdesmarais, line 111: Same question here, we are never handing errors

jaxdesmarais (changes requested):
I am a bit confused by the models directory here. The models should be the API layer and follow a pattern like `GenerateCustomerRecommendationsGraphQLBody`, `PayPalAccountPOSTEncodable`, etc and not prefixed with `BT`. These are all prefixed though with `BT` which is the pattern using for public types exposed to merchants and _would not_ live in the models directory. But none of them are exposed publicly. I am not sure which these are supposed to be, but either way need to be updated to match existing patterns.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodError.swift
rainaarya: Made the enum and its members public

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: Client needs to be internal. Nothing constructs it in this PR, but the future PRs will use this client for fetch/edit actions (i.e ViewModel of the UI layer PR).

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: Added a note in the initializer and both fetch methods.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: added

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: Both rails throw before post returns (GraphQL on any `errors[]`, BTHTTP on ≥400 and via `asError()` internally), and our parsers are failable, so a broken node throws `failedToParseSummary`

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: Same here, where BTHTTP throws before we get the body, and `BTPayPalCreditMessagingResult.init?` returns nil rather than building an empty result, so we throw `missingPreferredMessage`.

rainaarya:
> I am a bit confused by the models directory here. The models should be the API layer and follow a pattern like `GenerateCustomerRecommendationsGraphQLBody`, `PayPalAccountPOSTEncodable`, etc and not prefixed with `BT`. These are all prefixed though with `BT` which is the pattern using for public types exposed to merchants and _would not_ live in the models directory. But none of them are exposed publicly. I am not sure which these are supposed to be, but either way need to be updated to match existing patterns.

Have moved the response models to the module root so `Models/` holds only the API bodies. 

We noticed that `BTPayPalReturnURL`, `BTPayPalApprovalURLParser`, `BTPayPalFundingSource`, `BTVenmoAppSwitchReturnURL` etc, were also internal but BT-prefixed, so had kept ours consistent with those. If you prefer, we will drop the prefix if you'd rather these read as clearly internal only.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,122 @@
...
+    /// - Note: Requires a client token. Throws `BTPayPalSavedPaymentMethodError.invalidAuthorization` when initialized
+    ///   with a tokenization key, which carries no `paymentMethodIdJwt`.
+    func fetchPaymentMethod(
+        fundingInstrumentType: BTPayPalFundingInstrumentFetchType,
+        orderID: String? = nil,
+        merchantAccountID: String? = nil
+    ) async throws -> BTPayPalSavedPaymentMethodSummary {
+        guard apiClient.authorization.type == .clientToken else {
+            throw BTPayPalSavedPaymentMethodError.invalidAuthorization
+        }
```
buzzamus, line 47: There is some duplication in this file with the guard blocks. This guard/throw block is identical to the one on lines 100-102. There are also empty body blocks on lines 76-78 that match 112-114. Could we create some sort of helper functions to remove this repeated code in case exception names change?

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,122 @@
...
+    /// - Note: The message is additive. Callers are expected to hide the row when this throws rather than fail checkout.
+    /// - Note: Requires a client token. Throws `BTPayPalSavedPaymentMethodError.invalidAuthorization` when initialized
+    ///   with a tokenization key, since the PayPal API rail authenticates with the client token's bearer.
+    func fetchCreditPresentmentMessages(
+        amount: String,
+        currencyCode: String
+    ) async throws -> BTPayPalCreditMessagingResult {
+        guard apiClient.authorization.type == .clientToken else {
+            throw BTPayPalSavedPaymentMethodError.invalidAuthorization
+        }
```
buzzamus, line 107: This is the repeated guard statement for the client token check

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,122 @@
...
+        let (body, _) = try await apiClient.post(
+            "/v2/credit/fetch-presentment-messages",
+            parameters: parameters,
+            httpType: .payPalAPI
+        )
+
+        guard let body else {
+            throw BTPayPalSavedPaymentMethodError.emptyBodyReturned
+        }
+
```
buzzamus, line 120: repeated empty body check

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,122 @@
...
+            paymentMethodIDJWT: paymentMethodIDJWT,
+            orderID: resolvedOrderID,
+            merchantAccountID: merchantAccountID
+        )
+
+        let (body, _) = try await apiClient.post("", parameters: parameters, httpType: .graphQLAPI)
+
+        guard let body else {
+            throw BTPayPalSavedPaymentMethodError.emptyBodyReturned
+        }
```
buzzamus, line 81: repeated empty body check

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodError.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+
+public enum BTPayPalSavedPaymentMethodError: Int, Error, CustomNSError, LocalizedError, Equatable {
+
```
buzzamus, line 4: I am just trying to parse this all in my mind: This is a public error, but everywhere that it is currently thrown is in an internal struct. Will these be bubbled up to the public classes that merchants will interact with?

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
buzzamus: Actually, after reading the error message for `BTPayPalSavedPaymentMethodError.emptyBodyReturned` I don't think that is applicable here is it? It says it is for a`funding instrument details request`, but this seems to be an entirely different request. I think we may need a new error type for this, or change the response

buzzamus (commented):
I left a few comments/had a few questions. I am also curious if there will be a future task for adding the analytic events to this client? If so, can we also add some TODOs to show where we will be send

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: Pulled both into `validateClientTokenAuthorization()` and a private `post(_:parameters:httpType:)` that unwraps the body to reduce some duplication. 
I have kept the auth check as the first line of each method so a tokenization key still throws `invalidAuthorization` rather than `missingPaymentMethodIDJWT`.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: addressed above

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: Made the message text request agnostic rather than adding a second case, since the condition is identical.

```swift
case .emptyBodyReturned:
    return "An empty body was returned from the request."
```

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: addressed in earlier comments

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodError.swift
rainaarya: Yes, they could. The next UI PR #1854's `BTPayPalSavedPaymentMethodView` can surface them through its `completion: (BTPayPalAccountNonce?, Error?)` handler, so they need to be public.

rainaarya:
> I left a few comments/had a few questions. I am also curious if there will be a future task for adding the analytic events to this client? If so, can we also add some TODOs to show where we will be send

Yes analytics would be added in future PRs. The catalog is being finalized and needs review. I have added TODOs for some of the analytics events we plan to emit from the client (such as sticky-FI fetch, post-edit FI refresh, credit messaging)

Sources/BraintreePayPalSavedPaymentMethod/Models/PayPalCreditMessagingPOSTBody.swift
```diff
@@ -0,0 +1,55 @@
...
+    private let messagePlacements: [MessagePlacement]
+    private let flowContext = FlowContext()
+
+    enum CodingKeys: String, CodingKey {
+        case messagePlacements = "message_placements"
+        case flowContext = "flow_context"
+    }
+
+    init(amount: String, currencyCode: String) {
+        self.messagePlacements = [MessagePlacement(amount: MessagePlacement.Amount(currencyCode: currencyCode, value: amount))]
```
agedd: nit: can we newline this for better readability

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,138 @@
...
+        case .fiFromApprovedCheckout:
+            guard let orderID else {
+                throw BTPayPalSavedPaymentMethodError.missingOrderID
+            }
+
+            paymentMethodIDJWT = nil
+            resolvedOrderID = orderID
+        }
+
+        let parameters = PayPalFundingInstrumentDetailsGraphQLBody(
```
agedd: question:  it looks like `merchantAccountID` isn't validated against `fundingInstrumentType` the same way  1orderID/paymentMethodIDJWT` are - just would like to confirm if this is intentionally optional

Sources/BraintreePayPalSavedPaymentMethod/Models/PayPalCreditMessagingPOSTBody.swift
rainaarya: 👍 updated

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: Yeah it's intentional. `orderID`/`paymentMethodIDJWT` are identity fields where the API needs exactly the one matching the fetch type, whereas `merchantAccountID` is optional, valid for both.

When it's `nil`, the key is omitted and the default merchant account is used in the server. Also updated this in the docstring.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalCreditMessageItem.swift
```diff
@@ -0,0 +1,50 @@
...
+    let clickURL: URL?
+
+    /// The image to render for an `image` block.
+    let sourceURL: URL?
+
+    /// A non-unique identifier for the block, for example `"periodic_payment_count"` or `"paypal_logo"`.
+    let name: String?
+
+    /// Whether `clickURL` may be loaded in an embedded web view rather than an external browser.
+    let isEmbeddable: Bool
```
jaxdesmarais: If this is swift only we can use an optional bool correct?

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalCreditMessagingResult.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+/// The Pay Later message to display alongside a vaulted PayPal payment method.
+struct BTPayPalCreditMessagingResult: Equatable {
+
+    // MARK: - Internal Properties
```
jaxdesmarais, line 10: can we sort this section as: required > optional to match the rest of the SDK

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalPayer.swift
```diff
@@ -0,0 +1,30 @@
...
+/// The buyer's PayPal account.
+struct BTPayPalPayer: Equatable {
+
+    // MARK: - Internal Properties
+
+    /// The email address associated with the buyer's PayPal account.
+    let email: String?
+
+    /// Whether the buyer is allowed to change the funding instrument PayPal will charge.
+    let isEditable: Bool
```
jaxdesmarais: This can be optional in Swift only

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethod.swift
```diff
@@ -0,0 +1,42 @@
...
+    // MARK: - Initializer
+
+    /// Parses a single entry of the `paymentMethods` array.
+    /// - Returns: `nil` when the entry is not an object.
+    init?(json: BTJSON) {
+        guard json.isObject else {
+            return nil
+        }
+
+        self.type = json["type"].asString().flatMap(BTPayPalSavedPaymentMethodType.init(rawValue:))
```
jaxdesmarais: Why not give `BTPayPalSavedPaymentMethodType` and init so we can avoid this and instead pass the json to the init on that type? Something like `self.type = BTPayPalSavedPaymentMethodType(json: json["type"])` would be much cleaner here.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,139 @@
...
+    /// - Note: Requires a client token. Throws `BTPayPalSavedPaymentMethodError.invalidAuthorization` when initialized
+    ///   with a tokenization key, which carries no `paymentMethodIdJwt`.
+    func fetchPaymentMethod(
+        fundingInstrumentType: BTPayPalFundingInstrumentFetchType,
+        orderID: String? = nil,
+        merchantAccountID: String? = nil
+    ) async throws -> BTPayPalSavedPaymentMethodSummary {
+        // TODO: emit the sticky-FI and post-edit refresh analytics events once the catalog is approved.
+
+        try validateClientTokenAuthorization()
```
jaxdesmarais: I don't think this needs to be extracted out into it's own method. The code is very clear and the scope is quite small.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,139 @@
...
+
+        // The API rejects the request unless exactly the identity field matching the fetch type is sent.
+        let paymentMethodIDJWT: String?
+        let resolvedOrderID: String?
+
+        switch fundingInstrumentType {
```
jaxdesmarais, line 52: Can we call these both something that more clearly describes what they are to a developer? It's not at all clear in BT terms what `stickyFI` vs `fiFromApprovedCheckout` is or does. Can you share what these represent and we can maybe workshop a name? This will help future developers and maintainers of this code as well.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,139 @@
...
+        case .fiFromApprovedCheckout:
+            guard let orderID else {
+                throw BTPayPalSavedPaymentMethodError.missingOrderID
+            }
+
+            paymentMethodIDJWT = nil
+            resolvedOrderID = orderID
+        }
+
+        let parameters = PayPalFundingInstrumentDetailsGraphQLBody(
```
jaxdesmarais: It's a bit more verbose but it may be more clear to avoid extracting and setting `paymentMethodIDJWT` and `orderID` in the switch statements and instead set parameters as the top level property that we then set a value to depending on if we have an order ID or payment method ID JWT. Something like:
```swift
var parameters: PayPalFundingInstrumentDetailsGraphQLBody

switch fundingInstrumentType {
case .stickyFI:
            guard let jwt = (apiClient.authorization as? ClientTokenAuthorizationProviding)?.paymentMethodIDJWT else {
                throw BTPayPalSavedPaymentMethodError.missingPaymentMethodIDJWT
            }

            parameters = PayPalFundingInstrumentDetailsGraphQLBody(
            fundingInstrumentType: fundingInstrumentType,
                        paymentMethodIDJWT: paymentMethodIDJWT,
                        orderID: nil,
                        merchantAccountID: merchantAccountID
            )

case .fiFromApprovedCheckout:
            guard let orderID else {
                throw BTPayPalSavedPaymentMethodError.missingOrderID
            }

            parameters = PayPalFundingInstrumentDetailsGraphQLBody(
            fundingInstrumentType: fundingInstrumentType,
                        paymentMethodIDJWT: nil,
                        orderID: orderID,
                        merchantAccountID: merchantAccountID
            )
```
This allows our code to be a bit easier to reason about and has parameters (the reusable portion) extracted out while avoiding having `paymentMethodIDJWT` and `orderID` as properties only used by 1 case

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
```diff
@@ -0,0 +1,139 @@
...
+
+    /// The GraphQL rail reads the client token's `paymentMethodIdJwt` and the PayPal API rail authenticates with its bearer,
+    /// so neither works with a tokenization key.
+    private func validateClientTokenAuthorization() throws {
+        guard apiClient.authorization.type == .clientToken else {
+            throw BTPayPalSavedPaymentMethodError.invalidAuthorization
+        }
+    }
+
+    private func post(
```
jaxdesmarais: I don't think we need this, it adds additional logic to some already very nested cross module logic.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodSummary.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+
+#if canImport(BraintreeCore)
+import BraintreeCore
+#endif
+
+/// The funding instrument details for a vaulted PayPal payment method.
+struct BTPayPalSavedPaymentMethodSummary: Equatable {
+
+    // MARK: - Internal Properties
```
jaxdesmarais, line 10: Let's flip the order: required > optional

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodSummary.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    /// Parses the `paypalFundingInstrumentDetails` field of a `PaypalFundingInstrumentDetails` response.
+    /// - Returns: `nil` when the field is not an object.
+    init?(json: BTJSON) {
+        guard json.isObject else {
+            return nil
+        }
+
+        self.payer = BTPayPalPayer(json: json["payer"])
+        self.paymentMethods = json["paymentMethods"].asArray()?.compactMap(BTPayPalSavedPaymentMethod.init) ?? []
```
jaxdesmarais, line 27: Same question as one of the other types, can we provide an init on `BTPayPalSavedPaymentMethod` and avoid the compact map nesting?

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalCreditMessageItem.swift
rainaarya: you are right, have made them `Bool?`

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalPayer.swift
rainaarya: updated

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalCreditMessagingResult.swift
rainaarya: Done

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodSummary.swift
rainaarya: Done

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethod.swift
rainaarya: Added `init?(json:)` to both `BTPayPalSavedPaymentMethodType` and `BTPayPalCreditMessageItemType` so the models just pass the node through.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: Done, parameters is now assigned in each case, so the two single use optionals are gone.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: These were added at @buzzamus's request [in his earlier comment](https://github.com/braintree/braintree_ios/pull/1850#discussion_r3907251031) to reduce some duplication. I have inlined back, hope we are aligned.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: Similar to [earlier](https://github.com/braintree/braintree_ios/pull/1850#discussion_r3968071390). Have updated it back to previous implementation.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
buzzamus: I am fine with either, but putting the same check that is used in multiple places in a helper is a pattern a previous team of mine used. One time the checks changed and once only one spot was updated, which caused issues. It isolates it all to one spot and with a good descriptive name it makes it easy to tell what it is checking...but this probably won't change so, reverting works too

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodSummary.swift
rainaarya: `BTPayPalSavedPaymentMethod`  already has `init?(json:)`. compactMap here would pass to it when it runs each element through it. Since `self.paymentMethods` is an array.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalSavedPaymentMethodClient.swift
rainaarya: `stickyFI` is the instrument saved as default on the buyer's billing agreement (also what PayPal charges if nothing changes, resolved from the client token JWT). This is shown initially.

`fiFromApprovedCheckout` mode is used for the FI they pick/change during the edit FI flow in checkout page, which we will re-read after an edit by passing the orderId that we get from the nonce returned from`create_payment_resource` call, to the GQL. 

We could call them something else but the raw values would have to stay same as the GraphQL enum, since the API uses these.

jaxdesmarais: Yep, understand the underlying raw values wouldn't change. Thanks for the descriptions, that helps! I think it's especially important for these since there is logic around them vs some of the other types. Maybe: `stickyFI` > `buyerDefaultBillingAgreement` and `fiFromApprovedCheckout` > `buyerUpdatedBillingAgreement`? What do y'all think @agedd and @buzzamus ?

buzzamus: I agree with the suggestion @jaxdesmarais gave. I think keeping the raw values that interact with the GQL API/schema makes sense, but since we need to understand the code that references them and tell what they represent a name like `buyerDefaultBillingAgreement` and `buyerUpdatedBillingAgreement` make sense. Alternatives could be to put the descriptive word first such as `defaultBuyerBillingAgreement` or `updatedBuyerBillingAgreement`, but at that point we are splitting hairs really

rainaarya: Thanks all for your inputs! I have renamed it to `buyerDefaultBillingAgreement` / `buyerUpdatedBillingAgreement` and also updated the docstrings and test names to match.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalCreditMessageItem.swift
```diff
@@ -0,0 +1,50 @@
...
+    let clickURL: URL?
+
+    /// The image to render for an `image` block.
+    let sourceURL: URL?
+
+    /// A non-unique identifier for the block, for example `"periodic_payment_count"` or `"paypal_logo"`.
+    let name: String?
+
+    /// Whether `clickURL` may be loaded in an embedded web view rather than an external browser.
+    let isEmbeddable: Bool?
```
agedd, line 31: nit: could this default to `false` instead of being optional? it's a bit unclear what a caller should do differently for `nil` vs `false` here

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalPayer.swift
```diff
@@ -0,0 +1,30 @@
...
+/// The buyer's PayPal account.
+struct BTPayPalPayer: Equatable {
+
+    // MARK: - Internal Properties
+
+    /// The email address associated with the buyer's PayPal account.
+    let email: String?
+
+    /// Whether the buyer is allowed to change the funding instrument PayPal will charge.
+    let isEditable: Bool?
```
agedd, line 16: nit: same concerns as `isEmbeddable` - could this default to `false` instead of being optional

Sources/BraintreePayPalSavedPaymentMethod/Models/PayPalCreditMessagingPOSTBody.swift
```diff
@@ -0,0 +1,59 @@
...
+                case value
+            }
+        }
+    }
+
+    struct FlowContext: Encodable {
+
+        let attributes = ["BRAND_BRAINTREE", "EXPERIENCE_IOS_SDK"]
+        let channel = "MOBILE_APP"
+        let flowSpecifier = "EARLY_PRESENTMENT"
```
agedd, line 50: it seems these are hard-coded for early presentment/compact messaging - just checking if this is the only use case for this module or if these should instead be init params if another flow needs other values later

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalCreditMessagingResult.swift
agedd, line 10: nit: for result types, we typically keep our properties public and init internal. can we match this to our sdk pattern accordingly or is there a reason we would prefer to keep it as-is

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalCreditMessageItem.swift
rainaarya: Like other fields, I had made this optional in an [earlier comment](https://github.com/braintree/braintree_ios/pull/1850#discussion_r3960126337) since the types are Swift only. From a caller's perspective, I don't think there should be any distinction, so nil/false should behave the same.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalPayer.swift
rainaarya: same as above

Sources/BraintreePayPalSavedPaymentMethod/Models/PayPalCreditMessagingPOSTBody.swift
rainaarya: Yes, for View/Edit FI, the component always requests Treatment A (The "Or" version of text from the messaging) and any other forms are out of scope so we hardcode these. There's no other flow/caller for this.

Sources/BraintreePayPalSavedPaymentMethod/BTPayPalCreditMessagingResult.swift
rainaarya: Here, `BTPayPalCreditMessagingResult` is internal (it will only be used within our UI layer / ViewModel) and merchants/public do not need access to this. 

Since the overall struct is internal, `public` members here would still be capped at internal.

jaxdesmarais (approved):
Thanks for addressing feedback and comments!

buzzamus (approved):
Thanks for the changes! This looks good

PR #1851, opened 2026-08-14
Carry Over Venmo Buyer info from Vault Response

buzzamus:
### Summary of changes
- This PR carries over the external ID for the Venmo vault flow

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Discover where issue was, write unit tests

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/Sources/BraintreeVenmo/BTVenmoAccountNonce.swift
+++ b/Sources/BraintreeVenmo/BTVenmoAccountNonce.swift
@@ -35,8 +35,9 @@ import BraintreeCore
 
     // MARK: - Initializers
 
-    init(with nonce: String, username: String, isDefault: Bool) {
+    init(with nonce: String, username: String, isDefault: Bool, externalID: String? = nil) {
         self.username = username
+        self.externalID = externalID
         super.init(nonce: nonce, type: "Venmo", isDefault: isDefault)
     }
 
@@ -60,10 +61,13 @@ import BraintreeCore
     // MARK: - Internal Methods
 
     static func venmoAccount(with json: BTJSON) -> BTVenmoAccountNonce {
+        // The vault response does not include an `externalId` field, but `details.commonId`
+        // contains the same underlying Venmo account identifier.
         BTVenmoAccountNonce(
             with: json["nonce"].asString() ?? "",
             username: json["details"]["username"].asString() ?? "",
-            isDefault: json["default"].isTrue
+            isDefault: json["default"].isTrue,
+            externalID: json["details"]["commonId"].asString()
         )
     }
 }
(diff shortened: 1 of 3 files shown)


agedd (approved):
looks good! 🚀

Sources/BraintreeVenmo/BTVenmoAccountNonce.swift
```diff
@@ -35,8 +35,9 @@ import BraintreeCore
 
     // MARK: - Initializers
 
-    init(with nonce: String, username: String, isDefault: Bool) {
+    init(with nonce: String, username: String, isDefault: Bool, externalID: String? = nil) {
```
agedd, line 38: non-blocking: wonder if we can rename `externalID` to `commonID` instead to more closely align what the json response as the other properties do

jaxdesmarais (commented):
Ah, so this fixes the issue and will return the expected details in the nonce when vaulted? If so, we may want a changelog so we can point to the fix.

buzzamus:
> Ah, so this fixes the issue and will return the expected details in the nonce when vaulted? If so, we may want a changelog so we can point to the fix.

It doesn't return all of the buyer info since that would require it being added to the 2nd nonce in this particular flow from the GQL team, but this does satisfy the request to include the external ID for the merchant to use and tie to a vaulted venmo account

Sources/BraintreeVenmo/BTVenmoAccountNonce.swift
buzzamus: So, the reason for this is that it is known as externalID everywhere else, but in this flow it is returned as commonID, but commonID == externalID. This just keeps it consistent with the other developer facing uses of this value

jaxdesmarais:
> It doesn't return all of the buyer info since that would require it being added to the 2nd nonce in this particular flow from the GQL team, but this does satisfy the request to include the external ID for the merchant to use and tie to a vaulted venmo account

Ah - gotcha, I think we will want to expose this as a property on the nonce for merchants to access? Then we can add a changelog entry too. Will we also open the same PR on Android?

buzzamus:
> > It doesn't return all of the buyer info since that would require it being added to the 2nd nonce in this particular flow from the GQL team, but this does satisfy the request to include the external ID for the merchant to use and tie to a vaulted venmo account
> 
> Ah - gotcha, I think we will want to expose this as a property on the nonce for merchants to access? Then we can add a changelog entry too. Will we also open the same PR on Android?

I can review Android and see if this is needed and if so, make a PR for Android. This was just reported for iOS specifically and when I initially checked Android that was when I misunderstood and thought this was for all the buyer info.

external ID is already a public property on the nonce for merchants to access already, so this is just exposing it for this specific flow since it wasn't being populated. Venmo team confirmed common ID is external ID

jaxdesmarais (approved):

PR #1853, opened 2026-08-20
Update PA M3 Core Logic 

agedd:
### Summary of changes
- Aligns iOS Payment Actions M3 core logic with the backend design and cleans up leftover test-target config from module scaffolding.

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Identify where changes need to be made and why.

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

- @agedd


mzlangreder (approved):

Sources/BraintreePaymentActions/BTPaymentActionRequest.swift
```diff
@@ -5,14 +5,9 @@ import Foundation
...
-    // MARK: - Internal Properties
+    // MARK: - Initializer
     
-    /// The ID of the Payment Action to submit this payment method for.
-    let paymentActionID: String
-    
-    public init(paymentActionID: String) {
-        self.paymentActionID = paymentActionID
-    }
+    public override init() {}
```
buzzamus, line 10: I know everything keeps changing, so this may be hard to answer at the moment, but this class is now pretty bare. Is the idea that eventually there will be enough shared logic to warrant a base class?

agedd: yep great question - it's bare because `submitForPaymentAction` is `@objc-` exposed and takes in a generic `BTPaymentActionRequest` directly so it has to be a real class.

the `paymentMethodParameters()` , gets called once when building the graphQL body and throws an error if a subclass (card for m3 now, 3ds for m4, and maybe more PA-supported types in the future) does not override it.

Sources/BraintreePaymentActions/BTCreditCard.swift
```diff
@@ -21,7 +21,6 @@ public class BTCardPaymentActionRequest: BTPaymentActionRequest {
     
     /// Initialize a `BTCardPaymentActionRequest`
```
jaxdesmarais, line 22: It looks like we are calling this just `CreditCard` on Android. Can we decide on the naming an align the 2 platforms? @agedd @mzlangreder

Sources/BraintreePaymentActions/Models/BTCardPaymentActionRequest.swift
jaxdesmarais: Does this need `@objcMembers`?

Sources/BraintreePaymentActions/BTPaymentActionResult.swift
```diff
@@ -8,7 +8,7 @@ import Foundation
     
     let type: BTPaymentActionResultType
     let id: String
-    let serverAction: BTServerAction
+    let serverAction: BTServerAction?
```
jaxdesmarais, line 11: It looks like this is non-optional on Android currently. Is this change intentional?

Sources/BraintreePaymentActions/Models/BTCardPaymentActionRequest.swift
agedd: ah good catch, i cross-checked with the base class and where this card request is used and don't think so - mainly because `objcMembers` on the base class already covers what needs obj-c bridging and all the stored properties on this request are internal, rather than public

Sources/BraintreePaymentActions/BTPaymentActionResult.swift
agedd: agreed, optional is fine here. just checked on android and it looks like the platforms differ more than just null-handling - android uses a sealed class per outcome, so `serverAction` only exists where this is a `ServerActionRequired` case. 

i think we can't do the same here since `BTPaymentActionResult` needs Obj-C bridging, so we can't use swift enums with associated values, so decided to go with optional given this constraint

Sources/BraintreePaymentActions/BTCreditCard.swift
mzlangreder: I think since we're going with a single payment actions module handling multiple payment types, being explicit and using `CreditCard` will end up helping us in the future.

Sources/BraintreePaymentActions/Models/CreditCard.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 /// The card details to submit as the payment method for a Payment Action.
-public class BTCardPaymentActionRequest: BTPaymentActionRequest {
+public class CreditCard: BTPaymentActionRequest {
```
jaxdesmarais: I think we still want to prefix this with BT right? So we can give it `@objcMembers` as well?

Sources/BraintreePaymentActions/BTCreditCard.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 /// The card details to submit as the payment method for a Payment Action.
-public class BTCardPaymentActionRequest: BTPaymentActionRequest {
+@objcMembers public class BTCreditCard: BTPaymentActionRequest {
```
jaxdesmarais, line 4: not going to block the PR but before merge, `models` should only be used for the API layer internal shape. I think we want `BTCreditCard` to live outside of that since it's something merchants use. If you don't mind moving that before you merge in.

jaxdesmarais (approved):

Sources/BraintreePaymentActions/BTCreditCard.swift
agedd: ah yes, addressed in 7aaa9b47e43e4eda489b9fa2504d2aa7b8d22698

PR #1858, opened 2026-08-26
Add Basic Integration Tests For PA M3

agedd:
### Summary of changes
-  Add basic integration tests for PA M3 logic

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Brainstorming and formatting tests

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

- @agedd


IntegrationTests/BraintreePaymentActionsClient_IntegrationTests.swift
```diff
@@ -0,0 +1,154 @@
+import Foundation
+import XCTest
+@testable import BraintreeCore
+@testable import BraintreePaymentActions
+
+class BTPaymentActionsClient_IntegrationTests: XCTestCase {
```
buzzamus: I think you will need to add this file to the github workflow for our integration tests in order to have them run in CI

IntegrationTests/BraintreePaymentActionsClient_IntegrationTests.swift
```diff
@@ -0,0 +1,154 @@
...
+        paymentActionsClient = BTPaymentActionsClient(authorization: BTIntegrationTestsConstants.sandboxClientToken)
+    }
+    
+    // MARK: - Happy Path
+    
+    func testSubmitForPaymentAction_autoConfirmAutoCapture_returnsCompletedResult() async throws {
+        try XCTSkipIf(true, "Pending Flow 1 (AUTO confirm / AUTO capture) test PAN")
+        
+        let cardRequest = BTCardPaymentActionRequest(
+            cardNumber: "4111111111111111", // TODO: swap for the documented Flow 1 test PAN
```
buzzamus: oh I see these aren't necessarily ready to run in CI yet I take it?

IntegrationTests/BraintreePaymentActionsClient_IntegrationTests.swift
```diff
@@ -0,0 +1,154 @@
...
+            XCTAssertNil(result.serverAction)
+        } catch {
+            XCTFail("Unexpected error: \(error.localizedDescription)")
+        }
+    }
+    
+    func testSubmitForPaymentAction_flow2_manualConfirmAutoCapture_returnsServerActionRequiredConfirm() async throws {
+        try XCTSkipIf(true, "Pending Flow 2 (MANUAL confirm / AUTOMATIC capture) test PAN")
+        
+        let cardRequest = BTCardPaymentActionRequest(
```
buzzamus: I think this is named `BTCreditCard` now right?

jaxdesmarais (changes requested):
agree with @buzzamus comments thus far

IntegrationTests/BraintreePaymentActionsClient_IntegrationTests.swift
agedd: ah yep good catch, updated in 2920d7d4891b9c54584e38ee0e4b195fca2b00a6

IntegrationTests/BraintreePaymentActionsClient_IntegrationTests.swift
agedd: yep, still waiting on pa m3 core changes to be deployed to sandbox and actual PAN numbers we can test with

IntegrationTests/BraintreePaymentActionsClient_IntegrationTests.swift
agedd: ah yes, added this file to the `IntegrationTests` target membership and looks like `tests.yml` runs the whole `IntegrationTests` scheme so should be included now

buzzamus (commented):
This all looks good now except the integration test is failing due to no expectation set...I am assuming this is because the backend isn't ready yet, so I am not sure if we just need to wait for that? Is that correct?

jaxdesmarais:
I think backend is all set up - are we able to update this PR?

agedd:
> This all looks good now except the integration test is failing due to no expectation set...I am assuming this is because the backend isn't ready yet, so I am not sure if we just need to wait for that? Is that correct?

yep it's been updated and ready for review now

IntegrationTests/BraintreePaymentActionsClient_IntegrationTests.swift
agedd: updated now!

jaxdesmarais (approved):

buzzamus (approved):
These look great! Thanks for adding them

Demo/Application/Merchant API Client/BraintreeDemoMerchantAPIClient.swift
```diff
@@ -102,7 +102,7 @@ class BraintreeDemoMerchantAPIClient: NSObject {
         captureMethod: String = "AUTOMATIC",
         completion: @escaping (PaymentActionResponse?, Error?) -> Void
     ) {
-        guard var urlComponents = URLComponents(string: "https://braintree-sample-merchant.herokuapp.com/create_payment_action") else {
+        guard var urlComponents = URLComponents(string: BraintreeDemoSettings.currentEnvironmentURLString + "/create_payment_action") else {
```
buzzamus, line 105: nice 😎

PR #1863, opened 2026-09-01
Payment Actions M3 Cleanup

agedd:
### Summary of changes
- `setPaymentActionPaymentMethod` request body serializes to correct wire input format to match backend schema
-  `BTPaymentActionStatus.canceled.description` returns "CANCELLED" (double-L) to match backend `PaymentActionStatus` enum; public case name stays .canceled.
-  `BTPaymentActionAnalytics` created to house analytics constants that are fired on `submitForPaymentAction`

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
-  cross-checking and brainstorming

- [x] less than 30% 
- [ ] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

- @agedd


Sources/BraintreePaymentActions/BTPaymentActionStatus.swift
```diff
@@ -38,7 +38,7 @@ import Foundation
         case .requiresCapture: return "REQUIRES_CAPTURE"
         case .requiresCustomerAction: return "REQUIRES_CUSTOMER_ACTION"
         case .processing: return "PROCESSING"
-        case .canceled: return "CANCELED"
+        case .canceled: return "CANCELLED"
         case .expired: return "EXPIRED"
```
buzzamus, line 42: TIL both canceled and cancelled are correct spellings

buzzamus (approved):

Sources/BraintreePaymentActions/BTPaymentActionAnalytics.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+enum BTPaymentActionAnalytics {
+    
+    // MARK: - Payment Action Events
+    
+    static let paymentActionsSetPaymentMethodStarted = "card:payment-actions:set-payment-method:started"
+    static let paymentActionsSetPaymentMethodSucceeded = "card:payment-actions:set-payment-method:succeeded"
+    static let paymentActionsSetPaymentMethodFailed = "card:payment-actions:set-payment-method:failed"
```
jaxdesmarais: nit - can we reverse the order here for iOS and Android? Should be: `<module_name>:<method_name>:<optional_detail>:<result>`

jaxdesmarais: Same with the property name will need to be updated

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
```diff
@@ -70,24 +70,38 @@ import BraintreeCore
...
             let status = BTPaymentActionStatus.status(from: statusString)
+            apiClient.sendAnalyticsEvent(
+                BTPaymentActionAnalytics.paymentActionsSetPaymentMethodSucceeded
+            )
             return BTPaymentAction(id: paymentActionID, status: status)
         } catch {
-            // TODO: Replace with exact error type in a later PR.
+            apiClient.sendAnalyticsEvent(
+                BTPaymentActionAnalytics.paymentActionsSetPaymentMethodFailed,
+                errorDescription: BTPaymentActionError.decodingFailure.errorDescription
```
jaxdesmarais: should this be a decoding failure or the error from the catch?

Sources/BraintreePaymentActions/BTPaymentActionAnalytics.swift
agedd: yep, updated in 5882232118cd567095c52351f216ab78403f5b60

Sources/BraintreePaymentActions/BTPaymentActionsClient.swift
agedd: ah yep good catch, should be error from catch - updated in 5882232118cd567095c52351f216ab78403f5b60

Sources/BraintreePaymentActions/Models/BTCardPaymentActionPayload.swift
```diff
@@ -3,10 +3,23 @@ import Foundation
 /// The card payment method payload submitted to a Payment Action via `BTPaymentActionsClient`
 struct BTCardPaymentMethodPayload: Encodable {
     
-    let card: Card
+    let paymentMethodDetails: PaymentMethodDetails
+    
+    init(request: BTCreditCard) {
+        self.paymentMethodDetails = PaymentMethodDetails(request: request)
```
jaxdesmarais: It looks like some of this deviates from our existing model pattern. One being that we use the `BT` prefix which we typically only use for external types and shape these as the `POST/GQL body`. The 2nd being that we are using `request` in the init instead of the type. Is this intentional?

agedd: ah yes good catch on both counts, addressed this in 5947e93f669775f72356a0f1b9c1fb71802a414d

Sources/BraintreePaymentActions/Models/CardPaymentActionPayload.swift
```diff
@@ -1,12 +1,25 @@
 import Foundation
 
 /// The card payment method payload submitted to a Payment Action via `BTPaymentActionsClient`
-struct BTCardPaymentMethodPayload: Encodable {
+struct CardPaymentMethodPayload: Encodable {
```
jaxdesmarais, line 4: you'll want to change the filename too. I wish Swift did this automagically. :)

jaxdesmarais (approved):

PR #1864, opened 2026-09-01
Add Payment Actions To Demo App

agedd:
### Summary of changes
- Add Payment Actions To Demo App

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
To brainstorm and determine where to make changes to `BraintreeDemoMerchantAPIClient` to add new endpoints for Payment Actions.

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

- @agedd


Demo/Application/Features/PaymentActionsController.swift
```diff
@@ -0,0 +1,228 @@
+import UIKit
+import BraintreeCard
+import BraintreePaymentActions
+
+class PaymentActionsViewController: PaymentButtonBaseViewController {
```
jaxdesmarais: Should this just be a Swift UI view?

Demo/Application/Features/PaymentActionsView.swift
```diff
@@ -0,0 +1,226 @@
+import SwiftUI
+import BraintreeCard
+import BraintreePaymentActions
+
+final class PaymentActionsViewModel: ObservableObject {
```
jaxdesmarais: Is this a view or view model?

Demo/Application/Features/PaymentActionsView.swift
```diff
@@ -0,0 +1,226 @@
...
+    
+    // Card Form Fields
```
jaxdesmarais: I thought we had a swift UI card form but am I misremembering?

Demo/Application/Features/PaymentActionsView.swift
```diff
@@ -0,0 +1,226 @@
...
+    @Published var cardNumber: String = ""
+    @Published var expirationMonth: String = ""
+    @Published var expirationYear: String = ""
+    @Published var cvv: String = ""
+    @Published var postalCode: String = ""
+    
+    // UI State
+    @Published var progressMessage: String = ""
+    @Published var isPayEnabled: Bool = false
```
jaxdesmarais: nit:
```suggestion
    @Published var isPayButtonEnabled: Bool = false
```

Demo/Application/Features/PaymentActionsView.swift
```diff
@@ -0,0 +1,226 @@
...
+    @Published var isFieldsEnabled: Bool = false
```
jaxdesmarais: nit:

```suggestion
    @Published var isCardFieldsEnabled: Bool = false
```

Demo/Application/Features/PaymentActionsView.swift
```diff
@@ -0,0 +1,226 @@
...
+            }
+        }
+    }
+    
+    // MARK: Actions
+    
+    func tappedPay() {
+        progressMessage = "Submitting payment method for Payment Action.."
+        
+        guard !cvv.isEmpty else {
```
jaxdesmarais: Do we need to add all fields to this check? If we have a separate card fields view maybe we could share this logic?

Demo/Application/Features/PaymentActionsView.swift
```diff
@@ -0,0 +1,226 @@
...
+            case .confirm:
+                progressMessage = "Notifying server to confirm Payment Action \(result.id)..."
+                try await notifyServerToConfirm(paymentActionId: result.id)
+                showOrderConfirmation(paymentActionId: result.id)
+            case .capture:
+                // Authorized; capture is pending server-side.
+                showOrderConfirmation(paymentActionId: result.id)
+            case nil:
+                progressMessage = "Server action required but none was specified."
+            case .some:
```
jaxdesmarais: What is this case? Seems confusing to a merchant potentially to have to handle. We should likely have `confirm`, `capture` and maybe an `unknown`? Feels odd to return a `nil` or `some` action where `some` indicates unknown. I would expect a nil but not as a case, but maybe I am overthinking this from the merchant POV. Curious how Android is handling this @mzlangreder ?

Demo/Application/Features/PaymentActionsView.swift
```diff
@@ -0,0 +1,226 @@
...
+                progressMessage = "Server action unknown."
+            }
+            
+        case .paymentMethodRequired:
+            clearCardFields()
+            showDeclineMessage()
+            
+        case .customerActionRequired, .processing, .canceled, .expired, .unknown:
```
jaxdesmarais: These both feel like a ton more actions to handle where we have actions an additional actions. Wondering if there is anything we can do to flatten this or if this is purely documentation. cc: @mzlangreder

Demo/Application/Features/PaymentActionsView.swift
```diff
@@ -0,0 +1,226 @@
...
+            progressMessage = "Payment Action \(result.id): \(result.type)"
+        @unknown default:
+            progressMessage = "Payment Action \(result.id) case not handled."
+        }
+    }
+    
+    private func showOrderConfirmation(paymentActionId: String) {
```
jaxdesmarais: To match swift styling:
```suggestion
    private func showOrderConfirmation(paymentActionID: String) {
```

Demo/Application/Features/PaymentActionsView.swift
```diff
@@ -0,0 +1,226 @@
...
+        progressMessage = "Payment Action \(paymentActionId) complete ✅"
+    }
+    
+    private func showDeclineMessage() {
+        progressMessage = "Payment method declined. Please try another card."
+    }
+    
+    /// Asks the merchant server to confirm a Payment Action that requires it.
+    private func notifyServerToConfirm(paymentActionId: String) async throws {
```
jaxdesmarais: To match swift styling:

```suggestion
    private func notifyServerToConfirm(paymentActionID: String) async throws {
```

Demo/Application/Features/PaymentActionsView.swift
```diff
@@ -0,0 +1,226 @@
...
+        // TODO: Add confirmPaymentAction(id:) to BraintreeDemoMerchantAPIClient.
+        // Should hit the sample-merchant server's confirm endpoint for the given Payment Action id.
+        // try await BraintreeDemoMerchantAPIClient.shared.confirmPaymentAction(id: paymentActionId)
+    }
+    
+    private func clearCardFields() {
```
jaxdesmarais: Do we not need to clear expiry/postal?

Demo/Application/Features/PaymentActionsView.swift
jaxdesmarais: Ah, I see this file is both. We should likely separate the view model into it's own file but not sure if we've come across this yet. @buzzamus curious what your thoughts are for more complicated views or if this all just lives in the view?

buzzamus: Yeah, I think for this we should have a separate ViewModel that houses all of the business logic and then keep the view clean with a view model injected (which will make any UI tests we may eventually have to add easier too). I think that makes it easier to demo too since a dev can easily look at the ViewModel to see how the SDK is used and then view is simpler to show the calls. 

TBH This is sort of the first time in our demo we have had to use the MVVM pattern, but I think eventually once we are completely done with UIKit we could move over from MVC and entirely use MVVM. 

I do personally like skipping out on UIKit views entirely for sure.

Demo/Application/Features/PaymentActionsView.swift
agedd: ah yep, i pulled in the latest main to the feature branch and was able to use swift-ui `CardFormView`

Demo/Application/Features/PaymentActionsView.swift
agedd: ah yep great callouts, updated in a4f4041c4c291affbcc1308f030a089094a0abd7

Sources/BraintreePaymentActions/BTPaymentActionResult.swift
```diff
@@ -6,9 +6,9 @@ import Foundation
     
     // MARK: - Internal Properties
     
-    let type: BTPaymentActionResultType
-    let id: String
-    let serverAction: BTServerAction?
+    public let type: BTPaymentActionResultType
```
jaxdesmarais, line 9: You'll want to update the `MARK` here. I think also generally our result pattern is properties public and init internal.

jaxdesmarais (approved):

buzzamus (approved):
Looks good! Love the MVVM pattern usage

PR #1868, opened 2026-09-04
Convert Amex to Swift UI

agedd:
### Summary of changes
-  Convert Amex To Swift UI

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
brainstorming and fixing rendering the preview in canvas by adjusting the optimization levels

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

- @agedd


Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -996,6 +1000,7 @@
 				MARKETING_VERSION = 6.0.0;
 				PRODUCT_BUNDLE_IDENTIFIER = com.braintreepayments.Demo;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
```
buzzamus, line 1003: Is this intentional?

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -1016,6 +1021,7 @@
 					"$(inherited)",
 					"$(PROJECT_DIR)/../Frameworks/**",
 				);
+				GCC_OPTIMIZATION_LEVEL = 0;
```
buzzamus, line 1024: Same here, not sure if this was intended to be included or not

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -1025,6 +1031,7 @@
 				MARKETING_VERSION = 6.0.0;
 				PRODUCT_BUNDLE_IDENTIFIER = com.braintreepayments.Demo;
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
```
buzzamus, line 1034: same here

Demo/Application/Features/AmexView.swift
```diff
@@ -0,0 +1,101 @@
...
+                Task { await getRewards(for: "371544868764018") }
+            }
+            .opacity(loadingCardNumber == "371544868764018" ? 0.5 : 1.0)
+            .allowsHitTesting(loadingCardNumber == nil)
+
+            Button("Ineligible Card") {
+                Task { await getRewards(for: "378267515471109") }
+            }
+            .foregroundColor(.red)
+            .opacity(loadingCardNumber == "378267515471109" ? 0.5 : 1.0)
```
buzzamus: tiny nit take it or leave it: What about putting these card numbers into constants at the top or somewhere so they can easily be changed if needed in future. I highly doubt they will change, so i'm fine either way

Demo/Demo.xcodeproj/project.pbxproj
agedd: yep this was intentional, needed it to get the canvas preview building (as it was broken even on main).

Demo/Demo.xcodeproj/project.pbxproj
agedd: checked against `main` and it was already set to `0` in debug config - so it's likely just xcode touching the line without altering the value.

Demo/Application/Features/AmexView.swift
agedd: good idea, i've put these into constants at the top of the file so they're easier to update later

Demo/Demo.xcodeproj/project.pbxproj
agedd: same as https://github.com/braintree/braintree_ios/pull/1868#discussion_r3958814513

Demo/Demo.xcodeproj/project.pbxproj
buzzamus: ah that makes sense thanks!

Demo/Application/Features/AmexView.swift
buzzamus: Awesome thank you!

buzzamus (approved):
Looks good!

Demo/Application/Features/AmexView.swift
```diff
@@ -0,0 +1,107 @@
+import SwiftUI
+import BraintreeAmericanExpress
+import BraintreeCard
+import BraintreeCore
+
+private enum TestCardNumber {
+    static let valid = "371260714673002"
+    static let insufficientPoints = "371544868764018"
+    static let ineligible = "378267515471109"
+}
```
jaxdesmarais: Take it or leave it, but we can make this a bit more swift-y by making this a proper swift enum:
```swift
private enum TestCardNumber: String {
    case valid = "371260714673002"
    case insufficientPoints = "371544868764018"
    case ineligible = "378267515471109"
}
```
The we can set our `loadingCardNumber` as a `TestCardNumber` as well. That allows us to pass it directly into this method signature `private func getRewards(for cardNumber: TestCardNumber) async` as well and lets us remove `TestCardNumber` in most places to just use the dot syntax on the enum. You'll need to add `.rawValue` in a couple of places but I think the result is more common modern swift features.

buzzamus: oooh yeah I like the more swifty enum syntax here!

agedd: ah yes! i like this pattern, updated in 343918e7f521477ec04ac167521e757483761802

jaxdesmarais (approved):

PR #1873, opened 2026-09-14
M3 Integration Cleanup

jaxdesmarais:
### Summary of changes
Integration cleanup for the Payment Actions M3 feature branch. Refines the public BraintreePaymentActions API, fixes several bugs surfaced while wiring up the Demo app, and aligns the Demo app's Payment Actions screen with the rest of the app's UI conventions.

**SDK changes (Sources/BraintreePaymentActions)**
- Replaced the optional `serverAction` property on `BTPaymentActionResult` with a new `BTServerActionRequiredResult` subclass. Previously, `BTPaymentActionResult` carried an always-present `serverAction: BTServerAction?` that was only non-nil when `type == .serverActionRequired`. That's now modeled as a dedicated subclass carrying a non-optional `serverAction`, so callers only need to deal with the property when it's actually meaningful (checked via `as? BTServerActionRequiredResult`). `BTPaymentActionsClient.result(from:)` and the unit tests were updated accordingly.
- Added `Equatable` conformance to `BTPaymentActionResultType`, `BTPaymentActionStatus`, and `BTServerAction`.
- Moved `SetPaymentActionPaymentMethodGraphQLBody.swift` into `Sources/BraintreePaymentActions/Models/`, matching the SDK's convention that GraphQL/network body types live in a module's Models/ directory.
- Minor style cleanup (line breaks, spacing) to match the SDK style guide.

**Demo app changes**
Several fixes as the payment actions flow was not working in the demo app

- Fixed "Merchant account not found" error: `BraintreeDemoSettings.sandboxMerchantAccountID` was set to the sample-merchant server's top-level merchant_id rather than an actual merchant account ID vs merchant ID.
- Fixed "A payment action ID is required" error on submit: PaymentActionsViewModel was building its BTPaymentActionsClient from the app's generic demo authorization instead of the client token returned by fetchPaymentActionClientToken. Since the GraphQL mutation infers which Payment Action to act on entirely from the scope of the client token used to authenticate, the client is now (re)initialized with the fetched, payment-action-scoped client token before `tappedPay()` is allowed to proceed.
- Wired Payment Actions into the app's shared status bar: PaymentActionsView/PaymentActionsViewModel previously rendered their own inline progress/error to the app's existing bottom toolbar status bar used by every other feature. Replaced the view model's local `@Published progressMessage with an onProgress: (String?) -> Void` closure (mimics existing pattern), and `PaymentActionsViewController` now passes `progressBlock` through when embedding the view — so progress/error messages now surface in the same shared the Demo app.
- Removed now-dead code from PaymentActionsView (statusRow, statusKind, StatusKind) and the unused `paymentActionConfigText` display, and button layout.

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used to describe changes I want to make and iterate on desired outcome for result type and demo app updates

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist

### Authors
- @jaxdesmarais


Sources/BraintreePaymentActions/BTServerActionRequiredResult.swift
```diff
@@ -0,0 +1,18 @@
...
+@objcMembers public class BTServerActionRequiredResult: BTPaymentActionResult {
+
+    // MARK: - Public Properties
+
+    /// Required. The server-driven action to perform next.
+    public let serverAction: BTServerAction
+
+    // MARK: - Initializer
+
+    init(id: String, serverAction: BTServerAction) {
```
agedd, line 14: nit: can we add doc strings to the initializer

agedd (approved):
thanks for making these changes, this looks great! 🚀

Sources/BraintreePaymentActions/BTPaymentActionError.swift
```diff
@@ -21,10 +21,14 @@ public enum BTPaymentActionError: Error, CustomNSError, LocalizedError, Equatabl
...
-        case .decodingFailure: return 2
-        case .missingParameters: return 3
+        case .missingID:
+            return 0
+        case .missingStatus:
+            return 1
+        case .decodingFailure:
+            return 2
+        case .missingParameters:
+            return 3
```
buzzamus, line 31: Completely unrelated to the point of this PR, but I actually felt like the way it was was more readable...but assuming this makes it in line with our style guide?

Sources/BraintreePaymentActions/BTPaymentActionResult.swift
```diff
@@ -2,24 +2,23 @@ import Foundation
 
 /// The result of submitting a payment method to a Payment Action, mapped from the Payment Action's
 /// lifecycle status into the next action the merchant should take.
+///
+/// When `type == .serverActionRequired`, the result is actually a `BTServerActionRequiredResult`,
+/// which carries the required `serverAction` — cast to it to read that value.
```
buzzamus, line 7: This comment is very helpful!

buzzamus (approved):
This approach makes sense to me!

Sources/BraintreePaymentActions/BTPaymentActionError.swift
jaxdesmarais: Yeah, the newline is the general style we follow in the rest of the SDK. There is likely a swiftlint rule for this if we wanted to make it more "official"

Sources/BraintreePaymentActions/BTServerActionRequiredResult.swift
jaxdesmarais: It looks like the pattern used in the rest of the SDK for nonce and result types is to have no docstrings on the internal init and only have docstrings on the public types. Will leave as is and we can decide and change that pattern across the entire SDK if we want to have both with docstrings.

PR #1878, opened 2026-09-18
Remove CocoaPods Integration Artifacts

buzzamus:
### Summary of changes
- This PR removes the CocoaPods Integration artifacts and solely uses NPM instead.

Note: It appears the security workflow currently is dependent on CocoaPods, so that will fail and an investigation will need to be done to determine what updates are required for that

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**
Used to troubleshoot getting Demo app to work locally and to double check for all CocoaPods items were deleted

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- @buzzamus

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,9 @@ import BraintreePayPal
 ```
 
 ### CocoaPods
+
+⚠️ **Deprecation Notice:** CocoaPods support for `braintree_ios` is deprecated and will be removed in a future release, in line with the [official CocoaPods deprecation announcement](https://blog.cocoapods.org/CocoaPods-Specs-Repo/). Going forward, we recommend migrating to **Swift Package Manager** (primary) or **Carthage**. See [#1869](https://github.com/braintree/braintree_ios/issues/1869) for migration guidance and removal timing updates.
+
 ```
 # Includes Cards and PayPal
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -124,7 +124,7 @@ jobs:
         run: |
           set -o pipefail
           xcodebuild \
-            -workspace 'Braintree.xcworkspace' \
+            -project 'Demo/Demo.xcodeproj' \
```
agedd, line 127: nit: should this be `Braintree.xcodeproj` or is it intended to be set to demo 👀

agedd (approved):
looks good to me! just left one small nit

jaxdesmarais (approved):

.github/workflows/tests.yml
buzzamus: This is due to `Braintree.xcodeproj` only having module schemes now and since the workspace file is removed this is to trigger the tests

PR #1879, opened 2026-09-18
Fix tokenize call failure after returning back to Merchant/SDK 7.x

NastassiaRodzik:
### Summary of changes
- Wrap the return tokenize POST API call in a UIApplication background task so the request survives the suspend→active iOS app transition

Steps to test:
- In Demo app, perform and App Switch and Approve the purchase

Demo video of the functionality:
https://github.com/user-attachments/assets/852dc1d8-33ab-4629-b8e4-e4978f96ce0d

### AI Usage

**Which AI Agent Was Used?**
- [ ] Copilot 
- [x] Claude 
- [ ] Other (Type Name Here)

**How was AI used?**

- [ ] less than 30% 
- [x] 30 - 60% 
- [ ] 60 - 100% 

### Checklist
- [ ] Added a changelog entry

- NastassiaRodzik (nrodzik)

----
### Inner Source Process

### Inner Source Checklist
- [x] Provide steps to test the flows changed, if applicable in the summary
- [x] Demo video of the functionality, if applicable
- [x] All upstream dependencies are merged in and this PR can be released at any time; PRs should not be opened until this is true


vkspune (approved):

NastassiaRodzik:
Note on failing UnitTest: `testStart_withDisplayName_callsDelegateWithPaymentID() in BTLocalPaymentClient_IntegrationTests` has failed on this PR. However the work in this PR didn't touch local payments so I assume the failure is unrelated to changes introduced here.

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Braintree iOS SDK Release Notes
 
-## 7.12.0 (2026-09-16)
+## unreleased
 * Add support for Xcode 27 / iOS 27
+* BraintreePayPal
```
jaxdesmarais, line 4: Merge latest from main - the Xcode 27 changes were already released

Sources/BraintreeCore/UIApplication+BackgroundTaskManaging.swift
```diff
@@ -1,13 +1,24 @@
 import UIKit
 
-protocol BackgroundTaskManaging {
+/// :nodoc: This protocol is exposed for internal Braintree use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+///
+/// Used to mock `UIApplication`'s background task assertion API so that unit tests do not request
+/// real task assertions from the system.
+@_documentation(visibility: private)
+public protocol BackgroundTaskManaging {
```
jaxdesmarais, line 8: This should be moved outside of the `Analytics` directory if it's no longer just used for analytics.

UnitTests/BraintreePayPalTests/BTPayPalClient_Tests.swift
```diff
@@ -1912,7 +1913,213 @@ class BTPayPalClient_Tests: XCTestCase {
...
+
+        XCTAssertTrue(manager.didEndBackgroundTask)
+        XCTAssertEqual(manager.endCallCount, 1)
+    }
+
+    /// Regression test: when background time runs out the app suspends and the response never arrives, so
+    /// the expiration handler must end the wait. Without cancelling the request, the caller's `await` never
+    /// resumes and the merchant receives neither a nonce nor an error.
+    @MainActor
+    func testHandleReturn_whenBackgroundTaskExpires_failsWithExpirationError() async {
```
jaxdesmarais, line 2039: I believe an assertion on `endCallCount` would potentially reveal a bug here. I think we are likely missing a lock on `backgroundTaskID` in our code.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -272,7 +276,41 @@ import BraintreeDataCollector
...
+        do {
+            tokenizationResponse = try await tokenizationTask.value
+        } catch {
+            guard tokenizationTask.isCancelled else {
+                notifyFailure(with: error)
+                throw error
+            }
+
+            notifyFailure(with: BTPayPalError.returnBackgroundTaskExpired)
+            throw BTPayPalError.returnBackgroundTaskExpired
```
jaxdesmarais, line 313: What is the goal of this error vs returning the API level error from the catch for all cases? Are there races where we could have a legit error and the task canceled at the same time where we swallow the underlying error? Is this error only returned for cancel cases?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -272,7 +276,41 @@ import BraintreeDataCollector
             correlationID: contextID.flatMap { clientMetadataIDs[$0] }
         )
 
-        let (body, _) = try await apiClient.post("/v1/payment_methods/paypal_accounts", parameters: encodableParams)
+        var backgroundTaskID: UIBackgroundTaskIdentifier = .invalid
+        var cancellableTokenizationTask: Task<(BTJSON?, HTTPURLResponse?), Error>?
+
+        let endBackgroundTask = { [backgroundTaskManager] in
+            guard backgroundTaskID != .invalid else { return }
```
buzzamus, line 284: I think this could potentially cause a race condition since they aren't guaranteed to be on the same thread: `backgroundTaskID` is read and written from two different threads (the `beginBackgroundTask` expiration handler and the `defer { endBackgroundTask() }` at the end of `handleReturn`. the `beginBackgroundTask` is on the main thread, but `endBackgroundTask` is off the main thread. If both hit it at around the same time `endBackgroundTask` could get called twice with the same ID. This needs to be isolated to a single thread/actor to avoid the race condition.

buzzamus (changes requested):

Sources/BraintreePayPal/BTPayPalClient.swift
NastassiaRodzik: [Restricted handleReturn to MainActor] to resolve this (https://github.com/braintree/braintree_ios/pull/1879/commits/911cf964a24bbe5af83498699b58b89a8261ad35)

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Braintree iOS SDK Release Notes
 
+## unreleased
+* BraintreePayPal
+  * Request additional background execution time for the tokenization call made when returning from the PayPal app or browser, so the payment can finish if the buyer backgrounds the app immediately after approving
+  * Add `BTPayPalError.returnBackgroundTaskExpired` (code 15), returned when the background execution time runs out before tokenization completes. Previously the completion handler was never invoked in this case
```
```suggestion
* Fix bug where the return tokenization request could fail to complete after the app was backgrounded during the PayPal app switch or browser flow, resulting in a network connection timeout error
  * Add `BTPayPalError.returnBackgroundTaskExpired` (code 15), returned if background execution time runs out before the request completes
```

Sources/BraintreePayPal/BTPayPalClient.swift
NastassiaRodzik: The purpose of this error is to distinguish and report background task expiration. 
Good catch with swallowing the underlying error - fixed, let me know your thoughts.

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -653,7 +694,7 @@ import BraintreeDataCollector
                         return
                     }
 
-                    Task {
+                    Task { @MainActor in
```
buzzamus: I think we should wrap this in [weak self] here to avoid a strong reference. This only caught my attention now due to adding `@MainActor` on it

buzzamus: Actually, now that I think about it I think having `@MainActor` here and on line 455 are no longer needed since the function itself is now using `@MainActor`, but should have `[weak self]` here

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -411,7 +452,7 @@ import BraintreeDataCollector
                 return
             }
 
-            Task {
+            Task { @MainActor in
```
buzzamus: Same as comment on line 697 below I don't think we need `@MainActor` here anymore, but do need a weak self reference

buzzamus (commented):
Thanks for making the changes! Just a few small comments that came out of that addition as well

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -87,7 +91,7 @@ import BraintreeDataCollector
     
     /// Used for analytics purpose to determine the funding source of the flow i.e. credit, payLater
     private var fundingSource: BTPayPalFundingSource?
-
+
```
agedd: nit: can remove trailing whitespace

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -21,6 +21,10 @@ import BraintreeDataCollector
     /// to prevent calls to openURL. Subclassing UIApplication is not possible, since it enforces that only one instance can ever exist.
     nonisolated(unsafe) var application: URLOpener = UIApplication.shared
 
+    /// Defaults to `UIApplication.shared`, but exposed for unit tests to inject test doubles so that
+    /// background task assertions are not requested from the system during tests.
+    nonisolated(unsafe) var backgroundTaskManager: BackgroundTaskManaging = UIApplication.shared
```
agedd: can this drop `nonisolated(unsafe)` now that `handleReturn` is `@MainActor`

Sources/BraintreePayPal/BTPayPalClient.swift
NastassiaRodzik: I'd leave it here to ensure appSwitchCompletion(nonce, nil) is called on Main, to match existing behavior on lines 149/195 - what do you think?

Sources/BraintreePayPal/BTPayPalClient.swift
```diff
@@ -272,7 +277,43 @@ import BraintreeDataCollector
             correlationID: contextID.flatMap { clientMetadataIDs[$0] }
         )
 
-        let (body, _) = try await apiClient.post("/v1/payment_methods/paypal_accounts", parameters: encodableParams)
+        var backgroundTaskID: UIBackgroundTaskIdentifier = .invalid
+        var cancellableTokenizationTask: Task<(BTJSON?, HTTPURLResponse?), Error>?
+
+        let endBackgroundTask = { [backgroundTaskManager] in
```
agedd: nit(take it or leave it): can we rename to `endBackgroundTaskIfNeeded` instead as `endBackgroundTask` shadows the protocol

Sources/BraintreeCore/UIApplication+BackgroundTaskManaging.swift
```diff
@@ -0,0 +1,24 @@
...
+    func beginBackgroundTask(named: String?, expirationHandler handler: (() -> Void)?) -> UIBackgroundTaskIdentifier
+    func endBackgroundTask(_ identifier: UIBackgroundTaskIdentifier)
+}
+
+extension UIApplication: BackgroundTaskManaging {
```
agedd, line 14: nit: can we add  `// MARK: - BackgroundTaskManaging` before this

UnitTests/BraintreeCoreTests/BTWebAuthenticationSession_Tests.swift
```diff
@@ -97,8 +97,11 @@ class BTWebAuthenticationSession_Tests: XCTestCase {
             }
         }
         
-        // Release all threads at exactly the same time
-        DispatchQueue.global().asyncAfter(deadline: .now() + 0.1) {
+        // Release all threads at exactly the same time.
+        // Not `DispatchQueue.global()`: the blocks above park every worker in that pool, and this release
+        // is queued behind them, so it can only run once the pool grows past `threadCount`. A private
+        // queue targets the overcommit root queue and always gets a thread.
+        DispatchQueue(label: "start-barrier").asyncAfter(deadline: .now() + 0.1) {
```
agedd: do we need to call out the label here as `start-barrier` - can you clarify what is the intent here

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Braintree iOS SDK Release Notes
 
 ## unreleased
+* BraintreePayPal
+* Fix bug where the return tokenization request could fail to complete after the app was backgrounded during the PayPal app switch or browser flow, resulting in a network connection timeout error
```
agedd: nit: ensure this line nests under `BraintreePayPal`

Sources/BraintreePayPal/BTPayPalClient.swift
NastassiaRodzik: Task {} closures aren't stored on self, so there's no retain cycle, just a temporary strong reference to keep the client alive until tokenization finishes, but I can add it if there is a code convention.

UnitTests/BraintreeCoreTests/BTWebAuthenticationSession_Tests.swift
NastassiaRodzik: yeah the naming is not the best - changing it to match the codebase better. 
the intent behind this change is that private queue does not have a limitation of DispatchQueue.global() that has a limited pool of threads so the test will pass on both local computer and CI  .

Sources/BraintreePayPal/BTPayPalClient.swift
buzzamus: While it is most likely a non-issue in this block, I think the possibility does exist to keep `BTPayPalClient` and everything it holds in memory longer than the calling app may otherwise keep. We do call out in our CLAUDE.md file to follow this convention of always capturing `[weak self]`, so I think it makes sense to do it here as well

Sources/BraintreePayPal/BTPayPalClient.swift
buzzamus: oh good call out! Yep, I missed the completion callback portion

Sources/BraintreePayPal/BTPayPalClient.swift
NastassiaRodzik: done, added `[weak self]` in both places

buzzamus (approved):
Thanks for the updates!

jaxdesmarais (approved):

